PR #33, opened 2016-08-11
include doxygen configuration file

shajoezhu:


jeromekelleher:
Looks good, thanks @shajoezhu!

Two suggestions:

1. Can we change the project name to  "msprime" please? We don't use  "ms-prime" anywhere else so this would be confusing.

2. We should include running doxygen in the travis CI script so that this will stay current. Can you add this at the end of the 'script' section please?

shajoezhu:
Done :)

jeromekelleher:
Great, thanks @shajoezhu !

PR #58, opened 2016-10-15
Add 'file' option

petrelharp:
This allows two different methods for passing options in a file:
 1. doing '-f filename' will work if the *entire* command line is in the file
 2. inserting '@filename' will include commands in the file at the point it is inserted (this is the [fromfile](https://docs.python.org/3.5/library/argparse.html#fromfile-prefix-chars) option from argparse).

I have not figured out how to replicate `ms`'s behavior, allowing `-f` to coexist with other arguments: my implementation is straight from [this stackoverflow question](http://stackoverflow.com/questions/27433316/how-to-get-argparse-to-read-arguments-from-a-file-with-an-option-rather-than-pre), but the problem is that the `LoadFromFile` Action just calls `parse_args( )`, which causes an exception if the positional arguments aren't in the file.

### DIFF
--- a/msprime/cli.py
+++ b/msprime/cli.py
@@ -557,6 +557,16 @@ class IndexedAction(argparse._AppendAction):
         IndexedAction.index += 1
 
 
+class LoadFromFile (argparse.Action):
+    """
+    Argparse action class to allow passing a filename containing arguments
+    on the command line (for super-long argument files).
+    From http://stackoverflow.com/questions/27433316/how-to-get-argparse-to-read-arguments-from-a-file-with-an-option-rather-than-pre
+    """
+    def __call__ (self, parser, namespace, values, option_string = None):
+        with values as f:
+            parser.parse_args(f.read().split(), namespace)
+
 def get_mspms_parser():
     # Ensure that the IndexedAction counter is set to zero. This is useful
     # for testing where we'll be creating lots of these parsers.
@@ -692,6 +702,9 @@ def get_mspms_parser():
     group.add_argument(
         "--precision", "-p", type=positive_int, default=3,
         help="Number of values after decimal place to print")
+    group.add_argument(
+        "--filename", "-f", type=open, action=LoadFromFile,
+        help="Read command line arguments from this file.")
     return parser
 
 
--- a/msprime/cli.py
+++ b/msprime/cli.py
@@ -556,34 +556,39 @@ class IndexedAction(argparse._AppendAction):
             parser, namespace, (IndexedAction.index, values), option_string)
         IndexedAction.index += 1
 
+
 def convert_arg_line_to_args(arg_line):
     # from the docs on argparse.ArgumentParser.convert_arg_line_to_args
     return arg_line.split()
+
 
 def make_load_file_action(next_parser):
     """
     Argparse action class to allow passing a filename containing arguments
     on the command line (for super-long argument files).
-    From 
-        http://stackoverflow.com/questions/27433316/how-to-get-argparse-to-read-arguments-from-a-file-with-an-option-rather-than-pre
-    and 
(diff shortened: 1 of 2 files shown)


petrelharp:
[Stackoverflow](http://stackoverflow.com/questions/40060571/how-to-get-argparse-to-read-arguments-from-a-file-with-an-option-after-positiona/40062344) pointed out a  way to copy `ms`'s behavior: make two parsers, the first with just the positional arguments:
```
add_sample_size_argument(init_parser)
init_parser.add_argument(
    "num_replicates", type=positive_int,
    help="Number of independent replicates")
init_parser.add_argument(
    "-V", "--version", action='version',
    version='%(prog)s {}'.format(msprime.__version__))
```
and the second (`parser`) with everything else; and then we'd use them as:
```
args,more_args = init_parser.parse_known_args(arg_list)
parser.parse_args(args=more_args,namespace=args)
args.runner(args)
```

This uglifies the nice general code common to msp and mspms, though, so I haven't gone ahead and done it.

jeromekelleher:
Thanks a million for this @petrelharp, we should definitely support the ``-f`` option from ``ms``. Presumably lots of tools that automate calling ``ms`` and processing the results use this, so we should aim for 100% compatibility. In general, the motivation for ``mspms`` is to provide existing ``ms`` users with a smooth upgrade path. There are two guiding principles in my mind:

1. Absolute compatibility is the most important concern. If something works in ``ms``, the same thing should work in ``mspms`` (or it should blow-up noisily).
2. No new features (i.e., above what ``ms`` has) go into the ``mspms`` interface as it is already very complicated, and there are lots of nasty corner cases to be considered. 

With this in mind, I would prefer the two-parser option you have outlined here. It's OK if we lose some code sharing with ``msp``, we'll just need to refactor the test cases a little to deal with this. If you implement the parsers, I'm happy to work on updating the tests.

petrelharp:
Based on extremely light testing, I think this now works and matches 'ms' behavior (but see below), without messing with the msp interface, even.  I don't understand how the IndexedAction is working well enough to know if it's messing that up, though.

Ah, but at least one thing differs from ms: if, say, 'argfile.txt' specifies `-t 2`, and we run
```
ms 10 1 -f argfile.txt -t 10
```
then theta will be set to 10 (the last thing set), whereas ms throws an error.  This could be fixed by having the `LoadFromFile` action check for stuff that's already set in `namespace` and doing the appropriate thing.

jeromekelleher:
I like this @petrelharp, really nice and elegant. I don't think we need to worry too much about the last compatibility issue --- I doubt anyone's upstream code depends on ms's error behaviour, and so we just have to document that if a user provides conflicting options, the last value will be used. Having said that, if you think it's important to raise an error here I'm perfectly happy with that too. The main thing is that we're covering all the cases where ms would succeed.

Can you set the options to allow me to push commits to this PR please?

https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

There's a few minor style issues to be resolved and I'd like to add some test cases. I think it's best to do it all in one PR, so we can track the test coverage and so on. After that we're ready to merge I think.

Thanks again, this is a good addition to fill out our ms compatability!

petrelharp:
I don't think we need the error behavior particularly.

"Allow edits from maintainers" was already selected: are you not able to push to it?

jeromekelleher:
OK, I think that should cover it. I've changed things around a little bit to cover the error conditions more cleanly and to make sure we can't have '-f' options in the file (otherwise, we have to look out for recursion). I've also added some tests. Would you mind having a quick look at the tests and see if you can spot anything we've missed @petrelharp?

petrelharp:
Looks good to me.  Sorry for not being fully PEP8, btw.

I was thinking of adding the 'already specified' check, but then I noticed that check doesn't already happen on the command line.

jeromekelleher:
Great, thanks @petrelharp. I was also going to do the 'already specified' check when I noticed the same thing. If you think it matters we can sort it out fully in a follow-up.

Issue #74, opened 2016-11-02
Bug in subset() causing Python crash


jeromekelleher:
It looks like there is some memory corruption in the TreeSequence.subset() method. See here for a crash when running unrelated tests.

https://travis-ci.org/jeromekelleher/msprime/jobs/172700863

jeromekelleher:
Closing this issue as the crash has been fixed. See #79 for the underlying bug that exposed the incorrect error handling.

PR #76, opened 2016-11-02
Early version of subset-> bad records LibraryError

ashander:
And in subsequent updates these same records will cause more serious errors.

Running the commands below from this commit shows that the change to more
serious error was introduced by `4a87ead` (Added high-level interface and
tests.)

git cherry-pick 2595b
pip install -e . && python tests/subset_errors/reproduce_errors.py
git cherry-pick 4a87ead
pip install -e . && python tests/subset_errors/reproduce_errors.py

NB: the .tsv files commited here contain other records that will cause similar
errors

This may help shed some light on #74


ashander:
I'll try to reduce the size of this reproduce example, but I'm hoping it's helpful to have a reproducible way to get these errors. 

A bit more about the example records: They were developed from work with @petrelharp in attempting to fill msprime TreeSequence forward in time, then pull out a subset. Because some events result in single children, we create a second "phantom" sibling (all with the same id). This makes the records look strange but they seem to import fine with `load_txt` so I'm assuming they are "ok".

jeromekelleher:
This is extremely helpful, thanks @ashander. Working on it now...

jeromekelleher:
I don't think the example in the ``reproduce_error.py`` file is valid topologically @ashander. If you try to run ``ts.trees()`` on the original tree sequence, it detects a problem (correctly, I think) and raises an error. I use a Python version of the tree generation algorithm to look at things like this, as it's a lot more forgiving:

```python
def trees(records):
    M = len(records)
    I = sorted(range(M), key=lambda j: (records[j].left, records[j].time))
    O = sorted(range(M), key=lambda j: (records[j].right, -records[j].time))
    pi = [-1 for j in range(max(r.node for r in records) + 1)]
    chi = [[] for j in range(max(r.node for r in records) + 1)]
    j = 0
    k = 0
    while j < M:
        x = records[I[j]].left
        while records[O[k]].right == x:
            h = O[k]
            print("\tout:", records[h])
            chi[records[h].node] = []
            for q in records[h].children:
                pi[q] = -1
            k += 1
        while j < M and records[I[j]].left == x:
            h = I[j]
            print("\tin:", records[h])
            chi[records[h].node] = records[h].children
            for q in records[h].children:
                pi[q] = records[h].node
            j += 1
        yield pi, chi
```

This returns an iterator over the ``parent`` and ``children`` arrays. Use it like ``trees(list(ts.records()))``. When I run it on this example, I get far too many records with 0 as the left coordinate, for a sample of size 7 (which is what we infer given the input records). The trees returned are  topologically incorrect, with several roots.

I'm happy to help with getting your input data into coalescence records, so please ping back here or email me if you have any questions.

ashander:
Thanks! This function is indeed very helpful. I think what we were trying to do in producing this example wasn't made very clear, but I'm glad if it helped in debugging.

I will followup, I think on email with Peter, re getting our data into the msprime format. 
But I'll this close for now

PR #90, opened 2016-11-29
Updated 'subset' algorithm

jeromekelleher:
This implementation of the subset() algorithm should be much more robust, if somewhat slower.

@ashander and @hyanwong, can you try this out and see if it works for you please?

[Edited] 

**WARNING** The ``subset()`` method has been renamed to ``simplify()``. This seems like a more appropriate name given what we're now doing with the algorithm. Sorry to break your code!


ashander:
thanks! will do. cc @petrelharp

hyanwong:
Seems to work so far. The Makefile change accidentally has 8 spaces rather than a tab on the last line, though, which borks the make process.

ashander:
@hyanwong I don't see a Makefile change here. the tip builds fine with `pip install -e .`

hyanwong:
@ashander I think I messed up and left an old file hanging around.

jeromekelleher:
Yes, the makefile(s) shouldn't have any bearing on this @hyanwong, they are just for local development. If you want to install this version, you can just use ``python setup.py install --user`` from a checked out copy. 

Thanks for the ``pip install -e`` pointer @ashander, that's really useful!

hyanwong:
@jeromekelleher thanks. I'm using the makefile so that I can keep a local copy within a single directory (a submodule of our treeseq-inference repo) without polluting my python install and without using virtualenvs. That way I can just `import` it from the local folder, without having to do a python install.

jeromekelleher:
I see @hyanwong; you're appending the ``msprime`` subdirectory to the Python path to ensure you're using the local copy and not the installed one. That should work fine. 

There are no changes to the Makefile in this patch, so I think it is some local changes in your copy that caused your problems.

PR #95, opened 2016-12-01
automatic sample size determination?

petrelharp:
In `lib/tree_sequence.c`, when reading in records the sample size is set as follows:
```
        self->sample_size = GSL_MIN(self->sample_size, records[j].node);
```
which defines the sample size to be the smallest label on an internal node.  If there are then incompletely specified tips with a smaller label than any internal nodes, this will cause problems, I would think.

It seems like we should do one of the following things:

1. Manually specify the sample size, rather than determining it automatically.
2. Add the requirement that the smallest non-sampled label be an internal node to the list of requirements.

What do you think, @jeromekelleher ?

### DIFF
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -597,8 +597,7 @@ class TestWithVisuals(TopologyTestCase):
                       {0: 6, 1: 5, 2: 4, 3: 4, 4: 5, 5: 6, 6: -1, 7: -1}]
         ts = build_tree_sequence(records)
         tree_dicts = [t.parent_dict for t in ts.trees()]
-        print("Skipping test of sample size.")
-        # self.assertEqual(ts.sample_size, 3)
+        self.assertEqual(ts.sample_size, 4)
         self.assertEqual(ts.num_trees, 3)
         self.assertEqual(ts.num_nodes, 8)
         # check topologies agree:
@@ -661,11 +660,6 @@ class TestWithVisuals(TopologyTestCase):
                     self.assertEqual(t[k], a[k])
                 else:
                     self.assertEqual(a[k], msprime.NULL_NODE)
-        #  FAILS:
-        # # check can draw trees
-        # trees=ts.trees()
-        # for x in trees:
-        #   x.draw(path='test.svg')
         self.verify_simplify_topology(ts, [0, 1, 2])
 
     def test_single_offspring_records(self):
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -77,9 +77,9 @@ class TestRecordSquashing(TopologyTestCase):
     def test_single_record(self):
         records = [
             msprime.CoalescenceRecord(
-                left=0, right=1, node=2, children=(0, 1), time=1, population=0),
+               left=0, right=1, node=2, children=(0, 1), time=1, population=0),
             msprime.CoalescenceRecord(
-                left=1, right=2, node=2, children=(0, 1), time=1, population=0),
+               left=1, right=2, node=2, children=(0, 1), time=1, population=0),
         ]
         ts = build_tree_sequence(records)
         self.assertEqual(list(ts.records()), records)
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Nice examples, thanks @petrelharp. Also, I'm very impressed with your ASCII art skills! I think that for now we should do (2), i.e., explicitly require that the smallest internal node be equal to the sample size. How awkward is this for your use case?

In the future, I think we'll need to change things around a bit so that we explicitly specify the samples in some way. This might also involve removing the restriction that the samples are {0, .., n - 1}. One way we might do this is to require that a record with zero children is provided for each sample. We might also think about normalising the input by splitting up the input into 'node' (ID, time, population) and 'topology' (left, right, parent, children) records.

This would end up being quite a bit change though, which is why I don't want to get into it now. I would prefer to try and get the changes we've been working on stabilised, so that it's at least possible to import data in various forms. Once we have used this for a while and had some more experience with it, we can look to changing the interfaces a bit to make sample specification more natural. 

That's just my current take though --- if sample specification is a problem for your application then we can definitely look at figuring this out sooner rather than later.

tests/test_topology.py
```diff
@@ -523,3 +523,160 @@ def test_break_single_tree(self):
...
+        self.assertEqual(ts.num_trees, 3)
+        self.assertEqual(ts.num_nodes, 8)
+        # check topologies agree:
+        for a,t in zip(true_trees,tree_dicts):
+            for k in a.keys():
+                if k in t.keys():
+                    self.assertEqual(t[k],a[k])
+                else:
+                    self.assertEqual(a[k],-1)
+        ## FAILS:
```
jeromekelleher: Ah, didn't think of this. I guess we should open an issue on this and tackle it explicitly?

tests/test_topology.py
```diff
@@ -523,3 +523,160 @@ def test_break_single_tree(self):
...
+            msprime.CoalescenceRecord(
+                left=0.8, right=1.0, node=4, children=(2, 3), time=0.4, population=0),
+            msprime.CoalescenceRecord(
+                left=0.0, right=1.0, node=5, children=(1, 4), time=0.5, population=0),
+            msprime.CoalescenceRecord(
+                left=0.8, right=1.0, node=6, children=(0, 5), time=0.7, population=0),
+            msprime.CoalescenceRecord(
+                left=0.0, right=0.2, node=7, children=(0, 5), time=1.0, population=0),
+            ]
+        true_trees = [ {0: 7, 1: 5, 2: 4, 3: 4, 4: 5, 5: 7, 6:-1, 7:-1},
```
jeromekelleher: Perhaps it would be simpler to leave out the ``u: -1`` nodes here? You could just compare the lists of trees directly then without having to zip them. I previously included the ``root: -1`` for trees in these dictionaries, but I think it's better to leave them out since we are potentially dealing with trees that have many roots.

petrelharp: Simpler, maybe... but easier to check that I've got all the nodes in?  I think I'm comparing them in a way that ignores the -1's, so it's ok?

tests/test_topology.py
petrelharp: Up to you: I didn't know how central the tree-drawing stuff is.  (It is very nice...)

petrelharp:
(2) would be totally fine.  I realized that there is a simple strategy for my use case that will conform: suppose we want to take S samples eventualliy from a forwards-time simulation.  Start numbering individuals at S, assinging numbers incrementally increasing.  At the last generation, renumber the samples to be 0...S-1.  This will satisfy the requirements except in exceptional cases.

tests/test_topology.py
jeromekelleher: Up to you @petrelharp, I don't have a strong opinion about it.

tests/test_topology.py
jeromekelleher: OK, I've made an issue for this in #97. Let's delete this code here, as there will need to be separate tests for the tree drawing.

jeromekelleher:
Yes, this sounds like a good strategy @petrelharp, I think this should work well. OK, I think we can merge this if you remove the failing case and run ``flake8`` on the code. Unless you have some more cases you want to add?

petrelharp:
I'll have some more checks that will also test .simplify(); unless you'd rather those go elsewhere.

jeromekelleher:
Here is good, thanks!

FYI, you'll need to rebase this branch on upstream/master for the CI to pass. (A change in upstream components meant I had to change the CI a little bit.)

petrelharp:
This should be ready to merge now.

jeromekelleher:
Merging; thanks @petrelharp!

PR #98, opened 2016-12-06
moved division outside of sum

petrelharp:
Reading the code I noticed this obvious speedup: why not?

I also put in the definition, as it's easy to mix this up with the same thing divided by sequence length.

### DIFF
--- a/lib/tests.c
+++ b/lib/tests.c
@@ -3864,6 +3864,35 @@ verify_vargen(tree_sequence_t *ts)
 }
 
 static void
+verify_stats(tree_sequence_t *ts)
+{
+    int ret;
+    uint32_t sample_size = tree_sequence_get_sample_size(ts);
+    uint32_t *samples = malloc(sample_size * sizeof(uint32_t));
+    uint32_t j;
+    double pi;
+
+    CU_ASSERT_FATAL(samples != NULL);
+
+    ret = tree_sequence_get_pairwise_diversity(ts, NULL, 0, &pi);
+    CU_ASSERT_EQUAL_FATAL(ret, MSP_ERR_BAD_PARAM_VALUE);
+    ret = tree_sequence_get_pairwise_diversity(ts, NULL, 1, &pi);
+    CU_ASSERT_EQUAL_FATAL(ret, MSP_ERR_BAD_PARAM_VALUE);
+    ret = tree_sequence_get_pairwise_diversity(ts, NULL, sample_size + 1, &pi);
+    CU_ASSERT_EQUAL_FATAL(ret, MSP_ERR_BAD_PARAM_VALUE);
+
+    for (j = 0; j < sample_size; j++) {
+        samples[j] = j;
+    }
+    for (j = 2; j < sample_size; j++) {
+        ret = tree_sequence_get_pairwise_diversity(ts, samples, j, &pi);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Yes, good idea. The only potential downfall I can think of  is possible overflows, or more realistically, a loss of precision due to accumulating a very large number before doing the division. This seems unlikely now that I think of it though.

If you can fix up the pep8 violation (trailing whitespace in the Python code) then we're good to merge I think. I have a git commit hook set up to run this before each commit, which saves a lot of time. I just took the sample ``pre-commit`` hook and added 
```bash
exec flake8 --max-line-length 89 setup.py msprime tests
```
at the bottom.

petrelharp:
PEP8!  Right, sorry.  It is done.

jeromekelleher:
Hmm. Tests are really failing now:

https://travis-ci.org/jeromekelleher/msprime/jobs/181898256#L747

So, the failures are innocuous enough and can be fixed with an 'assertAlmostEqual' rather than assertEqual. This is making me nervous though... Could this loss of precision become significant? Is there a measurable speed up from moving the division outside of the loop?

petrelharp:
Huh, that's not how I thought floating point arithmetic worked?  Doing (a+b)/c instead of (a/c+b/c) doesn't change the relative orders of magnitudes of the things being added, which is what I thought was the important thing.  

By adding a print statement to the test, I see that with this change, the size of the difference is of the same order of magnitude as the difference that already exists to `simple_get_pairwise_diversity`.  For instance, on `master`, one set of pi values are:
```
pi1,pi2,pi3: 41.999999999999986 42.0 41.999999999999986
```
By this measure we are not losing any precision.

(That was produced by doing:
```
--- a/tests/test_highlevel.py
+++ b/tests/test_highlevel.py
@@ -452,6 +452,7 @@ class HighLevelTestCase(tests.MsprimeTestCase):
         pi1 = ts.get_pairwise_diversity()
         pi2 = simple_get_pairwise_diversity(haplotypes)
         pi3 = get_pairwise_diversity(ts)
+        print("pi1,2,3:",pi1,pi2,pi3)
         self.assertAlmostEqual(pi1, pi2)
         self.assertEqual(pi1, pi3)
         self.assertGreaterEqual(pi1, 0.0)
```
and running pytest with `-s`.)

jeromekelleher:
Yeah, you're right, it's fine. Can you change the ``assertEquals`` to ``assertAlmostEqual`` please? We can merge then.

jeromekelleher:
Hold on, I remember rebased and force pushed this to clean it up a bit, so your branch will be a mess now. I'll fix the asserts.

jeromekelleher:
OK, merging this. Thanks again @petrelharp!

jeromekelleher:
As a postscript to this one, I ran the pi computation on a 10 million sample human chr1 tree sequence I have lying around:
```
Before:
jk@holly$ python dev_tmp.py lib/tmp__NOBACKUP__/10000000.hdf5
pi =  153741.101922
took 73.387776 seconds

After:
jk@holly$ python dev.py lib/tmp__NOBACKUP__/10000000.hdf5
pi =  153741.101912
took 77.703859 seconds
```
So, the answers are the same by any reasonable definition. At first glance the new version is actually slower than the old one, but this isn't true. Over 99% of the time in this function is spent in ``sparse_tree_next``, so it makes no real difference what we do in the inner loop.

petrelharp:
Heh. =)

PR #105, opened 2016-12-15
Recurrent mutations

jeromekelleher:
Big refactor to enable recurrent mutations and fix various shortcomings in the data model.


jeromekelleher:
Merging this beast before it gets any bigger. The APIs and file format here are transitional, as explained in issue #120.

Issue #109, opened 2016-12-22
Add length field to VCF contig description


terhorst:
While not technically required by the VCF spec, adding a "length=" field to the VCF contig header is helpful for e.g. determining the number of nonsegregating sites.

PR #110, opened 2016-12-22
add length= field to vcf contig description

terhorst:
This resolves issue #109.

### DIFF
--- a/lib/msprime.h
+++ b/lib/msprime.h
@@ -430,8 +430,10 @@ typedef struct {
     char *vcf_genotypes;
     size_t vcf_genotypes_size;
     size_t record_size;
-    double sequence_length;
-    unsigned long last_position;
+    size_t num_mutations;
+    unsigned long contig_length;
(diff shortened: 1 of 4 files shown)


jeromekelleher:
This is great, thanks @terhorst! I've made a few changes to allow us to deal with cases where we can't just use the ``sequence_length`` directly (i.e., we are using the unit interval and we have lots of mutations). It's a bit messy, but I can't think of any other way to deal with the transition between real and integer coordinates, and it should do the right thing in the majority of cases.

There was also a minor issue with the ``snprintf`` that was flagged by valgrind. I tidied this up, and hopefully made it easier to extend in the future when we want to add more information to the header.

Would you mind having a quick look over this? If you're happy with the changes I'll merge.

terhorst:
This is a definite improvement over what I threw together :-) Thanks for
your help.

jeromekelleher:
Merged; thanks again @terhorst!

PR #113, opened 2017-01-05
general branch length statistics computation

petrelharp:
There are two different implementations of the same function in msprime/branch_stats.py - and, these are tested (for a few statistics) in tests/test_branch_stats.py .

The easier version (`branch_stats_node_iter()`) just uses the tracked_leaves option to the .trees() iterator.  The other version (`branch_stats()`) does it all by hand with .diffs() (and not .trees()).

Here is the general problem these solve, and the algorithm:

Let $A_1, ..., A_n$ be fixed sets of samples.
For a node $z$ in a tree, let $t(z)$ be the length of the branch above $z$, 
and let $X(z) = ( x_1(z), ... x_n(z) )$ be a list of integers, 
with $x_k(z)$ the number of samples in $A_k$ that subtend $z$.  
We will want to count the total length of all branches weighted by a function of $X$ 
which we encode by a function $f : X \to R$.  
Then for a given tree $T$ we define $L(T)$ to be this total length:
$$
   L(T) = \sum_z f(X(z)) t(z)
$$

Now, let $L$ be the sum of these along the tree sequence: if $T_i$ extends for length $u_i$ along the genome, then we want
$$
   S = \sum_i u_i L(T_i) .
$$

For instance: we get sequence length multiplied by average TMRCA between $a$ and $b$ with $A_1 = {a}$ and $A_2 = {b}$ and $f(x_1,x_2) = \text{xor}(x_1,x_2)$.

To compute this, we'll keep track of X(z) for each node of the current tree, and update these as we move along the tree sequence:

 0) initialize $X$
 1) $S += u_i * L(T_i)$
 2) move to the next tree and update $X.$

As for updating $X$, note that $X$ is additive: 
if node $z$ has children $y_1, ..., y_m$ then $X(z) = sum_j X(Y_j)$.  
So, to update we just need to propagate changes up from any node that has changed.
To avoid re-computing $f(X)$ we would also cache these and only update those that were changed: 
to do this, let $F(z) = f(X(z))$. Concretely, we can do this by:

  - For each record with parent $z$ and children $y_1, ..., y_m$ that is removed,

      - $L -= t(y_i) f(X(y_i))$ for each $i$ because these branches no longer exist
      - $f_\text{old} = f(X(z))$
      - $dx = X(y_1)+\cdots+X(y_m)$
      - $X(z) -= dx$ (note $X(z)$ is not 0 if $z\in A_i$ for some $i$)
      - $L += t(z) (f(X(z))-f_\text{old})$
      - for each ancestor $u$ of $z$ up the tree,
          do $f_\text{old}(u)=f(X(u))$ and $X(u) -= dx$ and $L += t(u) (f(X(u))-f_\text{old}(u))$ 

  - then, for each record with parent $z$ and children $y_1, ... y_m$ that is added,

      - $L += t(y_i) f(X(y_i))$ for each $i$ 
      - $f_\text{old} = f(X(z))$
      - $dx = X(y_1) + ... + X(y_m)$
      - $X(z) += dx$
      - $L += t(z) (f(X(z))-f_\text{old})$
      - for each ancestor $u$ of $z$ up the tree,
          do $f_\text{old}(u)=f(X(u))$ and $X(u) += dx$ and $L += t(u) (f(X(u))-f_\text{old}(u))$ 

After this we have $L(T)$ for the new tree, so we can do $S += L * u_i$.

Note that I've said this in terms of branch lengths, 
but by swapping out branch lengths for numbers of mutations, 
we'd be able to easily compute a large class of sequence statistics 
(e.g., Patterson's $f$ statistics, which involve four groups $A_1, ... A_4$).


petrelharp:
I did some testing with fairly large tree sequences, and the quick, simple version of divergence and the Y-statistic that I wrote for testing purposes works almost instantly.  I've added these to msprime/branch_stats.py for comparison.

petrelharp:
As for timing: already computing all pairwise divergences with 10 samples my general-purpose, compute-all-at-once code goes more quickly than traversing the `.trees()`:
```
import msprime
from timeit import default_timer as timer

ts = msprime.simulate(10, recombination_rate=1000)

A = [ [a] for a in ts.samples() ]
n = len(A)
def f(x):
    return [ float(x[j]*(1-x[k]) + (1-x[j])*x[k]) for j in range(n) for k in range(n) if j<k ] 

start = timer()

brlen = [ msprime.branch_length_diversity(ts,A[j],A[k]) for j in range(len(A)) for k in range(len(A)) if j<k ]

mid1 = timer()

bsv = msprime.branch_stats_vector(ts,A,f)

mid2 = timer()

ni = msprime.branch_stats_vector_node_iter(ts,A,f,method='length')

end = timer()

print("Should be zero:",max([abs(a-b) for a,b in zip(brlen,bsv)]))
print("Should be zero:",max([abs(a-b) for a,b in zip(brlen,ni)]))

print("branch_length_diversity():", mid1-start)
# 4.130114852916449
print("branch_stats_vector()", mid2-mid1)
# 3.1504349421011284
print("branch_stats_vector_node_iter()", end-mid2)
# 11.936302176909521
```
Doing 10 samples takes 3sec, doing 20 samples takes 17sec, and doing 40 samples takes 108sec.   Doing 100 takes a while.

This suggests to me that C-ifying the code in `branch_stats_vector` is worthwhile.

msprime/branch_stats.py
```diff
@@ -0,0 +1,229 @@
...
+            # y x  z
+            S += path_length(tr,x,xz_mrca)*tr.length
+        elif xz_mrca == yz_mrca:
+            #   /\
+            #  / /\
+            # z x  y
+            S += path_length(tr,x,xy_mrca)*tr.length
+    return S/ts.sequence_length
+
+def branch_length_f4(ts,A,B,C,D):
```
jeromekelleher: This method (and those above are really just for testing, so lets move these into the test_branch_stats.py file.

msprime/branch_stats.py
```diff
@@ -0,0 +1,229 @@
...
+    and averages this across the tree sequence ts, weighted by genomic length.
+
+    Doesn't do method='mutations'.
+    '''
+    out = branch_stats_vector(ts,leaf_sets,lambda x: [weight_fun(x)],method)
+    if len(out)>1:
+        raise ValueError("Expecting output of length 1.")
+    return out[0]
+
+def branch_stats_vector_node_iter(ts,leaf_sets,weight_fun,method='length'):
```
jeromekelleher: Ultimately, ``branch_stats`` (or something similar) will be a method of the tree sequence. I think we should offer the general interface as something like ``branch_stats(self, leaf_sets, weight_fun)``, and use the most efficient algorithm we have implemented in Python. So, let's move the ``branch_stats_vector_node_iter`` function into the TreeSequence, and move the less efficient implementations into the ``test_branch_stats.py``.

I think we should drop the 'method=whatever' argument for now. Once we have the ``num_branch_mutations`` method (#115), this will be easy to add.

tests/test_branch_stats.py
```diff
@@ -0,0 +1,358 @@
+
```
jeromekelleher: Please add the standard __future__ imports  here (important for py2/py3)

petrelharp:
I have:
 - PEP8-ified
- moved branch_stats and branch_stats_vector to the TreeSequence class, and all the other functions to test_branch_stats.py
- added __future__ stuff

... the tests aren't passing, but they aren't passing for me on master either: change in interface?

jeromekelleher:
I've updated this to use the new (but not final!) interface for importing records @petrelharp. Right now we're specifying ``samples`` as well as the records and mutations, but the plan is to specify a lists of ``nodes``, ``edgesets`` (== coalescence records - time and population information) and ``mutations``.

Another change that has happened is that mutations now have multiple nodes, since they can happen at many different locations on the tree. Therefore, the ``mutation.node`` attribute has changed to a list ``mutation.nodes``.

jeromekelleher:
Sorry, I lost track of this @petrelharp. I think this is pretty much ready to merge, right? There's some stuff we'd need to clean up before release, but that's a ways off at the moment. We can open some issues for these, but in the mean time we should merge so the code gets carried forward.

What do you think?

petrelharp:
Whoops - I never updated.  Yes, this is doing fine, and should be merged.

petrelharp:
There; I've rebased.

jeromekelleher:
This is a great addition thanks @petrelharp. I'm really looking forward to C-ifying this beautiful algorithm and watching it fly! (Might be a while though, unfortunately...)

PR #114, opened 2017-01-06
remove overly strict check in ts.diffs()

petrelharp:
The check was
```
assert(next_node_record < 2 * self->sample_size); 
```
and it caused the example in `tests/test_topology.py::test_many_single_offspring()` to not be able to produce a .diffs() iterator, failing with the error:
```
python3.5: lib/tree_sequence.c:2266: tree_diff_iterator_next: Assertion `next_node_record < 2 * self->sample_size' failed.
```

I don't think the check is vital, and will not be true for nonminimal tree sequences.

### DIFF
--- a/lib/tests.c
+++ b/lib/tests.c
@@ -3666,6 +3666,35 @@ test_nonbinary_tree_sequence_diff_iter(void)
 }
 
 static void
+test_unary_tree_sequence_diff_iter(void)
+{
+    int ret;
+    const char * text_records =
+        "2 10 4 2,3 0.071 0\n"
+        "0 2  5 1,3 0.090 0\n"
+        "2 10 5 1,4 0.090 0\n"
+        "0 7  6 0,5 0.170 0\n"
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Thanks for this @petrelharp, good spot. I'll need to write a few low-level tests which applies the diffs() algorithm on non-minimal tree sequences in the low-level tests to make sure there aren't any buffer overflows or whatever. I'll put them in when I get a chance over the next few days and merge this then.

jeromekelleher:
Looks good @petrelharp, I think the problem was a left-over assert for an old bound that hadn't been triggered. 

I've also added a new function ``cr`` in the test_topology.py file to make defining the tree sequences a bit less tedious. What do you think?

petrelharp:
Ah, that bound makes more sense.

`cr`: yes, very much tidier!

jeromekelleher:
Great, thanks @petrelharp. Merging.

PR #119, opened 2017-01-19
disambiguating BAD_COAL_RECORDS

petrelharp:
A slight disambiguation that I unfortunately had need of.


jeromekelleher:
Looks good to me, thanks @petrelharp. I think you just need to fix a failing C test case which is expecting the old generic error code.

We should try to make sure that all these codes get hit in the tests too. I'm happy to jump in to write a test case or two if it's not obvious how to do this.

petrelharp:
Huh, I thought I fixed those test cases - when I run pytest I only get an error that I don't have the module 'vcf' installed; I thought that meant the rest of everything ran OK.  (where do you get the vcf module, btw?)

jeromekelleher:
These are different tests @petrelharp, you run them by running the ``./tests`` executable in the lib directory. They aren't part of the Python test suite at all. Build them using ``make`` in the lib directory. You'll need to have CUnit and libconfig installed also.

For the vcf module, you want PyVCF. The ``requirements.txt`` file lists the modules we need for testing (probably should contain numpy though).

petrelharp:
hah; thanks.  I'm on it.

petrelharp:
OK, that passes now.

It would be great if you could write those test cases - it would take me some figuring out.   I'm also not sure when MSP_ERR_BAD_COALESCENCE_RECORDS itself will get triggered any more (but maybe that's OK).

jeromekelleher:
Will do @petrelharp, I'll take it from here.

Apologies for the lack of developer documentation, it's definitely something I have to do...

PR #128, opened 2017-01-25
Mutation generator

jeromekelleher:
Allow the mutation generator to be specified explicitly in simulate. Usage:

```python
mut_rng = msprime.RandomGenerator(mutation_seed)
mutation_generator = msprime.MutationGenerator(mut_rng, mutation_rate)
ts = msprime.simulate(
    sample_size=10, mutation_generator=mutation_generator, random_seed=tree_seed)
```

This opens up the API to specifying different mutation models, as well as having control of the random seed used for generating the mutations.

What do you think @hyanwong? This should fit your use-case where you want to throw different mutations on the same trees.

### DIFF
--- a/msprime/cli.py
+++ b/msprime/cli.py
@@ -158,6 +158,8 @@ class SimulationRunner(object):
         self._random_generator = msprime.RandomGenerator(seed)
         self._ms_random_seeds = ms_seeds
         self._simulator.set_random_generator(self._random_generator)
+        self._mutation_generator = msprime.MutationGenerator(
+            self._random_generator, self._mutation_rate)
 
     def get_num_replicates(self):
         """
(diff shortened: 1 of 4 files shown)


hyanwong:
Yes, that would be perfect. Thanks.

hyanwong:
Also I assume this allows flexibility e.g if you want recurrent / discretised / back mutations etc. Not sure how this plugs in to using bases (A,T,G,C) vs (1,0) for mutations. Could you just plug in e.g. a Jukes/Cantor mutation generator, and get out bases. Presumably you would need to specify a mosaic ancestral sequence in this case (can't assume it's all zeroes).

jeromekelleher:
> I also assume this allows flexibility e.g if you want recurrent / discretised / back mutations etc. Not sure how this plugs in to using bases (A,T,G,C) vs (1,0) for mutations. Could you just plug in e.g. a Jukes/Cantor mutation generator, and get out bases. Presumably you would need to specify a mosaic ancestral sequence in this case (can't assume it's all zeroes).

That's the plan all right. Actual bases (rather and 1s and 0s) will have to wait until the v5 file format and low-level replumbing is done, but the API here should be able to encapsulate arbitrarily complex mutation models.

jeromekelleher:
OK, going to merge this so and see how we get on with it. It'll remain undocumented for the moment until more of the details have been fleshed out.

PR #131, opened 2017-01-27
ts causes .simplify() to fail

petrelharp:
I've been running across such tree sequences.  Here is a small one, illustrated, causing
```
E       _msprime.LibraryError: Bad coalescence records in file: right coordinate not matching any left coordinate.
```

Haven't figured out *why*.

### DIFF
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -769,3 +769,67 @@ class TestWithVisuals(TopologyTestCase):
         self.verify_simplify_topology(ts, [0, 1])
         self.verify_simplify_topology(ts, [1, 2])
         self.verify_simplify_topology(ts, [2, 0])
+
+    def test_disconnected(self):
+        
+        # 6     5      7___    .   5       7___    .   5       7___    .   5       7     6 .
+        #       |     / \  \   .   |      / \  \   .   |      / \  \   .   |      / \    | .
+        # 5    12   13   10 11 .  12    13   10 11 .  12    13   10 11 .  12    13   10 11 .
+        #          / |\    \   .   |   / |\    \   .   |\   /|\    \   .   |\   /|\    \   .
+        # 4      14  | \   15  .   | 14  | \   15  .   | |14 | \   15  .   | |14 | \   15  .
+        #        /\  |  \   |  .   |  \  |  \   |  .   | |   |  \   |  .   | |   |  \   |  .
+        # 3    17 16 18  |  |  .  17  16 18  |  |  .  17 16  18  |  |  .  17 16  18  |  |  .
+        #       |        |  |  .   |         |  |  .   |         |  |  .   |         |  |  .
+        # 2    21       20 19  .   21       20 19  .   21       20 19  .   21       20 19  .
+        #      /|        |  |  .   /|        |  |  .   /|        |  |  .   /|        |  |  .
+        # 1  22 |        |  |  . 22 |        |  |  . 22 |        |  |  . 22 |        |  |  .
+        #     | |        |  |  .  | |        |  |  .  | |        |  |  .  | |        |  |  .
+        # 0   2 3        1  0  .  2 3        1  0  .  2 3        1  0  .  2 3        1  0  .
+        #
+        # ... continued:
+        #
+        # 6   .   5      6   7    .       5      6   7    .       5      6   7    .
+        #     .   |     / \  |    .       |     / \  |    .       |     / \  |    .
+        # 5   .  12    13 11 10   .      12    13 11 10   .      12    13 11 10   .
+        #     .   |\   /|\    \   .     / |\   /|\    \   .     / |\   /|     \   .
+        # 4   .   | |14 | \   15  .    /  | |14 | \   15  .    /  | |14 |     15  .
+        #     .   | |   |  \   |  .   /   | |   |  \      .   /   | |   |         .
+        # 3   .  17 16  18  |  |  .  |   17 16  18  |     .  |   17 16  18        .
+        #     .   |         |  |  .  |    |         |     .  |    |\              .
+        # 2   .   21       20 19  . 19    21       20     . 19   21 20            .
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I think this is possibly a problem related to the automatic sample-size detection. There's no node 4 in here, so the code thinks the same size is 5. If you renumber the internal nodes does this happen?

Automatic sample size detection is bad, it will die soon.

jeromekelleher:
Hmmm, no, there's more to it than that, ignore previous comment.

jeromekelleher:
This _is_ weird. Here are the diffs (after joining two redundant records). 

```
x =   0
        out =  []
        in  =  [(22, (2,), 1.0), (21, (3, 22), 2.0), (20, (1,), 2.0), (19, (0,), 2.0), (17, (21,), 3.0), (15, (19,), 4.0), (14, (16, 17), 4.0), (13, (14, 18, 20), 5.0), (10, (15,), 5.0), (7, (10, 11, 13), 6.0), (5, (12,), 6.0)]

x =   0.18
        out =  [(14, (16, 17), 4.0)]
        in  =  [(14, (16,), 4.0), (12, (17,), 5.0)]

x =   0.21
        out =  [(12, (17,), 5.0), (14, (16,), 4.0)]
        in  =  [(12, (16, 17), 5.0)]

x =   0.34
        out =  [(7, (10, 11, 13), 6.0), (17, (21,), 3.0)]
        in  =  [(17, (21,), 3.0), (7, (10, 13), 6.0), (6, (11,), 6.0)]

x =   0.43
        out =  [(6, (11,), 6.0), (7, (10, 13), 6.0)]
        in  =  [(7, (10,), 6.0), (6, (11, 13), 6.0)]

x =   0.75
        out =  [(12, (16, 17), 5.0), (15, (19,), 4.0)]
        in  =  [(12, (16, 17, 19), 5.0)]

x =   0.98
        out =  [(13, (14, 18, 20), 5.0), (17, (21,), 3.0)]
        in  =  [(17, (20, 21), 3.0), (13, (14, 18), 5.0)]
```

After we run simplify(), we get the following records:

```
CoalescenceRecord(left=0.0, right=1.0, node=4, children=(2, 3), time=2.0, population=0)
CoalescenceRecord(left=0.98, right=1.0, node=5, children=(1, 4), time=3.0, population=0)
CoalescenceRecord(left=0.0, right=0.18, node=6, children=(1, 4), time=5.0, population=0)
CoalescenceRecord(left=0.75, right=0.98, node=7, children=(0, 4), time=5.0, population=0)
CoalescenceRecord(left=0.98, right=1.0, node=7, children=(0, 5), time=5.0, population=0)
CoalescenceRecord(left=0.0, right=0.18, node=8, children=(0, 6), time=6.0, population=0)
CoalescenceRecord(left=0.18, right=0.43, node=8, children=(0, 1), time=6.0, population=0)
```
which have a right=0.43 but no corresponding left. Any ideas @petrelharp?

petrelharp:
Geez, good point about the missing '4'.  It was late.  There were two wierd things that happened around there.  One was the invisible recombination at 0.47, but removing that doesn't fix it.  The other is that 0 becomes disconnected - no longer in the same tree as 1.  Adding a root above 5,6,7 fixes the problem, so that's my guess.  (I've added a cr; uncommenting fixes the problem.)

jeromekelleher:
This is a doozy @petrelharp, it's not obvious to me why this is failing. I'm getting the same answer from two completely different implementations of the subset algorithm so it seems unlikely it's a simple bug. The most likely thing I think is that there must be some subtle aspect of working with many disconnected trees that the algorithm is not capturing.

Is it always on disconnected trees you see this? Any background info would be helpful in trying to figure this out. 

Also, how _do_ you generate these ASCII art trees? They're amazing!

jeromekelleher:
Ah, no, got it. The problem is I hadn't thought about unary paths to root in the simplify() algorithm, and had assumed that we'd always meet the path from another sample on the way upwards to root. Fixing...

jeromekelleher:
This isn't straightforward unfortunately... in the efficient algorithm used in tree_sequence.c we're using the assumption that paths cross eventually quite heavily, and I can't see a simple way to update the algorithm to allow for unary paths to root. I think we need to figure out how to track multiple roots efficiently first before this can be tackled properly, as at least we will know when we need to invoke the special case behaviour. I'm going to open an issue on this.

In the short term, all I can suggest is to either not run simplify() on the trees you are generating, or to somehow force a root for all local trees. I don't have any great ideas on how to do the latter, but you might have access to some information in your simulation that makes this easier.

petrelharp:
To provide better input I'll have to get my head around the simplify algorithm better

But - and I should have said this before - I think assuming the trees are not disconnected is fine - it would be easy to add a single record at the start saying that everyone in the initial generation is descended entirely from one ancestor a long time previous.

Hm: perhaps that provides a way to use the simplify algorithm: add a node at some remotely distant time that every root inherits from; simplify; then prune these out.

Punting to an issue is fine, though - should just keep track of these requirements.

petrelharp:
And: ascii trees?  Just vim... =)

jeromekelleher:
> But - and I should have said this before - I think assuming the trees are not disconnected is fine - it would be easy to add a single record at the start saying that everyone in the initial generation is descended entirely from one ancestor a long time previous.

OK great, let's do this for now if it doesn't muck up your code too much. The simplify() algorithm is subtle and I don't want to get bogged down in it again right now. As far as I can tell, unary paths will have to be dealt with in a special case, and just naively trying to handle the special case up front ended up breaking a bunch of other stuff. It should be fixed though, unless we do decide to insist on fully connected trees (which I think would probably be a mistake).

> And: ascii trees? Just vim... =)

I bow down to your superior vim skills!

ashander:
@petrelharp we should make a gif of you drawing some ascii trees in vim

petrelharp:
Heh, could do.

@ashander got me to say what I actually do: make a canvas of spaces, then
use 'r' a lot, basically.

PR #133, opened 2017-01-30
Tables update

jeromekelleher:


jeromekelleher:
I think this PR is nearly ready to merge, so it would be great to get some feedback @petrelharp, @ashander and @hyanwong. The changes here are all about making it easier and more efficient to load data into the tree sequence from external sources. A lot of low-level details have changed, but essentially there are two ideas here:

1. Rather than having node and sample information munged up in the coalescent records, we now have two concepts: ``nodes`` and ``edgesets``. More on this below.

2. To facilitate efficient movement of data between different components, we now have a bunch of ``XTable`` classes: NodeTable, EdgesetTable, MutationTable and MigrationTable. These are fairly low-level, and the majority of users should never see them.

### Nodes and edgesets

These are probably best explained with an example:
```python
        nodes = """\
        id  is_sample   time
        0   1           0
        1   1           0
        2   1           0
        3   0           0       # Non sample leaf
        4   0           0.4
        5   0           0.5
        6   0           0.7
        7   0           1.0
        """
        edgesets = """\
        0.0     0.2     4       2,3
        0.2     0.8     4       0,2
        0.8     1.0     4       2,3
        0.0     1.0     5       1,4
        0.8     1.0     6       0,5
        0.0     0.2     6       5
        0.0     0.2     7       0,6
        """
        ts = msprime.load_str(nodes, edgesets)
```
This is one of Peter's examples from tests/test_topology.py. Here we can see that the edgesets are just the old coalescent records with the population and time information removed. The nodes table then defines all the relevant information about the nodes in the tree sequence. When we define them, we can have a column for every attibute we are interested in. I've included the ``id`` column here just to make it a bit easier to read, but this isn't necessary. The ``is_sample`` column defines whether a particular node is a sample or not (note that this allows us to have samples that are not 0..n - 1; it hasn't been implemented yet though). The idea here would be to have a flexible parser which would allow the user specify the columns they are interested in, leaving out the ones they don't care about.

Another possible way to express this information is to make lists of records (also an example from the same file):
``` python     
        nodes = [msprime.Node(is_sample=True) for _ in range(3)]
        nodes += [
            msprime.Node(time=5.0),
            msprime.Node(time=4.0),
            msprime.Node(time=4.0),
            msprime.Node(time=3.0),
            msprime.Node(time=3.0),
            msprime.Node(time=2.0),
            msprime.Node(time=1.0),
            msprime.Node(time=1.0)]
        edgesets = [
            msprime.Edgeset(left=0.5, right=1.0, parent=10, children=(1,)),
            msprime.Edgeset(left=0.0, right=0.4, parent=10, children=(2,)),
            msprime.Edgeset(left=0.6, right=1.0, parent=9, children=(0,)),
            msprime.Edgeset(left=0.0, right=0.5, parent=9, children=(1,)),
            msprime.Edgeset(left=0.8, right=1.0, parent=8, children=(10,)),
            msprime.Edgeset(left=0.2, right=0.8, parent=8, children=(9, 10)),
            msprime.Edgeset(left=0.0, right=0.2, parent=8, children=(9,)),
            msprime.Edgeset(left=0.7, right=1.0, parent=7, children=(8,)),
            msprime.Edgeset(left=0.0, right=0.2, parent=7, children=(10,)),
            msprime.Edgeset(left=0.8, right=1.0, parent=6, children=(9,)),
            msprime.Edgeset(left=0.0, right=0.7, parent=6, children=(8,)),
            msprime.Edgeset(left=0.4, right=1.0, parent=5, children=(2, 7)),
            msprime.Edgeset(left=0.1, right=0.4, parent=5, children=(7,)),
            msprime.Edgeset(left=0.6, right=0.9, parent=4, children=(6,)),
            msprime.Edgeset(left=0.0, right=0.6, parent=4, children=(0, 6)),
            msprime.Edgeset(left=0.9, right=1.0, parent=3, children=(4, 5, 6)),
            msprime.Edgeset(left=0.1, right=0.9, parent=3, children=(4, 5)),
            msprime.Edgeset(left=0.0, right=0.1, parent=3, children=(4, 5, 7)),
        ]
        ts = msprime.load_records(nodes, edgesets)
```

For applications where we just want to build the tables and don't want to go though the intermediate record lists, we can just instantiate them directly:

```python
     nodes = msprime.NodeTable()
     nodes.add_row(flags=msprime.NODE_IS_SAMPLE, time=0.0, population=2)
     # etc
     edgesets = msprime.EdgesetTable()
     edgesets.add_row(left=0, right=1, parent=4, children=(0, 1, 3))
     # etc
     ts = msprime.load_tables(nodes=nodes, edgesets=edgesets)
```

### Decoupling components

Part of the motivation for these tables is to make the various components of the simulator and the tree sequence code less tightly coupled. For example, to generate mutations on an existing tree sequence, we can now do something like:

```python
            nodes = msprime.NodeTable()
            edgesets = msprime.EdgesetTable()
            mutations = msprime.MutationTable()
            ts.dump_tables(nodes=nodes, edgesets=edgesets)
            rng = msprime.RandomGenerator(seed),
            mutgen = msprime.MutationGenerator(rng, mutation_rate)
            mutgen.generate(nodes, edgesets, mutations)
             tsp = msprime.load_tables(nodes=nodes, edgesets=edgesets, mutations=mutations)
```

This interface is a bit preliminary, but the basic idea is there anyway.

### Next steps

I would like to merge this much soon and start working with the new APIs. It would be great to get any feedback you might have on the direction things are going in. This is a big change, so there may be some bumps along the road after merging...

petrelharp:
This makes sense.  And, it looks great!  I like removing the redundancy of time/locations from the coalescence records.  A couple of questions:

- I haven't yet tracked down: what is going on in your last example?  Is it: `ts` is an existing tree sequence; then you dump out its information to `nodes` and `edgesets`; then generate `mutations` on these; and load these into a new tree sequence, `tsp`?  This is because tree sequences are immutable?

- You've got the two interfaces: lists of things and tables.  Are you trying to decide which to do, or proposing that both interfaces exist?

- I don't think that samples have to be at time=0, right?  (your examples seemed to imply that)

hyanwong:
As with @petrelharp's point (1), I'm a little unsure why you need to pass in blank msprime.NodeTable(),  msprime.EdgesetTable() etc to ts.dump_tables. Wouldn't it make sense for this function to return these in a tuple, rather than pass in blank versions? Or have I misunderstood?

jeromekelleher:
Thanks for the feedback @petrelharp and @hyanwong, very helpful. Answers to your questions below (in topic-ish order):

> I haven't yet tracked down: what is going on in your last example? Is it: ts is an existing tree sequence; then you dump out its information to nodes and edgesets; then generate mutations on these; and load these into a new tree sequence, tsp? This is because tree sequences are immutable?

Yes, this is it exactly. Because a tree sequence is immutable, we dump out the data into tables to feed a mutation generator, and we then create a new tree sequence. We could make a short-hand method in the tree sequence for this which would return a new tree sequence.

> As with @petrelharp's point (1), I'm a little unsure why you need to pass in blank msprime.NodeTable(), msprime.EdgesetTable() etc to ts.dump_tables. Wouldn't it make sense for this function to return these in a tuple, rather than pass in blank versions? Or have I misunderstood?

This is a good point... I guess I'm imaging ``load_tables`` and ``dump_tables`` as a very low-level API that ordinary users will never see. The reason for passing them in as parameters is so that the same memory can be reused if you are doing something many times. The main reason for not returning the tables is API stability: if we return a tuple of tables (nodes, edgesets, ...) then we have a fixed number of tables to return and it's hard to update the API over time. (We could return a named tuple all right though, that would alleviate this issue.) We can also choose what we want to export using this interface. For example here, I don't want the mutations or migrations so I've not passed a MutationTable or MigrationTable.

It is a very good point all right though. Passing in mutatable parameters doesn't feel very Pythonic... I have been writing a lot of C recently.

> You've got the two interfaces: lists of things and tables. Are you trying to decide which to do, or proposing that both interfaces exist?

A bit of both really. I can see situations where either interface would be useful, so maybe it's best to have both? Certainly the tables interface is for large-scale stuff, but for putting together test cases and stuff this could end up being cumbersome. What do you think?

> I don't think that samples have to be at time=0, right? (your examples seemed to imply that)

Absolutely. Samples can be at any time.

petrelharp:
- Mutable parameters: I'm ok with that. Python does it both ways, no? Especially, as you say, this is low-level.  Also: you could have defaults e.g. `nodes=None` and then `if nodes is None: nodes=msprime.NodeTable()`.

- Tables versus lists versus both: I think sticking with one interface is sufficient and less cumbersome - maybe, tables? - and dealing with lists of things can happen by a function that turns these into a table (on the fly unless we need it a lot).

ashander:
Overall the decoupling looks great and makes sense.

* Mutable parameters for `dump_tables` and `load_tables`: here, the most user friendly way would be a combination of @petrelharp 's idea of defaults and returning a named tuple (instead of relying on the user to always set the parameters). But if this isn't compatible with using the same underlying memory, the current proposal seems fine for a low-level API

* for table vs lists: another way (along the lines of Peter's) is 1) table api, 2) list api, 3) core routine used by both. This would be similar to the old lower level `tree_sequence_load_records` (IIRC), which was used by the tables/ `load_records` routine but could also deal directly with a list of `CoalesenceRecords`.

jeromekelleher:
Thanks for the feedback everyone, this has been very helpful. On the specific points above:

- For dump_tables, I've left the API the way it is for the moment, but I think that returning a namedtuple of tables and using default values as suggested is the right way to go. This should cover all situations.

- I've removed ``load_records()`` following @petrelharp's arguments. I think the text format is more convenient for hand-building test cases, and that's the only real place I can see ``load_records`` being used. 

In the interest of actually using this code and getting some experience with working with it, I'm going to merge this now. The only major gotcha I think is that there is now a C-dependency on numpy --- let me know if you get hit by this. 

There is a bunch of stuff left to be done to push this change all the way through. I'll open some issues on these so we can track them and tag them appropriately so we can track them. Any help on knocking off some of these issues would be much appreciated!

@petrelharp and @ashander, it would be great if you could try this out on your code. Using the tables and the ``add_row`` methods with keyword arguments should be easy enough (sorry about the lack of documentation). Let me know if/when you hit any issues and I'll sort them out ASAP.

PR #153, opened 2017-02-28
General mutations

jeromekelleher:
General mutations via a "mutation type" table, which records the ancestral and derived state for each mutation type (i.e., A -> T, G -> C, etc). Each mutation then has a ``type`` field, which indexes into this table. Initially this only supports single characters, but will generalise well to short indels.

Hat-tip to @petrelharp for the idea.

(This is a work in progress, should be lots of updates here before merging)

## Update

As I was working through this I realised that the original plan won't work for back mutations, which I need. So, I've had to change the design quite a bit. We now have two concepts, *sites* and *mutations*. A site is a particular location on a genome, with a position and an ancestral state (telling us the state at the root of the tree). Mutations happen at a particular site over a particular node and have a derived state.

This will require quite a few changes API wise, but I'm hoping that it won't break much (or ideally any) existing code.

Note that this will change the APIs currently in master, as we're reverting from a mutation having a list of nodes back to a mutation having a single node.


jeromekelleher:
This works well enough to merge now. All existing code should continue to work, as they assume binary mutations. Nonbinary mutations, back mutations etc support is patchy, but should improve as this is used and updated.

PR #165, opened 2017-03-22
Xi-dirac ARG

shajoezhu:
This is an attempt ... most likely it won't work


lib/msprime.c
```diff
@@ -1620,6 +1622,7 @@ msp_recombination_event(msp_t *self)
     self->num_re_events++;
     /* We can't use the GSL integer generator here as the range is too large */
     l = 1 + (int64_t) (gsl_rng_uniform(self->rng) * (double) num_links);
+    printf( "l = %d, num_links = %d\n", (int)l, (int)num_links);
```
jeromekelleher: This is breaking the CI builds as it's outputting too much stuff.

lib/msprime.c
```diff
@@ -449,6 +450,7 @@ msp_set_simulation_model_dirac(msp_t *self, double psi)
         goto out;
     }
     self->model.params.dirac_coalescent.psi = psi;
+    self->model.params.dirac_coalescent.const_c = 1.0; // hard coded const_c value
```
jeromekelleher: What's this for? If it's not a parameter it shouldn't be in the model params struct. If it is a parameter we should just call it `c` and input it the same way as the others.

lib/msprime.c
```diff
@@ -2116,6 +2119,35 @@ msp_common_ancestor_event(msp_t *self, population_id_t population_id)
...
+    assert(b > 0);
+    assert(psi > 0);
+    assert(psi < 1);
+    assert(c >= 0);
+
+    unsigned int l, n_block;
+    n_block = GSL_MIN(b, 4);
+    double ret = 0;
+
+    for (l = 0; l < (unsigned)n_block; l++){
```
jeromekelleher: What's the cast for here? n_block and l are both already unsigned ints.

lib/msprime.c
```diff
@@ -2116,6 +2119,35 @@ msp_common_ancestor_event(msp_t *self, population_id_t population_id)
...
+msp_compute_lambda_Xi_dirac(unsigned int b, double psi, double c)
+{
+    assert(b > 0);
+    assert(psi > 0);
+    assert(psi < 1);
+    assert(c >= 0);
+
+    unsigned int l, n_block;
+    n_block = GSL_MIN(b, 4);
+    double ret = 0;
```
jeromekelleher: Keeping code and declarations separate is a good idea for portability (and clarity IMO). Here I would have
```c
unsigned int l, n_block;
double ret = 0;

assert(b > 0);
￼assert(psi > 0);
￼assert(psi < 1);
￼assert(c >= 0);

n_block = GSL_MIN(b, 4);

File: lib/msprime.c [hunk 5]
```diff
@@ -2116,6 +2119,35 @@ msp_common_ancestor_event(msp_t *self, population_id_t population_id)
 }
 
 
+/* This calculates the rate given by Eq (2) in the notes
+ */
+static double
+msp_compute_lambda_Xi_dirac(unsigned int b, double psi, double c)
```
jeromekelleher: This function isn't really a "method" of the msp "class", since we don't pass a reference to the msp object in with it. All functions that have an ``msp_`` prefix should  be methods of the ``msp_`` class. There are two ways to look at this: (1) either we make it a method, and don't pass psi and c (which we should be able to get from the msp object state); or (2) make this an independent function, and remove the ``msp_`` prefix.

lib/msprime.c
```diff
@@ -2392,6 +2426,31 @@ msp_get_common_ancestor_waiting_time(msp_t *self, uint32_t population_id)
...
+
+    if (self->model.type == MSP_MODEL_DIRAC){
+        mm_rate = msp_compute_lambda_Xi_dirac(n,
+                      self->model.params.dirac_coalescent.psi,
+                      self->model.params.dirac_coalescent.const_c);
+    }
+
+    if (mm_rate > 0.0) {
+        u = gsl_ran_exponential(self->rng, 1.0/mm_rate)/2.0;
+        ret = pop->initial_size * u;
```
jeromekelleher: Let's stay away from population structure for now. Assume 1 population with a fixed size of 1.

lib/msprime.c
```diff
@@ -2584,9 +2644,13 @@ msp_run_multiple_mergers_coalescent(msp_t *self, double max_time, unsigned long
             re_t_wait = gsl_ran_exponential(self->rng, 1.0 / lambda);
         }
         /* Common ancestors */
-        n = (double) avl_count(&self->populations[0].ancestors);
-        lambda = n * (n - 1.0);
-        ca_t_wait = gsl_ran_exponential(self->rng, 1.0 / lambda);
+        ca_t_wait = DBL_MAX;
+        for (j = 0; j < self->num_populations; j++) {
```
jeromekelleher: Here also, let's get the core functionality working for a single population before we start adding population structure.

lib/msprime.c
shajoezhu: my bad, this should be removed.

lib/msprime.c
shajoezhu: I am not sure about this constant c in equation (2), was meaning to come and discuss this with you

lib/msprime.c
shajoezhu: Yes, this is just a flag for where things are going wrong when I run `./main -v simulate dirac/dev.cfg -r 1`, It looks like recombination hasn't worked on this yet,

lib/msprime.c
shajoezhu: I see. I would prefer option (1) then. Since I will be able to parse psi through model.params.dirac_coalescent.psi directly

lib/msprime.c
shajoezhu: sounds good

lib/msprime.c
jeromekelleher: Well you need to comment it out it before submitting CI builds --- it's spewing out many megabytes of text and causing the builds to fail before they even finish running.

jeromekelleher:
I had a quick look at this @shajoezhu, and it looks like the problem is in ``msp_compute_lambda_Xi_dirac``. This is returning negative rates quite a lot, which leads to infinite CA waiting times. The eventual segfault is caused by calling ``msp_recombination_event`` even though there is an infinite re-event waiting time (there is also an infinite ca-event waiting time).

lib/msprime.c
```diff
@@ -2116,6 +2119,37 @@ msp_common_ancestor_event(msp_t *self, population_id_t population_id)
...
+        ret += gsl_sf_exp(gsl_sf_lnchoose(b, l) +
+                           gsl_sf_log(4.0) +
+                           (double)l * gsl_sf_log(psi/4.0) +
+                           (double)(b-l) * gsl_sf_log(1-psi));
+    }
+
+
+    ret = 1 - ret;
+    ret *= c*4.0/pow(psi,2.0);
+    ret += b * (b-1) / 2;
```
jeromekelleher: We need to be careful to cast the intermediate values to a double here, as b * (b - 1) will overflow. It might be safer to do something like
```c
msp_compute_lambda_Xi_dirac(msp_t *self, unsigned int num_ancestors)
{
     double b = num_ancestors;
```

You can safely remove the two casts in the loop also, since (b - l) will be promoted to double.

shajoezhu: i see! nice! I like this.

shajoezhu:
I was not sure about the (4)_\ell term either... Was meant to come and find you today and discuss this with you. I will come and find you tomorrow. I think it's going well. Moving towards the right direction

jeromekelleher:
Looks like good progress @shajoezhu. I think we need to fix up the beta coalescent code path so that it does something reasonable (i.e., error out earlier or complete successfully for some simpler process). This is causing problems at the moment and triggering asserts in the tests.

Other than that, I think all we need to do is to implement the four coalescence 'pots'. This should be easy: just create a local array Q[4], and assign participating lineages into these four pots with the right probability. We do then have to worry about the non events, where nothing coalesces as Bjarki pointed out. I would prioritise getting this done before propagating changes back to the Python interface (just set c to a constant in _msprimemodule.c to ensure the tests pass).

jeromekelleher:
This is looking close @shajoezhu, well done!

I made a mistake with the CircleCI config, as it wasn't building for PRs, but I should have fixed it now. Can you push an update to this branch please (maybe squash down a few of the commits?), so that it will build. I'd like to see the coverage report before we go any further, but I think we should be nearly ready to merge this.

jeromekelleher:
I can't figure out why we're not getting a codecov report on this @shajoezhu... Would you mind closing this PR and opening another one from the same branch?

Sorry for the hassle, but the codecov report is invaluable when trying to figure out what tests I need to write.

PR #180, opened 2017-04-09
Compute branch statistics in windows along the genome

petrelharp:
I've only tested the windowing with pairwise diversity, but I've not touched anything except the bookkeeping of average-by-sequence-length.

As it is, it is required to compute the statistics in a set of windows that entirely cover the chromosome; for applications we may want to relax this - but, rather than re-implementing this it may be easier (and useful for other things) to have a subset(ts,begin,end) function that returns the tree sequence restricted to a portion of the genome.  But if we might want to change this function to do different windows more generally, we should perhaps pass in `start=` and `end=` rather than the list of breakpoints of length 1 + # of windows (called `windows=`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 ``msprime`` is a Python API and command line interface for simulating the 
 coalescent with recombination. Please see the 
-[documentation](https://msprime.readthedocs.org/en/latest/>)
+[documentation](https://msprime.readthedocs.org/en/latest/)
 for further information and 
 [installation instructions](https://msprime.readthedocs.org/en/latest/installation.html).
 
@@ -15,5 +15,5 @@ to get started.
 
 If you use ``msprime`` in your work, please cite the [PLOS Computational
 Biology](http://dx.doi.org/10.1371/journal.pcbi.1004842) paper.
-See [here](https://msprime.readthedocs.org/en/latest/CITATION.html>) for
+See [here](https://msprime.readthedocs.org/en/latest/CITATION.html) for
 full citation details.
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2179,11 +2179,11 @@ class TreeSequence(object):
         and averages this across the tree sequence, weighted by genomic length.
         '''
         out = self.branch_stats_vector(leaf_sets, lambda x: [weight_fun(x)])
-        if len(out) > 1:
+        if len(out) != 1 or len(out[0]) != 1:
             raise ValueError("Expecting output of length 1.")
-        return out[0]
+        return out[0][0]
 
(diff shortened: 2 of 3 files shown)


petrelharp:
Whoops - this isn't right if a tree extends for more than one window.

petrelharp:
Fixed.

jeromekelleher:
This looks good to me @petrelharp, I'm happy to merge this whenever you're ready. First though, I think we need to rebase your branch off upstream/master and force push. Some of the CI checks haven't run, and I think this might be because the base for this is a bit old. Can you 
```
git fetch upstream
git rebase -i upstream/master
```
please?

Also, thanks for correcting the broken links!

petrelharp:
It is done.

jeromekelleher:
Ah, I understand now why the CI was running --- I didn't have CircleCI configured to run on PRs. Would you mind doing another force push? This should trigger the build. (I don't want to merge until we see the coverage report, as it's easier to fill in gaps in the tests now rather than later).

Sorry for the confusion. There's a lot of new CI stuff just come in recently and there are a few kinks to work out. There's a new section in the developer docs about it though, any feedback would be much appreciated.

petrelharp:
done.

jeromekelleher:
OK, cool, thanks @petrelharp. I'll add a few tests to fill out the coverage (hopefully this evening) and merge when it's done.

jeromekelleher:
I've made some minor updates to the interface @petrelharp and a few tests to cover the error conditions we expect to occur. Can you take a quick look to make sure these make sense please?

I changed a couple of ValueErrors here to asserts, as they were really sanity checks on the code rather than responding to legitimate error conditions.

petrelharp:
Looks good.  Thanks for those tests.

As for the interface - the function returns a matrix, effectively - a vector of statistics for each window - and so there's various convenience functions we could provide, that would effectively simplify the matrix to (a) a single vector if it has one column, (b) a single vector if it has one row, and (c) a single value if it has one row and column.  The thing that seems most convenient is being able to pass in a function that returns a single value rather than a vector, which is what we have.  But maybe this is unnecessary. Your opinion?

I guess that if the intention of this is just for us to use it to implement statistics, then perhaps this is unnecessary complications.

jeromekelleher:
I guess my take is that we don't want to get too bogged down in the interface details yet until we've used it for a while. Honestly, I haven't had a chance to think about the algorithm and interface properly, so I don't have an informed opinion about what it should look like.

The changes I made here we just very simple ones to make sure that we were covering all the possibilities with tests. If you're happy with these I'll merge the PR.

petrelharp:
Sounds good - go ahead and merge.

I am setting to work actually using it right now. =)

Issue #64, opened 2016-10-23
Make API more Pythonic by removing lots of 'get_'s


jeromekelleher:
Several aspects of the API are a little non-Pythonic. Mainly, this involves using methods like `obj.get_x()` rather than `obj.x`. 

For example, we should change:
- SparseTree.get_root() -> SparseTree.root
- SparseTree.get_index() -> SparseTree.index
- SparseTree.get_interval() -> SparseTree.interval
- TreeSequence.get_sample_size() -> TreeSequence.sample_size
- TreeSequence.get_num_mutations() -> TreeSequence.num_mutations

Also, we should change things like
- SparseTree.get_branch_length(u) -> SparseTree.branch_length(u)
- SparseTree.get_mrca(u, v) -> SparseTree.mrca(u, v)

Existing methods should all be marked as deprecated, removed from the documentation, and (eventually) removed entirely.

ashander:
I could help out as I familiarize myself with the API. I opened a PR with a list of candidate functions

ashander:
A question:  should the above be eg `SparseTree.root()`? (you don't intend to change them from instance methods, right?)

jeromekelleher:
This is great @ashander, thanks! I think we do want to make them properties; I'll go into more detail on how your PR.

shajoezhu:
sometimes I think github should have a 'like' buttom for things like this

jeromekelleher:
Finally finished this up in #384.

PR #190, opened 2017-04-16
statistics using branch length stats functions

petrelharp:
I've added a function that does divergence (=TMRCA) in windows.  But we could leave this pull request open to implement others (which I'll be doing sooner or later).  Also, to wait for the nomenclature to settle down.

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2312,7 +2312,7 @@ class TreeSequence(object):
         # move this division outside of f(x) so it only has to happen once
         # corrects the diagonal for self comparisons
         for w in range(len(windows)-1):
-            out[w] /= float(n[0] * n[1] * n[2])
+            out[w][0] /= float(n[0] * n[1] * n[2])
 
         return out
 
@@ -2338,7 +2338,7 @@ class TreeSequence(object):
         # move this division outside of f(x) so it only has to happen once
         # corrects the diagonal for self comparisons
         for w in range(len(windows)-1):
-            out[w] /= float(n[0] * n[1] * n[2] * n[3])
+            out[w][0] /= float(n[0] * n[1] * n[2] * n[3])
 
         return out
 
@@ -2363,15 +2363,14 @@ class TreeSequence(object):
 
         def f(x):
             return [float(x[0] * (x[0] - 1) * (n[1] - x[1]) * (n[2] - x[2])
-                          + (n[0] - x[0]) * (n[0] - x[0] + 1) * x[1] * x[2]
+                          + (n[0] - x[0]) * (n[0] - x[0] - 1) * x[1] * x[2]
                           - x[0] * (n[0] - x[0]) * (n[1] - x[1]) * x[2]
                           - (n[0] - x[0]) * x[0] * x[1] * (n[2] - x[2]))]
 
         out = self.branch_stats_vector(leaf_sets, weight_fun=f, windows=windows)
         # move this division outside of f(x) so it only has to happen once
-        # corrects the diagonal for self comparisons
         for w in range(len(windows)-1):
-            out[w] /= float(n[0] * (n[0]-1) * n[1] * n[2])
+            out[w][0] /= float(n[0] * (n[0]-1) * n[1] * n[2])
 
         return out
 
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2443,8 +2443,8 @@ class TreeSequence(object):
                 raise ValueError(
                     "elements of leaf_sets cannot contain repeated elements.")
             for u in U:
-                assert(type(u) is int)
-                assert(self.node(u).is_sample())
+                if not self.node(u).is_sample():
+                    raise ValueError("Not all elements of leaf_sets are samples.")
         num_windows = len(windows) - 1
         if windows[0] != 0.0:
             raise ValueError(
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2199,6 +2199,38 @@ class TreeSequence(object):
             leaves = list(samples)
         return self._ll_tree_sequence.get_pairwise_diversity(leaves)
 
+    def get_mean_tmrca(self, leaf_sets, windows):
+        """
+        Finds the mean time to most recent common ancestor between pairs of samples
+        from each group of leaves and in each window. Returns the upper triangle
+        (including the diagonal) in row-major order, so if the output is `x`, then:
+
+        j=0
+        for m in range(len(leaf_sets)):
+            for n in range(m,len(leaf_sets)):
+                trmca[m,n] = tmrca[n,m] = x[i][j]
+                j += 1
+
+        will fill out the matrix of mean TMRCAs in the `i`th window between (and
+        within) each group of leaves in `leaf_sets` in the matrix `tmrca`.  If
+        an element of `leaf_sets` has only one element, the corresponding
+        diagonal will be 0.
+
+        :param list leaf_sets: A list of sets of IDs of leaves.
+        :param iterable windows: The breakpoints of the windows (including start
+            and end, so has one more entry than number of windows).
+        :return: A list of the upper triangle of mean TMRCA values in row-major
+            order, including the diagonal.
+        """
+        nleaves = len(leaf_sets)
+        n = [len(x) for x in leaf_sets]
+
+        def f(x):
+            return [float(x[i]*(n[j]-x[j]) + (n[i]-x[i])*x[j])/float(n[i]*n[j])
+                    for i in range(nleaves) for j in range(i, nleaves)]
+
+        return self.branch_stats_vector(leaf_sets, weight_fun=f, windows=windows)
+
     def branch_stats(self, leaf_sets, weight_fun):
         '''
         Here leaf_sets is a list of lists of leaves, and weight_fun is a function
@@ -2235,7 +2267,7 @@ class TreeSequence(object):
         and averages this across the tree sequence, weighted by genomic length.
 
         It does this separately for each window [windows[i], windows[i+1])
-        and returns the values in a vector.
+        and returns the values in a list.
         '''
         if windows is None:
             windows = (0, self.sequence_length)
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2202,14 +2202,16 @@ class TreeSequence(object):
     def mean_pairwise_tmrca(self, leaf_sets, windows):
         """
         Finds the mean time to most recent common ancestor between pairs of samples
-        from each group of leaves and in each window. Returns the upper triangle
-        (including the diagonal) in row-major order, so if the output is `x`, then:
+        as described in mean_pairwise_tmrca_matrix (which uses this function).
+        Returns the upper triangle (including the diagonal) in row-major order,
+        so if the output is `x`, then:
 
-        >>> j=0
-        >>> for m in range(len(leaf_sets)):
-        >>>     for n in range(m,len(leaf_sets)):
-        >>>         trmca[m,n] = tmrca[n,m] = x[i][j]
-        >>>         j += 1
+        >>> k=0
+        >>> for w in range(len(windows)-1):
+        >>>     for i in range(len(leaf_sets)):
+        >>>         for j in range(i,len(leaf_sets)):
+        >>>             trmca[i,j] = tmrca[j,i] = x[w][k]
+        >>>             k += 1
 
         will fill out the matrix of mean TMRCAs in the `i`th window between (and
         within) each group of leaves in `leaf_sets` in the matrix `tmrca`.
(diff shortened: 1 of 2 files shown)


msprime/trees.py
```diff
@@ -2169,6 +2169,38 @@ def get_pairwise_diversity(self, samples=None):
             leaves = list(samples)
         return self._ll_tree_sequence.get_pairwise_diversity(leaves)
 
+    def get_mean_tmrca(self, leaf_sets, windows):
+        """
+        Finds the mean time to most recent common ancestor between pairs of samples
+        from each group of leaves and in each window. Returns the upper triangle
+        (including the diagonal) in row-major order, so if the output is `x`, then:
+
```
jeromekelleher: I think this is breaking the docs build, which is breaking the CI.

Can you try running ``make`` in the docs directory and fixing the issue please?

petrelharp:
sorry about that!  Fixed it.  I'll fuss around with the interface a bit
more before merging.

jeromekelleher:
So, contrary to my earlier prediction, the codebase actually has moved on quite a lot in the last couple of weeks. I think we should probably bring this up-to-date and merge it (any remaining stuff can be done in follow-ups). It looks like it might be a bit messy though, unfortunately...

ashander:
Not as bad as it looked, as this branched off an earlier point of `generalise-samples`. Now it's rebased to current master so should work to merge. 

But @petrelharp mentioned wanting to tweak the interface a bit before the merge. Still the case?

petrelharp:
I'd mostly like some input on what to call things, and what to write
individual functions for.

The exemplar is the `get_mean_tmrca` function, which takes a list of sets
of leaves and a list of breakpoints between windows, and returns a list of
(upper triangles of) matrices of average
total-time-to-most-recent-common-ancestor, pairwise between each of the
sets of leaves.

On the "what to call things" front:
 - Is "mean_tmrca" a good name for this?
     (a) Perhaps "mean" is unnecessary. But it is certainly different than
the `get_tmrca` function (but that's a function of trees, not tree
sequences).
     (b) It is actually more like 2*TMRCA, as it counts the total distance
from both samples up to the root.  An alternative would be to call it
`get_mean_tree_distance`, say.
- An alternative would be to call this `get_mean_divergence(...,
branch_lengths=True)` and use the same function as an interface to the
mutation-counting version of what is morally the same thing.  Otherwise,
we'll be implementing a separate function, `get_mean_divergence`, which
uses the same underlying code but measuring branch lengths in mutations
instead of time.

The last thing touches on 'what to write individual functions for'.
Options for each function are:
  - windowed?  (returning a list or a single value/single matrix of values)
  - measuring branch lengths in time or in mutations?
 ... so for any given function we could potentially have four versions.  My
inclination is to:
  - not implement non-windowed functions; so let the user take care of
simplifying back down to what they want if they're computing for the whole
genome (i.e., no windows, although the default window could be [0.0,
ts.sequence_length]
  - Implement time/mutation versions as separately named functions - this
has the downside of deemphasizing the relationship between a particular
pair but they are not totally equivalent (e.g., different units;
complications due to multiple mutations) and most people don't think about
them exchangeably.


My inclination is to change `get_mean_tmrca` to `get_mean_tree_distance`, I
guess - although it's too long.

What do you think?  When we settle this I can add a note on how to move
forward to the docs on these methods.

jeromekelleher:
I haven't thought this through properly @petrelharp, but the idea that I've had in the back of my head for this stuff is to somehow follow the pattern of the ``LdCalculator`` object in ``msprime/stats.py``. So, rather than having a lot of different methods on the tree sequence to calculate statistics, we'd have some classes that have an instance variable as a parameter and that have specific methods to compute the stats that we're interested in. This has a few advantages:

1) It avoids the namespace in the TreeSequence class becoming overcrowded with lots of different methods for computing stats (which, hopefully, there will be many).
2) It allows us to allocate a low-level C object to do the computations and to keep state between invocations. This is usually critical for large-scale efficiency. Also it allows us to split calculations across threads (see the [tutorial](https://msprime.readthedocs.io/en/stable/tutorial.html#working-with-threads))
3) It allows us to isolate the code for calculating statistics from the main tree sequence stuff.
4) If related functions are grouped together into a class then common parameters can be abstracted into the class constructor (e.g., leaf_sets and maybe windows).

I don't have a concrete proposal for what you have here now though. My plan was to add things in as we needed them, and to stand back and review the stats functions before we get to the next release. I was hoping then that some patterns would emerge and we could refactor them into a stats class as described above. I'm happy to do this refactoring step now if you like though. However, the interface will end up being driven by the low-level C objects that we end up with to some degree, so we can't really guarantee that things won't need to be rejiggered again when that step is taken on.

Some concrete points:
- I'd like to follow the pattern in LdCalculator, where we have the statistic and return type in the name (i.e., ``r2(a, b)`` for the r2 between a specific pair, ``r2_matrix`` for all pairwise r2 values.
- ``mean_tmrca_matrix`` seems good to me (but divide the current value by 2). Adding the new concept of tree distance seems confusing when it's just 2 * tmrca. This should return a numpy array like ``r2_matrix``.
- I'm leaning towards having a separate method for returning the same thing in terms of mutational distance.  I like ``divergence`` for this. Generally methods should describe what they return I think, and if you have different things being returned depending on the parameters then it's usually a sign that you need a different method.

Anyway, that's my 2c at this point. I'm not sure how helpful this has been!

jeromekelleher:
Hah, yes, just realised that divergence_time != 2 * tmrca when samples are not necessarily at time 0. Perhaps ``total_branch_length``? ``tree_distance`` may be better, but this could be confused with tree distance metrics (i.e., the distance between two different trees under Robinson-Foulds or whatever).

petrelharp:
Ah, good - I hoped you'd have the big picture.  All that sounds just right.
So, continue cluttering the tree sequence namespace for now, sort it out
later, eh?

The other reason for not going with tmrca is that it sounds like the mrca
of the entire leafset rather than just pairwise.  Hm, but good point about
tree distance. 'branch length' doesn't say *which* branches. hm: if we
divide by 2 it is still mean tmrca, where 'mean' also means picking a
random one of the 2 branches.  so maybe i am back to 'mean_pairwise_tmrca'
.  names may change in refactoring.

ok: will change name; add '_matrix' that returns a list of np.arrays, then
can merge.

jeromekelleher:
> So, continue cluttering the tree sequence namespace for now, sort it out
> later, eh?
I think so yeah. There's something to be said for avoiding premature refactoring!

> ok: will change name; add '_matrix' that returns a list of np.arrays, then can merge.
Sounds good to me.

ashander:
All looks /sounds good to me. I was going to advocate for a naming/namespace hierarchy that connects the 'windowed' operations together and separates from existing. But it makes sense to have that as a longer term plan and get this in soon.

One thing though, in naming these should we avoid from the outset the use of `get_` at python level? As flagged in https://github.com/jeromekelleher/msprime/issues/64 this is sort of non-pythonic. If that API change is a further away goal maybe using `get_` for additions now feels right, but I wanted raise the issue

jeromekelleher:
> One thing though, in naming these should we avoid from the outset the use of get_ at python level? As flagged in #64 this is sort of non-pythonic. If that API change is a further away goal maybe using get_ for additions now feels right, but I wanted raise the issue.

I agree, let's stop using ``get_`` prefixes for new API additions. For the next release I think we should have overhauled the whole API and updated the docs to only have the non-``get_`` methods/attributes (but keeping the existing getters as deprecated aliases, probably indefinitely).

ashander:
OK. I think the latest commit implements the agreement back in comment https://github.com/jeromekelleher/msprime/pull/190#issuecomment-298959042

But: I'm not sure what the final decision was on dividing the calculation by 2 so haven't implemented that yet.

jeromekelleher:
Sounds good @ashander, let me know when you would like feedback.

ashander:
Thanks @jeromekelleher it'll be good to get this in the main repo. I'd like some initial feedback on the items below (correspond to the first to checkboxes):

### 1. 

I'm not completely sure from the thread your final decision re the function, now called `mean_pairwise_tmrca`, and the factor of 2. I think it is to add a division by 2 per @petrelharp 2nd to last comment

>The other reason for not going with tmrca is that it sounds like the mrca
of the entire leafset rather than just pairwise.  Hm, but good point about
tree distance. 'branch length' doesn't say *which* branches. hm: if we
divide by 2 it is still mean tmrca, where 'mean' also means picking a
random one of the 2 branches.  so maybe i am back to 'mean_pairwise_tmrca'
.  names may change in refactoring.

### 2. 

Assuming the above naming is good for now, does `.mean_pairwise_tmrca_matrix` work also?

jeromekelleher:
@ashander: OK, cool.

1. I have to admit I've lost the train of thought a bit here. I'm happy to go with whatever you and @petrelharp think here. I'm assuming a later refactoring pass when we're properly nailing down the functionality and APIs for these stats, so I'm not overly concerned that we get it exactly right now.

2. ``.mean_pairwise_tmrca_matrix`` sounds good; let's go with this for now. We might think of a more elegant naming system during refactoring...

So, is it just a matter of fixing the pep8 failures and renaming and we are looking good to merge?

petrelharp:
It should be good to merge now.  We still may be writing _matrix-like functions, but need to work out the right syntax (since we probably don't want to do all three- or four-way combinations).

ashander:
@jeromekelleher is _was_ just a matter of those things, but :tada: @petrelharp added even more f-stats (f3, f4, Y). as he mentions these are not yet set up to do lots of comparisons so there are no corresponding `_pairwise` or `_matrix` methods. 

but the upshot: it's good to merge, and has [increased coverage](https://github.com/jeromekelleher/msprime/pull/190#issuecomment-294600038)

petrelharp:
OK - I actually vectorized Y in the right way (pass in leaf sets then a list of triples of indices) to compute many Ys efficiently.  This can be a template for how to do the others.

tests/test_branch_stats.py
```diff
@@ -72,45 +75,88 @@ def branch_length_diversity_window(ts, X, Y, windows):
...
-            #  / /\
-            # z x  y
-            S += path_length(tr, x, xy_mrca)*tr.length
-    return S/ts.sequence_length
-
-
-def branch_length_f4(ts, A, B, C, D):
+        if tr.interval[1] < begin:
+            continue
+        if tr.interval[0] > end:
```
jeromekelleher: Should this be >=?

tests/test_branch_stats.py
```diff
@@ -72,45 +75,88 @@ def branch_length_diversity_window(ts, X, Y, windows):
...
+        end = ts.sequence_length
+    assert(len(A) > 1)
+    for U in A, B, C:
+        if max([U.count(x) for x in set(U)]) > 1:
+            raise ValueError("A,B,C, and D cannot contain repeated elements.")
+    S = 0
+    for tr in ts.trees():
+        if tr.interval[1] < begin:
+            continue
+        if tr.interval[0] > end:
```
jeromekelleher: Also here.

tests/test_branch_stats.py
```diff
@@ -343,6 +486,16 @@ def test_windowization(self):
         ts = msprime.simulate(10, random_seed=self.random_seed, recombination_rate=100)
         self.check_windowization(ts)
 
+    def test_derived_functions(self):
+        '''
```
jeromekelleher: Can you change this to a '#' comment please? Docstrings for test functions are displayed on the console when you run ``nosetests -v``, which make it difficult to see what test is actually being run. This is the only time when docstrings are bad though!

msprime/trees.py
```diff
@@ -2235,14 +2434,17 @@ def branch_stats_vector(self, leaf_sets, weight_fun, windows=None):
...
+        and returns the values in a list.
         '''
         if windows is None:
             windows = (0, self.sequence_length)
         for U in leaf_sets:
             if len(U) != len(set(U)):
                 raise ValueError(
                     "elements of leaf_sets cannot contain repeated elements.")
+            for u in U:
+                assert(type(u) is int)
```
jeromekelleher: This looks more like an input check than a sanity check to me... The first check is actively bad I think, as it violates Python rules about allowing duck-typing. Is there any particular reason that it must be precisely an int?

The second check should be changed to a ValueError I think. 

My rule of thumb with asserts vs exceptions is, asserts should only be raised when a mistake happens _within_ msprime, not when the _user_ makes a mistake. I.e., asserts are just for sanity checking, not input validation. Bad inputs supplied by the user should always raise an exception.

jeromekelleher:
This looks fantastic, thanks @petrelharp. I've made a few very minor comments above. We're good to merge after that.

tests/test_branch_stats.py
petrelharp: actually, both should be >= if both intervals (begin,end] and (interval[0], interval[1]] are half-open.  but, it won't make a difference since in the edge case `this_length` will be zero (changing anyhow)

petrelharp:
I've added f2 and vectorized version of f3 and f4.  I've not unified tests of these things, which would be nice, but can happen when/if we add more stats of this form.  I've also unified the notation for vectorized versions - the argument to Y is now `indices=` not `triples=`.

Should be good to merge now.  Yay!

jeromekelleher:
Looks good @petrelharp, but there's a test failure on Python 2.7:

https://travis-ci.org/jeromekelleher/msprime/jobs/251537588#L878

(I don't know why it would fail on 2.7 in particular, it may be a randomness issue. Sometimes the RNGs on different versions of Python are different, and this leads to different test instances being run on the different versions. So, this could be the issue, if you're using simulated data for this test)

tests/test_branch_stats.py
```diff
@@ -197,6 +288,12 @@ def assertListAlmostEqual(self, x, y):
         for a, b in zip(x, y):
```
jeromekelleher: I should have spotted this before, but can we add a 
```
self.assertEqual(len(x), len(y))
```
here please? This will currently happily compare [2], [2, 3, 4, 5] as equal.

tests/test_branch_stats.py
```diff
@@ -287,54 +460,99 @@ def f(x):
...
+            A = [A_list[i] for i in index_list]
 
             def f(x):
                 return ((float(x[0])/len(A[0])-float(x[1])/len(A[1]))
                         * (float(x[2])/len(A[2])-float(x[3])/len(A[3])))
 
-            self.assertAlmostEqual(
-                    ts.branch_stats(A, f),
-                    branch_length_f4(ts, A[0], A[1], A[2], A[3]))
+            here_values = [branch_length_f4(ts, A[0], A[1], A[2], A[3],
```
jeromekelleher: The failure in py2 is caused by this list comprehension. The variable k is used within the comprehension, and this over writes the ``k`` from the loop above.

jeromekelleher:
The problem is a scoping issue in py27, it's simple to fix. I tried to push a fix, but I got a permission error.

Definitely good to merge after that!

tests/test_branch_stats.py
ashander: Ah good catch. And nice to know about the variable scoping difference w/r/t list comprehensions

ashander:
Should work :crossed_fingers: thanks @jeromekelleher !

jeromekelleher:
Merging! Thanks @petrelharp and @ashander.

petrelharp:
Glad you caught that!  Thanks, all.

PR #191, opened 2017-04-19
Update low level errors nodes edgesets

### LINKED ISSUE #146: Update low-level error checking and messages
Error messages in the low-level C code still refer to coalescence records. Update these symbolic constants with better names and update the error messages.

ashander:
Updates error codes to refer to nodes/edgesets versus coalescence records

closes #146

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -142,8 +142,10 @@ msp_strerror(int err)
         case MSP_ERR_BAD_EDGESET_NONMATCHING_RIGHT:
             ret = "Bad edgeset in file: right coordinate not matching any left coordinate.";
             break;
-        case MSP_ERR_BAD_COALESCENCE_RECORDS:
-            ret = "Bad coalescence records in file.";
+        case MSP_ERR_BAD_EDGESET:
+            ret = "Bad edgeset in file.";
+        case MSP_ERR_BAD_EDGESET_NO_LEFT_AT_ZERO:
+            ret = "Bad edgeset in file: no left coordinate starts at zero.";
             break;
         case MSP_ERR_ZERO_RECORDS:
             ret = "At least one record must be supplied";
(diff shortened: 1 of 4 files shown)


lib/msprime.c
```diff
@@ -133,17 +133,16 @@ msp_strerror(int err)
         case MSP_ERR_BAD_RECOMBINATION_MAP:
             ret = "Bad recombination map provided.";
             break;
-        case MSP_ERR_BAD_COALESCENCE_RECORDS_SAMPLE_SIZE:
-            ret = "Bad coalescence records in file: Samples must be labeled 0...n-1 and not be internal nodes.";
+        case MSP_ERR_BAD_NODES_SAMPLE_SIZE:
+            ret = "Bad nodes in file: No sampled nodes.";
```
jeromekelleher: Technically we need at least 2 samples I think, so maybe we should change this to ``MSP_ERR_INSUFFICIENT_SAMPLES`` and change the error message? Probably the "Bad nodes in file" bit doesn't add much and can be removed.

ashander: :+1:

ashander:
Thanks. I did that fixed and added a test for one of these. I also tried to write a test for samples < 2 but it didn't work as expected (so commented out)

jeromekelleher:
Your test case was perfect @ashander, there was a bug in the code. I've fixed this up and added another check (makes sure that samples are 0...n - 1, which we're still depending on in a few places).  

Thanks again!

lib/tree_sequence.c
```diff
@@ -719,10 +719,18 @@ tree_sequence_init_nodes(tree_sequence_t *self)
             self->sample_size++;
         }
     }
-    if (self->nodes.num_records > 0 && self->sample_size == 0) {
+    if (self->nodes.num_records > 0 && self->sample_size < 2) {
```
ashander: Ah! Oops no wonder that test didn't work. Thanks

PR #194, opened 2017-04-24
Initial changes for samples != {0 ,..., n - 1}.

jeromekelleher:
This is a work in progress to allow samples that are not 0, ..., n - 1. The initial impression is that this isn't too bad, but does break some important operations (haplotype generation is what I've found so far). 

The simplest thing may be to allow these types of tree sequences, but to disallow operations like haplotype generation on them. To run these operations you would need to run ``simplify`` to get back a tree sequence of the expected form.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -182,7 +182,6 @@ parse_sample_ids(PyObject *py_samples, tree_sequence_t *ts, size_t *num_samples,
     size_t j;
     Py_ssize_t num_samples_local;
     node_id_t *samples_local = NULL;
-    uint32_t n = tree_sequence_get_sample_size(ts);
 
     num_samples_local = PyList_Size(py_samples);
     if (num_samples_local < 2) {
@@ -201,12 +200,12 @@ parse_sample_ids(PyObject *py_samples, tree_sequence_t *ts, size_t *num_samples,
             goto out;
         }
         samples_local[j] = (node_id_t) PyLong_AsLong(item);
-        if (samples_local[j] < 0) {
-            PyErr_SetString(PyExc_ValueError, "sample IDs must be >= 0");
+        if (samples_local[j] < 0 || samples_local[j] >= tree_sequence_get_num_nodes(ts)) {
+            PyErr_SetString(PyExc_ValueError, "node ID out of bounds");
             goto out;
         }
-        if (samples_local[j] >= n) {
-            PyErr_SetString(PyExc_ValueError, "sample ids must be < sample_size");
+        if (! tree_sequence_is_sample(ts, samples_local[j])) {
+            PyErr_SetString(PyExc_ValueError, "Specified node is not a sample");
             goto out;
         }
     }
@@ -3666,8 +3665,42 @@ out:
 }
 
 static PyObject *
-TreeSequence_get_pairwise_diversity(TreeSequence *self, PyObject *args,
-        PyObject *kwds)
+TreeSequence_get_samples(TreeSequence *self)
+{
+    PyObject *ret = NULL;
+    node_id_t *samples;
+    PyObject *py_samples = NULL;
+    PyObject *py_int = NULL;
+    size_t j, n;
+    int err;
+
+    if (TreeSequence_check_tree_sequence(self) != 0) {
+        goto out;
+    }
+    n = tree_sequence_get_sample_size(self->tree_sequence);
+    err = tree_sequence_get_samples(self->tree_sequence, &samples);
+    if (err != 0) {
+        handle_library_error(err);
+    }
+    py_samples = PyList_New(n);
+    if (py_samples == NULL) {
+        goto out;
+    }
+    for (j = 0; j < n; j++) {
+        py_int = Py_BuildValue("i", (int) samples[j]);
+        if (py_int == NULL) {
+            Py_DECREF(py_samples);
+            goto out;
+        }
+        PyList_SET_ITEM(py_samples, j, py_int);
+    }
+    ret = py_samples;
+out:
+    return ret;
+}
+
+static PyObject *
+TreeSequence_get_pairwise_diversity(TreeSequence *self, PyObject *args, PyObject *kwds)
 {
     PyObject *ret = NULL;
     PyObject *py_samples = NULL;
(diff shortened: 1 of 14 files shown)


petrelharp:
That's a great idea.  Even if simplify() was the *only* think that worked
on them, that would be useful.

jeromekelleher:
OK, cool, working on it. Should have something to try out tomorrow.

jeromekelleher:
I think this should be working now @petrelharp and @ashander if you want to try it out. I've made the C-level changes required to make the basic infrastructure work, but I haven't tried it in Python yet, so there may be issues there.

Everything should work, except for ``variants()``, ``haplotypes()`` and newick trees. These should all throw and "Unsupported operation" error if called on a tree sequence with general samples. These could all be fixed easily enough probably, but I thought I'd bet the basic infrastructure in place first. I'll update this PR tomorrow with some Python test cases and merge fairly soon.

petrelharp:
Nice.  I've started testing it with ftprime, and am having issues, but
unclear yet where they are coming from. =)

jeromekelleher:
OK, this should be working pretty well now @petrelharp. I've done the changes for Python, and it mostly just works, except for ``haplotypes`` and ``variants``. The nice property that we have is that given any permutation of the nodes that preserves the order of the samples, we can create the corresponding tree sequence, then run ``simplify`` and get back precisely the original tree sequence (this is done in the new tests in ``test_topology.py``). So, we can look at ``simplify`` as a ``canonicalise`` function, which might actually be a useful way to think about it.

The downside here is that our current definitions of ``internal`` and ``leaf`` nodes don't really capture all the complexity. We probably need another definition for ``sample`` nodes, so we can distinguish non-sample leaves from samples.

petrelharp:
Awesome.

Remind me: where those are definitions recorded?

Also, remind me: must samples be leaf nodes?  Must they be at time 0?  Doing a bit of playing around I've found some cases where either we need to throw an error or .simplify() is giving the wrong thing, depending.

jeromekelleher:
> Remind me: where those are definitions recorded?

These are in msprime/trees.py. See 
https://github.com/jeromekelleher/msprime/blob/generalise-samples/msprime/trees.py#L513 for the updated definition.

> Also, remind me: must samples be leaf nodes? Must they be at time 0? Doing a bit of playing around I've found some cases where either we need to throw an error or .simplify() is giving the wrong thing, depending.

Hmm. At the moment, samples must be leaf nodes, yes. I think the counting algorithm do assume that they are leaves, but that could probably be fixed. Samples don't have to be at time 0, but I've probably never tested a tree sequence where there are no samples at time 0. Have you some examples where you are seeing problems?

I've updated this PR so that haplotypes and variants are also supported now, so I think this should be ready to merge soon. Should we merge now and look at your edge-cases later, or deal with these problems now before merging?

petrelharp:
Let me finish investigating -- later today.  I'll report back.

Re: definitions - how about adding a document to the docs somewhere writing down the requirements?  Could start from what we have at https://github.com/ashander/ftprime/blob/master/writeups/algorithm-notes.md - I would be happy to take a stab at it.

Re: samples must be leaves - this means that in practice to add samples we add a fake individual who copies entirely from the one we want to sample.  This is OK, if somewhat unnatural.  Also slightly wrong, as the sample must be at a slightly later time than the one we actually want to sample (but it can be arbitrarily close to right by taking the time difference small =).

jeromekelleher:
> Re: definitions - how about adding a document to the docs somewhere writing down the requirements? Could start from what we have at https://github.com/ashander/ftprime/blob/master/writeups/algorithm-notes.md - I would be happy to take a stab at it.

That's a great idea. Perhaps we could have a "Formal Model" page a bit like the "File Format" page? We could set stuff out properly there. (See the ``:math:`` tags usage in the API Documentation page.)

> Re: samples must be leaves - this means that in practice to add samples we add a fake individual who copies entirely from the one we want to sample. This is OK, if somewhat unnatural. Also slightly wrong, as the sample must be at a slightly later time than the one we actually want to sample (but it can be arbitrarily close to right by taking the time difference small =).

Dumb question: why not leave out the fake individual and mark the one you want to sample as sample? I'm missing something important I think!

petrelharp:
> > Re: samples must be leaves - this means that in practice to add samples
we add a fake individual who copies entirely from the one we want to
sample. This is OK, if somewhat unnatural. Also slightly wrong, as the
sample must be > > at a slightly later time than the one we actually want
to sample (but it can be arbitrarily close to right by taking the time
difference small =).
>
> Dumb question: why not leave out the fake individual and mark the one you
want to sample as sample? I'm missing something important I think!

Since if you want to sample a parent-offspring pair (say), then the parent
wouldn't be a leaf!

petrelharp:
I've started a placeholder for the doc; will work on it later.

jeromekelleher:
> Since if you want to sample a parent-offspring pair (say), then the parent wouldn't be a leaf!

I'm still not getting it though --- why sample the parent, and not just the offspring? I guess the parent might be carrying some ancestral material that you want to sample that is not present in the child? I'm still thinking backwards in time I think, so it's probably something simple I'm not following.

petrelharp:
> Since if you want to sample a parent-offspring pair (say), then the parent
> wouldn't be a leaf!
>
> I'm still not getting it though --- why sample the parent, and not just
> the offspring? I guess the parent might be carrying some ancestral material
> that you want to sample that is not present in the child? I'm still
> thinking backwards in time I think, so it's probably something simple I'm
> not following.
>
>
> Well, in the real world you might sample both parent and offspring - but
half the parent is genetic material not in the offspring, as you say.

Another reason: ancient DNA - probably whoever you sample will be parents
to everyone in the modern sample.

jeromekelleher:
OK, well let's think about making any node a sample. If we allow arbitrary nodes to be samples (the root, or a node not in any tree, e.g.), then I'm sure this will muck up the counting algorithms, which are about leaves. This can probably be fixed though.

The question then becomes, what do we actually mean by a sample? The coalescent simulations allow for ancient samples, but it does assume that these are not the ancestors of any of your other samples. Suppose I had sequenced all my family going back n generations, and I had somehow built a tree sequence out of all this information (and nothing else). What do we mean by sample then?

Also, shall I merge this PR or are there outstanding issues that that we should fix (interpretation of the concept of sample aside)?

petrelharp:
Apologies - I forgot to follow up on the actual errors.  I've just pushed some more tests in which internal nodes are samples.  The first two read in OK but then LibraryError when you try to .simplify().  The last one throws a C error and exits.  To fix this I added a check to the C for this situation, for my own edification - @jeromekelleher if this seems excessively cautious feel free to remove/modify.

petrelharp:
Either way I'm happy to merge.

ashander:
The stuff you just pushed @petrelharp is that d370ad8380ca6429d86af26ba76cfc4b1b22f60a on #199 ?

petrelharp:
Heh - yes, thanks, Jaime.  Er, I'll rebase that one to this one and update it.

jeromekelleher:
OK great, thanks @petrelharp  and @ashander. Merging.

Issue #189, opened 2017-04-14
no warning if numpy not present when installing pip


ashander:
When installing from github via `pip`, no warning shows up in a conda env that has no `numpy`. I think there should be one, based on [current code in setup.py](https://github.com/jeromekelleher/msprime/blob/e2a16a6bd86af71db45898d9b913073fa700e611/setup.py#L43).

```sh
env PYTHONNOUSERSITE=1 conda create -n msprime-numpy pip
source activate msprime-numpy
python -c 'import numpy'  # raises an error as expected
pip install git+https://github.com/jeromekelleher/msprime.git@master#egg=msprime  # no warning
```

example output
```sh
 $ env PYTHONNOUSERSITE=1 conda create -n msprime-numpy pip
Fetching package metadata .........
Solving package specifications: ..........

Package plan for installation in environment /home/jaime/miniconda3/envs/msprime-numpy:

The following NEW packages will be INSTALLED:

    ca-certificates: 2017.1.23-0      conda-forge
    certifi:         2017.1.23-py36_0 conda-forge
    ncurses:         5.9-10           conda-forge
    openssl:         1.0.2k-0         conda-forge
    pip:             9.0.1-py36_0     conda-forge
    python:          3.6.1-0          conda-forge
    readline:        6.2-0            conda-forge
    setuptools:      33.1.1-py36_0    conda-forge
    sqlite:          3.13.0-1         conda-forge
    tk:              8.5.19-1         conda-forge
    wheel:           0.29.0-py36_0    conda-forge
    xz:              5.2.2-0          conda-forge
    zlib:            1.2.11-0         conda-forge

Proceed ([y]/n)? y

Linking packages ...
[      COMPLETE      ]|#######<snip>###| 100%
#
# To activate this environment, use:
# > source activate msprime-numpy
#
# To deactivate this environment, use:
# > source deactivate msprime-numpy
#
 $ source activate msprime-numpy
 $ python -c 'import numpy'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'numpy'
 $ pip install git+https://github.com/jeromekelleher/msprime.git@master#egg=msprime
Collecting msprime from git+https://github.com/jeromekelleher/msprime.git@master#egg=msprime
  Cloning https://github.com/jeromekelleher/msprime.git (to master) to /tmp/pip-build-xap8rm9h/msprime
Collecting svgwrite (from msprime)
[...]

PR #195, opened 2017-04-24
Print 'warnings.warn' error on import and in CLIs

ashander:
@jeromekelleher :

1. How's the form and text of the warning?
2. Any advice/ideas to achieve the goal of not printing in CLIs? it's seeming a bit tricky

relates to #189

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,14 +23,14 @@ import subprocess
 import platform
 import os
 import os.path
-
-
-def warn(message):
-    print("Warning:", message)
+from warnings import warn
 
 
 CONDA_PREFIX = os.getenv("MSP_CONDA_PREFIX", None)
 IS_WINDOWS = platform.system() == "Windows"
+
+# NOTE: sadly these warnings won't work to warn users installing with pip
+# https://github.com/pypa/pip/issues/2933
 HAVE_NUMPY = False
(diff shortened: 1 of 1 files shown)


ashander:
Oops! Some issues in the changes to setup.py. Hopefully I've fixed them

jeromekelleher:
Thanks for this @ashander. You're right, this _is_ tricky. I've spent a bit of time playing with it, and I'm not sure now that there is any good way to do this with warnings. Because this happens at import time, it's very difficult to suppress the selectively suppress the warnings. The only time we get bad behaviour at the moment is from the tables API, which gives opaque error messages. For everything else, we get nice clear errors saying the reason we have an exception is that numpy is missing. We may just have to live with this for now I think.

For now, I'd suggest stripping out all the changes except to ``setup.py``, and merging that much (after squashing the commits). There are ways to improve the error messages from the Table subclasses, but they don't involve the warnings module. Sorry to send you down the wrong path!

ashander:
Sounds good, and done. Import messages are a bit annoying anyhow so nice to avoid them.

PS for future reference my import warning attempts are preserved here: https://github.com/ashander/msprime/tree/fix/add-import-warning-attempts

PR #199, opened 2017-04-27
checks for samples not at time 0

petrelharp:
I've added some tests; they look good.

I also removed the requirement that get_time() apply only to samples, as it was previously checking by comparing to num_samples - I can't see what this would break and the tests pass...

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2320,7 +2320,6 @@ class TreeSequence(object):
     def time(self, sample):
         return self.get_time(sample)
 
-    # TODO refactor these methods to take node arguments not samples.
     def get_time(self, sample):
         """
         Returns the time that the specified sample ID was sampled at.
@@ -2329,7 +2328,7 @@ class TreeSequence(object):
         :return: The time at which the specified sample was drawn.
         :rtype: int
         """
-        if sample < 0 or sample >= self.get_sample_size():
+        if sample < 0 or sample >= self.get_num_nodes():
             raise ValueError("Sample ID out of bounds")
         node = self.node(sample)
         return node.time
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2317,38 +2317,38 @@ class TreeSequence(object):
             time=time, population=population, name=name,
             is_sample=flags & NODE_IS_SAMPLE)
 
-    def time(self, sample):
-        return self.get_time(sample)
+    def time(self, u):
+        return self.get_time(u)
 
-    def get_time(self, sample):
+    def get_time(self, u):
         """
-        Returns the time that the specified sample ID was sampled at.
+        Returns the time that the specified ID was alive at.
 
-        :param int sample: The sample ID of interest.
-        :return: The time at which the specified sample was drawn.
+        :param int u: The individual ID of interest.
+        :return: The time at which the specified individual was alive at.
         :rtype: int
         """
-        if sample < 0 or sample >= self.get_num_nodes():
-            raise ValueError("Sample ID out of bounds")
-        node = self.node(sample)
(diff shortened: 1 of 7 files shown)


msprime/trees.py
```diff
@@ -2320,7 +2320,6 @@ def node(self, u):
     def time(self, sample):
         return self.get_time(sample)
 
-    # TODO refactor these methods to take node arguments not samples.
     def get_time(self, sample):
```
jeromekelleher: We should change the parameter here to ``u`` (also for the ``time`` method above), and change the documentation so that sample -> node.

jeromekelleher:
Hah, it's pretty cool that worked without any problems. So, there's no reason at all that time has to start at 0. It should also work if we have negative time I guess. Is there any point in allowing this?

petrelharp:
Hey, that's great.  Yes - then switching from forwards to reverse time
would just involve multiplying by -1, rather than having to go look up the
maximum time and subtract that.

msprime/trees.py
petrelharp: it is done.  also for get_population.

petrelharp:
OK - this is ready to merge (pending what @jeromekelleher wants to do about new check for internal samples).

ashander:
@petrelharp there were a couple issues breaking c-level tests. I was going to add comments here on the diff but wanted to make sure they'd work, so just pushed a PR to your `generalise-samples` branch (see above).

petrelharp:
Darn it, forgot to run those check, thanks.  I've checked and merged.

But - at first I couldn't check because compiling the tests gave me this error:
```
msprime/lib$ make
(...)
tree_sequence.c: In function ‘tree_sequence_write_hdf5_metadata’:
tree_sequence.c:2148:22: error: conversion to ‘herr_t {aka int}’ from ‘hid_t {aka long int}’ may alter its value [-Werror=conversion]
             status = dataspace_id;
```

Casting `dataspace_id` to `herr_t` makes this complie; I've done this but don't actually know if this might have bad effects. (@jeromekelleher)

ashander:
As a data point: I didn't run into that `herr_t` vs `hid_t` error/warning.

Also @petrelharp I think you may have forgotten to push your merge. In any case, since you gave me access to your fork, I just merged my commit to set if it'll work here on CI

jeromekelleher:
Thanks a million @petrelharp and @ashander, this is great. I've made a few minor changes:

- Move the test for internal nodes as samples to ``init_nodes``. This is just to try and keep all the error checking logic grouped together in some way. Currently there are two types of error checking 'cheap' error checks that we do everytime we load a tree sequence and more expensive checks that are done in ``tree_sequence_check``. The distinction between the two is somewhat arbitrary at the moment...

- Refactored the tests in ``test_topology`` a little bit.

Re the ``herr_t`` vs ``hid_t``  thing, this is because @petrelharp is using a different compiler. Casting should be harmless and is the right thing to do. Is the change included in this patch?

Unfortunately, I didn't spot that this PR was against the ``generalise-samples`` branch until after I had merged it. However, if you rebase this PR off master we should be able to merge it easily then.

ashander:
@jeromekelleher @petrelharp I rebased it and add that change (see https://github.com/jeromekelleher/msprime/pull/199/commits/2ef42a26b3896be8e2e3f5da6067c7849f8c5e4d ) which was there, but just on `petrelharp/master`

Note: I can't change the base of this to `master` so it looks like some of Jerome's commits from `generalise-samples` are still included here

petrelharp:
So that's where that commit went.  Whoops!  Thanks, Jaime.  I've also
changed the base.

jeromekelleher:
Merging; thanks @petrelharp and @ashander!

PR #203, opened 2017-05-29
added dump samples option to dump_text

petrelharp:
I wanted a way to get the information about the samples out easily, and figured it might be useful to others.  So, I added it as an option to dump_text - output is like dumping the NodeTable except with another column 'id'.

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -1753,8 +1753,6 @@ class TreeSequence(object):
         :param stream edgesets: The file-like object to write the EdgesetTable to.
         :param stream sites: The file-like object to write the SiteTable to.
         :param stream mutations: The file-like object to write the MutationTable to.
-        :param stream samples: The file-like object to write the subset of the NodeTable
-            describing the samples to, with an extra column, `id`.
         :param int precision: The number of digits of precision.
         """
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I see that this is a useful thing to have @petrelharp, but I'm a bit nervous about the interface. The idea of the ``dump_*`` and ``load_*`` methods is that they should be symmetrical methods for importing and exporting data. Since the ``samples`` output is never input anywhere, and is a subset of the nodes output, this seems like it breaks that symmetry to me, and we run the risk of the ``dump_text`` method getting more and more complicated.

Two suggestions occur to me:

1) Since the only difference between the output of the ``samples`` file is an ID column, we can easily add the ID column to the nodes file. This is probably a generally useful thing to have (but is still strictly redundant), we could add a parameter like ``include_id=False`` to ``dump_txt`` which includes an ID column in all the files. For your downstream application, you can then just filter out the rows in the samples file with ``is_sample`` = 0.

2) If you don't like the idea of filtering out the non-sample nodes, we can add an ``dump_samples_text`` method or something similar, which just outputs the samples.

What do you think?

petrelharp:
Meh, I was a bit nervous about adding more things too.  I could just make a
separate, quick function to do this - no need to clutter up the interface
with `include_id` I think.  It does seem like an important thing to obtain
from a simulation, though - the information about the samples - so maybe it
should be a separate method.  I am setting up a simulation, including
picking samples randomly, and although I could output sample info at that
point, it seems more certain to get the info straight from msprime to be
sure it is consistent.

So, I lean towards #2 or nothing but leave it up to your judegement?

jeromekelleher:
OK,  let's go with option (2) so. I'm sure exporting the samples in a tabular format will be a generally useful thing to have. I think we can leave out the ``is_sample`` column here though, since it will always be 1.

Would you mind coding it up? It's pretty much what you already have, plus a new test method ``test_dump_samples_text`` in ``TestTreeSequenceTextIO``.

petrelharp:
It is done.

msprime/trees.py
```diff
@@ -1742,10 +1742,20 @@ def dump_tables(
             mutations=mutations)
 
     def dump_text(
-            self, nodes=None, edgesets=None, sites=None, mutations=None, precision=6):
-        # TODO document.
+            self, nodes=None, edgesets=None, sites=None, mutations=None, samples=None,
```
jeromekelleher: We still have the ``samples`` parameter left over here.

petrelharp:
Heh, whoops -- so I do.

jeromekelleher:
Merging. Thanks again @petrelharp!

Issue #206, opened 2017-06-29
Multiple mutations on same branch in tree.draw() not being displayed


koelling:
It looks like tree.draw() cannot display multiple mutations on the same branch of the tree properly. Instead, only a single red square is displayed for all of them. Presumably this is because the squares are all drawn on top of each other at the exact same location?

![screen shot 2017-06-29 at 14 51 03](https://user-images.githubusercontent.com/369559/27690884-680f0914-5cda-11e7-9044-83f9d5bad834.png)

jeromekelleher:
Yes, well spotted. This is an easy fix: we just need to count the number of mutations over each node, and divide the distance along the branch in to this number of equal segments. The code would be here:

https://github.com/jeromekelleher/msprime/blob/master/msprime/trees.py#L294

Also, these red squares are embarrassingly ugly....

jeromekelleher:
Fixed in #207.

Issue #204, opened 2017-06-29
Return SVG string from tree.draw()


koelling:
It would be great if `tree.draw()` provided a way of returning the SVG string instead of only writing to a file. Maybe that could be the behaviour if `path == None` for example. This would make it possible to visualise trees directly in jupyter notebook using `IPython.display.SVG()`.

Here is a workaround wrapper function until then:

```python
import tempfile, IPython.display
def display_tree(tree, *args, **kwargs):
    with tempfile.NamedTemporaryFile() as f:
        tree.draw(f.name, *args, **kwargs)
        f.flush()
        IPython.display.display(IPython.display.SVG(filename=f.name))

#usage: display_tree(tree, show_times=True)
```

jeromekelleher:
This is a good idea @koelling, I completely agree. What we could do is return the SVG text and make the filename an optional keyword argument. This would mean that current code should continue to work, but also enable this use-case. What do you think?

koelling:
Sounds great!

jeromekelleher:
It's an easy fix if you would like to have a go...

PR #207, opened 2017-06-30
Changed to 'square' tree layout.

jeromekelleher:
This is a work in progress to improve tree drawing in msprime.

What do people think of "square" trees versus "triangular" trees? The consensus seems to be for square trees, so I think msprime's tree drawing should go along with this. Pinging @petrelharp and @ashander for opinions! 

Here's an example from the new code:

tree_0

@koelling, this fixes issue #206.

TODO:
1. Provide an option to rescale the branch lengths so it's easier to see the topology
2. Fix non binary tree handling
3. Return SVG as described in #204 
4. More?

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -271,10 +271,16 @@ class TreeDrawer(object):
     """
     A class to draw sparse trees in SVG format.
     """
-    def __init__(self, tree, width, height, show_times):
+    def __init__(
+            self, tree, width=200, height=200, show_times=False,
+            show_mutation_labels=False, show_internal_node_labels=True,
+            show_leaf_node_labels=True):
         self._width = width
         self._height = height
         self._show_times = show_times
+        self._show_mutation_labels = show_mutation_labels
+        self._show_internal_node_labels = show_internal_node_labels
+        self._show_leaf_node_labels = show_leaf_node_labels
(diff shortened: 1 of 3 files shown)


petrelharp:
I love it!

petrelharp:
I'm trying to resist the temptation to write something that will print them in ascii.

Oh, but it already exists: http://etetoolkit.org/

jeromekelleher:
> Oh, but it already exists: http://etetoolkit.org/

It's true, it does feel like a waste of effort making yet another tree viz method. I'm not all that keen on having a dependency on ete or dendropy or anything though, and I'm not sure how you would go about annotating their trees with mutations. Also, I'm hoping to write some real *tree sequence* visualisations some day, which will definitely need to be bespoke.

So, I'd be happy if you lost your fight against temptation @petrelharp!

petrelharp:
sounds good to me.  those ones you've got are great.

ashander:
Square looks good, and seems easier to sprinkle with mutations. Great stuff!

ashander:
Re @petrelharp comments: it would be great to have something that could be used either with `__str__` (or `__repr__` for even less typing) to printed an ascii tree in the REPL. but this is of limited utility (mostly for debugging and interaction with small tree examples).

PR #209, opened 2017-07-10
Bottomup simplify

petrelharp:
Here (in `simplify_work/`) is an implementation of the bottomup simplify algorithm that follows the python code in `algorithms.py` as closely as I could manage.

Good news: it runs!
```
msp simulate 10 trees.ts
python3 simplify_algorithms.py simplify trees.ts 3
```

Bad news: it isn't right if there's more than one tree.  Also maybe it falls in an infinite loop sometimes.  But, this is just debugging.  @jeromekelleher : is this close enough to the C to enable easy translation, once it works?

Main points:

- This is different to msprime in that recombination and coalescence events happen at the same time.  To do this, I wrote my own `remove_ancestry` function, kinda like `recombination_event`, and basically used `merge_ancestors` as-is.
- But, to associate ancestors in the internal state to nodes of the edgesets we're reading in, I had to add an additional data structure, `A`, that stores for each node ID in the input sequence the ID of the head ancestral segment.  This then has to be updated in `merge_ancestors`.
- The only other thing is getting the initial state right.
- I don't think we actually need the Fenwick tree here, since we don't need to choose segments proportional to their length.  But maybe I'm missing something else we need them for.

I'm not clear how this eventually gets into NodeTables and EdgesetTables. I think that happens later on, so I shouldn't worry?

I'll finish tracking down the bugs, but a go-ahead would be helpful.


simplify_work/simplify_algorithms.py
```diff
@@ -0,0 +1,305 @@
...
+        self.segment_stack = []
+        self.segments = [None for j in range(self.max_segments + 1)]
+        for j in range(self.max_segments):
+            s = Segment(j + 1)
+            self.segments[j + 1] = s
+            self.segment_stack.append(s)
+        self.P = [SearchablePopulation(id_) for id_ in range(N)]
+        self.C = []
+        self.L = FenwickTree(self.max_segments)
+        self.S = bintrees.AVLTree()
```
jeromekelleher: I don't think we need S here. This is used to track the amount of ancestral material present in each segment, so that we know when Hudson's algorithm is complete.

jeromekelleher:
Thanks @petrelharp, this is fantastic. I'm convinced this is the right way to do simplify, and the implementation looks at exactly the right level for porting to C.

The algorithm is subtle, and I'll need to really think hard about this again to give detailed feedback. I'll try and get to this in a couple of days.

In the mean time though:

- You definitely don't need a Fenwick tree. This is just for selecting recombination breakpoints (as you say).
- Don't worry about the tables for now, these will be created at the end when the full simplify pass is done. It's too hard to create them all in one go.

petrelharp:
This appears to work, or at least to pass the ascii-illustrated tests I borrowed from `test_topology.py`:
```
cd simplify_work
nosetests3 test_simplify.py
```

I realized I was thinking unclearly about one or two things; the final resolution is hopefully well-enough described in `simplify_algorithm.md`.  In particular,  the "population" `P[0]` was not doing anything and serving pretty much the same purpose as the dictionary `A` where I kept track of which input node went with which internal ancestor; now they are the same.  (but they're both still there, to allow easier re-use of code...)

Unfortunately, `merge_ancestors` still needs to be modified somewhat to allow assigning the resulting ancestor the correct input node label: either it needs to return the head segment of the merged ancestor (the current strategy) or to take an extra argument so it can do that assignment itself.  The second way might be clearer, as we could then pass the optional argument onto `pop.add(alpha)` (currently commented out).

Input on how to proceed?

petrelharp:
I've added tests for the case when input is overlapping records.  They don't pass, but I think it is working out of the box (yay!) except for not merging adjacent identical records, which I am hoping is dealt with elsewhere?

petrelharp:
Ah, good.  Got it working on all the tests, by merging adjacent identical records on output.  This is not totally robust, but also a problem you've dealt with more correctly in the main code, i think.

So: should be good to translate over.

jeromekelleher:
Fantastic! This is superb @petrelharp, I'm very happy you got to the end of this one. Some minor notes:

1) For the A map, I would just use flat array in the C code. It's rarely worth bothering with a hash table for these when you can bound the number of keys and the keys are dense. (I haven't thought about the reverse ID map properly though, this might be more complicated.)

2) WRT ``merge_ancestors``, we were never going to be able to use it exactly as-is (nor would it be good idea to try I think). So, I'm perfectly happy to copy the bulk of the code from ``msprime.c`` into ``simplifier.c``,  get rid of the unnecessary stuff and tweak for the new application. Things like ``defrag`` are unnecessary here and can be cut out.

How would you like to proceed from here? I'd like to take a pass through the algorithm (I could spend a couple of hours on it tomorrow), and see if I can push it closer to the minimum design needed for the C implementation. Once at that point I'm confident I could translate it into actual C code pretty quickly. However, if you'd like to do the C coding I'm more than happy to provide pointers.

petrelharp:
You taking a pass through sounds great.  I have done enough looking at the C code that it's starting to come into focus, but I'd much rather look at what you do with it.  And, it'd be good to have another pair of eyes on it.

You mention the 'reverse ID map' - from segments back to input node IDs?  The only time we need to do this is, sort of, is to assign a new ancestor in `merge_ancestors`, I think; at all other times we're accessing ancestors by their IDs.

Oh, one other note: we should probably be using S to decide when to stop, but I hadn't gotten around to figuring out how to do that.

jeromekelleher:
I've made a another pass at this @petrelharp, and I think it's looking great. I've gotten rid of the text IO and started updating the Node/Edgeset tables directly as the new values are allocated. This is very easy because of the nice properties of the bottom-up algorithm.

The only thing we need to do before C-ifying it is to add support for Sites and Mutations. It would be nice if we could tackle multiple and back mutations straight away, but we can keep it simple with just infinite sites mutations for now. I feel like it should be pretty simple to handle mutations here, as we just have to look at see what mutations are above a given node when we allocate a new one. What do you think?

Also, have we tested this on a tree sequence with multiple roots?

Once this is converted to C code, what would you like to do with the stuff in the ``simplify_work`` directory? I was thinking of replacing the implementation in ``simplify_tree_sequence`` function in ``tests/highlevel.py`` with the ``Simplifier`` class. I guess we can move the tests into ``test_topology.py``. What about the algorithm description, where do you think the right home for this is?

petrelharp:
I'm very impressed how much cleaner this is now!

**Mutations:**

I think I have the basic set-up for doing sites and mutations (including back-mutations).  Although: it doesn't work because I don't know the right way to do a few things.  I think we need to do two things:

- (`record_mutations`) Every time we process the input edgesets for an input_id, go through the resulting (merged) ancestral segments, copying over any input mutations found on these segments except now associated with the output node of that segment.  This requires:
    - finding all (sites, mutations) that affect a given input node and fall in a given [left, right) interval

- (`update_ancestral_state`) Whenever the algorithm finishes on a particular segment, reassign ancestral states at all sites in that segment.  This requires
    - finding the last mutation above a given input node at a given site, if any.

For output, I'm keeping a dict of Sites, indexed by position, each of which has a list of Mutations; then these get assigned indices and nodes respectively on output.  I am not remembering how site and mutation tables need to be ordered, so perhaps am not doing this right.

petrelharp:
I've added the multiple roots test from https://github.com/jeromekelleher/msprime/pull/131 - it seems a bit overly complicated, but there it was.  It fails, but on
```
    assert self.num_used_segments == 0
```
unsuprisingly.  Not sure what to do about this - check that number of used segments is equal to the number of roots in the tree?

petrelharp:
About future homes for things:
 - SImplifier and tests: sounds good.
 - the algorithm description we should store away for the methods paper: maybe start a new repo for that?

jeromekelleher:
Yes, this seems like the right way to do mutations, excellent. I'll try and finish this up tomorrow, and get a working C implementation. I hope we'll be able to merge early next week then.

Re. the assertion on ``self.num_used_segments`` --- we can just get rid of this. It's pretty clear that we're freeing segments when we should be, so this is an unnecessary complication.

jeromekelleher:
OK, so I've failed to implement all of this in C today. However, I did make some decent progress and will hopefully pick it up early next week and finish it off.

# Gappy tree sequences 

While the algorithm works when we have multiple roots, it was producing a non-conforming tree sequence in the super-awkward test example. The non-conformance was due to the NONMATCHING_RIGHT requirement, where every edgeset must have some matching left/right coordinate somewhere else. When we simplify the awkward example, it gets rid of a bunch of unary edgesets, and this ends up leaving a hole in the tree sequence. 

After hacking the algorithm a bit to force in some unary edgesets for the remaining ancestral material when simplify completes, I ended up with quite an ugly mess, which also contradicted what you would expect for simpler cases. So, I think the right thing to do is remove the requirement that there can't be gaps in the tree sequence. We'll have to think about the consequences of this a bit more and make sure that 'gappy' tree sequences don't break any of our code, but I think it should mostly be OK.

# Sites/mutations are hard

I spent a long time today trying to figure out how to do mutations properly in the new algorithm, and I'm not sure I've found the right approach. Certainly the test cases are finding ways in which the current draft breaks. I'm going to make another pass at this on Monday and see if I can figure out the right approach.

In terms of progress, I've moved the code into its more permanent home in the tests directory. Once I've implemented the new simplify in C, then we can port the remaining test cases into ``test_topology.py`` and remove the ``simplify_work`` directory.

petrelharp:
Without thinking about it substantially, it sounds to me like maybe the right thing to do is to require that trees have only one root.  This is easy to guarantee in simulations, and prevents bad things like what we'd have to do when determining haplotuypes in such situations (assume that *all* roots have the same `ancestral_type`).

petrelharp:
An issue we haven't discussed, but maybe you've already figured out: I think the plan is to let simplify work on tables that are not good tree sequences, not only because they have not been sorted but also because edgesets may be overlapping.  Currently, we'd do this by
```
msprime.sort_tables(...)
ts.load_tables(...)
ts.simplify()
```
... but this requires letting tables with overlapping edgesets into a tree sequence.  Currently, I believe that msprime will load such edgesets without complaint, but: is this the plan going forward?  (seems like the best option to me.)  We should make a note in the documentation about this if so.

jeromekelleher:
I hadn't thought of this @petrelharp... I think you're right, and we allow overlapping edgesets into the tree sequence. Simplify should deal with this fine (I think). We should be able to detect overlapping edgesets during tree generation, and so we can fail there. This would ensure safety I think, because pretty much everything depends on tree generation.

I don't think we can detect overlapping edgesets cheaply at import time, so I think allowing them (and failing on tree generation) is the best solution all round. Even better if simplify will get rid of them. I havent' checked any of this though --- will definitely need to formalise it and write a bunch of topology tests!

petrelharp:
Well, `single_childify()` will help with the tests.

jeromekelleher:
OK, this is a C version of simplify. Unfortunately I wasn't able to implement sites/mutations, so this does break a lot of test cases. I think the current code passes all the topological tests, but I would treat it with suspicion right now. I does need quite a bit more work before it'll be ready to merge. In terms of performance, there's a lot of ``check_state`` calls in there at the moment that are verifying various things, so I wouldn't read too much into the current performance.

TODO:

1. Add sites/mutations
2. Fix various test cases
3. Clean up, fix segment leaks, etc.

petrelharp:
yay!  I will investigate.

petrelharp:
A note from @jeromekelleher that I am recording here:
> I think the right path is to make a function simplify_tables, which simplifies a given set of tables in place. > The edgeset table can definitely be updated in place, which will save quite a lot of memory.

petrelharp:
I need to figure out generally what to do about internal nodes that are samples. With overlapping generations, in forwards time we will want to run simplify on tables like this. A workaround is to place phony samples - but since offspring must come strictly after parents, this is not right either.  (this would work if offspring could be contemporaneous with parents)  But, I don't know what it would take to remove this restriction.

Naively removing the check for internally sampled nodes does not work, even for this tree:
```
nodes = six.StringIO("""\
id      is_sample   population      time
0       1           -1              1.00000000000000
1       1           -1              0.00000000000000
2       1           -1              0.00000000000000
""")
edgesets = six.StringIO("""\
id      left            right           parent  children
0       0.00000000      1.00000000      0       1,2
""")
ts = msprime.load_text(nodes=nodes, edgesets=edgesets)
tss = ts.simplify()
for x in tss.dump_tables():
  print(x)
```
produces
```
id	flags	population	time
0	1	-1		1.00000000000000
1	1	-1		0.00000000000000
2	1	-1		0.00000000000000
3	1	-1		1.00000000000000
id	left		right		parent	children
0	0.00000000	1.00000000	3	1,2
```

petrelharp:
Writing up things I saw the suggestions for `simplifier_remove_ancestry`; implemented them in the python version.

petrelharp:
Whoops - that last commit should only be over on https://github.com/petrelharp/msprime/pull/5 - not here.

jeromekelleher:
I've updated this with a more streamlined algorithm for ``remove_ancestry``. While this isn't really any simpler than @petrelharp's, it does have much tighter inner loops which should pay off efficiency wise. Also, it should make life a bit easier when we're splitting mutations among segments, as it's fairly clear where we need to do this now.

Segment leaks are fixed, edgeset squashing and children sorting are both implemented now also. Hopefully it's just a case of implementing sites/mutations now and we're nearly done.

jeromekelleher:
OK, we're getting there! It's taken a while to add the algorithm for sites/mutations, but mostly because the current approach doesn't feel right to me. Basically, we're now attaching a sorted list of mutations to every ancestral segment. As we go backwards in time, segments pick up the mutations that are relevant to them, and when we have a coalescence we know the correct output node and have all the mutations on hand. Currently, this involves a lot of AVL tree manipulation, which seems like overkill. However, we'll see how much time it really takes with a bit of profiling. It might turn out to be OK.

Anyway, the good news is that the algorithm now works for infinite sites mutations (and possibly some forms of recurrent mutations); yay! It's currently broken for back mutations, but I'll try and sort this out tomorrow.

jeromekelleher:
Definitely nearly there now... 

## Good news:

- Code is working for recurrent/back mutations when we have 0/1 mutations. For more general ACGT type mutations, will need to introduce the ``parent_mutation`` idea so that we can correctly capture the ancestral state. This is also the case for when we have mutations above the root. Neither of these seems hard now though, as the logic for deciding what the ancestral state is seems easy once we have links to the parent mutation.

- Nearly all tests are passing, and the C code is looking reasonably hardened now.

## Bad news:

- We have a **big** performance drop off from the old sideways algorithm. It's clear that there is a massive bottleneck at the moment in keeping the overlap counts updated, so I'll look into removing these from the algorithm tomorrow. It's such a massive bottleneck right now though I'm not too worried about performance; it's very obvious where we're going wrong, and I'm sure it'll be possible to fix.

## Next steps:

I would like to fix the performance bottlenecks in the code here, add a Python ``simplify_tables`` interface and then merge (hopefully tomorrow). Before merging I would like to delete the ``simplify_work`` directory --- is this OK @petrelharp? The Python version of the algorithm is now in ``tests/__init__.py``. A simple CLI for this Python version of simplify is available as ``python tests/__init__.py``, and there are some tests for it in ``test_topology.py``.

Are there some tests in the ``simplify_work`` directory that should be ported back into the test suite @petrelharp?

Also, the current version of the algorithm doesn't support non-leaf samples. I'll either implement this tomorrow, or open an issue about it.

petrelharp:
I agree, it seems like we can get rid of overlap counts: we might not know
when we can stop tracking a given segment, but that's OK if the performance
gain is worth it.

`simplify_work` - yeah, go for it.  Sorry I haven't cleaned it up so far.
If I'm missing something there I'll excavate it.  I don't think there are
tests we need to move over from there, either.

happily,
  Peter

jeromekelleher:
> I agree, it seems like we can get rid of overlap counts: we might not know when we can stop tracking a given segment, but that's OK if the performance gain is worth it.

A quick update here. It seems that by roughly pulling out the stuff for maintaining overlap counts it almost (but not quite) works. It looks like topologically it works just fine, but there's some complication with mutations. But, performance is **massively** improved, such that it's now properly bloody fast. It now takes about 80 seconds to run through a 1.7G file with 10^7 samples (and a large chunk of that is building the indexes for the output tree sequence). This takes 230 seconds with the sideways algorithm. There's probably a bit more to be squeezed out with a bit of profiling, and hopefully fixing up the mutation problem won't be too hard/slow things down.

jeromekelleher:
Update: 

- Sites are working well now. We'll have to look at it again when we're doing ACGT mutations, but I don't think it'll be hard.
- Ancestral samples in simplify is working. There are still some slightly odd aspects to this, so I'm not sure if it works properly yet.
- Performance is excellent; I didn't bother doing any profiling as it was too quick to get an decent feel for what's going on. We should have a look again when some very large forwards time simulations are simplified down though.

I would like to convince myself that the ancestral sample stuff is working properly and then merge. We can make an issue for the remaining stuff that needs  to be fixed up and take things on piece by piece.

petrelharp:
I've added mutations to a visual python test - doesn't work yet, but I don't think we expect it to yet.

In doing that, I came across one question: should a requirement for mutation tables be that there is only one mutation per unique (node, site) combination?

jeromekelleher:
> In doing that, I came across one question: should a requirement for mutation tables be that there is only one mutation per unique (node, site) combination?

I would have said yes, but I have a habit of over constraining things. Can you think of a reason why we would want multiple mutations on a branch? I guess if someone wanted to look at specific mutations then this could be useful.

jeromekelleher:
> I've added mutations to a visual python test - doesn't work yet, but I don't think we expect it to yet.

The code should work for 0/1 mutations so long as the ancestral state is 0. I see that one of the sites here has an ancestral state of 1 --- would you mind changing that please?

I'd rather sort out the ancestral state stuff now and revisit sites/mutations more thoroughly later. There's quite a bit of work to be done to fully sort it out.

petrelharp:
Flipped that ancestral state.

jeromekelleher:
Update:

- Added simplify_tables interface. Use this like 
```python
msprime.simplify_tables(samples=[0, 1], nodes=node_table, edgesets=edgeset_table)
```

This will simplify the specified tables in place. The samples argument is a numpy array, so the most efficient way to use this is to have a numpy array for your samples (dtype=np.int32). Lists etc are fine though, the just get converted on the fly.

- In the process of testing this, I found myself needing some more sophisticated table tools. I added two things:
1. A copy method for all tables that returns a deep copy. (i.e., ``node_copy = node_table.copy()``)

2. An equality function for tables, which basically does a byte-by-byte comparison of all the values. So, we should have ``node_table.copy() == node_table``.

I think this is good enough to merge now. There's a lot still to be done, but what's here is reasonably solid and we can make incremental progress from this point with smaller PRs.

Thanks for the algorithm @petrelharp, it's awesome!

jeromekelleher:
>> In doing that, I came across one question: should a requirement for mutation tables be that there is only one mutation per unique (node, site) combination?
> I would have said yes, but I have a habit of over constraining things. Can you think of a reason why we would want multiple mutations on a branch? I guess if someone wanted to look at specific mutations then this could be useful.

Another quick point here while I remember it. Actually, allowing for multiple mutations from a given site on a given branch would make our lives a bit easier when it comes to simplify. We then won't need to worry about coalescing mutations that occur on different branches on the path rootward.

PR #210, opened 2017-07-10
Add ability to colour tree drawings

hyanwong:
May need some work for cases where leaf nodes are not 0..n

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -274,7 +274,7 @@ class TreeDrawer(object):
     def __init__(
             self, tree, width=200, height=200, show_times=False,
             show_mutation_labels=False, show_internal_node_labels=True,
-            show_leaf_node_labels=True):
+            show_leaf_node_labels=True, node_colours=[]):
         self._width = width
         self._height = height
         self._show_times = show_times
@@ -289,11 +289,17 @@ class TreeDrawer(object):
         self._tree = tree
         self._x_coords = {}
         self._y_coords = {}
+        try:
+            self._colour = {node:col for node,col in node_colours.keys()}
+        except:
+            self._colour = {node:col for node,col in enumerate(node_colours)}
+            
         for u in tree.nodes():
             scaled_t = tree.get_time(u) * self._y_scale
             self._y_coords[u] = height - scaled_t - y_padding
         self._leaf_x = 1
         self._assign_x_coordinates(self._tree.get_root())
+        self._assign_colours(self._tree.get_root())
         self._mutations = []
         node_mutations = collections.defaultdict(list)
         for site in tree.sites():
@@ -316,7 +322,7 @@ class TreeDrawer(object):
         code as text.
         """
         dwg = svgwrite.Drawing(size=(self._width, self._height), debug=True)
-        lines = dwg.add(dwg.g(id='lines', stroke='black'))
+        lines = dwg.add(dwg.g(id='lines', stroke='dimgrey'))
         labels = dwg.add(dwg.g(font_size=14, text_anchor="middle"))
         for u in self._tree.nodes():
             v = self._tree.get_parent(u)
@@ -343,8 +349,12 @@ class TreeDrawer(object):
                     dy=dy))
             if v != NULL_NODE:
                 y = self._x_coords[v], self._y_coords[v]
-                lines.add(dwg.line(x, (x[0], y[1])))
-                lines.add(dwg.line((x[0], y[1]), y))
+                if self._colour.get(v):
+                    lines.add(dwg.line(x, (x[0], y[1]), stroke=self._colour[v]))
+                    lines.add(dwg.line((x[0], y[1]), y, stroke=self._colour[v]))
+                else:
+                    lines.add(dwg.line(x, (x[0], y[1])))
+                    lines.add(dwg.line((x[0], y[1]), y))
         for x, mutation in self._mutations:
             r = 3
             dwg.add(dwg.rect(
@@ -371,6 +381,22 @@ class TreeDrawer(object):
             self._x_coords[node] = self._leaf_x * self._x_scale
             self._leaf_x += 1
 
+    def _assign_colours(self, node):
+        """
+        Assign colours up the tree, making parents the same
+        colour as their children, unless there is a disagreement
+        in which case no colour is set
+        """
+        if self._tree.is_internal(node):
+            children = self._tree.get_children(node)
+            for c in children:
+                self._assign_colours(c)
+            if node not in self._colour:
+                colours = set([self._colour.get(c) for c in children])
+                if len(colours)==1:
+                    single_colour = colours.pop()
+                    if single_colour is not None:
+                        self._colour[node] = single_colour
 
 # TODO:
 # - Pickle and copy support
@@ -637,7 +663,7 @@ class SparseTree(object):
     def draw(
             self, path=None, width=200, height=200, show_times=False,
             show_mutation_labels=False, show_internal_node_labels=True,
-            show_leaf_node_labels=True):
+            show_leaf_node_labels=True, node_colours=[]):
         """
         Returns a representation of this tree in SVG format.
 
@@ -650,6 +676,7 @@ class SparseTree(object):
         :param bool show_mutation_labels: If True, show labels for mutations.
         :param bool show_internal_node_labels: If True, show labels for internal nodes.
         :param bool show_leaf_node_labels: If True, show labels for leaf nodes.
+        :param list branch_colours: A list giving colours for nodes in the tree. If this is a .
         :return: A representation of this tree in SVG format.
         :rtype: str
         """
@@ -660,7 +687,8 @@ class SparseTree(object):
                 self, width=width, height=height, show_times=show_times,
                 show_mutation_labels=show_mutation_labels,
                 show_internal_node_labels=show_internal_node_labels,
-                show_leaf_node_labels=show_leaf_node_labels)
+                show_leaf_node_labels=show_leaf_node_labels,
+                node_colours=node_colours)
         svg = td.draw()
         if path is not None:
             with open(path, "w") as f:
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -676,7 +676,7 @@ class SparseTree(object):
         :param bool show_mutation_labels: If True, show labels for mutations.
         :param bool show_internal_node_labels: If True, show labels for internal nodes.
         :param bool show_leaf_node_labels: If True, show labels for leaf nodes.
-        :param list branch_colours: A list giving colours for nodes in the tree. If this is a .
+        :param list branch_colours: A list giving colours for nodes in the tree. If this is an array such as ['red','blue']the indexes of the array correspond to the nodes to colour. If a dict, the keys give the node numbers for each colour.
         :return: A representation of this tree in SVG format.
         :rtype: str
         """


msprime/trees.py
```diff
@@ -650,6 +676,12 @@ def draw(
         :param bool show_mutation_labels: If True, show labels for mutations.
         :param bool show_internal_node_labels: If True, show labels for internal nodes.
         :param bool show_leaf_node_labels: If True, show labels for leaf nodes.
+        :param list branch_colours: A list giving colours for edges in the tree.
+            If this is an array such as ['red','blue'], the indexes of the array
+            correspond to the node whose edges are coloured. If a dict, the keys
+            give the node numbers for each colour. Unspecified edges are coloured
```
jeromekelleher: This is over-complicated I think. It's sufficient to specify that the ``branch_colours`` is a dictionary, indexed by the younger node. Any use cases for the array input can be trivially remapped to a dict. The consensus colour thing seems far too complicated to me.

msprime/trees.py
```diff
@@ -637,7 +663,7 @@ def get_sample_size(self):
     def draw(
             self, path=None, width=200, height=200, show_times=False,
             show_mutation_labels=False, show_internal_node_labels=True,
-            show_leaf_node_labels=True):
+            show_leaf_node_labels=True, node_colours=[]):
```
jeromekelleher: ``branch_colours`` is a better name here, as ``node_colours`` might be used to colour the nodes themselves at some point. Also a default value of None is safer here as it allows more freedom to expand the API with mutually exclusive options at some later point.

msprime/trees.py
```diff
@@ -289,11 +289,17 @@ def __init__(
         self._tree = tree
         self._x_coords = {}
         self._y_coords = {}
+        try:
+            self._colour = {node:col for node,col in node_colours.keys()}
+        except:
```
jeromekelleher: Blanket swallowing all exceptions is a very bad idea. Something catastrophic could have happened in this block which would be silently passed over.

msprime/trees.py
```diff
@@ -289,11 +289,17 @@ def __init__(
         self._tree = tree
         self._x_coords = {}
         self._y_coords = {}
+        try:
+            self._colour = {node:col for node,col in node_colours.keys()}
```
jeromekelleher: It's simplest and most consistent if we use ``self._branch_colours`` as the instance variable here. Then we just say
```
self._branch_colours = {}
if branch_colours is not None:
    self._branch_colours = node_colours
```

msprime/trees.py
```diff
@@ -289,11 +289,17 @@ def __init__(
...
+        except:
+            self._colour = {node:col for node,col in enumerate(node_colours)}
+            
         for u in tree.nodes():
             scaled_t = tree.get_time(u) * self._y_scale
             self._y_coords[u] = height - scaled_t - y_padding
         self._leaf_x = 1
         self._assign_x_coordinates(self._tree.get_root())
+        self._assign_colours(self._tree.get_root())
```
jeromekelleher: This should no longer be needed, right?

jeromekelleher:
The circleCI tests failed here because of pep8 violations:
```
flake8 --max-line-length 89 setup.py msprime tests
msprime/trees.py:293:33: E231 missing whitespace after ':'
msprime/trees.py:293:46: E231 missing whitespace after ','
msprime/trees.py:295:33: E231 missing whitespace after ':'
msprime/trees.py:295:46: E231 missing whitespace after ','
msprime/trees.py:296:1: W293 blank line contains whitespace
msprime/trees.py:396:32: E225 missing whitespace around operator
msprime/trees.py:403:1: E302 expected 2 blank lines, found 1
msprime/trees.py:682:82: W291 trailing whitespace
```

I've found it useful to install a git commit hook at that runs this check before allowing me to commit. It saves time waiting for CI runs to finish and getting back failures for this sort of thing.

msprime/trees.py
```diff
@@ -371,6 +381,22 @@ def _assign_x_coordinates(self, node):
...
+        Assign colours up the tree, making parents the same
+        colour as their children, unless there is a disagreement
+        in which case no colour is set
+        """
+        if self._tree.is_internal(node):
+            children = self._tree.get_children(node)
+            for c in children:
+                self._assign_colours(c)
+            if node not in self._colour:
+                colours = set([self._colour.get(c) for c in children])
```
hyanwong: The assumption here is that not all nodes need to have a colour. So self._colour[c] may fail, in which case we simply treat it as colour=None, and it does not get a colour assigned. In the sag, that means it simply inherits the colour from the line group, rather than having an explicit override.

msprime/trees.py
hyanwong: Fine. Note that this also works for internal branches

msprime/trees.py
hyanwong: We need to percolate colours up the tree, otherwise we may never spot coloured samples if they are on a small branch. My description is probably overly complicated, though. It's simply that if all children share the same colour, their parent should get that colour too (I think we need this functionality). If they don't all share the colour, we simply drop any bespoke colours for the parent node. That means the parents will take the default SVG colour for branches, which I have currently set to "dimgrey".

Agree about colour specs. They are fed straight into the SVG text at the moment, so anything that is acceptable to SVG is acceptable here. I don't do any sanity checking, so there is potential for SVG hacking, but since this is all currently client side, I'm not too worried.

msprime/trees.py
```diff
@@ -343,8 +349,12 @@ def draw(self):
                     dy=dy))
             if v != NULL_NODE:
                 y = self._x_coords[v], self._y_coords[v]
-                lines.add(dwg.line(x, (x[0], y[1])))
-                lines.add(dwg.line((x[0], y[1]), y))
+                if self._colour.get(v):
```
hyanwong: OK - want to check for None too, so it would have to be `if v in self._colour and self._colour[v] is not None`

msprime/trees.py
jeromekelleher: > We need to percolate colours up the tree, 

This should be done by the client. If we just give the ability to colour branches, the client code can decide what to do. What you're suggesting is too specific to the current use-case. People might want to use this for all sorts of things, and by making this over-complicated and over-specific you exclude this. Keep it simple: an optional colour, per branch. Nothing else.

hyanwong: Hmm, but for most cases, clients won't know what the numbers / names of internal nodes are, and this changes for each tree in the tree sequence. So I can't see how this would be possible for a client to work out, TBH.

My need I think is the common one, which is to colour samples (tips) in such a way as to see whether they cluster. For this, you need to colour higher-level branches, without knowing beforehand what the tree shape is.

I've made the colouring more general, so that *if you do know a specific internal node*, you can choose to override the inherited colour. But I can't think that this would be used very often.

jeromekelleher: It's not hard. Suppose I want to colour the clade containing nodes 0 and 1:
```python
for t in ts.trees():
     branch_colours = {u: "red" for u in  t.nodes(t.mrca(0, 1))}
     t.draw(branch_colours=branch_colours)
```
There's no way we can predict all possible ways in which people might want to colour upwards form the leaves, and I **strongly** believe that it's a bad idea to try. This general interface allows the client full power to colour branches any way they want.

hyanwong: Well, if you feel strongly about it, happy to drop that functionality, although I think your solution to this use case should be documented. But I don't quite see how your solution works? What happens to all the intermediate internal nodes between 0 and 1?

More generally, if you can suggest an easy msprime way to calculate the nodes I want to colour, then fine. You have a better grasp of the msprime functions than me, so perhaps you could suggest? I guess you know what I want to do: take (say) nodes {1,2,5,6,9,10} and colour in such a way that we can see immediately whether they cluster. This requires working out which nodes give rise to only those tips, and colouring those with the same colour.

hyanwong: Actually, I see that we have `tree.leaves(u)`, so we could do something like:

```
colour_nodes = set([1,2,5,6,8,9])
branch_colours = {u: "red" if set(t.leaves(u)) <= colour_nodes for u in t.nodes())}
```

although I imagine that for large trees this might be unacceptably slow, since it re-calculates the leaf sets for each node.

jeromekelleher: > Well, if you feel strongly about it, happy to drop that functionality, although I think your solution to this use case should be documented. But I don't quite see how your solution works? What happens to all the intermediate internal nodes between 0 and 1?

The nodes iterator returns all nodes in the subtree u in preorder. So, all branches in the subtree will be coloured. Try it out.

> More generally, if you can suggest an easy msprime way to calculate the nodes I want to colour, then fine. You have a better grasp of the msprime functions than me, so perhaps you could suggest? I guess you know what I want to do: take (say) nodes {1,2,5,6,9,10} and colour in such a way that we can see immediately whether they cluster. This requires working out which nodes give rise to only those tips, and colouring those with the same colour.

You can colour them using precisely the same algorithm in client code. Please just remove this complexity, I'm not going to merge the patch otherwise.

We can make a cookbook of clever ways to colour trees in the documentation, and there will be myriad of different ways that people will want to do it.

hyanwong: > The nodes iterator returns all nodes in the subtree u in preorder. So, all branches in the subtree will be coloured. Try it out.

Oh, I see. Thanks.

> You can colour them using precisely the same algorithm in client code.

Had just noticed. Sorry.

> We can make a cookbook of clever ways to colour trees in the documentation

Yes, as long as we can remember how to do it, that seems fine.

hyanwong: p.s. where should I put this in the documentation? It doesn't feel quite right to me to insert it into the .py file. What's your preference?

jeromekelleher: Put the interface description into the Python source, same as the other arguments. There's no point in writing any other documentation yet --- we need to use this for a bit before it'll be clear what the best usage patterns are.

msprime/trees.py
```diff
@@ -289,17 +289,13 @@ def __init__(
         self._tree = tree
         self._x_coords = {}
         self._y_coords = {}
-        try:
-            self._branch_colours = {node: col for node, col in branch_colours.keys()}
-        except AttributeError:
-            self._branch_colours = {node: col for node, col in enumerate(branch_colours)}
-        
+        self._branch_colours = branch_colours or {}
```
jeromekelleher: Following pep8, I think it's much clearer to explicitly compare with ``None`` when doing this sort of thing. Please change to
```python
self._branch_colours = {}
if branch_colours is not None:
    self._branch_colours = branch_colours
```
Ditto for node colours.

msprime/trees.py
```diff
@@ -327,7 +323,10 @@ def draw(self):
         for u in self._tree.nodes():
             v = self._tree.get_parent(u)
             x = self._x_coords[u], self._y_coords[u]
-            dwg.add(dwg.circle(center=x, r=3))
+            if u in self._node_colours and self._node_colours[u] is not None:
```
jeromekelleher: Why are we comparing with ``None`` here? The input map is from nodes to colours.

msprime/trees.py
```diff
@@ -349,9 +348,9 @@ def draw(self):
                     dy=dy))
             if v != NULL_NODE:
                 y = self._x_coords[v], self._y_coords[v]
-                if v in self._branch_colours and self._branch_colours[v] is not None:
-                    lines.add(dwg.line(x, (x[0], y[1]), stroke=self._branch_colours[v]))
-                    lines.add(dwg.line((x[0], y[1]), y, stroke=self._branch_colours[v]))
+                if u in self._branch_colours and self._branch_colours[u] is not None:
```
jeromekelleher: Ditto here.

msprime/trees.py
```diff
@@ -676,12 +658,10 @@ def draw(
...
         :param bool show_internal_node_labels: If True, show labels for internal nodes.
         :param bool show_leaf_node_labels: If True, show labels for leaf nodes.
-        :param list branch_colours: A list giving colours for edges in the tree.
-            If this is an array such as ['red','blue'], the indexes of the array
-            correspond to the node whose edges are coloured. If a dict, the keys
-            give the node numbers for each colour. Unspecified edges are coloured  
-            according to the consensus colour of their children, or use the default
-            colour if there is disagreement.
+        :param list branch_colours: A dict giving colours for edges in the tree.
+            Keys correspond to node numbers in the tree.
```
jeromekelleher: Need to explain which node here (i.e., the child or parent node). Also, the param type is incorrect, should be 'dict'.

hyanwong: good points.

msprime/trees.py
hyanwong: Oh, yuck. Didn't know that about pep8 verbosity. Will change against my normal programming preference for short-circuiting.

msprime/trees.py
hyanwong: It seems not unreasonable to be able to specify that a branch or node should take the "default" colour. I thought is would be useful if that could be done either way: by omitting that node from the dict, or by including all nodes, but setting some to a value of None.

jeromekelleher: I disagree, this is pointless complexity. Please change.

hyanwong: Thought you'd say that.

hyanwong: There was another reason, when I was percolating colours (or absence of them) through nodes, and it was useful to have a flag to mark a default colour setting. But more generally, I dislike the code bombing out when I do something I would expect to work. As a user, I would hope that setting something to None would be covered in a sensible way (and indeed would expect it from an external library). But you seem to have a different approach, whereby the code drops out more often than I, as a user, would hope. Either way it's your codebase, though. Just a user's perspective.

jeromekelleher: No, I fundamentally disagree with this. If a user makes a mistake, then the library should NOT try to infer what they meant by this mistake and should not try to accommodate it. Libraries should fail early and fail noisily so that the user can fix their code. Simplicity of interface is absolutely key to simple, intuitive and reliable software.

*Please* change this.

There are still no tests, and CI is still failing.

PR #213, opened 2017-07-17
Document requirements for valid tables

petrelharp:
I've made a start at this - not in correct .rst format, apologies - but wanted to write the algorithmic requirements down to make sure I'm not assuming untrue things.  Is this the right place to write these requirements down?


petrelharp:
In particular, I want to check about
```
def EdgesetTable ...
    It is not required that all records corresponding to the same parent be
    adjacent in the table.
```

msprime/trees.py
```diff
@@ -159,6 +177,34 @@ def __str__(self):
...
+
+    Requirements: to describe a valid tree sequence, a `EdgesetTable` (and
+    corresponding `NodeTable`, to provide birth times) must satisfy:
+        1. each list of children must be in sorted order,
+        2. any two edgesets that share a child must be nonoverlapping, and
+        3. the birth times of the `parent` in an edgeset must be strictly
+            greater than the birth times of the `children` in that edgeset.
+    Furthermore, for algorithmic requirements
+        4. the the table must be sorted by birth time of the `parent`, and
+        5. any two edgesets corresponding to the same `parent` must be nonoverlapping.
```
jeromekelleher: Is this an algorithmic requirement? If I have 
```
0  1   1    3,5
0  1   1    3,4
```
then these are contradictory. This must be true in general for any edgeset with a given parent and overlapping intervals I think.

msprime/trees.py
```diff
@@ -159,6 +177,34 @@ def __str__(self):
...
+    the `parent` and the `children` on that interval.
+
+    Requirements: to describe a valid tree sequence, a `EdgesetTable` (and
+    corresponding `NodeTable`, to provide birth times) must satisfy:
+        1. each list of children must be in sorted order,
+        2. any two edgesets that share a child must be nonoverlapping, and
+        3. the birth times of the `parent` in an edgeset must be strictly
+            greater than the birth times of the `children` in that edgeset.
+    Furthermore, for algorithmic requirements
+        4. the the table must be sorted by birth time of the `parent`, and
```
jeromekelleher: Also, the leftmost coordinate must be 0.

msprime/trees.py
```diff
@@ -159,6 +177,34 @@ def __str__(self):
...
+        5. any two edgesets corresponding to the same `parent` must be nonoverlapping.
+    It is an additional requirement that the complete ancestry of each sample
+    must be specified, but this is harder to verify.
+
+    It is not required that all records corresponding to the same parent be
```
jeromekelleher: This is true I think. The requirement is that edgesets are listed in nondecreasing order of parent node time. Thus, if we have nodes with equal times, then these can be arbitrarily ordered.

jeromekelleher:
Thanks @petrelharp, it is important that we get this stuff properly documented soon. I think the right thing to do is to generalise the current "Tree sequence file format" page, to contain a more helpful discussion about tree sequence IO in general. We could call this new page "Tree Sequence Formats", and it might look like
```
Tree Sequence Formats
    - Definitions
    - Text format
    - Tables API
    - HDF5 format
```

I think this would be a helpful flow, because we can give definitions at the top, and then have some immediate examples in the text format (which should be easy to understand). Once the text format is understood, then the Tables API will be easy to follow. The HDF5 format just takes this and formalises it into a HDF5 hierarchy.

If we gather all the examples and definitions together here, then we can keep the API documentation for the various methods short and simple, and provide links to this page for details and examples.

What do you think?

msprime/trees.py
petrelharp: Your example there is disallowed by (2.) No multiple inheritance.  The reason I'm including (5.) and calling it an algorithmic requirement is because 

        0.0     1.0     4       0,
        0.0     2.0     4       1,

is not contradictory but is disallowed.

msprime/trees.py
petrelharp: I ran some quick non-exhaustive checks and did not come across any errors, so I think it is not currently a requirement.

Hm.  I agree, this would be useful!  And, "gathering all edges for a given parent" is something that is not yet efficiently implemented, right?

I was thinking that not having this requirement would make the forwards-time algorithm easier, since we could just push every new individual's record onto the output stack, but in fact I was wrong: with overlapping generations we'll still need to sort on parent birth time.

I am already advocating for having two sets of requirements, "strict" and "relaxed", with "relaxed" tree sequences not needing to satisfy the last requirement, but with `simplify` and `simulate` guaranteed to output strict tree sequences, but methods other than `simplify` not guaranteed to work on relaxed ones.  So one option would be to add more requirements to `strict` ones - I think that edgesets are already left-to-right ordered in output, no?

So, how about this: there will need to be a method that orders edgesets appropriately: by parent birth time at least; possibly by left endpoint as well.  The only question is whether that is the user's responsibility or is the first step in simplify().  Either is fine, although it seems nicer and cleaner to put it in simplify, as otherwise we'll have to check if its true anyhow?  What do you think?

petrelharp:
Sounds good -- I made a small start at this, but changing "Text format" to "Working with tables" (feel free to change).  I can work more on this.

Something that should be put in, that I am not certain on: I am guessing that working with numpy arrays and using the `.set_columns` method would be much faster than lots of `.add_rows` - is this right?  Or do you want to not encourage using `.set_columns`?

msprime/trees.py
jeromekelleher: I see, fair enough. Just out of curiosity, how would you interpret these edgesets? My take would be that the second contradicts the first, since we only allow one parent->children mapping on a given interval. Do you see these as being additive, so that these could also be interpreted as 
```
 0.0     1.0     4       0,1
 1.0     2.0     4       1,
```

msprime/trees.py
jeromekelleher: > I am already advocating for having two sets of requirements, "strict" and "relaxed", with "relaxed" tree sequences not needing to satisfy the last requirement, but with simplify and simulate guaranteed to output strict tree sequences, but methods other than simplify not guaranteed to work on relaxed ones. So one option would be to add more requirements to strict ones - I think that edgesets are already left-to-right ordered in output, no?

I see the attraction of having strict and relaxed tree sequences, but I worry about the consequences. Not guaranteed to work is easy to say, but I don't think that's enough. We would have to say, "guaranteed to fail predictably with a meaningful error". Just saying the behaviour is undefined definitely isn't good enough, and if we write efficient algorithms in C, then we need to be able to lean heavily on assumptions about the input. If we have to test that these assumptions are true before we run the algorithm, then this could be a heavy cost that we end up paying over and over again. Also, for every algorithm, we'd have to think, what are the possible input variations here, and how would they possibly end up leading to (e.g.) memory violations and so on. It just seems error prone to me to allow for needless variation.

I also like the idea of canonicalisation. When you have a complex combinatorial object, then I think it makes sense to remove as much symmetry as possible from it. Pick a single representation that has useful properties, and stick with it. This allows for meaningful definitions of equality, as well as the algorithmic considerations above.

> So, how about this: there will need to be a method that orders edgesets appropriately: by parent birth time at least; possibly by left endpoint as well. The only question is whether that is the user's responsibility or is the first step in simplify(). Either is fine, although it seems nicer and cleaner to put it in simplify, as otherwise we'll have to check if its true anyhow? What do you think?

Well, if we insist that all edgesets for a given parent are adjacent and they are ordered left-to-right within a parent, and edgesets parents are ordered by non-decreasing time (which we can check cheaply on import) then this first part of simplify is done. It's just:
```python
edgesets = ts.edgesets()
parent_edgesets = [next(edgesets)]
last_parent = parent_edgesets[0].parent
for edgeset in edgesets:
    if edgeset.parent != last_parent:
          yield parent_edgesets
          parent_edgesets = []
          last_parent = edgeset.parent
     parent_edgesets.append(edgeset)
yield parent_edgesets
```
The ``parent_edgesets`` list can be transformed directly into a segment chain, since the edgesets are guaranteed to be in left-to-right order. Therefore, we can implement simplify with no sorting **at all** (well, except for sorting children within an edgeset, but that's very cheap).

This seems to me like a very useful property to be able to provide, and I'm sure this will be useful in other algorithms besides simplify. Plus, we could also add a ``parent`` argument to the edgesets iterator which returns all edgesets for a given parent. This could either be implemented by skimming forward until we reach the parent in question, or by keeping an array of offsets to the start of the edgesets for each parent (a modest extra memory cost).

I know it's awkward for users doing the required sorting to get the input in the required form, but perhaps we could provide method to do this? I.e., given node, edgesets, site and mutation tables, if we run ``canonicalise`` on them, they do all the required sorting. That way, if you know that your tables have the required form then you can avoid this cost. If you are doing something awkward then you don't have to worry about what the precise requirements are and just fill your tables as suits your application. Running ``canonicalise`` will sort these tables in the correct orders, and they can then be imported into a valid tree  sequence.

jeromekelleher:
> Something that should be put in, that I am not certain on: I am guessing that working with numpy arrays and using the .set_columns method would be much faster than lots of .add_rows - is this right? Or do you want to not encourage using .set_columns?

I see this as being a more advanced usage of the tables API. The page starts at the simplest approach, moves into more efficient/advanced. This progression to me is, text -> tables API, adding row-by-row -> tables API using set_columns -> HDF5.

Using ``set_columns`` is something that will be documented, and is quite efficient. It's not a simple thing to use though (have to have your numpy arrays in the right type, ragged arrays must be flattened, character strings must be converted to ints), so it's definitely a more advanced usage then ``add_row``, which takes care of a lot of tedious details for you.

If you are sequentially generating the records in Python, there won't be much difference between using ``add_row`` and ``set_columns``. Where ``set_columns`` really wins is when some other low-level component has the information in a bunch of binary arrays, and you want to create a tree sequence from them. Them, ``set_columns`` is really good.

msprime/trees.py
petrelharp: Proposal (rationale below):
 - require that edgeset tables be sorted by parent ID and left endpoint to be loaded into a tree sequence
 - provide a method to sort edgeset tables
This avoids having two types of tree sequence, and only requires checking for proper order at the time the tables are loaded.

Rationale: thinking again, I wasn't actually wanting less strict tree sequences, just less strict edgeset tables.  So: should provide a method to sort tables (just edgeset tables:  "sorting" a node table requires re-indexing the edgeset table - no need to implement this unless we actually have a use case) and assume sortedness in tree sequences.

I set about to actually do this, and realized that doing it non-hackily requires either:
 - a way to reorder the .children array, or
 - a way to pull out specific edgesets from a table
I then set out to implement the latter, as it seemed like simple copying over of tree_sequence_get_edgeset and related things... but eventually discovered it was not so straightforward (e.g. a tree sequence's edgeset is not a edgeset_table_t ...).

So, here's a hacky implementation.  I'd put it in the code if it wasn't so terrible and didn't really belong in this pull request:
```
def sort_edgesets(edgesets, nodes):
    """
    Returns a copy of the EdgesetTable sorted by parent ID (and hence, by
    birth time of parent), and for each parent, sorted by left endpoint.
    """
    out = msprime.EdgesetTable()
    temp_ts = msprime.load_tables(nodes=nodes, edgesets=edgesets)
    for (u, l, k) in sorted(zip(edgesets.parent, edgesets.left, 
                                range(edgesets.num_rows))):
        # out.add_row(*edgesets.get_edgeset(k))
        out.add_row(*temp_ts._ll_tree_sequence.get_edgeset(k))
    return out
```
If we had a `get_edgeset(k)` method for edgeset tables, this could be an `EdgesetTable` method and wouldn't need the `nodes` argument.

This has the additional property that edgesets appear in the order the parents appear in the node table - so, in order by time + parent id?  (which is the same as by parent id)  I'm not sure if this would actually be useful for anything, since parent ID already gives the index, but it's something that can be checked internally from only the edgeset without having to consult the NodeTable, so maybe this should be the requirement.

petrelharp:
I've added something for sites and mutations, but i am probably missing some requirements on these tables.

msprime/trees.py
jeromekelleher: This sounds great to me @petrelharp, I like it a lot. I think we get the best of both worlds here: users can fill the edgeset table in whatever order they like, and we can take advantage of nice sortedness properties.

So, the question is, should we provide a similar method for sites/mutations (I'm thinking about having mutation nodes listed in pre-order again so we can do useful things with them)? Or bundle them all together into a `sort_tables` method, so that we allow for other things in the future as we think of them?

I'm going to open another issue on this so we can keep track of it a bit better.

jeromekelleher:
This is great stuff @petrelharp, thanks so much for doing it. We can merge whenever you like, just let me know when you think it's ready.

msprime/trees.py
petrelharp: a single method, I'd say.  especially as some sorts of sortedness requires comparing between tables.

petrelharp:
I'm good to merge this as long as we remember to update it with the resolution of #214.

petrelharp:
OK, I'm good to merge this.

jeromekelleher:
Merging. Thanks @petrelharp!

PR #215, opened 2017-07-22
sort_tables implementation

### LINKED ISSUE #214: Add ``sort_tables`` function
As discussed in the [comments](https://github.com/jeromekelleher/msprime/pull/213#discussion_r127903251) for PR #123, it would be very useful to have a function that would take a set of tables describing a valid (but non-canonical) tree sequence, and sort the tables such that it is canonical.

Issues:
1. Should we sort the nodes, such that samples are 0...n and we order nodes by time after that? (I think probably not because of complications with ancestral samples and so on, but it's worth thinking about.)
2. We sort sites by increasing position (duplicate sites are not allowed)
3. Do we sort mutation nodes by some topological ordering? I like the idea of pre-order, as it's what naturally happens if you do a Markov generation process down the tree. It would be easy enough to implement here: we just make a tree sequence out of the ``nodes`` and ``edgesets`` tables and then do the required tree traversals for the sites where we have > 1 mutation.

jeromekelleher:
This is the bare-bones implementation of sort_tables. It just sorts edgesets for now, but sorting sites is a simple addition. 

TODO:
- ~~Need to add input checking (nodes and edgesets cannot be null) + low level tests of API.~~
- ~~Add Python API~~
- ~~High level tests to check lots of permutations of the input nodes to ensure sorting works OK.~~

Closes #214. 

cc. @petrelharp

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3246,7 +3246,7 @@ TreeSequence_load_tables(TreeSequence *self, PyObject *args, PyObject *kwds)
         }
     }
     if ((mutations == NULL) != (sites == NULL)) {
-        PyErr_SetString(PyExc_TypeError, "Must specify both mutations and mutation types");
+        PyErr_SetString(PyExc_TypeError, "Must specify both site and mutation tables");
         goto out;
     }
     err = tree_sequence_load_tables_tmp(self->tree_sequence,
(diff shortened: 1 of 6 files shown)


jeromekelleher:
OK, I think this is ready to merge. There are two things that I haven't done:

1. Sort the mutations within a given site. This is an ongoing discussion, and we can update ``sort_tables`` when we know what the right thing to do is.
2. Anything with migrations. These are very much second class citizens at the moment, and the correct sorting strategy will be part of finishing the implementation.

PR #218, opened 2017-07-26
Implement generalized statistics computation with mutations

petrelharp:
To do here:

1. move the `branch_stats` stuff to an object, like LD calculator
2. switch the branch stats machinery over from taking a single argument (number of nodes in the leaf set below the branch) to two (number below and number not below)
3. implement all the same stuff in terms of mutations, as another class

The second thing will have the advantage that the functions will be more re-usable: currently we have to define a custom function each time that knows about the number of nodes in the leaf set; this way it will not.  It may be a bit more awkward as we need to pass a list of tuples, though.

Another note: currently we have eg a method, `ts.f4` that does *not* compute the f4 statistic - it computes the branch length version of it.  Were it to stay under TreeSequence we should rename this `branch_length_f4` or something; but this way - as the branch length version and the mutation version will be under separate classes - both version can be called just `f4`.


petrelharp:
Also to-do: implement covariance for both.

petrelharp:
I am changing "branch stats" to "tree stat" generally.  Hopefully this is a better name.

ashander:
I've rebased this on current master.  It'd gotten out of sync with the change from leafs -> samples. 
Hopefully it'll be mergable soon and not fall behind again!

Todo: 

- [ ] update leafs -> samples in SiteStatCalc https://github.com/jeromekelleher/msprime/pull/218/files#diff-71c56ecf7eccce3069862f1392a93974R718 (and in tests?)

jeromekelleher:
Sorry, I'd completely forgotten about this. I think we should try and merge this as soon as we can before it gets clobbered by the edgeset->edge change (which is big in terms of LOC affected).

This all looks good to me, I think we just need to bump up the test coverage a little bit. See 

https://codecov.io/gh/jeromekelleher/msprime/pull/218/src/msprime/stats.py?before=msprime/stats.py

for the current coverage on a line-by-line basis. The main thing is to get some coverage on the SiteStatCalculator I think.

ashander:
Thanks! Turns out SiteStatCalculator is a bit of a placeholder for to-be-implemented stuff. We've moved that to a different branch and removed it here. So coverage should be good to go

ashander:
Alright so this is now good. 

There's a conflict with Jerome's WIP on #267 but I just checked and happily it's just a single merge conflict in one commit --- in 09a13c1... Updated various topology tests. So if we merge this it should be a very quick rebase

petrelharp:
Awesome.  Thanks, @ashander.  Must get to the SiteStatCalculator soonish, though.

jeromekelleher:
Excellent, thanks @ashander and @petrelharp. Merging.

PR #222, opened 2017-08-04
Add option to output vcf in compressed format

DomNelson:
Requested by a member of my lab, and seems it would be generally useful for people who need VCF output as part of a pipeline.

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2853,7 +2853,7 @@ class TreeSequence(object):
                 u for u in samples if self.get_population(u) == population_id]
         return samples
 
-    def write_vcf(self, output, ploidy=1, contig_id="1"):
+    def write_vcf(self, output, ploidy=1, contig_id="1", compresslevel=None):
         """
         Writes a VCF formatted file to the specified file-like object. If a
         ploidy value is supplied, allele values are combined among adjacent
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Thanks for the update @DomNelson, it looks good. However, I'm not sure there's much point in adding this option since it can easily be reproduced by the client, e.g.:

```python
import gzip
with gzip.open("output.vcf.gz", "wb") as vcf_file:
    tree_sequence.write_vcf(vcf_file, 2)
```

The advantage then is that the user has full flexibility on how they use the compression levels and so on, and we don't have to complicate the interface.

However, I think that many users might not realise that this option is available to them, and so we should add this snippet to the docs for the method.

Also, I think it would be worth adding a snippet showing how to pipe the data into bcftools so that the data can be encoded directly as bcf. Something like:
```python
import os
import subprocess

read_fd, write_fd = os.pipe()
write_pipe = os.fdopen(write_fd, "w")
with open("output.bcf", "w") as output:
    proc = subprocess.Popen(
        ["bcftools", "view", "-O", "b"], stdin=read_fd, stdout=output)
    ts.write_vcf(write_pipe)
write_pipe.close()
os.close(read_fd)
proc.wait()
if proc.returncode != 0:
    raise RuntimeError("bcftools failed with state:", proc.returncode)
```

I haven't tested this at scale, but I think it should work. This is a useful addition to the docs I think, since people will often want to write to BCF without bothering with the VCF. I'd rather not add BCF generation as an actual function of the library, as this adds a dependency on bcftools, which would cause all sorts of headaches.

What do you think @DomNelson, should we add this information about compressed VCF to the docs for the ``write_vcf`` method?

jeromekelleher:
@DomNelson --- should we follow this up, or would you prefer to close the PR?

DomNelson:
How about we close for now - I'll open a new one implementing your suggestions later

PR #223, opened 2017-08-10
Allow ancestral samples

petrelharp:
I've removed checks for internal samples and added some tests.

The only thing I've encountered so far that needs fixing is `leaf_lists`.  I've made a change that (hopefully) makes it so `leaves()` returns a list of all *samples* below (and, including) the node.  But, maybe this should have a different name, then?  Grrr.  What do you think, @jeromekelleher?

Still needs some more tests.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3964,7 +3964,7 @@ SparseTree_init(SparseTree *self, PyObject *args, PyObject *kwds)
     TreeSequence *tree_sequence = NULL;
     node_id_t *tracked_leaves = NULL;
     int flags = 0;
-    uint32_t j, n, num_tracked_leaves;
+    uint32_t j, num_tracked_leaves, num_nodes;
     PyObject *item;
 
     self->sparse_tree = NULL;
@@ -3978,7 +3978,7 @@ SparseTree_init(SparseTree *self, PyObject *args, PyObject *kwds)
     if (TreeSequence_check_tree_sequence(tree_sequence) != 0) {
         goto out;
     }
-    n = tree_sequence_get_sample_size(tree_sequence->tree_sequence);
+	num_nodes = tree_sequence_get_num_nodes(tree_sequence->tree_sequence);
     num_tracked_leaves = 0;
     if (py_tracked_leaves != NULL) {
         if (!flags & MSP_LEAF_COUNTS) {
@@ -4000,8 +4000,8 @@ SparseTree_init(SparseTree *self, PyObject *args, PyObject *kwds)
             goto out;
         }
         tracked_leaves[j] = (node_id_t) PyLong_AsLong(item);
-        if (tracked_leaves[j] >= n) {
-            PyErr_SetString(PyExc_ValueError, "leaves must be < sample_size");
+        if (tracked_leaves[j] >= num_nodes) {
+            PyErr_SetString(PyExc_ValueError, "leaves must be valid nodes");
             goto out;
         }
     }
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3978,7 +3978,7 @@ SparseTree_init(SparseTree *self, PyObject *args, PyObject *kwds)
     if (TreeSequence_check_tree_sequence(tree_sequence) != 0) {
         goto out;
     }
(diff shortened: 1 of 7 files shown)


petrelharp:
I found one more errant check (in initializing tracked_leaves), and added some python tests.  If we're OK with the "leaf"-related functions actually referring to samples, then I think perhaps this is ready to merge.

Note that `SparseTree.is_internal(u)` is currently equal to `not Sparse_tree.is_leaf(u)`, contradicting my interpretation of "leaves" above.

_msprimemodule.c
```diff
@@ -3978,7 +3978,7 @@ SparseTree_init(SparseTree *self, PyObject *args, PyObject *kwds)
     if (TreeSequence_check_tree_sequence(tree_sequence) != 0) {
         goto out;
     }
-    n = tree_sequence_get_sample_size(tree_sequence->tree_sequence);
+	num_nodes = tree_sequence_get_num_nodes(tree_sequence->tree_sequence);
```
jeromekelleher: Indentation is off here

tests/test_topology.py
```diff
@@ -106,6 +106,48 @@ def assert_variants_equal(self, ts1, ts2):
         v2 = list(ts2.variants(as_bytes=True))
         self.assertEqual(v1, v2)
 
+    def check_num_leaves(self, ts, x):
+        """
+        Compare against x, a list of tuples of the form
+            (tree number, parent, number of leaves)
```
jeromekelleher: Minor nitpick here, but I don't think multiline docstrings should be indented like this. It may even break downstream processing tools (like sphinx) since this indentation is significant. See https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation

Also (less importantly), I think pep8 says that multiline block comments shouldn't be indented either:

> Comments generally apply to some (or all) code that follows them, and are indented to the same level as that code. Each line of a block comment starts with a # and a single space (unless it is indented text inside the comment).

< /nitpick >

tests/test_topology.py
```diff
@@ -106,6 +106,48 @@ def assert_variants_equal(self, ts1, ts2):
...
+            while k < j:
+                t = next(tss)
+                k += 1
+            self.assertEqual(nl, t.num_leaves(node))
+
+    def check_num_tracked_leaves(self, ts, tracked_leaves, x):
+        k = 0
+        tss = ts.trees(leaf_counts=True, tracked_leaves=tracked_leaves)
+        t = next(tss)
+        print(tracked_leaves)
```
jeromekelleher: We should remove all debug prints before merging.

jeromekelleher:
On the one hand I'm delighted that we could make such a fundamental generalisation with so few changes. This really is great!

On the other hand, I'm slightly dismayed by the terminological mess that I've gotten us into. I think that we have to try to sort this out now, or it'll only get worse in the future. We can't go on with having a leaf mean "sample, which might not be an external node". I think the way to proceed is to systematically change everything that is not user visible (e.g. C code, internal Python code) to change things like ``leaf_lists`` to ``sample_lists``, ``tracked_leaves`` to ``tracked_samples`` etc. At the API level then we need to be careful to make sure we don't break anyone's code. I'm guessing we can do something like
```python
tss = ts.trees(sample_counts=True, tracked_samples=tracked_samples)
```
as the recommended and documented interface, but keep ``tracked_leaves`` and ``leaf_counts`` as deprecated aliases for these. This way, we won't break code.

However, this _is_ a lot of work and will be a slog to do. Because of the size of the diff, I would rather do this after  #209 has been merged. I'm happy to merge this now, and pick up the terminology refactor afterwards, or we can leave this branch as is for a while if you prefer. If you would like to try out this code on your application @petrelharp, we may as well merge now I think.

petrelharp:
Thanks for the nitpicking, and sorry I missed those!

Re: leaves - I keep trying to convince myself that "leaf" could be a good alternative name for "sample", but no, not really.  So, that sounds good - let's merge it now.  I've started an issue for this (#225).  I think we can keep `is_leaf`, but may want to add `is_sample` to `SparseTree`.

jeromekelleher:
Merged; thanks @petrelharp. Will discuss terminology over in #225.

Issue #230, opened 2017-08-14
unused node (or something) causes bad trees


petrelharp:
Here is a MWE.  Reading this in and simplifying causes it to forget what happens to input node 11 past 0.9; for some reason, removing unused node 3 fixes this.
```
import msprime
import six

n = """\
id    is_sample    population    time
0     0    0        6.0
1     0    0        2.0
2     0    0        2.0
3     0    0        2.0
4     0    0        2.0
5     0    0        1.0
6     0    0        1.0
7     0    0        1.0
8     1    0        0.0
9     1    0        0.0
10    1    0        0.0
11    1    0        0.0
12    1    0        0.0
"""

e = """\
id    left   right  parent  children
0     0.0    0.8    5       9
1     0.3    1.0    5       10
2     0.0    1.0    6       8
3     0.0    0.3    6       10
5     0.0    0.9    7       11
6     0.0    1.0    7       12
7     0.8    1.0    7       9
4     0.9    1.0    1       11
9     0.4    1.0    1       6
11    0.0    0.4    4       6
12    0.0    1.0    4       7
8     0.0    1.0    0       1,2,4,5
"""

ts = msprime.load_text(six.StringIO(n), six.StringIO(e))

sts = ts.simplify()
stst = sts.trees()
for t in stst:
    print(t.parent(3))
```

The true trees should be
```
0.0 to 0.3:
    {12:7, 11:7, 10:6, 9:5, 8:6, 7:4, 6:4, 5:0, 4:0, 2:0, 1:0}
0.3 to 0.4:
    {12:7, 11:7, 10:5, 9:5, 8:6, 7:4, 6:4, 5:0, 4:0, 2:0, 1:0}
0.4 to 0.8:
    {12:7, 11:7, 10:5, 9:5, 8:6, 7:4, 6:1, 5:0, 4:0, 2:0, 1:0}
0.8 to 0.9:
    {12:7, 11:7, 10:5, 9:7, 8:6, 7:4, 6:1, 5:0, 4:0, 2:0, 1:0}
0.9 to 1.0:
    {12:7, 11:1, 10:5, 9:7, 8:6, 7:4, 6:1, 5:0, 4:0, 2:0, 1:0}
```

jeromekelleher:
Superb, thanks @petrelharp. I'll sort this out ASAP.

jeromekelleher:
[...]

PR #231, opened 2017-08-14
test with WF simulator

petrelharp:
There is currently a bug (#230) - as far as I can tell, in the simplify machinery, although maybe I'm being dense.


jeromekelleher:
Looks good, thanks @petrelharp.

Given that the simulator is so simple, I think it's probably best to just put all this into one file --- say, ``tests/test_wright_fisher.py``? Otherwise there's a lot of chasing references around to try and find out where the code actually is. (This contradicts what I said before, I know, but I had imagined it would need a lot more code and be more complicated.)

petrelharp:
I've rebased this to master and moved the machinery.

This will have ancestral samples, thanks to overlapping generations.  I've bumped up the survival probability for some of the tests to 0.9 to ensure this happens a lot.

jeromekelleher:
Great, thanks @petrelharp. As you're off for a while, shall I take it from here and see if I can get to the bottom of these ancestral samples bugs?

petrelharp:
That'd be wonderful.  Also note the `simplify_tables` test that fails for mysterious (to me) reasons.

jeromekelleher:
A note on progress here. The tests are passing here after a tweak to the simplify algorithm, but I'm not happy that it's fully correct, as we're leaking segments (which is nearly always a sign of a bug in these algorithms). Unless overlapping generations is on the critical path for anyone, I'd like to spend a bit more time and try to fully iron out the details here. If someone needs this code ASAP we can merge now and work on the algorithm in another PR.

jeromekelleher:
Progress here. While I'm still not happy with the underlying algorithm here, I can't find any cases where I can break it. Therefore, in the interest of moving on and making progress I think we should merge this much and try it out with various inputs. Hopefully the assertions will catch things if anything weird happens.

What do you think @petrelharp? Basically I've tweaked the simplify algorithm a bit, and done more testing with your WF framework. I've disabled mutation generation, as I didn't have time to go through it properly and write tests for it. I think we could add this as a TODO issue (write tests for WF simplify with mutations). As for the algorithm, hopefully it's correct now and it's just not handling ancestral samples very elegantly. I'm sure this inelegance will become clear when we start writing it up and hopefully a solution will be found.

jeromekelleher:
Merging this as it's a substantial improvement over what's currently in master.

PR #232, opened 2017-08-14
Adds checks for bad trees to iterator.

jeromekelleher:
Some topological properties can only be checked at run time while we are generating the trees. This adds support for these checks.

I _think_ this should be comprehensive, but I'd like to get a second opinion here. Would you mind taking a look over this and reviewing please @petrelharp? I'm not at all sure about the terminology I've chosen here to describe the different errors, so please suggest something else if you can improve them.

There were a couple of incidental changes here also:

1. I added a check to make sure num_samples > 1. There's no deep reason for this, but I got some segfaults when I accidentally put in < 2 samples in an example. If we relax this assumption we need to track down these problems.

2. Removed an unnecessary array and sort left over from older checks.


jeromekelleher:
Excellent, thanks @petrelharp. Merging.

PR #233, opened 2017-08-14
Allows overlapping edgesets in simplify

### LINKED ISSUE #230: unused node (or something) causes bad trees

jeromekelleher:
Closes #230; cc @petrelharp.

This fixes the bug given in #230. The underlying cause here was that I had assumed that the input number of edgesets would always be <= the output number. However, with overlapping edgesets this isn't the case.

However, investigating has revealed a more subtle algorithmic bug. When we have situations like 
```  
        id      is_sample   time
        0       1           0
        1       1           0
        2       0           1
     
        0       2       2       0
        1       3       2       1
```
We're not outputting the correct tree sequence in either the Python of C version. Currently looking into this, and will hopefully fix and merge soon.


lib/msprime.h
```diff
@@ -666,6 +666,10 @@ typedef struct {
     node_table_t input_nodes;
     size_t *node_name_offset;
     size_t num_input_sites;
+    /* Keep a copy of the input edgesets for simplicity. We cannot modify the
+     * edgeset table in place because we may have more output edgesets than input.
```
petrelharp: Oh, right.  Yes, it's true. darn!

jeromekelleher:
Ah, so some odd stuff if coming out of the woodwork here! The following took me by surprise:
```python
        nodes = """\
        id      is_sample   time
        0       1           0
        1       1           0
        2       0           1
        """
        edgesets_before = """\
        0       2       2       0
        1       3       2       1
        """
        # We resolve the overlapping edgesets here. Since the flanking regions
        # have no interesting edges, these are left out of the output.
        edgesets_after = """\
        1       2       2       0,1
        """
```
So, it turns out that we can produce invalid tree sequences from inputs that are valid. This doesn't matter for ``simplify_tables``, but it is a good reason to remove the requirement that the leftmost coordinate is zero at some point.

There is still a bug in here, which occurs when we have overlapping unary edgesets and an internal sample. I think this is probably a symptom of not quite handling the internal samples case correctly in simplify, so hopefully we'll be able to find some more examples of this using the WF simulator and tackle it properly.

So, unless you object I'd like to merge this much @petrelharp. What do you think?

petrelharp:
I'm OK merging this; we'll address any more things popping out of the
woodwork separately.  (I'll hook up the WF sims to include internal samples
to help find those.)

PR #246, opened 2017-08-25
Sample terminology

### LINKED ISSUE #225: "tracked leaves" should be "tracked samples"
As mentioned in #223, to change the name, keeping these as deprecated aliases:

- `tracked_leaves`
- `leaf_counts`
- `leaf_lists`

To *not* change:

- `is_leaf` and `is_internal`

but perhaps we should add a `is_sample` function.

jeromekelleher:
Closes #225. Also adds much better support for internal samples in the test cases.

TODO: Some work to do with simplify on internal sample examples.


jeromekelleher:
Would you mind having a quick look over this @petrelharp or @ashander?  It's a bit change, and I've basically just done s/leaf/sample everywhere, so it would be great to get another pair of eyes on the diff.

lib/tests.c
```diff
@@ -2646,7 +2646,7 @@ test_simplest_unary_records(void)
 }
 
 static void
-test_simplest_non_sample_leaf_records(void)
+test_simplest_non_sample_sample_records(void)
```
petrelharp: I think this should still be as before, "test_simplest_non_sample_leaf_records"

lib/tests.c
```diff
@@ -7265,7 +7268,7 @@ main(int argc, char **argv)
         {"test_simplest_records", test_simplest_records},
         {"test_simplest_nonbinary_records", test_simplest_nonbinary_records},
         {"test_simplest_unary_records", test_simplest_unary_records},
-        {"test_simplest_non_sample_leaf_records", test_simplest_non_sample_leaf_records},
+        {"test_simplest_non_sample_sample_records", test_simplest_non_sample_sample_records},
```
petrelharp: also, here.

msprime/trees.py
```diff
@@ -684,7 +684,7 @@ def time(self, u):
     def get_time(self, u):
         """
         Returns the time of the specified node in generations. Returns 0 if u
-        is a leaf or is not a node in the current tree.
+        is a sample or is not a node in the current tree.
```
petrelharp: I think that this should read "Returns 0 if u is not a node in the current tree."

msprime/trees.py
```diff
@@ -669,7 +669,7 @@ def get_children(self, u):
         """
         Returns the children of the specified node as a tuple :math:`(v, w)`.
         For internal nodes, this tuple is always in sorted order such that
-        :math:`v < w`. If u is a leaf or is not a node in the current tree,
+        :math:`v < w`. If u is a sample or is not a node in the current tree,
```
petrelharp: I think this should read "if u is a leaf".

msprime/trees.py
```diff
@@ -697,7 +697,7 @@ def population(self, u):
 
     def get_population(self, u):
         """
-        Returns the population associated with the specified node. For leaf
+        Returns the population associated with the specified node. For sample
```
petrelharp: I am not sure this distinction needs to be made?  If not, could delete this sentence.

msprime/trees.py
```diff
@@ -2246,7 +2311,7 @@ def diffs(self):
         tuples, and these describe the records that must be applied to create
         the tree covering the current interval. These records are returned in
         time-increasing order, such that the records affecting the lowest parts
-        of the tree (i.e., closest to the leaves) are returned first.
+        of the tree (i.e., closest to the samples) are returned first.
```
petrelharp: how about instead of "closest to the leaves" say "closest to the present day"

petrelharp: or if not, "closest to the leaves"

_msprimemodule.c
```diff
@@ -122,9 +122,9 @@ typedef struct {
 typedef struct {
     PyObject_HEAD
     SparseTree *sparse_tree;
-    leaf_list_node_t *head;
-    leaf_list_node_t *tail;
-    leaf_list_node_t *next;
+    node_list_t *head;
+    node_list_t *tail;
+    node_list_t *next;
 } LeafListIterator;
```
petrelharp: By my reading this should be `SampleListIterator`, but this is never externally visible, so not a big deal.

tests/test_topology.py
```diff
@@ -1094,7 +1094,7 @@ def test_simplest_non_degenerate_case(self):
         self.assertEqual(t.time_dict, {0: 0, 1: 0, 2: 0, 3: 0, 4: 1, 5: 2})
 
     def test_two_reducable_trees(self):
-        # We have n = 4 and two trees, with some unary nodes and non-sample leaves
+        # We have n = 4 and two trees, with some unary nodes and non-sample samples
```
petrelharp: should be "non-sample leaves"

tests/test_topology.py
```diff
@@ -1105,7 +1105,7 @@ def test_two_reducable_trees(self):
         5       0           1
         6       0           2
         7       0           3
-        8       0           0   # Non sample leaf
+        8       0           0   # Non sample sample
```
petrelharp: Non sample leaf

tests/test_topology.py
```diff
@@ -1177,7 +1177,7 @@ def test_one_reducable_tree(self):
...
```
petrelharp: Again

tests/test_topology.py
```diff
@@ -1357,7 +1357,7 @@ def test_partial_non_sample_external_nodes(self):
         0       1           0
         1       1           0
         2       1           0
-        3       0           0.2  # Non sample leaf
+        3       0           0.2  # Non sample sample
```
petrelharp: Again

tests/test_topology.py
```diff
@@ -1415,7 +1415,7 @@ def test_partial_non_sample_external_nodes_2(self):
         4       0           0.5
         5       0           0.7
         6       0           1.0
-        7       0           0    # Non sample leaf
+        7       0           0    # Non sample sample
```
petrelharp: Again

tests/test_topology.py
```diff
@@ -1468,7 +1468,7 @@ def test_single_offspring_records(self):
         0   1           0
         1   1           0
         2   1           0
-        3   0           0       # Non sample leaf
+        3   0           0       # Non sample sample
```
petrelharp: Again

petrelharp:
Whew, looked through all that.  Now I need to go take a walk under the sampley branches of the trees outside...

msprime/trees.py
jeromekelleher: Well spotted. Opening an issue on this.

jeromekelleher:
Fixed all those. This was a huge help, thanks @petrelharp! I'm glad we've fixed this terminological mess, it would just have been horribly confusing to try and carry it forward.

PR #247, opened 2017-08-26
Documentation improvement

petrelharp:
In writing things up I've come upon some things that need fixing in the docs.  Here they are - there may be more.


petrelharp:
whoops - already got one of these =)

PR #258, opened 2017-09-11
ASCII trees

jeromekelleher:
This is a work in progress to render trees in ASCII. The current output looks like this:

```
        8              
    ----+---           
    |       |          
                       
    1       7          
        ----+--        
        |      |       
                       
        3      6       
            ---+--     
            |     |    
                       
            2     5    
                --+-   
                |   |  
                       
                0   4  
```

It turns out to be easy enough to do, and we can reuse the majority of the existing tree layout code.

I think the right thing to do here API wise though is to add a ``format="ascii"`` parameter to the existing ``tree.draw`` method. I guess we can just ignore any arguments that don't make sense for a particular rendering backend.

Any thoughts @petrelharp, @ashander?

### DIFF
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -1120,16 +1120,15 @@ class SparseTree(object):
                 yield stack.pop()
 
     def _inorder_traversal(self, u):
-        stack = [u]
-        k, j = NULL_NODE, NULL_NODE
-        while stack:
-            v = stack.pop()
-            if self.is_internal(v) and v != k and v != j:
-                children = self.get_children(v)
-                j = stack[-1] if stack else NULL_NODE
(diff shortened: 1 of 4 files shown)


petrelharp:
awesome.

it could be even fancier with these:
  https://en.wikipedia.org/wiki/Box-drawing_character

jeromekelleher:
> awesome.

> it could be even fancier with these:
>  https://en.wikipedia.org/wiki/Box-drawing_character

Cool, great idea. Should be easy to do. I guess we could use ``format="unicode"`` to use these symbols and ``format="ascii"`` to fall back on plain-ole-ascii for times when unicode is problematic.

ashander:
This is great!

PR #264, opened 2017-09-15
Added pickle support for tables.

### LINKED ISSUE #263: pickling tables
I noticed that the various table structures cannot be pickled.  An example is here, where I have added a failing unit test: https://github.com/molpopgen/msprime/commit/8ba795bd70cc34d805d70a2f6ef20c09a52dfbcd

I ran into this while attempting to use concurrent.futures to run forward simulations and return the simplified tables.  It did not work, and at least part of the reason is that the futures need to be instances of pickleable types.

There are several use cases for forward simulations where pickling would be desirable.

jeromekelleher:
Closes #263.

### DIFF
--- a/msprime/__init__.py
+++ b/msprime/__init__.py
@@ -31,3 +31,4 @@ from msprime.formats import *  # NOQA
 from msprime.trees import *  # NOQA
 from msprime.stats import *  # NOQA
 from msprime.exceptions import *  # NOQA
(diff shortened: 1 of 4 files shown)


jeromekelleher:
NB. This PR moves the tables functionality into a new ``tables.py`` file. Pinging @petrelharp and @molpopgen in case you are wondering where all the code went.

PR #267, opened 2017-09-22
Change edgesets -> edge

jeromekelleher:
It seems that the idea of an 'edgeset' was a premature denormalisation. With the greatly expanding generality of the tree sequences handled by msprime, it makes sense to fix this mistake before it becomes any more difficult.

This is going to result in rather a large mess I'm afraid, but the changes are mostly straightforward and result in cleaner code.

This should also lay the groundwork for more general tree sequences, handling multiple roots, zero edges and various other edgecases that we're not handling properly now.


jeromekelleher:
Some progress here. The tests should all now pass, as I've triaged all the outstanding issues. Main problems left:

1. Decide what the correct order for edges is now.
2. Implement a diffs() method to mimic the old interface (i.e., returning the full children lists at every tree).
3. Patch up/rewrite the Newick generator
4. Make a method to squash adjacent edges, and apply it to the output of simulate and simplify.

jeromekelleher:
More progress. The ordering for edges is now fixed as (time[parent], parent, child, left), and this has been propagated to all the relevant places. The order was chosen as it's the natural order for squashing edges that come out of the simulation and simplify. This turns out to be a bit tricky for stuff written using the older text format, as this ordering would sometimes be violated. I've fixed this by putting in an automatic ``sort_tables`` at the end of ``load_text``. I figure anyone using text isn't too bothered about performance anyway, and it's convenient to just be able to write edges in whatever order you want.

TODO:

1. I seem to have some bugs with simplify which need sorting
2. Make an algorithm to compute edgesets reasonably efficiently, so that we can support old code using the diffs() iterator and so on.

jeromekelleher:
OK! This should be pretty much done now. The only real casualties of the edgeset->edge transition are

- diffs(). This has been superseded by ``edge_diffs`` which should be easier to use and more useful.
- newick_trees(). This has been superseded by tree.newick(), which is more general and much easier to use.

These two were too awkward to keen around, and I doubt anyone is using them anyway. Other than that, there are still some edge cases for simplify (tree sequences with holes, various funky topologies) that are not working but I'd rather follow these up in PRs rather than keep this monster diff alive any longer.

jeromekelleher:
Merging.

PR #269, opened 2017-10-05
Proper handling of multiple roots

jeromekelleher:
Changes required to handle multiple roots in a systematic way. The roots of a tree are defined as the unique endpoints of paths from the samples to roots (nodes where parent[u] = -1). 

The low-level data structures are changed so that we have a new attribute ``left_root``. This is the leftmost root. Roots are considered to be sibs, so ``tree.right_sib[root]`` is the root to the right of root.

The high-level API exposes this interface, but also keeps the ``SparseTree.root`` attribute. When there is only one root, this works as expected (therefore not breaking code, and doing what you would expect in the majority of cases). There is also a `roots()` iterator, for cases where we have a forest.

The current algorithm can't handle internal samples/non samples leaves, so this is the next step.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -4056,14 +4056,14 @@ out:
 }
 
 static PyObject *
-SparseTree_get_root(SparseTree *self)
+SparseTree_get_left_root(SparseTree *self)
 {
     PyObject *ret = NULL;
 
     if (SparseTree_check_sparse_tree(self) != 0) {
         goto out;
     }
-    ret = Py_BuildValue("i", (int) self->sparse_tree->root);
+    ret = Py_BuildValue("i", (int) self->sparse_tree->left_root);
 out:
     return ret;
 }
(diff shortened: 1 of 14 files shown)


jeromekelleher:
I think this is looking pretty good now. The basic algorithm was simple enough, but internal samples complicated things quite a bit. I don't think this will slow down the tree transition algorithm appreciably, especially in the nominal case where all leaves are samples.

A root is defined as a unique path from the samples to nodes with parent[u] == -1. Thus a tree with no edges has n roots, one for each sample.

The API updates are as follows:

- A new attribute ``tree.roots`` that returns a list of the roots. I flip-flopped a bit on whether this should be an iterator like samples(), nodes(), etc, but it seems more convenient to just have a list ``tree.roots`` rather than having to do list(tree.roots()) all the time. Most of the time, this will be a one-element list. 

- The existing attribute ``tree.root`` is still there, and works as expected when there is one root. When there are multiple roots, it raises a ValueError.

- To get at the low-level algorithmic way of doing things, there is also a new attribute ``tree.left_root``, which allows you to get the full set of roots via ``tree.right_sib[root]``.

What do you think @petrelharp and @ashander? If this API seems sensible to you I'll finish this up, document it and merge.

petrelharp:
And also `tree.num_roots`, I see (good).

Looks good!  This means I should get back to the statistics calculations
with mutations.

PR #270, opened 2017-10-06
Revert "Adds a edge_start parameter to sort_tables"

jeromekelleher:
Reverts jeromekelleher/msprime#268

It turns out that we don't actually need this for what we thought we needed it for (incrementally updating tables for forward time simulations). Doesn't seem much point in carrying forward this feature then.

@petrelharp what do you think?

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -7402,18 +7402,15 @@ msprime_sort_tables(PyObject *self, PyObject *args, PyObject *kwds)
     migration_table_t *migrations = NULL;
     site_table_t *sites = NULL;
     mutation_table_t *mutations = NULL;
-    Py_ssize_t edge_start = 0;
 
-    static char *kwlist[] = {"nodes", "edges", "migrations", "sites", "mutations",
-        "edge_start", NULL};
+    static char *kwlist[] = {"nodes", "edges", "migrations", "sites", "mutations", NULL};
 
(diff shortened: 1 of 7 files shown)


petrelharp:
I think the strategy you proposed there is only guaranteed to work with nonoverlapping generations.  If we sorted by offspring birth time (i.e., time that the edge is generated) rather than parent birth time, then we'd have the property that chunks of history generated at separate times could be sorted separately and appended, but that would be less convenient elsewhere.

It is too bad that simulation is producing edges in almost exactly the reverse order that we want. I suppose we could implement a directionality to the tables, but that doesn't seem worth it.

jeromekelleher:
> I think the strategy you proposed there is only guaranteed to work with nonoverlapping generations. If we sorted by offspring birth time (i.e., time that the edge is generated) rather than parent birth time, then we'd have the property that chunks of history generated at separate times could be sorted separately and appended, but that would be less convenient elsewhere.

I was thinking this all right. The only place we really use the current ordering-by-parent is in simplify. It might be worth thinking this through a bit to see if it really is essential we process edges in parent order rather than child order.

OK, let's keep this functionality in for now until we're sure that the ordering can't be changed.

jeromekelleher:
Nope, it turns out this is actually a useful feature and we should keep it.

https://github.com/molpopgen/fwdpy11_arg_example/pull/8

Closing.

PR #274, opened 2017-10-11
Finish simplify implementation

jeromekelleher:
Find and squash all remaining simplify bugs!


jeromekelleher:
I finally fixed the nagging segment leak @petrelharp; I'm pretty sure we're topologically correct now with simplify! It turns out to have been a subtle issue with adding in internal samples that are parents in edges. I've fixed it up now and it gets rid of the ugly special case at the end of the merge_ancestors loop. This is almost identical to the simulation algorithm again now, except for the case where we must insert unary edges for internal samples.

petrelharp:
Ah, I see.  Damn corner cases!  Good job tracking that down.

Question:  `insert_ancestral_sample` seems wrong, since it looks like it just discards any information (ancestral segments) already stored in that ancestor.  (in the python code, it [overwrites `A[sample_id]`](https://github.com/jeromekelleher/msprime/blob/bb74d129a1d201e90653b92c5f15ea342d067954/tests/__init__.py#L838))  So, when called [here](https://github.com/jeromekelleher/msprime/blob/83f4e36eb28d7168f961bbbad398ea5f0d13dc6a/lib/table.c#L2489), if there had been segments inheriting from `parent` that had *not* coalesced during [`merge_ancestors`](https://github.com/jeromekelleher/msprime/blob/83f4e36eb28d7168f961bbbad398ea5f0d13dc6a/lib/table.c#L2482), then it seems like we'd forget about them.

But, it [looks like this is being tested](https://github.com/jeromekelleher/msprime/blob/83f4e36eb28d7168f961bbbad398ea5f0d13dc6a/lib/tests.c#L4385), with[this example](https://github.com/jeromekelleher/msprime/blob/83f4e36eb28d7168f961bbbad398ea5f0d13dc6a/lib/tests.c#L178), so it must be right?  What am I missing?

jeromekelleher:
> Question: insert_ancestral_sample seems wrong, since it looks like it just discards any information (ancestral segments) already stored in that ancestor. (in the python code, it overwrites A[sample_id]) So, when called here, if there had been segments inheriting from parent that had not coalesced during merge_ancestors, then it seems like we'd forget about them.

This is exactly the issue that took me ages to figure out. The subtlety occurs when we're inserting a sample that is an edge parent. When this happens, there are two sources of ancestral material that we have to take into account:

1. That inherited from younger samples, in the usual way
2. The new ancestral material inserted covering 0 to L.

Rather than try to merge these together in a complicated way, we're now waiting until the end of the event which may introduce new ancestral material, and then replacing this with a segment covering the full interval. By doing this, the assumptions of ``merge_ancestors`` are kept, and we avoid edge leaks by freeing the already extant segments for the parent. There is no information lost, as we are just replacing some set of segments [(l1, r1, x), ... (lk, rk, x)] with a single segment (0, L, x). 

The algorithm was correct before, but only by accident really. Took me a long time to figure out how it worked. Thanks for looking over the code!

jeromekelleher:
Merging this much as it corrects the topological issues with simplify. Working on mutational issues in a follow up.

petrelharp:
That bit makes sense, but what I don't see is where we catch any coalescences: suppose the ancestor `x` already has a single segment, with descendant `y`; but the ancestor is a sample, so we need to output an edge from `x` to `y` -- since output happens in `merge_ancestors`.

jeromekelleher:
That's why we map a new node ID for the parent before running ``merge_ancestors``. Once there is an output ID mapped to the input ID, the machinery in ``merge_ancestors`` can operate as usual (with the added code for outputting unary edges leading to samples). We run through all the edges first, allocating new nodes for any sampled children, and then allocate a node for the parent (if it is a sample). This  gives a sensible time ordering for the output nodes.

petrelharp:
Gotcha.

PR #248, opened 2017-08-26
mutations in WF test

petrelharp:
I've added mutations to the Wright-Fisher testing framework.

This is not critical, I just did this in the process of writing up the WF simulation method to include mutations.

PR #275, opened 2017-10-13
Simplify with recurrent mutations, etc.

jeromekelleher:
Fix up the mess in simplify caused by handling of recurrent mutations, etc.

This adds a ``parent`` column to the mutation table, which is the ID of a given mutation's parent mutation. Hopefully this will make clearing up this mess easier, and clear the way for nice general methods for statistics.

### DIFF
--- a/lib/table.c
+++ b/lib/table.c
@@ -2131,8 +2131,6 @@ simplifier_record_mutations(simplifier_t *self, node_id_t input_id)
      * the ancestry segments and the list of positions. This will be more efficient
      * unless we have huge numbers of mutations per node */
 
-    /* simplifier_print_state(self, stdout); */
-
     if (avl_count(avl_tree) > 0) {
         while (seg != NULL) {
             search.position = seg->left;
@@ -2463,7 +2461,7 @@ simplifier_output_sites(simplifier_t *self)
     mutation_id_t input_mutation, mapped_parent ,site_start, site_end;
     site_id_t num_input_sites = (site_id_t) self->input_sites.num_rows;
     mutation_id_t num_input_mutations = (mutation_id_t) self->input_mutations.num_rows;
-    mutation_id_t num_output_mutations, num_output_site_mutations;
+    mutation_id_t input_parent, num_output_mutations, num_output_site_mutations;
     node_id_t mapped_node;
     bool keep_mutation, keep_site;
     bool filter_zero_mutation_sites = (self->flags & MSP_FILTER_ZERO_MUTATION_SITES);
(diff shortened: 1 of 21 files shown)


jeromekelleher:
OK, I _think_ this is nearly done. It would like to say it's been fun, but it hasn't; this is a _hard_ and subtle problem!

There are two outstanding issues I could use your input on @petrelharp:

1. The ``filter_invariant_sites`` flag is now misleading. What I have implemented is really ``filter_zero_mutation_sites``, because actually getting all invariant sites in general is hard; we can't be sure of this unless we actually create the tree and look at the sample genotypes, essentially. I think we should rename this flag to ``filter_zero_mutation_sites``
2. More fundamentally, it appears that we have to make a choice between either having the samples provided to simplify be mapped to 0, ..., n - 1 in the output tree sequence **or** we enforce that edges are listed in increasing order of parent ID in the output edgesets. I think we have to let go of the parent ordering within edgesets, as it's just too much trouble.

What do you think?

jeromekelleher:
I've updated this to disable the check for sorted parents, and it all seems to work OK. If you agree with this @petrelharp I'll push through the change more rigorously.

Also note I've cherry picked in the WF mutations from #248.

petrelharp:
(1) `filter_zero_mutation_sites` - sure, sounds good, although `filter_invariant_sites` is not wrong, since all sites filtered *are* invariant sites.  A shorter name would be `filter_unmutated_sites`.

(2) I'm not bothered by a lack of sorted parents.  Still sorted by parent time, though?

jeromekelleher:
> (1) filter_zero_mutation_sites - sure, sounds good, although filter_invariant_sites is not wrong, since all sites filtered are invariant sites. A shorter name would be filter_unmutated_sites.

Well ``filter_invariant_sites`` would imply to me that it should filter all invariant sites, so I think we should avoid that. We can add further flags in the future to filter out other types of sites,  so I think it's good to be precise here (even at the expense of it being a bit wordy).

> (2) I'm not bothered by a lack of sorted parents. Still sorted by parent time, though?

Yes, definitely. And parents should be grouped together, otherwise simplify won't work.

jeromekelleher:
OK, I've updated those points @petrelharp and I think we're ready to merge.

lib/table.c
```diff
@@ -790,6 +796,12 @@ mutation_table_append_columns(mutation_table_t *self, size_t num_rows, site_id_t
             num_rows * sizeof(node_id_t));
     memcpy(self->derived_state + self->total_derived_state_length, derived_state,
             total_derived_state_length * sizeof(char));
+    if (parent == NULL) {
+        /* If parent is NULL, set all parents to the null mutation */
+        memset(self->parent + self->num_rows, 0xff, num_rows * sizeof(mutation_id_t));
```
petrelharp: Why is this the null mutation?

jeromekelleher: Why do we set the parent to the null mutation by default, or why is 0xff the null mutation?

petrelharp: Sorry, the second -- why 0xff?

jeromekelleher: The null mutation/node is -1, which is all 1s at the bit level. 0xf is hex for all 1s. Memset takes a byte as it's argument, so you have to specify 0xff to fill a buffer with 1 bits. It's not that obvious, and is a departure from using NULL_MUTATION where possible but it is convenient and fast.

petrelharp:
thanks!

Issue #286, opened 2017-10-19
Method to compute `parent` column in `MutationTable`


petrelharp:
Write a function that does this.  Maybe two of them, one sideways and one backwards, for comparison.

Once this is written, we may also want to add  flags to `load_tables` to:
 - verify the parent column if present, and
 - compute the `parent` column on the fly, for sources where this is not already easily available.

petrelharp:
It turns out we need to do this *now*, cc @bhaller.

jeromekelleher:
This is done in ``table_collection_compute_mutation_parents``.

Issue #122, opened 2017-01-23
accomodate recurrent mutations in branch_stats algorithm


petrelharp:
Thoughts:

There's only quite such an easy correspondance between branch length and # of mutations if we assume all mutations are distinguishable.  So to exactly match the branch length statistics, we'd ignore the fact that there may be double mutations.  But to match what we see in the sequence (like we actually want to do): take divergence: a mutation that occurs on branches leading to both samples should be counted zero times, not twice. But, if the mutations returned by branch_mutations know where else on the tree they occur, then this is do-able with a bit more bookkeeping. Like for instance: first update the X's for each edge; then iterate over the mutations, evaluating (and caching) f(\sum_{branches} X) for each mutation to get the weight for that mutation.  Or maybe just using the original algorithm but then going and looking individually at all the recurrent mutations to adjust.

jeromekelleher:
With the most recent merge, we also have to consider back mutations, and eventually > 2 alleles. I'm not sure what the best approach is, but it is something we need to consider.

petrelharp:
Related question: let X and Y be two independent, randomly chosen leaves of a tree, with P(X=Y)=0.  Let $d(x,y)$ be the Hamming distance between the sequences at leaves $x$ and $y$, sequences that have evolved on the tree, and we know where/when/what all the mutations were.  How do we efficiently calculate $E[d(X,Y)]$?  

In the infinite sites model this is just $\sum_e P(e \in X->Y) m(e)$, where $m(e)$ is the number of mutations on  edge $e$, and $e \in X->Y$ means that edge $e$ is on the path from $X$ to $Y$.  But with backmutation, is there a way to calculate it without summing over all pairs of possible endpoints?

jeromekelleher:
Good question @petrelharp, I don't know the answer to that one. 

I expect in general that many things will be a lot harder when we don't have the nice 1 site == 1 mutation relationship. However, we can test this on a site-by-site basis (each site has a reference to the list of mutations), so that a simple efficient algorithm can be used if the site has a single mutation, and a less efficient algorithm otherwise. I expect most sites will have this single mutation property.

I haven't really worked through the consequences of having multiple mutations per site yet, I'm just trying to represent the data that I have at the moment. LD and pi calculations will error out at moment if called on tree sequences with > 1 mutation per site.

[...]

PR #288, opened 2017-10-21
python version of compute_mutation_parent

petrelharp:
Here's an implementation, addressing #286 , in python.  Issues:

- interface.  As written, it could be called with (valid, no-mutations tree sequence and associated sites, mutations tables that don't have a mutation.parent column) to produce a correct parent column for the mutation table, to then be loaded into a tree sequence.  Maybe this is not a use case.  Maybe we just want to change this over to checking an existing parent column.

- tree sequence or table?  Maybe it should just work on tree sequences, to do a check for correct mutation.parent that was not feasible at import time.  Never mind tables.

- efficiency.  Currently this (a) gets a list of nodes having mutations at the current site, and (b) from each, traverses up the tree until it hits the root or another node in the list.  I need to at least efficiently check if the current node is in the list of mutated nodes.  Maybe there is a more efficient way to do the whole thing.

- nodes.  Is it allowed to have more than one mutation on a given node? I think not, but the code makes this assumption; perhaps we should check that here.

What do you think, @jeromekelleher? I can try to move it over to C next.


petrelharp:
I merged #289 into here to use it for testing.

msprime/trees.py
```diff
@@ -1779,3 +1779,60 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+    cur_site = mutations.site[mut_start]
+    cur_pos = sites.position[cur_site]
+    num_muts = 1
+    while ((mut_start + num_muts < mutations.num_rows)
+           and (mutations.site[mut_start + num_muts] == cur_site)):
+        num_muts += 1
+    for t in ts.trees():
+        left, right = t.interval
+        if right < cur_pos:
+            next
```
jeromekelleher: Does this work? I've never seen the next function used like this before

tests/test_wright_fisher.py
```diff
@@ -259,6 +265,31 @@ def test_with_mutations(self):
...
+    def test_with_recurrent_mutations(self):
+        # actually with only ONE site, at 0.0
+        N = 10
+        ngens = 100
+        tables = wf_sim(
+            N=N, ngens=ngens, mutation_rate=10.0, deep_history=False,
+            nloci=1, seed=self.random_seed)
+        self.assertEqual(tables.sites.num_rows, 1)
+        self.assertEqual(tables.sites.position, 0.0)
+        self.assertGreater(tables.mutations.num_rows, 0)
```
jeromekelleher: We should make sure running ``haplotypes`` succeeds here as well, i.e., load the pre-simplify tables into a tree sequence, and get the haplotypes.

msprime/trees.py
```diff
@@ -1779,3 +1779,60 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+            mut_start = mut_start + num_muts
+            if mut_start >= mutations.num_rows:
+                break
+            cur_site = mutations.site[mut_start]
+            cur_pos = sites.position[cur_site]
+            num_muts = 1
+            while ((mut_start + num_muts < mutations.num_rows)
+                   and (mutations.site[mut_start + num_muts] == cur_site)):
+                num_muts += 1
+    return new_parent
```
jeromekelleher: This looks good, but I wonder if we can make it a bit simpler. Why not use something like this:
```python
mutation_parent = {}
for t in ts.trees():
    for site in t.sites():
         # Also assumes that there is one mutation per node, which we can't do.
         node_map = {mut.node: mut for mut in site.mutations}
         for mut in site.mutations:
              u = t.parent(mut.node)
              while u != NULL_NODE and u not in node_map:
                   u = t.parent(u)
               if u != NULL_NODE:
                   # mutation.id currently doesn't exist, but it should. We can keep track of 
                   # the mutation IDs here with a simple counter for the moment I guess.
                   # This is what is would look like if we did have mutation IDs.
                   mutation_parent[mutation.id] = node_map[u].id
```
This is just a sketch to illustrate the overall structure, I haven't thought it through properly or ran it or anything.

jeromekelleher:
ps. Ignore the 'codacy' CI check. I was just playing with it to see if it was any good, and didn't realise it would be turned on for PRs. I've turned it off now and it should go away when you push next.

petrelharp:
re: multiple mutations on a node - makes sense; I mostly wanted to assume this on a first pass.  Fixing up...

msprime/trees.py
petrelharp: hah, no. I meant `continue`.

petrelharp:
In your (good) suggestion you're assuming that everything is loaded into the tree sequence.  This seems fine, but FYI: implies that we can load tables that don't have a (valid) `mutation.parent` column.

msprime/trees.py
petrelharp: `mutation.id` *does* seem to exist - `site.id` does not, was that what you were thinking?

petrelharp:
Wait a sec here.  If we don't have the mutation parent information then we *can't* properly deal with more than one mutation per node.

On the plus side, I brought over your Jukes-Cantor implementation from #122 to `tests/tsutils.py` (and modified it some).

I have not switched over to your simpler implementation.

msprime/trees.py
jeromekelleher: Sorry, keep getting confused with this. We should have 'id' on all of these objects, but there's some tricky compatibility issues with existing code so I've avoided fixing the issue. Something that needs to be sorted out before release.

jeromekelleher:
> Wait a sec here. If we don't have the mutation parent information then we can't properly deal with more than one mutation per node.

That's true... So, does this mean that we can't compute the parent column? I'm not sure this is true if the requirement that a child mutation must always have an ID < its parent holds. Then, for multiple mutations along a branch, we know that the mutations must be listed in order that they happen.

> In your (good) suggestion you're assuming that everything is loaded into the tree sequence. This seems fine, but FYI: implies that we can load tables that don't have a (valid) mutation.parent column.

I think this is a reasonable assumption. There are some inconsistencies that we can only detect at run-time, and I think faulty parent references are one of these. Inconsistent edges (where the same child refers to different parents over an interval) for example, can only be detected when we try to generate the trees.

petrelharp:
> That's true... So, does this mean that we can't compute the parent column? I'm not sure this is true if the requirement that a child mutation must always have an ID > its parent holds. Then, for multiple mutations along a branch, we know that the mutations must be listed in order that they happen.

That'd do it.  Since we only need this in the case where parent mutation hasn't been computed, and then only if there's more than one mutation per node, I'm a bit loathe to make it a strict requirement.  We could alternatively require it for *this* function only.  No way to check, of course.

jeromekelleher:
> That'd do it. Since we only need this in the case where parent mutation hasn't been computed, and then only if there's more than one mutation per node, I'm a bit loathe to make it a strict requirement, because then we'd have to check for it, etc. We could alternatively require it for this function only. No way to check, of course. What do you think?

We're already checking for this in ``tree_sequence_check``. I think it's a useful property, and has some power to ensure that the input makes sense, but I wouldn't complain too much if we removed it.

tests/test_wright_fisher.py
petrelharp: It is [already being tested](https://github.com/jeromekelleher/msprime/blob/master/tests/test_wright_fisher.py#L364) that this works.

Strangely!  This should *not* work, because in laying down mutations I [don't check the ancestral state](https://github.com/jeromekelleher/msprime/blob/master/tests/test_wright_fisher.py#L117).

jeromekelleher: That's why we should run it, to make sure that the pre-simplify mutations make sense. We're not testing 'haplotypes' here, rather we're checking that the input data makes sense.

petrelharp: Yes, I agree. I did insert code to make sure that haplotypes ran, and they did; but removed it when I noticed it was being tested elsewhere.  With correct links: [here](https://github.com/petrelharp/msprime/blob/mut_parent/tests/test_wright_fisher.py#L328) we run WF sims with three loci and mutation rate = 5.0, so are getting multiple muts per site.

So, I claim it *does* work, but is almost certainly violating the parent != child mutation condition?  I have not verified that, so maybe something is not as I expect.  (Later: now, am translating kidz -> school.)

jeromekelleher:
OK, looks good @petrelharp. How about we move the function into ``tests/tsutils.py`` for now? We're only using it for testing at the moment anyway. We can add a C version of this to the low-level checks at some point if it seems like a good thing to do.

petrelharp:
sounds good.

oh, perhaps **that's** why msprime does not complain when loading WF-simulated tables with parent state = mutated state -- because the WF sim does not compute the parent mutation?  **Nope:** I've just added it, and still no complaints.

tests/test_wright_fisher.py
petrelharp: I've added in parent mutation to the WF sim and am now [loading and checking haps](https://github.com/petrelharp/msprime/blob/7b7983d58fa8773f4e927f11945405ab0dd3b617/tests/test_wright_fisher.py#L302) and no complaints.  I've checked, and in this WF sim there are *lots* of consecutive mutations with the same state, e.g.
```
id      site    node    derived_state   parent
0       0       10      A       -1
1       0       10      T       0
2       0       10      C       1
3       0       10      T       2
4       0       10      T       3
```

I think that maybe you are not actually checking for the parent != child mutation property?

jeromekelleher:
> oh, perhaps that's why msprime does not complain when loading WF-simulated tables with parent state = mutated state -- because the WF sim does not compute the parent mutation? Nope: I've just added it, and still no complaints.

I haven't updated haplotypes/variants to use the parent mutation at all. There are some checks in there, but it's all just 'it happens to work' at the moment I think.

petrelharp:
Well, OK -- I think this is good to merge then.  (just force-pushed flake8 fixes btw)

jeromekelleher:
Great, thanks @petrelharp. Merging.

petrelharp:
This version makes the assumption that mutations come in order by node time, which might not be true; the fix is eash (see #286)

Issue #244, opened 2017-08-25
Add mutations to Wright-Fisher tests


jeromekelleher:
The tests in ``tests/test_wright_fisher.py`` currently only check topological properties, and mutations have been disabled in the WF simulator. Enable mutations in the simulator and write some tests.

petrelharp:
I think that this is no longer true: for instance, [here](https://github.com/jeromekelleher/msprime/blob/master/tests/test_wright_fisher.py#L364) we check haplotypes pre- and post-simplify?

However: it is currently doing infinite sites.  We should modify this to allow recurrent mutations.

PR #289, opened 2017-10-21
enabled recurrent mutations in WF tests

petrelharp:
This is working in the simplest possible way.  There's not even any checking that the derived mutation is different from the parental mutation.  This addresses #244.


petrelharp:
merged this into #288 and fixed flake8 errors there

PR #290, opened 2017-10-22
Implementation of site statistics

petrelharp:
I've started an implementation of the site-based statistics computation, as described in #122.

Hardly begun, obviously.

### DIFF
--- a/msprime/stats.py
+++ b/msprime/stats.py
@@ -595,68 +595,6 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
         interested in.
     """
 
-    def tracked_sample_tree_stat_vector(self, sample_sets, weight_fun, windows=None):
-        if windows is None:
-            windows = (0, self.tree_sequence.sequence_length)
-        for U in sample_sets:
-            if len(U) != len(set(U)):
-                raise ValueError(
-                    "elements of sample_sets cannot contain repeated elements.")
-            for u in U:
-                if not self.tree_sequence.node(u).is_sample():
-                    raise ValueError("Not all elements of sample_sets are samples.")
-        num_windows = len(windows) - 1
-        if windows[0] != 0.0:
-            raise ValueError(
-                "Windows must start at the start of the sequence (at 0.0).")
-        if windows[-1] != self.tree_sequence.sequence_length:
-            raise ValueError("Windows must extend to the end of the sequence.")
-        for k in range(num_windows):
-            if windows[k + 1] <= windows[k]:
-                raise ValueError("Windows must be increasing.")
-        for U in sample_sets:
-            if max([U.count(x) for x in set(U)]) > 1:
-                raise ValueError("elements of sample_sets",
-                                 "cannot contain repeated elements.")
-        tr_its = [self.tree_sequence.trees(
-                        tracked_samples=x,
-                        sample_counts=True) for x in sample_sets]
-        n = [len(U) for U in sample_sets]
-        n_out = len(weight_fun([0 for a in sample_sets]))
-        S = [[0.0 for j in range(n_out)] for _ in range(num_windows)]
-        T = [0.0 for j in range(n_out)]
-        window_num = 0
-        for k in range(self.tree_sequence.num_trees):
-            for j in range(n_out):
-                T[j] = 0.0
-            trs = [next(x) for x in tr_its]
-            root = trs[0].root
-            for node in trs[0].nodes():
-                if node != root:
-                    x = [tr.num_tracked_samples(node) for tr in trs]
-                    nx = [a - b for a, b in zip(n, x)]
-                    w = [a + b for a, b in zip(weight_fun(x), weight_fun(nx))]
-                    for j in range(n_out):
-                        T[j] += w[j] * trs[0].branch_length(node)
-            tr_len = (min(windows[window_num + 1], trs[0].interval[1])
-                      - max(windows[window_num], trs[0].interval[0]))
-            for j in range(n_out):
-                S[window_num][j] += T[j] * tr_len
-            while (window_num < num_windows and
-                   trs[0].interval[1] >= windows[window_num + 1]):
-                # wrap up the last window
-                window_length = windows[window_num + 1] - windows[window_num]
-                for j in range(n_out):
-                    S[window_num][j] /= window_length
-                # start the next
-                window_num += 1
-                if window_num < num_windows:
-                    tr_len = (min(windows[window_num + 1], trs[0].interval[1])
-                              - max(windows[window_num], trs[0].interval[0]))
-                    for j in range(n_out):
-                        S[window_num][j] += T[j] * tr_len
-        return S
-
     def tree_stat_vector(self, sample_sets, weight_fun, windows=None):
         '''
         Here sample_sets is a list of lists of samples, and weight_fun is a
@@ -709,8 +647,6 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
 
         S = [[0.0 for j in range(n_out)] for _ in range(num_windows)]
         L = [0.0 for j in range(n_out)]
-        # print("sample_sets:", sample_sets)
-        # print("n_out:",n_out)
         N = self.tree_sequence.num_nodes
         X = [[int(u in a) for a in sample_sets] for u in range(N)]
         # we will essentially construct the tree
@@ -724,8 +660,6 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
             length = interval[1] - interval[0]
             for sign, records in ((-1, records_out), (+1, records_in)):
                 for edge in records:
-                    # print("Record (",sign,"):",node,children,time)
-                    # print("\t",X, "-->", L)
                     dx = [0 for k in range(num_sample_sets)]
                     if sign == +1:
                         pi[edge.child] = edge.parent
@@ -735,8 +669,6 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
                     dt = (node_time[pi[edge.child]] - node_time[edge.child])
                     for j in range(n_out):
                         L[j] += sign * dt * w[j]
-                    # print("\t\tchild:",child,"+=",sign,"*",wfn(X[child]),
-                    #    "*(",node_time[pi[child]],"-",node_time[child],")","-->",L)
                     if sign == -1:
                         pi[edge.child] = -1
                     old_w = wfn(X[edge.parent])
@@ -747,8 +679,6 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
                         dt = (node_time[pi[edge.parent]] - node_time[edge.parent])
                         for j in range(n_out):
                             L[j] += dt * (w[j]-old_w[j])
-                        # print("\t\tnode:",node,"+=",dt,"*(",wfn(X[node]),"-",
-                        #   old_w,") -->",L)
                     # propagate change up the tree
                     u = pi[edge.parent]
                     if u != -1:
@@ -764,12 +694,8 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
                                 dt = (node_time[pi[u]] - node_time[u])
                                 for j in range(n_out):
                                     L[j] += dt*(w[j] - old_w[j])
-                                # print("\t\tanc:",u,"+=",dt,"*(",wfn(X[u]),"-",
-                                #    old_w,") -->",L)
                             u = next_u
                             next_u = pi[next_u]
-                    # print("\t",X, "-->", L)
-            # print("next tree:",L,length)
             while chrom_pos + length >= windows[window_num + 1]:
                 # wrap up the last window
                 this_length = windows[window_num + 1] - chrom_pos
@@ -848,8 +774,6 @@ class SiteStatCalculator(GeneralStatCalculator):
         S = [[0.0 for j in range(n_out)] for _ in range(num_windows)]
         if num_sites == 0:
             return S
-        # print("sample_sets:", sample_sets)
-        # print("n_out:",n_out)
         N = self.tree_sequence.num_nodes
         # initialize: with no tree, each node is either in a sample set or not
         X = [[int(u in a) for a in sample_sets] for u in range(N)]
@@ -860,9 +784,6 @@ class SiteStatCalculator(GeneralStatCalculator):
         ns = 0  # this will record number of sites seen so far
         s = next(sites)
         # index of *left-hand* end of the current window
-        tabs = self.tree_sequence.dump_tables()  # FOR DEBUGGING
-        print(tabs)
-        print("sample sets:", sample_sets)
         window_num = 0
         while s.position > windows[window_num + 1]:
             window_num += 1
@@ -873,8 +794,6 @@ class SiteStatCalculator(GeneralStatCalculator):
             # update the tree
             for sign, records in ((-1, records_out), (+1, records_in)):
                 for edge in records:
-                    # print("Record (",sign,"):",node,children,time)
-                    # print("\t",X, "-->", L)
                     dx = [0 for k in range(num_sample_sets)]
                     if sign == +1:
                         pi[edge.child] = edge.parent
@@ -904,9 +823,6 @@ class SiteStatCalculator(GeneralStatCalculator):
                     while s.position > windows[window_num + 1]:
                         window_num += 1
                 nm = len(s.mutations)
-                print("---------")
-                print(s)
-                print("X", X)
                 if nm > 0:
                     U = {s.ancestral_state: list(n)}
                     for mut in s.mutations:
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I haven't looked at this in detail yet, but I'm curious about how useful the "parent mutation" is here. Is it handy or indispensable?

msprime/stats.py
```diff
@@ -711,3 +720,155 @@ def wfn(x):
...
+                        while u != -1:
+                            for k in range(num_sample_sets):
+                                X[u][k] += dx[k]
+                            u = next_u
+                            next_u = pi[next_u]
+                    # print("\t",X, "-->", L)
+            # print("next tree:",L,length)
+            site_pos = sites.position[site_num]
+            start_mut = next_mut
+            assert mutations.site[start_mut] == site_num
```
petrelharp: Here I'm assuming that every site has some mutations.  Is this something we assume, @jeromekelleher?  It'd be easy to not assume it if  not already.

msprime/stats.py
```diff
@@ -711,3 +720,155 @@ def wfn(x):
...
+                    if p != -1:
+                        parent_muts.append(p - start_mut + 1)
+                    else:
+                        parent_muts.append(0)
+                    next_mut += 1
+                nm = next_mut - start_mut
+                for j in range(nm):
+                    if parent_muts[j] != -1:
+                        for k in range(num_sample_sets):
+                            U[parent_muts[j]][k] -= X[mut_nodes[j]][k]
```
petrelharp: Here's the place where we use the "mutation parent" information. @jeromekelleher

petrelharp:
The "parent mutation" info gets used as we outlined over at #122 - to convert from `X`, the number of samples below each mutated node, to `U`, the number of samples below each mutated node *that inherit that mutation*.  (I've tagged in the code where it happens.) In #122 we discussed either traversing up the mutation-tree to do this or down it; I realized we don't have to do either, and can update in whatever order they come: if `X[j]` but the number of nodes below `j` and let `p[j]` be the parent of `j`; then first initialize `U[j] = X[j]` and then for each `j`, subtract `X[j]` from `U[p[j]]`.

This algorithm is quite nice (although my implementation could be greatly simplified - next on the list!); the alternative is that we'd have to traverse up the node-tree on the fly to construct `p[]`.  This wouldn't be hard, since we've gone along constructing the tree as well.

msprime/stats.py
jeromekelleher: We can have sites that have zero mutations; this turns out to be useful for me when doing inference.

jeromekelleher:
Excellent, thanks @petrelharp. So, by having a parent mutation field we're avoiding a O(tree depth) upward traversal for each mutation to find (essentially) it's parent, right? This seems worthwhile.

petrelharp:
It does seem useful, although it's only happening at sites with >1 mutation.   This also means the extra space cost could be decreased quite a bit by not storing the null parent mutations somehow.

I do wonder if it's worth making users deal with it - but, since we have a method to compute it on the fly, perhaps this is no big deal, especially if we make this an option to load_tables.

jeromekelleher:
> So, current proposal is: leave in mutation parents, since they are useful, but automatically detect in load_tables whether there is a parent column in the mutation table, and if not, calculate it? Or, actually, to know we can do this but to leave out this option until someone actually wants to use it.

Great, this an excellent plan.

petrelharp:
To-do:

- test with recurrent mutations
- SFS
- covariance
- TMRCA/divergence/whatever

petrelharp:
Grrrr, noticed I wasn't actually testing the SiteStats -- had `mutation_rate` at default (0.0).

petrelharp:
Finally worked out what the problem is - and it's inconsistent definitions for multiallelic sites.  Got to think this out a bit more carefully.

jeromekelleher:
> Finally worked out what the problem is - and it's inconsistent definitions for multiallelic sites. Got to think this out a bit more carefully.

Interesting.. what are the competing definitions?

petrelharp:
Still mulling it over: see discussion in #122 (put it there with the rest of the implementation discussion).

petrelharp:
Realized that the allele count bit can be much simplified; and am working on the C version [elsewhere](https://github.com/petrelharp/msprime/tree/update_other_stats).

I've removed the canceling factors of two and included evaluating the weighting function in the branch length statistic...  but have not debugged everything.  There may be still issues with multiple mutaitons per branch.

jeromekelleher:
> Realized that the allele count bit can be much simplified; and am working on the C version elsewhere.

I had a quick look at this and it looks good. Let me know if/when you would like more detailed feedback on it.

hyanwong:
@petrelharp: I would be interested in being able to calculate some of the stats (especially the SFS and values based on it, like Watterson's theta, etc) on a TS with no variation. Without mutations, such stats would have to be entirely branch-length based. I.e. you would need to calculate the total length of terminal branches, then branches with 2 tips, with 3 tips, etc. This would be different for each tree (although you could probably use the diffs between trees to update the statistic).

From a brief inspection of the code, it seems like this diff-based approach isn't how stuff is implemented at present. Should I try to mock up some code which will do it? If so, should I fork from `petrelharp:site_statistics` or from `msprime:master`?

hyanwong:
@petrelharp: for my own purposes I have cobbled together a basic incremental branch-based SFS calculator in python. By incremental I mean that it uses the diffs between successive trees to calculate how the SFS changes at each breakpoint. If you set `check_method = True` at the top of the file, it will run and compare the output with the full branchwise SFS generated by looking at the entire tree at each breakpoint, which is obviously slow, but useful to verify. This works OK on some test treeseqs I have generated from ftprime (which can have >2 children per node), but I suspect it will fail if there are multiple roots (in this case, we don't know what the branch length is at the root of each tree, so I'm not sure if calculating the full SFS is possible anyway). Running with `check_method = False` is tolerably fast.

FYI, the way it works is that I calculate the SFS once at the start, then update it by looking at all the changed edges and their ancestors, removing branch lengths from the SFS for changed edges in the out tree, then adding branch lengths back in for changed edges in the in tree. 

I mention all this as I wondered if this would be a useful check on your methods (or possibly another angle for an algorithm).

[SFS.py.txt](https://github.com/jeromekelleher/msprime/files/1433588/SFS.py.txt)

hyanwong:
NB, once you have the SFS, then maybe the following are useful:

```
def H(n):
    """Returns an approximate value of n-th harmonic number.

       http://en.wikipedia.org/wiki/Harmonic_number
    """
    # Euler-Mascheroni constant
    gamma = 0.57721566490153286060651209008240243104215933593992
    return gamma + math.log(n) + 0.5/n - 1./(12*n**2) + 1./(120*n**4)

def Watterson(SFS, n):
    """
    Calculate the (mutationless) Watterson estimator
    """
    return sum(SFS.values()) / H(n)

def nucleotide_diversity(SFS, n):
    """
    Calculate the (mutationless) nucleotide diversity
    2 sum(i(n−i)*ξi) / n / (n-1)
    """
    return 2 * sum([i*(n-i)*SFS[i] for i in SFS.keys()]) / n / (n-1)
```

petrelharp:
Hey, thanks.  As you probably saw, we have two versions of this algorithm: a non-working one in code, and a theoretical one not coded.  I'll be interested to compare to yours, but it'll be not 'til next week (when I'll be back on this branch).

hyanwong:
Thanks @petrelharp. By the way, what should we call the distribution of total branch lengths binned according to the number of tips? It's not strictly the "site frequency spectrum", as it isn't based on sites, nor on their frequencies. Essentially, it's the probability distribution of which the SFS is a sample. Or more poetically, it's a summary spectrum of the true history, of which the SFS is the only visible shadow*. I guess "tip count spectrum" or something similar would be more appropriate, but I guess there must be a formal name for it.

*echoes of Plato's cave here, methinks.

petrelharp:
Nice about the "shadow".  =)

Options:
 - 'expected sfs'
 - 'tree frequency spectrum'
 - 'induced split distribution' since it's the distribution on splits induced by the distribution on trees
 - ... ?

This is a question that needs answering more generally: I've been referring to these sorts of things as the "branch length version" of the usual stats, which isn't very satisfactory.

petrelharp:
This is working.  Before merging:

- I should make a pass through the interface and check that things are consistent.
- The underlying `tree_stat_vector` functions I wrote at a low level, thinking to mirror what we'd do in C.  But, in C we can still use tree iterators and `num_tracked_samples`, which would substantially simplify the code.  Should I rewrite using these?

jeromekelleher:
> The underlying tree_stat_vector functions I wrote at a low level, thinking to mirror what we'd do in C. But, in C we can still use tree iterators and num_tracked_samples, which would substantially simplify the code. Should I rewrite using these?

Yes, this would be good I think. Making code simpler is nearly always a good thing!

msprime/stats.py
```diff
@@ -711,3 +727,162 @@ def wfn(x):
...
+                        S[window_num][j] /= window_length
+                    # may need to advance through empty windows
+                    while s.position > windows[window_num + 1]:
+                        window_num += 1
+                nm = len(s.mutations)
+                print("---------")
+                print(s)
+                print("X", X)
+                if nm > 0:
+                    U = {s.ancestral_state: n.copy()}
```
ashander: This breaks on python2, try `list(n)`

tests/test_tree_stats.py
```diff
@@ -889,56 +934,255 @@ def test_case_2(self):
...
+        ts = tsutil.jukes_cantor(orig_ts, num_sites=3, mu=3,
+                                 multiple_per_node=True, seed=self.seed)
+        branch_tsc = msprime.BranchLengthStatCalculator(ts)
+        py_branch_tsc = PythonBranchLengthStatCalculator(ts)
+        site_tsc = msprime.SiteStatCalculator(ts)
+        py_site_tsc = PythonSiteStatCalculator(ts)
+
+        for t in ts.trees():
+            print(t.interval)
+            print(msprime.drawing.draw_tree(t, format='unicode'))
```
ashander: This causes an exception on py27

```  
File "/home/travis/build/jeromekelleher/msprime/msprime/drawing.py", line 78, in draw_tree
    raise ValueError("Unicode tree drawing not supported on Python 2")
ValueError: Unicode tree drawing not supported on Python 2
```

but it's nice to have drawing for debug output. maybe enclose in a try/catch?

jeromekelleher: Use ``format='ascii'`` if you want to use py27. I guess we could add another format called 'text' which chooses unicode on py3 and ascii on py27.

petrelharp: thx @ashander - I have fixed these (deleting the second one)

petrelharp:
@jeromekelleher - I sat down to rewrite the basic `tree_stat_vector` functions using multiple `tracked_node` iterators, as that's basically what they are doing, and recalled that I'd already done that, in the testing code.  Here's the [original](https://github.com/petrelharp/msprime/blob/site_statistics/msprime/stats.py#L597) and the (much  more concise! version is [here](https://github.com/petrelharp/msprime/blob/site_statistics/tests/test_tree_stats.py#L267).

How's that look?  I still need to rewrite the site-based stats version like this.

petrelharp:
I did some testing on timing of the two, like so:
```
import msprime

nsamples = 4000
ts = msprime.simulate(nsamples, recombination_rate=10.0)

sc = msprime.BranchLengthStatCalculator(ts)


sample_sets = [list(range(0,int(nsamples/4))), 
               list(range(int(nsamples/4),int(nsamples/2))), 
               list(range(int(nsamples/2),int(nsamples*3/4))), 
               list(range(int(nsamples*3/4),nsamples))]

print(sc.f4(sample_sets, windows=list(k/100 for k in range(0,101))))
```
and with the [this function](https://github.com/petrelharp/msprime/blob/site_statistics/msprime/stats.py#L597), newly added.

The tracked sample version is **much** slower: 
```
# low level
elapsed: 0:00.84 / kernel: 0.47 / user: 0.89 / mem: 57216
# tracked sample
elapsed: 0:44.19 / kernel: 0.48 / user: 44.23 / mem: 57328
```

... so I vote to leave in the "low level" version.  And merge once I double check the interface and mark things as "beta".

cc @ashander that current stats computation method is as fast as we should expect

ashander:
Wow that is slower. Please do keep the 'low level' version in!

The current version works fine for our purposes as long as we keep the
total number of samples down

jeromekelleher:
> ... so I vote to leave in the "low level" version. And merge once I double check the interface and mark things as "beta".

Agreed. So we do need the diffs for this calculation---good to know.

petrelharp:
Really what we need is the *tracked samples* diffs.  At least that's the main difference I see.

petrelharp:
Since 'divergence' and 'TMRCA' are the same thing, except for a factor of 2, and divergence is more commonly calculated (maybe); I've removed the factor of 2 and renamed the 'tmrca' functions to be 'divergence'.  I realize this breaks @ashander's code.  So, we should probably put the tmrca functions back in, but trivially as just multiplying divergence by 2?  But, this seems bloat-ey.  What do you think?

petrelharp:
pending answer to the previous query, I say this is ready to merge.

hyanwong:
I haven't followed this thread much, but just to point out that TMRCA is not necessarily the same as divergence over two, if you have samples that exist at different times (e.g. with ancient DNA samples). I think Jerome is allowing this sort of thing in msprime.

ashander:
I'd vote for non broken code, but it's easy to fix ;)

Oh but this seems to argue for having only divergence as it's more general:

I haven't followed this thread much, but just to point out that TMRCA is
> not necessarily the same as divergence over two, if you have samples that
> exist at different times (e.g. with ancient DNA samples). I think Jerome is
> allowing this sort of thing in msprime.
>

So I'm happy to make some small fixes downstream if we ditch TMRCA

petrelharp:
Ah, right. Another good reason to remove `tmrca` -- because in our usage it was "mean tmrca", where "mean" implied among other things the average over which of the two branches you were choosing -- and so "mean tmrca" is divergence divided by two.

petrelharp:
hm, maybe those tests will satisfy codecov.

jeromekelleher:
I'd vote for keeping it simple and having just "divergence". It's easy enough to add in the tmrca equivalents later if it turns out they are useful (but very difficult to remove them if it turns out they are not!).

Looks like you need to bump the patch coverage up a bit to keep codecov happy @petrelharp. From a quick look at the [report](https://codecov.io/gh/jeromekelleher/msprime/pull/290/diff) it looks like you can get this by getting rid of some dead code and by doing some tests on the input checking

msprime/stats.py
```diff
@@ -408,9 +405,18 @@ def f4_vector(self, sample_sets, windows, indices):
                 raise ValueError("All tuples in indices should be of length 4.")
         n = [len(x) for x in sample_sets]
 
+        # def f(x):
```
jeromekelleher: No point in keeping dead code lying around.

msprime/stats.py
```diff
@@ -157,53 +157,50 @@ def get_r2_matrix(self):
...
+    implemment different `tree_stat_vector()` methods, but given that
+    general-purpose function, many statistics are computed in the same way.
     """
 
     def __init__(self, tree_sequence):
         check_numpy()
         self.tree_sequence = tree_sequence
 
-    def mean_pairwise_tmrca(self, sample_sets, windows):
+    def divergence(self, sample_sets, windows):
```
jeromekelleher: I much prefer this way of putting it. Way cleaner and easier to type!

msprime/stats.py
```diff
@@ -581,13 +583,28 @@ def tree_stat_windowed(self, sample_sets, weight_fun, windows=None):
...
 
+
+class BranchLengthStatCalculator(GeneralStatCalculator):
+    """
+    Class for calculating a broad class of tree statistics.  These are all
+    calculated using :meth:``BranchLengthStatCalculator.tree_stat_vector`` as the
+    underlying engine.  This class requires the `numpy
+    <http://www.numpy.org/>`_ library.
+
+    NOTE: this interface is still in beta, and may change in the future.
```
jeromekelleher: Let's put in a sphinx [..warning](http://www.sphinx-doc.org/en/stable/markup/para.html) in here (and the other classes) to make sure it's very obviously a warning.

ashander:
Looking good! I noticed today that empty sample sets can be passed and will cause a cryptic divide by zero error. So, consider merging this to your `site_statistics` https://github.com/petrelharp/msprime/pull/6

Or ... if you all want to be done with this monster; I can re-propose it as a PR to msprime after

jeromekelleher:
Superb, thanks @petrelharp!. Merging.

hyanwong:
> Options:
> 
> 'expected sfs'
> 'tree frequency spectrum'
> 'induced split distribution' since it's the distribution on splits induced by the distribution on trees
> ... ?

For some reason my mind wandered back to this recently, @petrelharp, and I wondered about "idealised SFS". I'm still not sure if "frequency" is the right terminology, though. I haven't read through your recent bioarxiv paper fully yet - perhaps it is mentioned there?

petrelharp:
It is not mentioned there, yet.

PR #296, opened 2017-10-31
Wf

DomNelson:
Includes updates to discrete-time Wright Fisher model to use new simulation model API. 

I considered a more descriptive name for `msp_recombine` to differentiate it from `msp_recombination_event`, but the only difference is that multiple events are possible... maybe ok as-is, where `event` implies only a single breakpoint?

Change to `dev.cfg` is to avoid sampling internal nodes - this currently doesn't work under the Wright Fisher model, failing the `overlap_count` check in `msp_verify_overlaps()`. Still looking into why this happens, not sure if you want to hold off on merge until that's properly implemented?


algorithms.py
```diff
@@ -266,7 +277,7 @@ def __init__(
             for k in range(sample_size):
                 j = len(self.nodes)
                 x = self.alloc_segment(0, self.m, j, pop_index)
-                self.L.set_value(x.index, self.m - 1)
+                # self.L.set_value(x.index, self.m - 1)
```
jeromekelleher: It looks like this will break the code for the standard coalescent case. We need to make sure both code paths are working here before we merge as other people are developing algorithms using this file.

algorithms.py
```diff
@@ -324,9 +335,59 @@ def free_segment(self, u):
...
+                anc = pop.remove(i)
+                parent = np.random.choice(cur_inds)
+                if parent not in offspring:
+                    offspring[parent] = []
+                offspring[parent].append(anc)
+
+            ## Draw recombinations in children and sort segments by
+            ## inheritance direction
+            for children in offspring.values():
+                need_merge = True if len(children) > 1 else False
```
jeromekelleher: ``need_merge == len(children)`` is more direct here.

algorithms.py
```diff
@@ -851,4 +1003,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    ts = main()
```
jeromekelleher: So yes, it looks like the changes here have broken the standard coalescent code path. We need to make a separate control flow for the WF code (probably adding a command line argument --model=dtwf is the simplest). But we definitely need to be sure that the standard coalescent code path works before merging; as I say, other people are developing algorithms against the code.

lib/dev.cfg
```diff
@@ -20,15 +20,19 @@ model = {
...
+    # psi = 0.99;
+    # c = 0.1;
 
     # The 'Beta' multiple mergers coalescent
     #name = "beta";
     #alpha = 2.2;
     #truncation_point = 10.0;
+
+    # The discrete time Wright Fisher
+    name = "wright_fisher";
```
jeromekelleher: Let's make this 'dtwf' for consistency. (We might want to put in a forwards time wright fisher in the future, for example)

lib/msprime.c
```diff
@@ -1589,6 +1589,86 @@ msp_defrag_segment_chain(msp_t *self, segment_t *z)
...
+    segment_t s1, s2;
+    segment_t *seg_tails[] = {&s1, &s2};
+
+    if ( self->recombination_rate > 0 ) {
+        mu = 1.0 / self->recombination_rate;
+        k = 1 + (uint64_t) x->left +
+            (uint64_t) gsl_ran_exponential(self->rng, mu);
+    } else {
+        mu = INFINITY;
+        k = INT64_MAX;
```
jeromekelleher: This should probably be UINT64_MAX, since k is a uint64_t. I'm not clear on how this interacts with the code below. Would you mind putting in a quick comment here describing what happens in the zero recombination case. 

Come to think of it, why don't we just skip the function entirely when rho = 0?

lib/msprime.c
```diff
@@ -2238,6 +2318,26 @@ msp_initialise(msp_t *self)
 
 /* Given the specified rate, return the waiting time until the next common ancestor
  * event for the specified population */
+/* static double */
```
jeromekelleher: Why us this commented out?

lib/msprime.c
```diff
@@ -2429,6 +2529,127 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
...
+        // Iterate through offspring of parent k, adding to avl_tree
+        for ( k = 0; k < N; k++) {
+            for (s = parents[k]; s != NULL; s = s->next) {
+                node = s->node;
+                x = (segment_t *) node->item;
+                avl_unlink_node(&pop->ancestors, node);
+                msp_free_avl_node(self, node);
+                
+                // Recombine ancestor
+                ret = msp_recombine(self, x, &u[0], &u[1]);
```
jeromekelleher: We should rename this to ``msp_dtwf_recombine``

lib/msprime.c
```diff
@@ -2429,6 +2529,127 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
...
 
+/* List structure for collecting segments by parent */
+typedef struct _segment_list_t {
+    avl_node_t *node;
+    struct _segment_list_t *next;
+} segment_list_t;
+
+/* Performs a single generation under the Wright Fisher model */
+static int WARN_UNUSED
+msp_wright_fisher_generation(msp_t *self)
```
jeromekelleher: Let's rename this to msp_dtwf_geneation

lib/msprime.c
```diff
@@ -2429,6 +2529,127 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
...
+    }
+    if (segment_mem != NULL){
+        free(segment_mem);
+    }
+    return ret;
+}
+
+/* The main event loop for the Wright Fisher model. */
+static int WARN_UNUSED
+msp_run_wright_fisher(msp_t *self, double max_time, unsigned long max_events)
```
jeromekelleher: Rename to ``msp_dtwf_run``. It really helps when methods are named consistently. We should try to make all methods that go along the dtwf code path start with the ``msp_dtwf`` prefix.

lib/simulation_tests.c
```diff
@@ -659,6 +659,72 @@ test_demographic_events(void)
     gsl_rng_free(rng);
 }
 
+static int
+get_num_children(int node, size_t num_edges, edge_t *edges)
+{
+    int num_children = 0;
+    edge_t edge;
+
+    for (int i = 0; i < (int) num_edges; i++) {
```
jeromekelleher: If you declare i as a ``size_t`` then you avoid needing to cast.

lib/simulation_tests.c
```diff
@@ -659,6 +659,72 @@ test_demographic_events(void)
...
+        edge = *(edges + i);
+        if ( edge.parent == (int) node ) {
```
jeromekelleher: More canonical: 
```c
if (edges[j]->parent == node) {
    num_children++;
}
```

jeromekelleher:
Looks like good progress here @DomNelson. Ping me when you would like some more detailed feedback.

DomNelson:
Whew! Ok @jeromekelleher all tests are passing now, but I may have over-complicated things.

I tried to match `test_dtwf_multi_locus_simulation` to its namesake as much as I could, but it wasn't easy to follow the strategy of stepping through single events, since many events can happen simultaneously in discrete time. The loop will stop at the first step than exceeds `max_events`, but to get the actual number of events I added another argument to `msp_run_dtwf`. Then to be able to access *that* value I had to bypass `msp_run` in the testing, which meant adding a few things to `msprime.h`.

Also, the end result is that the check becomes less meaningful, since `num_events` is mostly incremented alongside the values it's being tested against, except for `num_migrations`. Might be a lot of trouble for nothing.

Another testing strategy - comparing number of lineages to the expected number of coalescence events, like in the single locus test - is also tricky, since `num_re_events` doesn't map to number of lineages when back-and-forth recombination is possible. It could be done though, or something checking that lineages are created/destroyed in a way consistent with recombination and coalescence... but I haven't thought of a clean way to do it yet.

What do you think? Any other tests you'd like to see?

jeromekelleher:
Thanks for this @DomNelson, I think we're really close to mergeable now. I've made a quick pass through the code and made a few changes. Mainly I've simplified the interface by removing the requirement that we track the total number of events in the DTWF case. Passing around pointers like this gets complicated quickly, and I don't think it's worth it. All you need for a WF model is to be able to run it for a certain number of generations, so I've changed the interface to do this.

I've changed the tests a bit too to change the testing strategy. All we're doing now is just doing very basic counts on the CA and RE events to make sure they are > 0, and the same when we run with the same seed. I've run out of time to do the updates here, but I hope it won't be to hard for you to fix.

There's one minor stylistic niggle that it would be good to sort out. You have a tendency to sometimes put in extra spaces around conditions like,
```C
if ( some_stuff == other_stuff ) {
}
```
This should be 
```C
if (some_stuff == other_stuff) {
}
```
I've changed some of these, but there are others remaining. 

After that, I think we're mergable!

DomNelson:
Ok @jeromekelleher I think I've got it straightened out. After some digging, there was an issue with `msp_reset()` initializing to a slightly different state, so the even sequential calls to `msp_run(max_time=DBL_MAX)` would give different results starting from the same random seed. Resetting fully from `msp_alloc()` fixed this, and stepping one generation at a time also gives consistent results now.

Anything else you'd like before the merge?

jeromekelleher:
Superb, thanks @DomNelson, we're ready to merge I think. I'm uneasy about msp_reset() not working though. What's different between this and the standard coalescent case?

DomNelson:
One possible reason for a difference between a reset and re-allocation is because `msp_reset()` adds samples to an avl_tree:

https://github.com/jeromekelleher/msprime/blob/master/lib/msprime.c#L2161

and `msp_alloc()` puts them sequentially into an array:

https://github.com/jeromekelleher/msprime/blob/master/lib/msprime.c#L745

To be honest I still don't fully understand the complete initialization process so there may be more going on here, but when iterating through sample IDs at the start of the simulation they are initially ordered [0, 1, 2, ... ] and after a reset they're shuffled as [4, 5, 1, 2, ... ] (approximately). 

If this were just swapping node IDs then no problem, but there seems to be some underlying structure that isn't completely reset, so that the ordering of, say, drawing common ancestor events still follows the original [0, 1, 2, ...] sequence.

This was a pretty baffling issue initially - I printed out the sequences of random numbers drawn and they would be identical up to some point and then start to diverge, and the divergence wasn't even at a consistent time / location in the code. What I think now is that they stay consistent until the re-ordering leads to a different number of recombinations drawn, which can happen if different segments are being merged, and then the whole symmetry breaks down. 

Does this seem plausible? I still haven't tracked down what `msp_reset()` isn't resetting, and how this explicitly leads to different results, and I also haven't worked out whether the consistent results in the coalescent model contradict my theory altogether. I'm going to try a few more tests and see if I can be more specific.

DomNelson:
@jeromekelleher this is a very basic example but could give an idea of what I'm thinking about. In the first generation the nodes are swapped but everything else - recs and merges - is the same. Then in the second generation, things have been shuffled again (now the order is different even given the ID swaps) but the number / size of merges stays the same, leading to diverging results.

I may be missing the forest for the trees here, but this is what got me thinking that the initial states in each case may not be strictly equivalent.

f3fb7df5-8514-4f41-86d2-623b02d790ee
f8dffe23-d57e-4f22-aca1-b9dbafb1d2c3

jeromekelleher:
Sorry @DomNelson, I sent you down a wrong path there. It turns out that there are subtle reasons we can't get exactly the same random path after a reset, which you've found. I've updated the code with some new tests which check for determinism where possible. I'll have another quick look through this and merge ASAP.

jeromekelleher:
OK, merging. Thanks for the great work @DomNelson!

ps. The reason I'm so paranoid about determinism from the simulations is hard experience. There is absolutely nothing worse than not being able to reproduce bugs that occur rarely and stochastically. It sounds odd, but it is **vital** that stochastic simulations are deterministic!

PR #305, opened 2017-11-09
Specific error for recomb maps with nonzero last rate.

jeromekelleher:
Better error for the case where recombination maps are malformed.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -668,7 +668,9 @@ class RecombinationMap(object):
                 # Rate is expressed in centimorgans per megabase, which
                 # we convert to per-base rates
                 rates.append(rate * 1e-8)
-            assert rate == 0
+            if rate != 0:
+                raise ValueError(
+                    "The last rate provided in the recombination map must zero")
         finally:
             f.close()
         return cls(positions, rates)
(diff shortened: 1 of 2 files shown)


agladstein:
In reference to the issue https://github.com/jeromekelleher/msprime/issues/13#issuecomment-343102177
That genetic map is a modified version of the HapMap genetic map so that the positions are relative to the length of the chromosome (I believe). We were using that map for MaCS, but I don't know why it was modified so the rates are relative to the length. We don't need that specific map to work, just a genetic map. 

```
start_pos   end_pos    recom_rate
0.0     0.000108446888448       2.981822
0.000108446888448       0.000130914068846       2.082414
0.000130914068846       0.000800462533344       2.081358
0.000800462533344       0.00204302844717        3.354927
0.00204302844717        0.00204312075605        2.887498
0.00204312075605        0.00204637966111        2.885864
0.00204637966111        0.00205797445999        2.883892
```

jeromekelleher:
Thanks for the explanation @agladstein. I think the simplest thing to do would probably be to use the original HapMap genetic map. I haven't looked at the type of input MaCS accepts.

agladstein:
Sounds good.
Do you have a link to a genetic map that you have used for testing msprime?

jeromekelleher:
These are surprisingly hard to find! I think these are the maps:

https://genome.sph.umich.edu/w/images/0/09/Genetic_map_HapMapII_GRCh37.tar.gz

agladstein:
Thanks so much! We will try it.
@raflopjr

Issue #306, opened 2017-11-09
Add ``variant_matrix`` method for efficiently getting all variants


jeromekelleher:
It's often useful to get all of the variant data out in one go. It's possible to do this with the ``variants`` iterator at the moment, but it's not quite as efficient as it could be. Ideally, we would have a ``TreeSequence.variant_matrix`` method which returns a numpy array filled with the variants. This should use the numpy C API and fill the memory allocated by numpy directly.

jeromekelleher:
Done in #309, but needs documentation.

Issue #301, opened 2017-11-06
Add ``variant.alleles`` for non-binary mutations


jeromekelleher:
Currently we use the ``variants`` iterator like:

```python
V = np.empty((ts.num_sites, ts.num_samples), dtype=np.int8)
for v in ts.variants():
    V[v.index] = v.genotypes
```

So, at the end we have a matrix of zeros and ones in the obvious way for biallalic sites. 

We need to update these semantics in a backward compatible way for more general mutations.  One possibility is to have an array ``v.alleles`` associated with each variant (which gives the text state of each allele), and then view the genotypes array as indexes into this allele array. So, we might have

v.alleles = ["A", "CT", "G"]
v.genotypes = [0, 0, 1, 0, 1, 0, 2, 0]

If the 8 samples, 4 have the "A" allele, 2 have the "CT" allele and 1 has the "G" allele. The ordering is important here because we want to maintain the current semantics for binary mutations. We must have the zeroth allele as the ancestral state. After that, I guess the ordering of the alleles can be arbitrary (or maybe there's a more useful ordering we can provide?).

Questions:

1. What is the most useful ordering to have for alleles?
2. How do we implement this efficiently?

jeromekelleher:
#328 adds the frameworks for this. Close this issue when documentation is updated.

PR #309, opened 2017-11-12
Full genotype matrix

jeromekelleher:
Issue #306 

Much, much more efficient way of getting all the genotypes out of msprime.

Any thoughts on the terminology here @petrelharp? I've called it ``genotype_matrix`` for consistency with the ``variant.genotypes`` array.  Holding off on documenting this until we finalise the actual semantics of the genotypes/alleles as discussed over in #301.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3671,8 +3671,6 @@ out:
     return ret;
 }
 
-
-
 static PyObject *
 TreeSequence_get_num_trees(TreeSequence *self, PyObject *args)
 {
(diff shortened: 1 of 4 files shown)


petrelharp:
Looks very nice.  If I was to ask for one piece of functionality it would be `start' and `end' parameters so that if the whole matrix is too big, we could pull it out in chunks and deal with the genotypes one chunk at a time.

jeromekelleher:
OK, thanks @petrelharp. I'm not sure about the start and end params though. Probably in this case a user would be better off using the ``variants`` iterator.

Merging this even though it hasn't hit the codecov patch diff requirement, as it's not possible to test the code that's not hit.

PR #310, opened 2017-11-15
minor adjustments to simplify

petrelharp:
I removed some spurious logic (replacing with equivalent `assert`).  Perhaps don't merge quite yet as more simpification may come when finishing up writeup in `ftprime_ms`.

### DIFF
--- a/lib/table.c
+++ b/lib/table.c
@@ -2263,6 +2263,13 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
     /* check if the input ID is a sample */
     output_id = self->node_id_map[input_id];
     coalescence = sample_node = self->is_sample[input_id];
+    if (sample_node) {
+        assert(self->ancestor_map[input_id] != NULL);
+        x = self->ancestor_map[input_id];
+        assert(x->left == 0.0);
+        assert(x->right == self->sequence_length);
+        assert(x->node == output_id);
+    }
 
     head_sentinel.left = 0.0;
     head_sentinel.right = 0.0;
@@ -2363,7 +2370,7 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
         z->next = alpha;
         z = alpha;
     }
-    if (self->ancestor_map[input_id] == NULL) {
+    if (!sample_node) {
         self->ancestor_map[input_id] = head->next;
     } else {
         /* There is already ancestral material present for this node, which
@@ -2375,11 +2382,6 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
             simplifier_free_segment(self ,x);
             x = x->next;
         }
-        x = self->ancestor_map[input_id];
-        assert(self->is_sample[input_id]);
-        assert(x->left == 0.0);
-        assert(x->right == self->sequence_length);
-        assert(x->node == self->node_id_map[input_id]);
     }
     if (defrag_required) {
         ret = simplifier_defrag_segment_chain(self, input_id);
--- a/lib/table.c
+++ b/lib/table.c
@@ -2243,9 +2243,9 @@ static int WARN_UNUSED
 simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
 {
     int ret = MSP_ERR_GENERIC;
-    bool coalescence = false;
+    bool coalescence, sample_node;
     bool defrag_required = 0;
-    node_id_t v;
+    node_id_t output_id;
     double l, r, next_l;
     uint32_t j, h;
     avl_node_t *node;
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Looks good. Ping me when it's good to go.

petrelharp:
some more possible improvements?  merge if it looks good to you.

lib/table.c
```diff
@@ -2260,6 +2260,17 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
         goto out;
     }
 
+    /* check if the input ID is a sample */
+    output_id = self->node_id_map[input_id];
+    coalescence = sample_node = self->is_sample[input_id];
```
jeromekelleher: I don't understand this bit. Why is coalescence not always initially false? I'm missing something here I think, as if we can pull this outside the inner loop, then perhaps we should have separate simpler loops to take care of these cases. Perhaps we should deal with the case of ``is_sample`` separately?

Also, I'm not a big fan of multiple assignments (a = b = c). It's clear here what it means, but I have to think about it to make sure it's something more subtle.

lib/table.c
```diff
@@ -2260,6 +2260,17 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
...
+    if (sample_node) {
+        assert(self->ancestor_map[input_id] != NULL);
```
petrelharp: I moved these `assert`s up top from down below since they do not depend on the processing of the edges.

lib/table.c
```diff
@@ -2304,39 +2315,36 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
                     goto out;
                 }
             }
-            v = self->node_id_map[input_id];
-            if (v >= 0 && self->is_sample[input_id]) {
+            if (sample_node) {
```
petrelharp: we want to do this only if this node is a sample

lib/table.c
```diff
@@ -2304,39 +2315,36 @@ simplifier_merge_ancestors(simplifier_t *self, node_id_t input_id)
...
                 if (ret != 0) {
                     goto out;
                 }
                 /* The node on this segment must be remapped to the parent so that we
                  * have the correct nodes further up in the tree. */
-                alpha->node = v;
+                alpha->node = output_id;
             }
         } else {
             if (!coalescence) {
```
petrelharp: whereas this bit we want to do if the node is a sample **or** we've assigned this ancestor an `output_id` already because we found a coalescence here

lib/table.c
petrelharp: Here's the logic:

The `output_id` was previously named `v`, assigned either [here](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2307) or [here](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2330).  It could turn out that this ancestor has an output ID because:

- either we [encounter a coalescence](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2324) 
- or [discover that this input node is a sample](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2318) and so already has an output ID.

The logic above had us checking if `self->node_id_map[input_id]` existed several times in different places, but for essentially the same reason. 

Previously, we did:

- [this bit](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2308) only if the node in question was a sample.  This is the flag `sample_node`.
- [this bit](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2321) only if a coalescence has not been encountered yet **and** the node was [not a sample](https://github.com/jeromekelleher/msprime/blob/master/lib/table.c#L2321).  This is the flag `coalescence`.

The `coalescence` flag is assigned also if `sample_node` is because if it's a sample node, there will be edges output, and hence coalescence.

petrelharp:
i certainly don't mind splitting out that double assignment, obviously.

lib/table.c
jeromekelleher: This is the bit I don't agree with. Why would we record a node if it's a sample? We've already done this in the sample case, haven't we? The coalescence flag is  for the case where a coalescence has happened, and we want to **avoid** recording a node mapping multiple times.

lib/table.c
jeromekelleher: I don't think coalescence flag is right, and could introduce subtle errors. Put it another way, what do we gain by not setting this to ``false``?

The other changes are great though, and definitely clarify. I'm happy to merge if we set ``coalescence = false`` rather than tying it to the sample status of the input node.

jeromekelleher:
OK, I get it now. Very nice! I've pushed an update where we pull more of the logic out of the inner loop and try to make it more explicit. What do you think?

lib/table.c
petrelharp: Sorry, I meant to say that "this bit we **DO NOT** want to do if the node is a sample **or** we've already assigned an ancestor due to previously finding a coalescence." =\

petrelharp:
That looks good to me! I don't know how to evaluate the tradeoff between checking a boolean in the loop versus preassigning the output id, but it's minor either way.  This is nice and clear!

jeromekelleher:
Excellent; merging!

PR #311, opened 2017-11-16
total branch length with multiple roots

petrelharp:
Easy fix.  Maybe slightly slower?


jeromekelleher:
Looks good to me. The number of roots is typically small, so this will be as fast in practise.

Incidentally, if we defined the branch length over a root as zero we wouldn't need to do this. I can't see a good reason not to do this...

Merging this in any case.

PR #313, opened 2017-11-17
compute SFS

petrelharp:
I knocked together SFS calculations by modifying tree_stat_vector.  Everything seems to check out, agreeing with naive python versions, but I'm a bit nervous that I'm not testing things properly.  I think we should merge it, though, if nothing turns up.

@hyanwong - see if this does what you want?

### DIFF
--- a/msprime/stats.py
+++ b/msprime/stats.py
@@ -771,7 +771,7 @@ class BranchLengthStatCalculator(GeneralStatCalculator):
                         pi[edge.child] = edge.parent
                     dx += sign * X[edge.child]
                     dt = (node_time[pi[edge.child]] - node_time[edge.child])
-                    if X[edge.childe] > 0:
+                    if X[edge.child] > 0:
                         L[X[edge.child] - 1] += sign * dt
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Looks good to me. There's a lot of duplicated code here which we should try to refactor at some point; it seems the basic structure of a lot of these algorithms is the same, with only the internal stats updates changing a bit. No hurry for now, but something to think about.

Looks like some real test failures in the CI, so probably a logic bug somewhere though.

petrelharp:
Refactoring: yes, definately!  But, not quite yet.

petrelharp:
Hm, ok - looks like the only problem *may* have been in the 0th entry of the SFS, which was meaningless anyhow.  I've removed that entry, and added more ts to the checks.

I'm leaving the `n`th entry in, though...

petrelharp:
how's that?  @hyanwong, want to take a look?

petrelharp:
i think we can merge this.

jeromekelleher:
Merging.

PR #318, opened 2017-11-20
print node table prints `is_sample` instead of `flags`

petrelharp:
Note that this is still printing the flags; only the column header is changed.  This seems OK.


jeromekelleher:
Looks good, merging.

PR #320, opened 2017-11-22
Release GIL during table operations.

jeromekelleher:
Adds a 'lock' to each table object, which should ensure thread safety in
Python. Any access to a table operate which changes its state should
set these locks before releasing the GIL.

Threads that try to access a table while it's locked will raise and exception.

I think this is safe, but it needs some better tests.

cc @molpopgen


jeromekelleher:
Merging this. FYI @molpopgen, you should now be able to run simplify and sort_tables with the GIL released. Let me know how it goes for you.

molpopgen:
I'll give this a try once we get the ftprime paper submitted.  I don't want to mess with the existing code until then.  Thanks, though--looking forward to testing.

Issue #284, opened 2017-10-19
Provenance and metadata


jeromekelleher:
We need some method to record provenance and metadata for tree sequences. I propose two things:

1. Add a ``ProvenanceTable`` that records where a particular tree sequence came from and what operations have been performed on it.
2. Add a 'metadata' column to (e.g.) ``NodeTable`` and ``SiteTable`` to allow us to record metadata about particular sites and nodes.

This is increasingly important now that a number of different programs can produce tree sequences. It is particularly important when storing real data, which we will be doing shortly.

## ProvenanceTable

There is already a half-hearted effort to so this in mpsrime, where we record a list of 'provenance strings' for each tree sequence. The idea here is to  allow us to store some information that would allow the specific simulation to be reproduced. An example of a provenance string for msprime is
```json
{
  "environment": {
    "python": {
      "implementation": "CPython",  "version": ["2", "7", "12" ]
    },
    "os": {
      "node": "empire",
      "release": "4.4.0-97-generic",
      "machine": "x86_64",
      "version": "#120-Ubuntu SMP Tue Sep 19 17:28:18 UTC 2017",
      "system": "Linux"
    },
    "gsl": {"version": [ 2, 1]},
    "hdf5": { "version": [ 1, 8,16]}
  },
  "version": "0.4.1.dev723+ng503704e",
  "command": "simulate",
  "parameters": {
    "TODO": "encode simulation parameters"
  },
  "software": "msprime"
}
```
The format of the JSON isn't important here, but the idea is for a particular program to encode all the necessary information to reproduce a particular simulation. 

My proposal is to create a ProvenanceTable to explicitly store these provenance strings, and to provide an easy API for adding new rows. This would have two columns:

```
timestamp                                record
2017-10-19T07:08:49+00:00   {"environment"... "command":"simulate", "parameters":{...}, "software": "ftprime"}
2017-10-19T07:08:50+00:00   {"environment"... "command":"simplify", "parameters":{"samples":[0, 1]}, software:"msprime"}
```

Here we ran a simulation in ftprime, and then ran 'simplify' on the resulting tree sequence. The idea is that every time we make a copy of a tree sequence and run some operation on it, we also copy the provenance table and add a new row describing the operation. This should provide a useful ledger of the operations that have been performed to create a particular tree sequence, and in principle, make it fully reproducible.

[...]

PR #322, opened 2017-11-29
Change ragged column lengths to offsets

jeromekelleher:
Given that an actual release is around the corner, we need to fix up any remaining warts on the tables API so that we're not stuck with them. A big wart on the current API IMO is that we are dealing with ragged columns by passing the length of each item as a separate array. This PR changes this to passing an array of offsets instead. This is a more efficient approach, and will allow for easier access of tables by row, as well as simplifying the tree sequence C code.

To be specific, if we have a ragged column like ``site.ancestral_state``, we currently have two arrays, ``ancestral_state`` and ``ancestral_state_length``. I propose change this to ``ancestral_state`` and ``ancestral_state_offset``. Concretely, if we had 3 ancestral states ['A', 'CCAC', 'G'], the current encoding is
```python
ancestral_state = [A, C, C, A, C, G]
ancestral_state_length = [1, 4, 1]
```
and the proposed encoding is
```python
ancestral_state = [A, C, C, A, C, G]
ancestral_state_offset = [0, 1, 5, 6]
```
Note that 

1. ancestral_state_offset[0] is always equal to zero
2. The length of ancestral_state_offset is num_rows + 1

This formulation means that we can find the ancestral state for a given row ``j`` using
```python
a = ancestral_state[ancestral_state_offset[j]: ancestral_state_offset[j + 1]]
```

The code here is a WIP on the C side of things, where there are quite few tedious changes. However, I do think that this is the right approach for ragged columns, and we should fix this before we release 0.5.0. 

Any thoughts @petrelharp?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,15 @@
 
 SRC=_msprimemodule.c
 
+ext3: ${SRC}
+	python3 setup.py build_ext --inplace
+
 ext2: ${SRC}
 	python setup.py build_ext --inplace
 
 ext2-coverage: ${SRC}
 	rm -fR build
 	CFLAGS="-coverage" python setup.py build_ext --inplace
-
-ext3: ${SRC}
-	python3 setup.py build_ext --inplace
 
 figs:
 	cd docs/asy && make 
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -2731,7 +2731,7 @@ ProvenanceTable_init(ProvenanceTable *self, PyObject *args, PyObject *kwds)
         PyErr_NoMemory();
         goto out;
     }
-    /* Take the default size increments for timestamp and provenance */
+    /* Take the default size increments for timestamp and record */
     err = provenance_table_alloc(self->provenance_table,
             (size_t) max_rows_increment, 0, 0);
     if (err != 0) {
@@ -2750,19 +2750,19 @@ ProvenanceTable_add_row(ProvenanceTable *self, PyObject *args, PyObject *kwds)
     int err;
     char *timestamp = "";
     Py_ssize_t timestamp_length = 0;
-    char *provenance = "";
-    Py_ssize_t provenance_length = 0;
-    static char *kwlist[] = {"timestamp", "provenance", NULL};
+    char *record = "";
+    Py_ssize_t record_length = 0;
+    static char *kwlist[] = {"timestamp", "record", NULL};
 
     if (!PyArg_ParseTupleAndKeywords(args, kwds, "s#s#", kwlist,
-                &timestamp, &timestamp_length, &provenance, &provenance_length)){
+                &timestamp, &timestamp_length, &record, &record_length)){
         goto out;
     }
     if (ProvenanceTable_check_state(self) != 0) {
         goto out;
     }
     err = provenance_table_add_row(self->provenance_table,
-            timestamp, timestamp_length, provenance, provenance_length);
+            timestamp, timestamp_length, record, record_length);
     if (err != 0) {
         handle_library_error(err);
         goto out;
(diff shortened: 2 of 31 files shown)


jeromekelleher:
> My thoughts: ￼ ￼ ￼
> Too bad about the tedium.

OK, cool, I'll chip away at this so. A small part of me was hoping you'd come up with a good reason for why this is a terrible idea and I wouldn't have to go through the tedium!

petrelharp:
Sadly, no.  At least with the switch to edges we won't want to do this for
`children` also...

jeromekelleher:
OK, I think this is covers the changes needed to replace `x_length` columns with `x_offset` columns in the tables API. This also affects the low-level tree sequence code, and makes dealing with these columns substantially less horrible.

The practical upshot of this change for users is that it will break people's code (those that are tracking master anyway and using sites/mutation tables), and will make their tree sequence files unreadable (unless we write a converter). In the interest of minimising this disruption, I think I should add the remaining changes that we need before merging this PR. These are:

1. Change the (unused) EdgeTable.name column to ``metadata``.
2. Add a `metadata` column to the site table.
3. Add a ProvenanceTable.

These changes were discussed in #284.

I need 1, 2 and 3 for the stuff I'm doing with real data. I'm happy to add a ``metadata`` column to MutationTable too, but it seems less urgent to me and easy to tack on in future if it's needed.

Any thoughts @petrelharp, @ashander, @molpopgen?

petrelharp:
Sounds good.  Just so I'm clear, the point of the ProvenanceTable is to say
what version of msprime produced the tree sequence?

Do you have a README accumulating of important changes?  It'd be easy to at
least describe what you'd need to do to convert.

jeromekelleher:
> Sounds good.  Just so I'm clear, the point of the ProvenanceTable is to say what version of msprime produced the tree sequence?

It's to record the version and environment information for *every* program that has produced a particular tree sequence. Suppose I ran a large simulation with ftprime, and then made a tree sequence of a subset of it. Then, the provenance table should have two rows (or maybe three): rows. One recording version information about ftprime, and parameters used to generate the simulation, and a second recording the version of msprime used to run ``simplify``. It's to give an audit trail for where a tree sequence came from, and to make them reproducible (in principle at least; I wouldn't get too hung up on actually making it happen too soon).

I'm feeling the need for this at the moment, as I generate tree sequences from data, (with lots of intermediate states), and it would be very handy to be able to see where they came from.

> Do you have a README accumulating of important changes?  It'd be easy to at least describe what you'd need to do to convert.

No --- it'd be quicker to write a converter to be honest. Are there many people working with version 9 files (that can't be easily reproduced) do you think? I've been careful to make sure that released file format versions are upgradeable, but I've got through 6 or 7 versions since the last release.

ashander:
> It's to record the version and environment information for every program that has produced a particular tree sequence. Suppose I ran a large simulation with ftprime, and then made a tree sequence of a subset of it. Then, the provenance table should have two rows (or maybe three): rows. One recording version information about ftprime, and parameters used to generate the simulation, and a second recording the version of msprime used to run simplify. It's to give an audit trail for where a tree sequence came from, and to make them reproducible (in principle at least; I wouldn't get too hung up on actually making it happen too soon).

This seems useful. I'm currently running into reproducibility issues, which I think are downstream from simplify. But while chasing them down it has crossed my mind that simplify could cause trouble. Knowing were a given sequence comes from would be nice to eliminate some potential issues, so +1! The format in #284 seems reasonable.

> Are there many people working with version 9 files (that can't be easily reproduced) do you think? I've been careful to make sure that released file format versions are upgradeable, but I've got through 6 or 7 versions since the last release.

Not me. I can reproduce or use old msprime versions where needed. 

Speaking of this, though, it would be nice to tag or otherwise document the file format versions even on the master branch. (At least going forward.) This would aid in finding the last commit that works with a given file format. I had to do this recently with git bisect and `python -c 'import msprime; msprime.load("mytree.hdf5")' but it was tedious. Now that I look I see going backwards through `git blame` on [these lines](https://github.com/ashander/msprime/blame/8d659b6c407a04a63b7d2e261a8a911a8346afb2/lib/msprime.h#L40-L41) would've been a better bet

petrelharp:
> Are there many people working with version 9 files 

Nothing that can't be easily re-done that I know of...

jeromekelleher:
> Speaking of this, though, it would be nice to tag or otherwise document the file format versions even on the master branch. (At least going forward.) This would aid in finding the last commit that works with a given file format. I had to do this recently with git bisect and python -c 'import msprime; msprime.load("mytree.hdf5")' but it was tedious. Now that I look I see going backwards through git blame on these lines would've been a better bet

Yes, that is annoying, sorry about that. These tags are a good idea, and could be done retrospectively too using git blame on the lines you pointed out.

jeromekelleher:
Phew, this has turned into a biggie. I think I'm nearly done now though, just need to add the `metadata` column to the SiteTable and that should be it. 

Provenance handling is hugely improved now, and should actually be useable (although the details of what's recorded needs to be worked out).

jeromekelleher:
OK, I think this is finally ready to go.  This updates the HDF5 file format version to 10,  and so any existing version 9 files will have to be recreated. Pinging @petrelharp, @ashander, @molpopgen and @hyanwong to alert you to this.

Here are the changes:

1. Changed all string columns to use the colname, colname_offsets convention as described above. 
2. Added ProvenanceTable, which contains two string columns, ``timestamp`` and ``record``.
3. Changed ``Node.name`` to ``Node.metadata``, and added ``Site.metadata`` and ``Mutation.metadata``.
4. Changed ``sort_tables`` to return the ``node_map`` array rather than pass it in as an argument (if you don't use node_map, then you're code isn't affected). This was motivated by the difficulties with releasing the GIL.

There is various work left to be done on the metadata and provenance side of things (mostly: update the text versions and document the new APIs), but the basic infrastructure is now in place and I'd like to merge.

petrelharp:
yay!

jeromekelleher:
OK, merging. One more warning that version 9 files will no longer work @petrelharp, @ashander, @molpopgen and @hyanwong!

Issue #253, opened 2017-08-29
Change default behaviour of ``variants`` iterator to return a copy of the genotypes data


jeromekelleher:
It's currently a nasty gotcha when people store a reference to the ``genotypes`` array, not realising that the same array is used for every site. We should change the default behaviour here as it has tripped too many people up to be worth the default performance gains. We can add a flag to keep the present behaviour for those that want it.

jeromekelleher:
#328 changes this to always return a new object. It seems unlikely that anyone actually depends on the old semantics, so it seems OK to change.

Issue #169, opened 2017-03-27
Binary mutations not 


hyanwong:
When I have some ancestral states as 1 not 0, mutations are not considered binary. E.g. with the attached files, try:

```
ts = msprime.load_text(nodes=open("TSnodes.txt"), edgesets=open("TSedges.txt"), sites=open("TSsites.txt"), mutations=open("TSmuts.txt"))
for v in ts.variants(as_bytes=True):
 print(v.position, v.genotypes.decode(), sep="\t")
```

I get

`_msprime.LibraryError: Only binary mutations are supported for this operation.a`

[TSedges.txt](https://github.com/jeromekelleher/msprime/files/871821/TSedges.txt)
[TSmuts.txt](https://github.com/jeromekelleher/msprime/files/871820/TSmuts.txt)
[TSnodes.txt](https://github.com/jeromekelleher/msprime/files/871822/TSnodes.txt)
[TSsites.txt](https://github.com/jeromekelleher/msprime/files/871823/TSsites.txt)

jeromekelleher:
Thanks, I'll sort this out. In the mean-time you can get at the sequences using the haplotypes() method (which is basically the transpose of the variants).

jeromekelleher:
I remember now why I didn't finish this. I've not quite figured out what the interface for variants() should be in the general case (arbitrary alphabet, lots of complicated mutations and > 2 alleles), so I postponed implementing it. (For simplicity, I'm treating anything that doesn't have 0 as ancestral and 1 as derived as nonbinary.)

Do you need the variants method, or will haplotypes do the trick?

hyanwong:
I can use haplotypes() by sorting the array and transposing it. But how to I get the positions of all the variants?

jeromekelleher:
This is the easiest way:
 ```python
        A = np.zeros((n, m), dtype='u1')
        for j, h in enumerate(ts.haplotypes):
            A[j] = np.fromstring(h, np.uint8) - ord('0')
```

hyanwong:
Can this now be closed?

jeromekelleher:
No, this is still not sorted. Hopefully we'll resolve these issues using the approach outline in #301.

jeromekelleher:
Sorted out now in #328.

PR #328, opened 2017-12-19
Add 'alleles' list to Variant object.

jeromekelleher:
Issue #301

This should be the last structural change before 0.5.0, and put us in a position to finalise everything to do with sites and mutations.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -6060,6 +6060,9 @@ VariantGenerator_next(VariantGenerator *self)
         goto out;
     }
     if (err == 1) {
+        /* Copy the genotypes into the local buffer */
+        memcpy(self->buffer.buf, var->genotypes,
+                self->variant_generator->num_samples * sizeof(char));
         ret = make_site(var->site);
     }
 out:
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -536,25 +536,53 @@ out:
     return ret;
 }
 
+#ifdef HAVE_NUMPY
+
+static PyObject *
+make_alleles(variant_t *variant)
+{
+    PyObject *ret = NULL;
+    PyObject *item, *t;
+    size_t j;
+
+    t = PyTuple_New(variant->num_alleles);
+    if (t == NULL) {
+        goto out;
+    }
+    for (j = 0; j < variant->num_alleles; j++) {
+        item = Py_BuildValue("s#", variant->alleles[j], variant->allele_lengths[j]);
+        if (item == NULL) {
+            Py_DECREF(t);
+            goto out;
+        }
+        PyTuple_SET_ITEM(t, j, item);
+    }
+    ret = t;
+out:
+    return ret;
+}
+
 static PyObject *
 make_variant(variant_t *variant, size_t num_samples)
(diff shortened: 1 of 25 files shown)


jeromekelleher:
This is getting ready to merge, and is real progress towards handling general mutational patterns effectively. To recap, the basic idea is that we now have an ``alleles`` tuple that is attached to the variant object returned by ``variants``. The values in the ``genotypes`` arrays are then indexes into this array. So, if we have a site with observations (indexed by sample order) ["AA", "C", "T", "G"] and the ancestral state was "G", then we'd have 
```python
variant.alleles == ("G", "AA", "C", "T")
variant.genotypes == [1, 2, 3, 0]
```
The ordering of the alleles is decided by the ordering of the underlying mutations. The ancestral state will always be the zeroth allele. This seems quite nice because we can now support any arbitrary pattern of mutations at a given site, but without breaking anyone's existing code that assumes binary mutations.

I've also updated the ``variants()`` iterator to return a new object every time for the genotypes array (see #253). 

Another tweak I made was to remove the mutation generator from the WF simulator. This was causing problems because the mutations were not consistent (i.e., we could have A->A transitions), and it seemed like a lot of work to make them so. It seemed a lot simpler to just follow our own advice and use the Jukes-Cantor generator on the resulting tree sequence.

I need to do a few more tests just to be completely that it all works for some corner cases, but I think it's ready to merge after that.

@petrelharp, do you have any thoughts here?

jeromekelleher:
For the record, this should close 
- #253
- #301 
- #169

petrelharp:
This sounds good, but I'd like to be clear about the problems with the WF mutation generator.  Where do  problems occur if there are A->A transitions?

jeromekelleher:
> This sounds good, but I'd like to be clear about the problems with the WF mutation generator. Where do problems occur if there are A->A transitions?

When generating haplotypes and variants, we check if the derived state has already been set for each sample it subtends, and raise an error if it does. See

https://github.com/jeromekelleher/msprime/blob/variant-alleles/lib/vargen.c#L154

It's debatable whether this is a good idea, but the original thinking (ie., specifying an A->A transition) would be the result of a mistake, and so it's a useful error checking feature. The reason this wasn't an issue before is because the haplotype generator mistakenly wasn't checking for this, in this general string case.

I'd imagine that we can add a flag to the variant and haplotype generators to turn off this error checking if we do want to allow these mutations.

jeromekelleher:
Ah, looking at this code again and I realise that we are assuming a (weak) tree pre-ordering for the mutations. So, the question is, do we continue to rely on this ordering, or rejigger things a bit to use the mutation.parent field?

petrelharp:
re: A->A transitions - I fall on the side of allowing mutations that don't change anything, if this doesn't actually break anything, but could go either way.

Which code is assuming the tree ordering?

petrelharp:
About assuming ordering in `vargen_update_site` - I agree, it is correct.  Should perhaps insert a comment saying that we use this fact here.

jeromekelleher:
OK, merging this. I've moved discussion of the degenerate mutations issue over to #329.

Issue #332, opened 2017-12-21
Text table formats for awkward data


jeromekelleher:
One of the last remaining details to clear up before 0.5.0 is to decide how to handle the text data format for recent extensions. There are a few thorny issues:

1) How do we encode the metadata column? In principle this contains binary data, so we cannot just print it out as an utf-8 stream. On the other hand, if the column does contain text, it's a lot less useful to see a base64 encoded string than your actual JSON metadata. I think the best option here is to provide a boolean flag to force binary encoding of the metadata, but assume that the metadata is text by default. 

2) How do we deal with empty strings for ancestral and derived state (and indeed, metadata)? These are perfectly valid values now, and will cause problems with a white-space delimited file. The only real solution I can see here is to add a option to insist on tab-delimiting (allowing us to detect zero length strings), but to allow more lenient white-space delimiting by default, hopefully failing reasonably gracefully when a file with zero-length strings is provided.

Any thoughts on this @petrelharp, @ashander, @molpopgen?

petrelharp:
Text by default = good.

Empty strings: the problem is in text-based representations of site and mutation tables?  The only other option would be to quote everything, which would be annoying; I support strict tab-delimited files.

molpopgen:
Are we talking about representation in files, or when printed to the screen?  For the metadata, any text-based representation in files has to be a non-starter, as it would cause reasonable unit tests to fail due to the presence of floating-point values in the original data.

I'm imagining that meta-data will always be binary.  I really cannot see using JSON for this, as it has big overhead for lots of instances of small objects.   JSON would also have the same problems in causing test failures due to precision loss.  You would not be able to iterate over your simulated population and assert that you can find every mutation position or selection coefficient in your list of JSON objects.  Such a task is guaranteed to fail at some point.

I think some functional programming ideas would be handy here.  For example, asking the user to supply a callback that converts the binary metadata into something human-readable:

```python
def metatdata2txt( data ):
      x = pickle.loads(data)
      return x.__repr__() # or str(x), or some "{} {}".format() thing....
```

jeromekelleher:
Thanks for this @molpopgen, these are convincing arguments. Lets base64 encode metadata by default then for text formatting. We can add options for custom decoding for display later, but by default we should ensure the integrity of the data.

molpopgen:
[...]

PR #333, opened 2018-01-01
Text metadata

jeromekelleher:
This PR deals with the awkward corner cases for text formats discussed in #332, and finalises the text formats for 0.5.0. Changes:

1. Makes metadata ASCII safe by base64 encoding the data. This has the downside of making text metadata unreadable, but we can deal with that in the future by providing custom metadata decoders. For now, this approach ensures the integrity of of binary metadata, which is more important.
2. Deals with empty strings for ancestral and derived state by providing a ``sep`` argument to ``load_text``. If ``sep`` is not provided, we use a relaxed whitespace delimited method (which is the default). If ``sep`` is specified, a strict delimiting algorithm is used. (This is done by providing the sep argument directly to str.split, so we inherit these semantics from Python's built in method). The upshot is that we can reliably round-trip data by using ``sep="\t"`` in ``load_text``. We can also use other values as delimiters too (except comma, annoyingly, because of a left-over from edgeset parsing).

Once this is merged, I think the tables API is ready for documenting and release.

Any thoughts @petrelharp, @molpopgen?

### DIFF
--- a/lib/table.c
+++ b/lib/table.c
@@ -1667,14 +1667,8 @@ typedef struct {
     site_table_t *sites;
     mutation_table_t *mutations;
     migration_table_t *migrations;
-    /* sorting edges */
-    edge_sort_t *sorted_edges;
-    /* sorting sites */
+    /* Mapping from input site IDs to output site IDs */
     site_id_t *site_id_map;
-    site_t *sorted_sites;
-    mutation_t *sorted_mutations;
-    char *ancestral_state_mem;
-    char *derived_state_mem;
 } table_sorter_t;
 
 static int
@@ -1736,11 +1730,7 @@ table_sorter_alloc(table_sorter_t *self, node_table_t *nodes, edge_table_t *edge
     self->mutations = mutations;
     self->sites = sites;
     self->migrations = migrations;
-    self->sorted_edges = malloc(edges->num_rows * sizeof(edge_sort_t));
-    if (self->sorted_edges == NULL) {
-        ret = MSP_ERR_NO_MEMORY;
-        goto out;
-    }
+
     if (self->sites != NULL) {
         /* If you provide a site table, you must provide a mutation table (even if it is
          * empty */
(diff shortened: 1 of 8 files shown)


jeromekelleher:
Bumping this for @petrelharp --- I'm not sure if mentions in an edited comment trigger a notification. I think this is the last real code update before the doc-a-thon.

jeromekelleher:
> Except one minor thing: my inclination is to make sep ="\t"the default, at least in load_text if not in the various parse_* functions. This is because (a) having sep=None seems like an annoying gotcha when implementing insertions; (b) it is more intuitive to have round-trip-idempotent as the default; and (c) it seems easier to debug the case where you have sep="\t" but need a different value of sep (here the error should be apparent on the first line) than the case where you have sep=None but need sep="\t" (here the error may only appear far down the file).

Good points... I guess the reason for making sep=None by default was to avoid breaking existing code (including a large chunk of tests). My feeling is that insertions would be a relatively rare use case, and having a more relaxed algorithm by default was more useful. Also, having to specify ``sep=None`` to access the more convenient whitespace based splitting algorithm seems counterintuitive.

I agree with your arguments though. Perhaps a different API would be better? I don't see any value in allowing separators other than "\t" and general whitespace, so perhaps we should make the argument be ``strict=True`` by default?

petrelharp:
Good point about `None`, but no need to remove ability to use other separators, I'd say. I'd vote for one of the following two things:

1. Leave as-is, but make sure to say in the `load_text` documentation that if you've got indels, you must specify the `sep` argument.
2. or... Change the default in `load_text` to `sep="\t"`, add `sep=None` to everywhere it gets called in the tests (as necessary), and say in the documentation that `sep=None` gives you whitespace-splitting but it breaks on indels.

jeromekelleher:
I'd lean more towards (1), but have one more thing to bring up:

> Good point about None, but no need to remove ability to use other separators, I'd say. I'd vote for one of the following two things:

The annoying thing about allowing other separators is that we can't allow comma (which would surely be the most commonly used after tabs) because of our vestigial support for edgesets. The code will break in   mysterious ways at the moment if someone tries to use comma as the delimiter. Also, we should consider the metadata, which is base64 encoded. So, if someone tried to use +, = or / as the delimiter, things would also break in mysterious ways. We don't have a lot of useful delimiters left after that, and we have quite a few corner cases to check if we're going to code it up in a proper defensive manner. There are probably other subtle gotchas lying in wait too.

So, I don't really see what we gain by allowing a general 'sep' argument. I think that just allowing whitespace and having a ``strict`` option (defaulting to true) would be quite a bit simpler and solve the problems you've pointed out.

petrelharp:
Ah, I see.  So:

1. as above but also mention that `sep` can only be whitespace, or
2. Have the two options be `strict=True|False`, with `True` meaning `sep="\t"` and `sep=None` otherwise; with `strict=True` by default.

... but, (1) sounds good.

molpopgen:
Is it possible to "skip" the meta data upon printout?  What I mean is, in the absence of a custom decoder, take a cue from the default `__repr__` for Python instances of C/C++ types and simply write the meta-data as:

`<msprime.TableType.metadata_gibberish at XXX>`

where XXX may be the index of the gibberish in the table.  Then, if a custom decoder were present, you may simply output its return value.

This may help prevent pages of unreadable text being vomited out at the end user.

jeromekelleher:
> Is it possible to "skip" the meta data upon printout? What I mean is, in the absence of a custom decoder, take a cue from the default __repr__ for Python instances of C/C++ types and simply write the meta-data as:

Well there's a distinction here between printing out for human viewing and interchanging. This PR is all about interchange, where we need to be able to round-trip the metadata through the text format so we have to always base64 encode it. I think you're right about this for the ``__str__`` implementation though.

jeromekelleher:
Hmmm. I still don't see what you see as the benefit of having an arbitrary ``sep`` argument though @petrelharp... Rather than hold up this PR (and the documentation process) I'm going to merge what we have move the discuss to issue #337.

Issue #323, opened 2017-11-29
Tracking updates to documentation for tables api


ashander:
below done in #334 
- [x] NodeTable and EdgeTable documentation (some done in #321)
- [x] tables I/O (eg `load_tables` , `dump_tables`) 
- [x] tables manipulation (eg `simplify_tables`) --- #287
-  [x] define sort order (done in #321)

still todo
- [x] update error messages https://github.com/jeromekelleher/msprime/issues/280
- [x] add requirements related to mutations (see https://github.com/jeromekelleher/msprime/issues/291 )

 (feel free to add suggestions below)

ashander:
also
-  [x] define sort order (done in #321)
- [ ] update error messages https://github.com/jeromekelleher/msprime/issues/280

ashander:
and

- [ ] add requirements related to mutations (see https://github.com/jeromekelleher/msprime/issues/291 )

PR #334, opened 2018-01-02
Docstring updates: load_tables, simplify_tables, EdgeTable

ashander:
contributing to #323

### DIFF
--- a/msprime/tables.py
+++ b/msprime/tables.py
@@ -119,14 +119,11 @@ def _pickle_node_table(table):
 
 class EdgeTable(_msprime.EdgeTable):
     """
-    TODO Update docs to relfect EDGES.
-
-
     Class for tables describing all edges in a tree sequence, of the form
         left	right	parent	child
         0.0     0.4     3       0
         0.0     0.4     3       2
-        0.4     1.0     3       0,
+        0.4     1.0     3       0
         0.4     1.0     3       1
         0.4     1.0     3       2
         0.0     0.4     4       1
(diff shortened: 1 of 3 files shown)


ashander:
looks like I'll need to add some tests to load_tables.

msprime/tables.py
```diff
@@ -603,15 +600,14 @@ def sort_tables(*args, **kwargs):
...
-    .. todo:: Update this documentation to describe the keyword arguments and
-       combinations that are allowed.
-
-    :param NodeTable nodes:
-    :param EdgeTable edges:
-    :param MigrationTable migrations:
-    :param SiteTable sites:
-    :param MutationTable mutations:
-    :param int edge_start: The index in the edge table where sorting starts.
+    :param NodeTable nodes (required).
```
jeromekelleher: I don't this this is valid sphinx markup. It should be something like
```
:param NodeTable nodes: The tree sequence nodes (required).
```
See http://www.sphinx-doc.org/en/stable/domains.html for examples of the required markup formats.

msprime/tables.py
```diff
@@ -649,6 +646,7 @@ def simplify_tables(*args, **kwargs):
     :param MutationTable mutations: The MutationTable to be simplified.
     :param bool filter_invariant_sites: Whether to remove sites that have no
         mutations from the output (default: True).
+    :param sequence_length: The length of the sequence.
```
jeromekelleher: Need a type here (float I guess?)

msprime/trees.py
```diff
@@ -818,8 +818,41 @@ def load(path):
...
+def load_tables(nodes, edges, sequence_length=None, migrations=None, sites=None,
+                mutations=None, provenances=None):
+    """
+    Loads a tree sequence from the table objects provided.  If sequence_length
+    is 0 or not provided, it is inferred to be equal to the largest right value
+    in the edges.
+
+    :param NodeTable nodes: The NodeTable.
+    :param EdgeTable edges: The EdgeTable.
+    :param sequence_length: The length of the sequence.
```
jeromekelleher: Type missing.

msprime/trees.py
```diff
@@ -818,8 +818,41 @@ def load(path):
...
+    :param MigrationTable migrations: The MigrationTable.
+    :param SiteTable sites: The SiteTable.
+    :param MutationTable mutations: The MutationTable.
+    :param ProvenanceTable provenances: The ProvenanceTable.
+    :return: A :class:`msprime.TreeSequence` consistent with the tables.
+    :rtype: TreeSequence
+    """
+    if nodes is None:
```
jeromekelleher: I don't think there's much point in doing these checks here, as it's all done already by the low-level module. I think it's better to just stick with the existing way of packing the (*args, **kwargs) --- otherwise, we're doing the same thing twice in two places.

jeromekelleher:
Actually, I just spotted the code changes to sort_tables, and I think we should revert these before merging. I realise the *args, **kwargs approach isn't ideal for Sphinx documentation, but we can figure out some workaround for this.

ashander:
Made these changes and built docs locally without issues. Thanks!

ashander:
Oops, I squashed but then just realized we have an existing solution in sphinx to list the arguments for `*args, **kwargs` functions. I updated that just now for simplify_tables

jeromekelleher:
Merging. Thanks @ashander!

PR #335, opened 2018-01-02
Update docs for tables API.

jeromekelleher:
This is a work in progress, which I'm opening as a PR for early feedback.

The documentation for the tables API (and the formal tree sequence requirements) are all currently on the File Formats page. This PR splits the various bits:

- Move all the actual function and class definition docs to the API page
- Move the 'working with tables' bit to the tutorial
- Change the 'file formats' page into the 'Tree sequence interchange' page, where we define the tree sequence requirements and discuss the high-level  stuff about the various interchange methods.

The idea is to link them all together using hyperlinks (the structure of this is only starting to come together), so we can refer to (for example) the basic requirements for tree sequences without spelling them out several times.

How does this sound to you @ashander and @petrelharp? Any violent (or mild!) disagreements with this approach?


petrelharp:
Sounds good. I am fuzzy on the implementation. Will the docs still be there in `help( )` in python?

docs/interchange.rst
```diff
@@ -0,0 +1,478 @@
...
+   right)``.
+2. Has the ``population`` of any such ancestor matching ``source``, if another
+   ``migration`` does not intervene.
+
+
+******************
+Text table formats
+******************
+
+.. todo Stuff about tables API.
```
ashander: er, about the text formats? (that is `TreeSequence.dump_text` and `TreeSequence.load_text`)

docs/api.rst
```diff
@@ -188,3 +188,23 @@ population genetics statistics from a given :class:`.TreeSequence`.
...
+
+***********
+Tables API
+***********
+
+.. autofunction:: msprime.sort_tables(nodes, edges[, migrations, sites, mutations, edge_start])
+
+.. autofunction:: msprime.simplify_tables(samples, nodes, edges[, migrations, sites, mutations])
+
+
```
ashander: add `msprime.tree_sequence.load_text` and `msprime.tree_sequence.load_tables`

ashander:
> Sounds good. I am fuzzy on the implementation. Will the docs still be there in help( ) in python?

Yes. I don't think Jerome's proposing stripping out any docstrings. Also as I commented, some of the material in docstrings can and should be better integrated into the docs.

jeromekelleher:
Excellent. Thanks for the feedback @ashander, we're definitely on the same page on where this is going. I'll see what I get done today and merge it, so that the new basic structure is in place.

jeromekelleher:
Unfortunately, I've run out of time here and things are still in flux.

The good news is I've figured out how to make the docstrings for the table classes show up. I've partially done the docs for the NodeTable class. The idea is that we move all the definitional stuff away from the docstrings, but we link to all the relevant information. The docstrings then document the actual attributes and methods of the class.

The docs are live at http://msprime.readthedocs.io/en/docathon/ so you can see how it's developing. See:

http://msprime.readthedocs.io/en/docathon/api.html#tables-api

http://msprime.readthedocs.io/en/docathon/interchange.html#data-model

For works in progress. The idea is to put all the definition stuff into the "Data model" section. The layout of this section is still in flux, but hopefully the basic idea is clear.

Unfortunately I'm at a conference for the next few days, so I won't get a chance to do any real work on this. Comments are welcome in the mean time though (and I'm happy to merge as-is if someone else wants to take this up).

ashander:
Looks like a great start! The overall plan makes sense to me. I won't have
time to take it up before a few days from now anyway but will look through
and leave some comments if needed

jeromekelleher:
I've made some updates here to the documentation for NodeTable, which I think is almost right:

http://msprime.readthedocs.io/en/docathon/api.html#msprime.NodeTable

The external links for the general stuff haven't been filled in yet, but I think the general outline for the API docs for this table should be about right now. Would you mind taking a look through this (at the link above) @petrelharp, @ashander and @molpopgen?

I've added row-access for the NodeTable here because it felt too tricky to explain what a table is without having this concept. The example I've put in here perhaps belongs somewhere else, but I don't know how to communicate the fact that we support ``len``, ``str``, ``_getitem__``, negative indexing and so on without doing something like this. Any ideas? I guess we could have a general discussion at the top of the section that does this, using one example table.

Any input would be much appreciated, as I'll be using what's here as a template for the rest of the tables, and I'd much rather sort out any issues before applying the template!

ashander:
Sure thing. I will take a look today

petrelharp:
I can look after Jaime; tomorrow afternoon?

msprime/tables.py
```diff
@@ -53,33 +54,86 @@ def text_decode_metadata(encoded):
...
+    :vartype metadata_offset: numpy.ndarray, dtype=np.uint32
     """
+
     def __str__(self):
         time = self.time
         flags = self.flags
         population = self.population
         metadata = unpack_bytes(self.metadata, self.metadata_offset)
-        ret = "id\tis_sample\tpopulation\ttime\tmetadata\n"
+        ret = "id\tflags\tpopulation\ttime\tmetadata\n"
```
ashander: why s/is_sample/flags here, and throughout? I had found the earlier change to `is_sample` clarifying

msprime/tables.py
```diff
@@ -53,33 +54,86 @@ def text_decode_metadata(encoded):
...
+    :ivar time: The array of time values.
+    :vartype time: numpy.ndarray, dtype=np.float64
+    :ivar flags: The array of flags values.
+    :vartype flags: numpy.ndarray, dtype=np.uint32
+    :ivar population: The array of population IDs.
+    :vartype population: numpy.ndarray, dtype=np.int32
+    :ivar metadata: The flattened array of binary metadata values.
+    :vartype metadata: numpy.ndarray, dtype=np.int8
+    :ivar metadata_offset: The array of offsets into the metadata column.
+    :vartype metadata_offset: numpy.ndarray, dtype=np.uint32
```
ashander: should http://msprime.readthedocs.io/en/docathon/interchange.html#sec-encoding-ragged-columns be cross-referenced in the class-level docstring too?

msprime/tables.py
```diff
@@ -53,33 +54,86 @@ def text_decode_metadata(encoded):
...
+        2
+
+    :warning: The numpy arrays returned by table attribute accesses are **copies**
+        of the underlying data. In particular, this means that you cannot edit
+        the values in the columns by updating the attribute arrays.
+
+        **NOTE:** this behaviour may change in future.
+
+    :ivar time: The array of time values.
+    :vartype time: numpy.ndarray, dtype=np.float64
```
ashander: These and other docs below do a nice job laying out the dtypes. This info is duplicated in several places. 

An alternative is to put the type this info in code somewhere, then and referring? Something like below (slightly abusing the namedtuple) which could be a static member of `NodeTable`.

```py
 NodeTableTypes = NodeTableRow(flags=np.uint32, time=np.float64,
                                 population=np.int32, metadata=np.int8)
```

At the end of the day, though, keeping the class, row, and 2 column interfaces up to date is maybe less than figuring out and maintaining a code API to the type info

molpopgen:
I'll take a look tomorrow.

msprime/tables.py
jeromekelleher: There's a distinction here between printing out the **table** and the text IO format. For the text IO format I think it's right that we have ``is_sample``, but it doesn't make any sense to have ``is_sample`` here and I think it would be very confusing. Users can put whatever they want into the flags column, and they should see the same value coming back.

msprime/tables.py
jeromekelleher: Yeah, I was trying to figure out something like this. The short version is the user will want to have the dtype information on hand, and it's too much trouble trying to figure out an automatic way of doing it. The dtypes can't change very much, so hopefully it won't be a maintenance problem.

msprime/tables.py
jeromekelleher: Good idea. I'll see what I can do.

jeromekelleher:
OK, more updates here. I think the basic structure is looking good now. I've updated various things to make  adding documentation easier, and I hope the frameworks are mostly in place. 

The [data model](http://msprime.readthedocs.io/en/docathon/interchange.html#data-model) section is still a mess and needs a structure and narrative.

I'm reasonably happy with the [loading data](http://msprime.readthedocs.io/en/docathon/api.html#loading-data) section, as this gets together all the different IO methods.

I think the [tables API](http://msprime.readthedocs.io/en/docathon/api.html#tables-api) header and template (based on NodeTable) is looking pretty good and we can probably just apply this to all the tables.

What do we think? If we agree the general direction is good then I'd like to merge this much and we can start filling in the various bits in smaller chunks.

msprime/tables.py
ashander: Sounds good

msprime/tables.py
ashander: Gotcha, mostly.

Because I had only ever used `flags` for flagging samples, and hadn't been keeping careful track of changes in their internal use, I had the vague impression that `flags`, a formerly dual-purpose column had been split into `metadata` and `is_sample`. I think further updates to the node table (e.g., https://github.com/ashander/msprime/blame/e044d3d6183de7e6839d2499efdc0255027f2f5e/docs/file-format.rst#L51-L64 )  and metadata definitions will reduce my confusion and clarify this.

ashander:
LGTM. @petrelharp just started teaching a class and will take a look this afternoon

msprime/tables.py
jeromekelleher: Noted, thanks. Definitely need to clear this up definitively somewhere.

petrelharp:
Generally looking good.  Comments:

There are two sections titled "Tables API": [here](http://msprime.readthedocs.io/en/docathon/api.html#tables-api) and [here](http://msprime.readthedocs.io/en/docathon/interchange.html#sec-tables-api).  I'm not sure what the second one is doing, but it's linked to from the top of the first one.  They should have different names, I think.

- I think there should be definitions of Mutation and Site in the list of definitions [here](http://msprime.readthedocs.io/en/docathon/interchange.html#data-model)

- More generally, I'm not sure where the paragram that says an overview sthg like  "A tree sequence can be stored in a collection of five tables: Node, Edge, Site, Mutation, and Migration.  The first two store the genealogical relationships that define the trees; the next two describe where mutations fall on those trees; and the final table describes how lineages move across space.  The last three may be empty."  Maybe should be at the start of the Tables API section, and refers to the Data Model section for more details?

In the [NodeTable] doc:

- I don't see where a `NodeTable`'s `flags` are defined.  The example over at [the definition](http://msprime.readthedocs.io/en/docathon/interchange.html#sec-node-table-definition) is confusing in that the printed out version has `is_sample` but then it talks about `flags`.

- The documentation in `append_columns` should maybe say that "Appends the specified arrays to the end of the columns of this table.  This allows many new rows to be added at once." -- since at first maybe someone would think this means it would actually add entirely new columns (analogously to `add_rows`).

- perhaps note that `set_columns` does not require the new columns to be the same length as the old ones.

jeromekelleher:
> There are two sections titled "Tables API": here and here. I'm not sure what the second one is doing, but it's linked to from the top of the first one. They should have different names, I think.

Done. I've change it to "Binary interchange". The idea is to discuss the high-level ideas here, which would clutter up the API docs.

> I think there should be definitions of Mutation and Site in the list of definitions here

Done. I've generally expanded the list of definitions.

> More generally, I'm not sure where the paragram that says an overview sthg like "A tree sequence can be stored in a collection of five tables: Node, Edge, Site, Mutation, and Migration. The first two store the genealogical relationships that define the trees; the next two describe where mutations fall on those trees; and the final table describes how lineages move across space. The last three may be empty." Maybe should be at the start of the Tables API section, and refers to the Data Model section for more details?

I've put this in at the end of the definitions section.

> I don't see where a NodeTable's flags are defined. The example over at the definition is confusing in that the printed out version has is_sample but then it talks about flags.

I've gone into as much detail as I can here. Does this help? (Also, what do you think @ashander?)

> The documentation in append_columns should maybe say that "Appends the specified arrays to the end of the columns of this table. This allows many new rows to be added at once." -- since at first maybe someone would think this means it would actually add entirely new columns (analogously to add_rows).

Done.

I think the frameworks are largely in place now, so I'm going to merge this much. There's a bunch of stuff still to be resolved (like: how do we express all the tree sequence requirements, what do we put in the tutorial), but I think it's going well.

ashander:
This looks great to merge!

> I've gone into as much detail as I can here. Does this help? (Also, what do you think @ashander?)

the details about flags vs metadata seem quite helpful and clarify things a _lot_. one question:

how does or could a user of e.g. `.set_columns` set a flag other than `IS_SAMPLE`?

jeromekelleher:
> how does or could a user of e.g. .set_columns set a flag other than IS_SAMPLE?

You'd need to bitwise OR the various flags that you want together to get the overall value. Luckily we only have one flag for now so we don't need to worry about this!

ashander:
Ah, doh! That **bitwise** keyword has been there the whole time but because I'd only ever seen `IS_SAMPLE` I hadn't absorbed it. Thanks

molpopgen:
I know I'm late to this party, but I've discovered something.  I cannot find a description of what `ancestral_state_offset` is.

My p-doc's PR is not working for me due to a change from `ancestral_state_length` to `..._offset`.

I've grepped for it in msprime/ lib/ and docs/, but cannot find it referred to in any docstrings.

ashander:
I think this set out the template but only claims to document NodeTable.
Others are still todo.


The offset relates to ragged columns capability introduced in in
https://github.com/jeromekelleher/msprime/pull/322

jeromekelleher:
> I know I'm late to this party, but I've discovered something. I cannot find a description of what ancestral_state_offset is.

This is still missing all right, thanks for the heads up. The *offset arrays are offsets into the flattened list of strings. So, if I had a column like ``somecol = ["a", "bc", "d"]``, then we'd have encoded it like
```python
somecol = ["a", "b", "c", "d"]
somecol_offset = [0, 1, 3, 4]
```

Note that the length of somecol_offset is the number of rows + 1.

Thus, the column value for row j is ``somecol[somecol_offset[j]: somecol_offset[j + 1]]``.

molpopgen:
Thanks.  I'm worried that we are having an issue updating these columns during a forward sim.  We want ancestral/derived to be 0/1, but the ancestral states are coming out as -1.

Something may be amiss here, for example:

https://github.com/molpopgen/fwdpy11_arg_example/blob/ancestral_sample/practice/prototype_mutations_history_gc.py#L168

cc @DL42

PR #352, opened 2018-01-16
Experimental tree sequence viz.

jeromekelleher:
This is a very rough start. Just opening a PR here so I don't forget about it --- seems useful to be able to visualise the full tree sequence at once.

Example:

tmp

### DIFF
--- a/msprime/drawing.py
+++ b/msprime/drawing.py
@@ -86,6 +86,26 @@ def draw_tree(
     return td.draw()
 
 
+def draw_tree_sequence(tree_sequence, width=None, height=None, format=None):
+    if format is None:
+        format = "SVG"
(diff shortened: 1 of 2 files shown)


petrelharp:
Useful for making this talk, like, right now!

jeromekelleher:
Yeah, this is pretty handy. I think what we need to do here is (a) add an option to tree.draw() to make node heights consistent across a tree sequence; and (b) then figure out how to put a few of these together (might already have done that here). I wish I had time to play with this sort of fun stuff...

petrelharp:
Another useful option for `tree.draw()` would be a little "progress bar" down at the bottom that says how far along the sequence you are. Recording these ideas for future self with more time...

PR #360, opened 2018-01-19
Document objects

jeromekelleher:


petrelharp:
Note: currently the [interchange docs](http://msprime.readthedocs.io/en/latest/interchange.html) omit the sortedness property that edges are sorted by parent ID among parents with the same time value.  We don't currently use this fact, I think, but it should be consistently documented.  Maybe just a note saying that there may be other sortedness requirements, see `sort_tables` for the definitive list.

jeromekelleher:
> Note: currently the interchange docs omit the sortedness property that edges are sorted by parent ID among parents with the same time value. We don't currently use this fact, I think, but it should be consistently documented. Maybe just a note saying that there may be other sortedness requirements, see sort_tables for the definitive list.

We don't require this AFAIK. We did at one point, but it made it too difficult to output edges in the required form from simplify. There was some quirk with internal samples, which made this too hard to do (without lots of sorting within simplify). So, nondecreasing parent time overall, and with parents with equal IDs contiguous seemed like the best option.

petrelharp:
Ok.  So, that's a property that `sort_tables` produces but that we do not require.  That should be noted in the docs for `sort_tables` perhaps, since that's where I'd tend to go look for what the definitive list of requirements is.

PR #370, opened 2018-01-24
Added node labels text & cleaned up API.

### LINKED ISSUE #367: Change node_label_text to node_labels in tree.draw
Part of #294.

The proposed ``node_labels`` parameter is a dictionary mapping node IDs to a string label. If a node is not in the dictionary, the empty string is used. If the node_labels parameter is not supplied or None, the default labels are used.

This replaces the current parameters: ``times``, ``internal_node_labels``, ``leaf_node_labels`` and ``node_label_text``. I don't think many people are currently using this API, so I think it's better to clean it up now.

jeromekelleher:
Partial cleanup of the tree drawing API. This gives much more flexible handling of node labels, and gets rid of a bunch of old confusing parameters.

Closes #367.


jeromekelleher:
Merging this for now. Mutation rendering needs work though.

PR #403, opened 2018-02-06
First steps to updating C API.

jeromekelleher:
This separates the tables API from the overall msprime code, so that we can isolate the code needed for using the tables API.

This should not be merged until after 0.5.0 has been released.


jeromekelleher:
This code isn't ready to merge yet @petrelharp, but it might be useful to you. I've isolated the HDF5 code into tables.c/h now, so it's possible to dump to a HDF5 file without making a tree sequence object. There's a fair bit of work to be done before this can be merged into master, but the basic code should work now. Here's the basic idea:

```c
    table_collection_t tables;
    
    ret = table_collection_alloc(&tables, MSP_ALLOC_TABLES);
    if (ret != 0) {
        handle_error("table_collection_alloc", ret);
    }
    /* Need to set the sequence length to use the HDF5 output below */
    tables.sequence_length = 1.0;

    /* Create a simple chain of nodes, with 0 as the only sample. */
    for (j = 0; j < 10; j++) {
        ret = node_table_add_row(&tables.nodes, j == 0, j, 0, NULL, 0);
        if (ret < 0) {
            handle_error("add_node", ret);
        }
        if (j > 0) {
            ret = edge_table_add_row(&tables.edges, 0, 1, j, j - 1);
            if (ret < 0) {
                handle_error("add_edge", ret);
            }
        }
    }
    /* Write the state out to file */
    ret = table_collection_dump(&tables, "tmp.hdf5", 0);
    if (ret != 0) {
        handle_error("dump", ret);
    }
```

The API is definitely not finalised, but it should work well enough for you to output HDF5 from simulations and get an idea of how things are scaling.

petrelharp:
Awesome.  cc @jgallowa07 .

jeromekelleher:
OK, this has turned into a bit of a beast but I think it's time to merge it and start making smaller units of work to follow up on. The commit message should explain what has changed here.

The main thing is I think this is ready to start embedding into other code bases now so we can see how things are performing and how the API shapes up. We only need to copy a handful of files and you should have the tables API working (see the makefile for the ``table_example.c`` and ``table_example_cpp.cc`` for example of using this in C and C++). The full example looks like
```c
int
main(int argc, char **argv)
{
    int j, ret;
    table_collection_t tables;
    node_id_t samples[] = {0};

    /* Allocate a table collection, with all the internal tables initialised
     * using default alloc sizes */
    ret = table_collection_alloc(&tables, MSP_ALLOC_TABLES);
    if (ret != 0) {
        handle_error("table_collection_alloc", ret);
    }
    /* NB: must set the sequence_length !! */
    tables.sequence_length = 1.0;

    /* Create a simple chain of nodes, with 0 as the only sample. */
    for (j = 0; j < 10; j++) {
        ret = node_table_add_row(&tables.nodes, j == 0, j, 0, NULL, 0);
        if (ret < 0) {
            handle_error("add_node", ret);
        }
        if (j > 0) {
            ret = edge_table_add_row(&tables.edges, 0, 1, j, j - 1);
            if (ret < 0) {
                handle_error("add_edge", ret);
            }
        }
    }

    /* Write the state out to file */
    ret = table_collection_dump(&tables, "tmp.hdf5", 0);
    if (ret != 0) {
        handle_error("dump", ret);
    }

    /* Useful debugging feature */
    table_collection_print_state(&tables, stdout);

    ret = table_collection_simplify(&tables, samples, 1, 0, NULL);
    if (ret != 0) {
        handle_error("simplify", ret);
    }
    /* After simplify, we only have 1 node left and no edges */
    node_table_print_state(&tables.nodes, stdout);
    edge_table_print_state(&tables.edges, stdout);

    /* Clean up. This should usually also be done in the error handling case,
     * but since this is a simple standalone program and we can exit on
     * error. */
    table_collection_free(&tables);
    return 0;
}
```

The API will almost certainly change a bit before it's finalised in tskit, but I don't think it'll change all that much.

Have you any thoughts here @petrelharp, @molpopgen? I'll merge this soon and make some issue for the follow up stuff (like adding the TableCollection object into Python and updating the high-level APIs, hopefully backward compatibly) unless someone objects.

molpopgen:
This looks good (although I've not tested anything).

For the C++ example, you may consider wrapping the main execution loop in a try/catch block, and the catch section frees the tables upon error.  This would allow a non-zero return status to trigger an exception.  

C++ allows you to throw any type as an exception:

```cpp
#include <stdexcept>
#include <iostream>

using namespace std;

int
main()
{
    try
        {
           // pretend we got a bad return value from msprime
            throw - 1;
        }
    catch (int x)
        {
            cout << "caught error code: " << x << '\n';
            //We can free up and C-allocated stuff here.
        }
    catch (exception & e)
        {
            cout << "caught something else: " << e.what() << '\n';
        }

   //Compare to what happens if we throw an exception:
    try
        {
            throw std::runtime_error("We gotta go!");
        }
    catch (int x)
        {
            cout << "caught error code: " << x << '\n';
        }
    catch (exception & e)
        {
         //we go here instead            
         cout << "caught something else: " << e.what() << '\n';
        }
}
```

lib/table_example_cpp.cc
```diff
@@ -0,0 +1,68 @@
+#include "tables.h"
+
+#include <iostream>
+#include <stdlib.h>
```
molpopgen: This should be:

```cpp
#include <cstdlib>
```

lib/table_example_cpp.cc
```diff
@@ -0,0 +1,68 @@
...
+
+using namespace std;
+
+
+static void
+handle_error(string msg, int err)
```
molpopgen: Needs `#include <string>`.

petrelharp:
Looks good.  We'll change over the SLiM stuff to use this.

jeromekelleher:
Thanks @molpopgen, I'll implement these changes to the cpp example before merging. One question: what's idiomatic way to raise errors in this case? Should it be something like
```cpp

static void 
raise_exception(string msg, int err)
{
      throw (Some c++ for Exception("{}:{}".format(msg, mps_strerror(err)))
}

....
    ret = edge_table_add_row(&tables.edges, 0, 1, j, j - 1);
    if (ret < 0) {
          raise_execption("add_edge", ret);
     }
```

And then stick a try-catch block around main above, as you suggest?

molpopgen:
```cpp
#include <sstream>
#include <utility>
#include <stdexcept>

void raise_exception(std::string msg, int err) noexcept(false)
{
std::ostringstream m(std::move(msg);
m << ": " << err;
throw std::runtime_error(m.str());
}
```

You can make the above static, or in an empty namespace, as you see fit.

And the the try/catch block around main.

jeromekelleher:
> And the the try/catch block around main.

Done, thanks!

Issue #349, opened 2018-01-15
0.5.0 release plan


jeromekelleher:
After #346 is merged, I think we've covered most of the outstanding documentation requirements for the tables API. I think it's time to release 0.5.0b1.

- 0.5.0b1 is the first beta, and with documentation for all the tables API. Use this to (a) allow people play with tables API; and (b) get feedback from existing users on whether we've broken their code.
- 0.5.0b2 should include the remaining release items for the 0.5.0 milestone, and should follow in a week or two.
- 0.5.0 should follow quickly after that.

This is all assuming that things go smoothly, of course!

Pinging @petrelharp, @molpopgen and @ashander to keep you in the loop.

petrelharp:
yay!!

jeromekelleher:
> awesome. let me know if there's any easy-to-delegate contribs to docs or fixes

It would be super helpful if you take a look at the "Processing Results" section of the docs now at see if you can spot any issues. I think I have most things sorted out now, but another pair of eyes on things like the ``variants`` iterator would help a lot.

http://msprime.readthedocs.io/en/latest/api.html#processing-results

I'll probably roll another beta tomorrow so we can try out these API updates. After that it's a matter of cleaning up the last few niggling issues, and writing up the release notes. It will probably need a third beta all right though.

PR #404, opened 2018-02-07
Suggested slight reorg/edit for 'Processing Results'

ashander:
Per #349.

The docstring on variants made sense. I thought the organization of the section could be clearer and so suggested these edits

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -139,32 +139,52 @@ Variable recombination rates
     :members:
 
 
-*******************
-Processing results
-*******************
+************************
+Using simulation results
+************************
 
 The :class:`.TreeSequence` class represents a sequence of correlated trees
 output by a simulation. The :class:`.SparseTree` class represents a single
-tree in this sequence.
+tree in this sequence. 
+These classes are the interfaces used to interact with the trees
+and mutational information stored in a tree sequence returned from a simulation.
+There are also methods for loading data into these objects, either from the native
+format, e.g., :func:`msprime.load`, or from another format, e.g.,
+:func:`msprime.load_text` compatible with the `Tables API <sec_tables_api>`.
+
++++++++++++++++++
+Top level-classes
++++++++++++++++++
+
+
+.. autoclass:: msprime.TreeSequence()
+    :members:
+
+.. autoclass:: msprime.SparseTree()
+    :members:
+
 
 +++++++++
 Constants
 +++++++++
 
-.. data:: msprime.NULL_NODE = -1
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -139,83 +139,69 @@ Variable recombination rates
...
+tree in this sequence. 
+These classes are the interfaces used to interact with the trees
+and mutational information stored in a tree sequence returned from a simulation.
+There are also methods for loading data into these objects, either from the native
+format, e.g., :func:`msprime.load`, or from another format, e.g.,
+:func:`msprime.load_text` compatible with the `Tables API <sec_tables_api>`.
+
++++++++++++++++++
+Top level-classes
++++++++++++++++++
```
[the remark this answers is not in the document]
jeromekelleher: It's for sidebar navigation really. This is to make it quick to navigate in to the TreeSequence and SparseTree classes from the top.

docs/api.rst
```diff
@@ -1,4 +1,4 @@
-.. _sec_api:
+.. _sec-api:
```
jeromekelleher: Why the change here? I changed the separators to be underscores a while back so they will autocomplete easily in vim.

docs/api.rst
```diff
@@ -139,83 +139,69 @@ Variable recombination rates
...
 
 +++++++++
 Constants
 +++++++++
 
-.. data:: msprime.NULL_NODE = -1
+These constants are used in :class:`.TreeSequence` and :class:`.SparseTree`
+instances to define special values of various variables.
+
+.. data:: msprime.NULL_NODE == -1
```
jeromekelleher: What's the difference with == here?

docs/api.rst
ashander: Oops. That's an error! Fixed

docs/api.rst
ashander: I changed to == because the single = looked like assignment, to me. With equality I can paste it into an interpreter and see 

```
In [4]: msprime.NULL_NODE == 1
Out[4]: True
```
(Of course, I could paste single = in too, but we want to discourage changing these manually in an imported module ;)

Happy to change back if it reads weirdly.

ashander:
Great. CI failed due to that unintended label change so should be good to go pending =/== resolution. Let me know about that and I'll change if needed and squash these

jeromekelleher:
Let's keep it as == if you think it reads better (I don't have a strong opinion).  We're good to merge after squashing.

PR #412, opened 2018-02-24
allow human-readable metadata for text input

petrelharp:
For testing purposes, we are recording SLiM IDs as node metadata in text tables.  Of course, this led to an error, because `load_text()` expected metadata to be Base64-encoded.  So, I added a flag that allows input from human-readable text files.  No corresponding output; that would be a bigger can of worms.

### DIFF
--- a/msprime/tables.py
+++ b/msprime/tables.py
@@ -96,16 +96,14 @@ class NodeTable(_msprime.NodeTable):
     :vartype metadata_offset: numpy.ndarray, dtype=np.uint32
     """
 
-    def __str__(self, encoding='utf8', base64_metadata=True):
+    def __str__(self):
         time = self.time
         flags = self.flags
         population = self.population
         metadata = unpack_bytes(self.metadata, self.metadata_offset)
         ret = "id\tflags\tpopulation\ttime\tmetadata\n"
         for j in range(self.num_rows):
-            md = metadata[j]
-            if base64_metadata:
-                md = base64.b64encode(md).decode(encoding)
+            md = base64.b64encode(metadata[j]).decode('utf8')
             ret += "{}\t{}\t{}\t{:.14f}\t{}\n".format(
                 j, flags[j], population[j], time[j], md)
         return ret[:-1]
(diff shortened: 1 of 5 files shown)


jeromekelleher:
I agree we need to fix this, but I think we should do it in a slightly different way.  

First we need to add an 'encoding' parameter to these functions that defaults to utf8. That way, we can take out the shared ``tokens[metadata_index].encode('utf8')`` and do it once. Then, I guess we could add another parameter, ``base64_metadata=True``, and so end up with something like:

```python
          if metadata_index is not None and metadata_index < len(tokens):
               metadata =  tokens[metadata_index].encode(encoding)
               if base64_metadata:
                    metadata = base64.b64decode(metadata)                    
```

Then we can get rid of the ``text_[de|en]code_metadata`` functions from tables.py. This approach should work symmetrically then for ``dump_text``. If we also add an ``encoding`` and ``base64_metadata`` parameters here it should all work fine. 

We will need to make it clear that embedded whitespace within metadata that is not base64 encoded will break things though.

petrelharp:
That sounds better, yes.  So, add both `metadata_encoding` and `metadata_base64` parameters to `dump_text()` and to `load_text()`.

jeromekelleher:
Just "encoding", I think because the whole line is using utf8, or utf16 or whatever, not just metadata.

petrelharp:
This is perhaps done.

The only hiccup I ran into is the `__str__`s in `tables.py`.  See what you think.

msprime/tables.py
```diff
@@ -112,15 +96,18 @@ class NodeTable(_msprime.NodeTable):
     :vartype metadata_offset: numpy.ndarray, dtype=np.uint32
     """
 
-    def __str__(self):
+    def __str__(self, encoding='utf8', base64_metadata=True):
```
jeromekelleher: I don't think we should change the signature of ``__str__``, as this could have some nasty unintended consequences. In this case we just have to print out **something**. Assuming utf8 and always writing it out base64 encoded is probably the right thing to do here.

tests/test_highlevel.py
```diff
@@ -1523,7 +1523,6 @@ def convert(v):
             self.assertEqual(str(node.is_sample()), splits[1])
             self.assertEqual(convert(node.time), splits[2])
             self.assertEqual(str(node.population), splits[3])
-            self.assertEqual(msprime.text_encode_metadata(node.metadata), splits[4])
```
jeromekelleher: Should replace this test rather than just delete it I think. We could defined a local function (in the test module) called ``base64_encode`` which will have the same definition as the old ``text_encode_metadata`` function.

petrelharp:
There, that should do it.

msprime/trees.py
```diff
@@ -1339,18 +1359,24 @@ def load_text(nodes, edges, sites=None, mutations=None, sequence_length=0, stric
...
         False, a relaxed whitespace splitting algorithm is used.
+    :param string encoding: Encoding used for text representation.
+    :param bool base64_metadata: If True, metadata is encoded using Base64
+        encoding; otherwise, as plain text.
     :return: The tree sequence object containing the information
         stored in the specified file paths.
     :rtype: :class:`msprime.TreeSequence`
     """
-    node_table = parse_nodes(nodes, strict=strict)
+    node_table = parse_nodes(nodes, strict=strict, encoding=encoding,
```
jeromekelleher: Nothing essential, but I do wince a bit every time I see code visually indented like this. My objection is that it makes it harder than it needs to be to change code --- if we decided to change the ``node_table`` variable to ``nodes``, say, then you'd also have to reindent. Or, if we changed the name of ``parse_nodes``, we couldn't just do a global search and replace, we'd have to reindent all the visual indents as well.

Anyway, it doesn't really matter, so feel free to ignore. I just had to get that one off my chest!

jeromekelleher:
Looks good to me, thanks @petrelharp.

PR #413, opened 2018-02-25
changed node_table_print_state to output is_sample

petrelharp:
Currently, the way to output a NodeTable as text from C seems to be the function `node_table_print_state`.  This outputs `flags` as a column, rather than `is_sample`, which is what is required for `load_text()`.  The function is only used for debugging at the moment; so changing to output `is_sample` is harmless.  At first I removed `flags` entirely, but realized that we can safely output *both* columns, without loss of generality.

### DIFF
--- a/lib/msprime.h
+++ b/lib/msprime.h
@@ -985,6 +985,7 @@ int node_table_append_columns(node_table_t *self, size_t num_rows, uint32_t *fla
 int node_table_clear(node_table_t *self);
 int node_table_free(node_table_t *self);
 void node_table_print_state(node_table_t *self, FILE *out);
+void node_table_dump_text(node_table_t *self, FILE *out);
 bool node_table_equal(node_table_t *self, node_table_t *other);
 
 int edge_table_alloc(edge_table_t *self, size_t max_rows_increment);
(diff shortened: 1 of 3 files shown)


jeromekelleher:
This is still a bit of a hack though, since there's a bunch of other stuff in there that's not part of the text format. I think the right approach is probably to leave ``x_table_print_state`` alone, and instead create a new function ``x_table_dump_text(x_table_t *table, FILE *out)``. These can just copy most of the code from ``print_state``, and give the text format that we want.

It's pretty straightforward to do other than testing, which would require writing a parser, by right. This is really tedious though, so I wouldn't bother. I'd just call the code from the tester like ``print_state`` is called now, at least making sure that it's all memory safe and so on. We could hook up an undocumented Python interface to allow us test it as well, if we liked. (I'd rather not make the C code the main text output method, as changing from C streams to Python streams is a bit painful and ends up being quite restrictive.)

I'm happy to do it if you don't have time for the C legwork @petrelharp. It wouldn't take long.

petrelharp:
Oh, that's a good idea.  I can do that, np.  And good point about not needing to test it particularly much.

petrelharp:
This is done, hopefully correctly. The main convenience in this from my point of view was actually not including the header stuff - having extra columns in was no problem, really.

jeromekelleher:
I changed the function signature to allow for error returns @petrelharp, and added error checking. Does this look OK to you?

PR #416, opened 2018-02-27
Improved extend method for Fenwick trees

JereKoskela:
The code I wrote out in my original email was wrong, which is why this is slightly different. 

The main difference to earlier (apart from the more efficient way in which the expanded entries are populated) is that we can't use the fenwick_alloc_buffers method in fenwick_expand anymore since the two callocs would defeat the whole point. Hence I've done the mallocing directly in fenwick_expand using realloc to keep things as in-place as possible.

This passes your tests and simulation_tests runs, and I've made sure that the new for loop behaves as intended.

### DIFF
--- a/lib/fenwick.c
+++ b/lib/fenwick.c
@@ -75,16 +75,21 @@ fenwick_alloc(fenwick_t *self, size_t initial_size)
 int WARN_UNUSED
 fenwick_expand(fenwick_t *self, size_t increment)
 {
-    int ret = -1;
+    int ret = MSP_ERR_NO_MEMORY;
     size_t j, n, k;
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Very nice indeed, thanks @JereKoskela! I've added a little bit of paranoia around the reallocs (if realloc fails, and you assign directly into the old pointer, then you can leak the old memory). I've also added a simple extra test.

If these additions look good to you, I'll merge.

JereKoskela:
Looks good to me. Merge away.

jeromekelleher:
Superb; thanks @JereKoskela!

PR #419, opened 2018-03-06
Redefine Y3

ashander:
In some other work @petrelharp and I have realized it is more natural to define
Y3 so that it's 0 in a panmictic population. We do this based on the old definition
(which we now call lowercase y3):

```
with sample_sets are A, B, and C, then define 'y3(A;B,C)' as the mean total
length of any edge in the tree between a and the most recent common ancestor of
b and c, where a, b, and c are random draws from A, B, and C respectively; or
the density of mutations segregating a|bc.
```

Then the new definition of Y3 is Y3(A; B,C) = y3(A; B,C) - 0.5 * (Y(B; A,C) + Y(C; A,B)

I've taken a path of least resistance in updating the tests, preferring
redefining Y3 and y3 as above, and using the definition just given. This has
mostly worked.  But there are a few issues mostly relating to the changes to
overlapping-samples corrected versions Y2 and Y1

- [ ] the [obvious way to change Y2](https://github.com/jeromekelleher/msprime/pull/419/files#diff-71c56ecf7eccce3069862f1392a93974R345) now implies Y2(A;B) = - Y2(B;A) but this is not what we thought earlier 
- [ ] a related issue is that with the new Y2 a group with a single
  sample can no longer be used in the first argument A (this is currently causing the only error in the tests)
- [x] is there an analogous change we should make to Y1? I can't see one that would make sense. **no**

### DIFF
--- a/msprime/stats.py
+++ b/msprime/stats.py
@@ -275,13 +275,14 @@ class GeneralStatCalculator(object):
 
     def Y3_vector(self, sample_sets, windows, indices):
         """
-        Finds the 'Y' statistic between three sample_sets.  The sample_sets should
-        be disjoint (the computation works fine, but if not the result depends
-        on the amount of overlap).  If the sample_sets are A, B, and C, then the
-        result gives the mean total length of any edge in the tree between a
-        and the most recent common ancestor of b and c, where a, b, and c are
-        random draws from A, B, and C respectively; or the density of mutations
-        segregating a|bc.
+        Finds the 'Y' statistic between three sample_sets.  The sample_sets
+        should be disjoint (the computation works fine, but if not the result
+        depends on the amount of overlap).  If the sample_sets are A, B, and C,
+        then define 'y3(A;B,C)' as the mean total length of any edge in the tree
+        between a and the most recent common ancestor of b and c, where a, b,
+        and c are random draws from A, B, and C respectively; or the density of
+        mutations segregating a|bc.
+        Then 'Y3' is y3(A;B,C) - 1/2 y3(B;A,C) - 1/2 y3(C;A,B).
 
         The result is, for each window, a vector whose k-th entry is
             Y(sample_sets[indices[k][0]], sample_sets[indices[k][1]],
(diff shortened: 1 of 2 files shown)


msprime/stats.py
```diff
@@ -389,20 +391,6 @@ def Y2(self, sample_sets, windows):
         return self.Y2_vector(sample_sets, windows, indices=[(0, 1)])
 
     def Y3(self, sample_sets, windows):
-        """
```
petrelharp: This was redundant, but shouldn't it have some documentation? Or is there some trick to documenting both in the same place?

msprime/stats.py
```diff
@@ -355,10 +358,9 @@ def Y1_vector(self, sample_sets, windows):
         """
         Finds the 'Y1' statistic within each set of samples in sample_sets. The
         sample_sets should be disjoint (the computation works fine, but if not
```
petrelharp: Hm, yes - now as defined `Y1(A) = 0`.  Just as `f1(A) = 0`.  We could define `y1(A)` but I might vote to leave this one out.

petrelharp:
Yep; with the redefinition, `y1(A) = 0` always.  I think this is fine; and it can be removed.

msprime/stats.py
ashander: Uh, yeah meant to comment. I will look into trick. I was deleting for now as the content should duplicate `Y3_vector`

tests/test_tree_stats.py
```diff
@@ -1086,12 +1119,17 @@ def test_case_2(self):
...
+        branch_true_Y = 0.2*4 + 0.2*(4+2) + 0.2*4 + 0.2*2 + 0.2*(5+1) - 0.5 * (
+            (0.2*4 + 0.2*2 + 0.2*4 + 0.2*2 + 0.2*4) +
+            (0.2*(5+1) + 0.2*2 + 0.2*(5+1) + 0.1*(5+3) + 0.1*(4+2) + 0.2*4)
+        )
 
         # site stats
-        # Y(0;1, 2)
-        site_true_Y = 1
+        # Y(0;1, 2) = 0|12 - 0.5 * (1|02 + 2|01)
+        site_true_Y = 1 - 0.5 * (1 + 1)
```
ashander: NB: I'm sure this is true by symmetry but I can't see it easily from the trees drawn above. Perhaps looking closely at the sites table would help (it did for recurrent muts example)

jeromekelleher:
I don't have much of an opinion on this, but seems good to me. Ping me when you want a review/merge.

ashander:
Thanks @petrelharp for taking a look. Two things:

- I'm a bit unclear on the upshot for Y1, due in part to lowercase/uppercase confusion and my notation. I think you've said this: leave as before. with new notation this is Y1=y3(A;A,A) (using lowercase y3 defined above in this PR)
- Any thoughts on Y2 issues

jeromekelleher:
@petrelharp, @ashander, do we want to merge this in or close the PR? I'm hoping to get tskit pulled out of msprime in the coming weeks, so now's the time to get it merged in if we want it.

jeromekelleher:
Closing this as it's out of date (would need to be moved over to tskit).

PR #430, opened 2018-03-15
Prototype hybrid simulation model

DomNelson:
Implements a working (yet statistically untested) hybrid model. It extends the existing DTWF model with an extra parameter, the number of DTWF generations to be simulated before switching to the coalescent.


lib/msprime.c
```diff
@@ -2238,10 +2252,18 @@ msp_dtwf_generation(msp_t *self)
         avl_init_tree(&Q[i], cmp_segment_queue, NULL);
     }
 
+    num_links = fenwick_get_total(&self->links);
+    ret = msp_sanity_check(self, num_links);
+    if (ret != 0) {
+        goto out;
+    }
+    msp_verify_segments(self);
```
jeromekelleher: This will kill performance, so we'll need to remember to remove this before merging.

jeromekelleher: Also, might as well use msp_verify() if you're doing this, as this should check some other stuff too. (this is exactly the right thing to do while developing the algorithms though!)

lib/msprime.c
```diff
@@ -2346,17 +2371,38 @@ msp_dtwf_generation(msp_t *self)
...
         self->time++;
         ret = msp_dtwf_generation(self);
         if (ret != 0) {
             goto out;
         }
     }
+    /* printf("dtwf simulations complete\n"); */
+    /* printf("%lu ancestors remaining\n", msp_get_num_ancestors(self)); */
+
+    transition_pop_size = 1; //TODO: Verify correct value
```
jeromekelleher: This is a bit awkward all right. I think it's probably safer to call run(time) at the top level and then change the model in the simulator. There'll be all sorts of odd side effects otherwise I think.

jeromekelleher:
Also, would you mind squashing these commits down a bit before I start working on it @DomNelson? A lot of these commits don't have anything to do with the end code, so it would be good to squash them out.

DomNelson:
Sorry just saw these updates! Yep I'll work on cleaning things up a bit - I'm all for your suggested changes.

DomNelson:
Ok I'm bracing myself - bit of a rebase noob here. I've squashed the commits, just hoping I haven't created a whole new alternate history in the process. Fingers crossed...

jeromekelleher:
Oops, that is a bit of a mess! I had a quick look, and I'm not sure how you would rebase this. I think the simplest thing to do is to get the diff from upsteam/master, make a patch from it, and apply this to a fresh branch checked out from upstream/master: 

https://www.thegeekstuff.com/2014/03/git-patch-create-and-apply/

It might be easier to make a new PR based on this branch (or, you can force push over this one; up to you).

Rebasing is a black art, and everyone messes it up! The formula that I've adopted that works most of the time is:
```
git fetch upstream
git checkout upstream/master
git checkout -b workbranch
# Do your work, commiting regularly. When the time comes to make a PR:
git fetch upstream
git rebase -i upstream/master
# Squash down the commits using the editor and push to origin
git push origin/workbranch
```

DomNelson:
Thanks! Sounds like a plan - I'll create a new PR with the patched branch.

PR #450, opened 2018-04-04
get simplify tables to remove duplicate sites

petrelharp:
Since sites are supposed to be unique, when outputting from a simulation we ought to check for each new mutation whether the site has been mutated before or not.  This information is not necessarily available in the simulator, if for instance the site has mutated before but those mutations have been lost, so the site is no longer variable. The simulator could be made to remember this, e.g., by maintaining a list of positions currently in the site table and checking for membership at addition of each new mutation, However, 

- this is a lot of checking to do, esp as most sites will be new;
- duplicate sites can be easily removed when `sort_tables` is called

This PR modifies `sort_tables` to remove additional entries from the site table that have the same position, after first checking that ancestral state is the same for each such duplicate entry (and throwing an error otherwise). If a simulator wishes to have different ancestral states, then it will have to keep track of this.

FWIW, this modification passes existing tests; but I have not added tests of the new, site-removing functionality; I wanted to get feedback first.

### DIFF
--- a/lib/tables.c
+++ b/lib/tables.c
@@ -1995,6 +1995,8 @@ table_sorter_sort_sites(table_sorter_t *self)
             sizeof(*ancestral_state_mem));
     char *metadata_mem = malloc(self->sites->metadata_length *
             sizeof(*metadata_mem));
+    site_id_t jj;
+    double last_position;
 
     if (sorted_sites == NULL || ancestral_state_mem == NULL || metadata_mem == NULL) {
         ret = MSP_ERR_NO_MEMORY;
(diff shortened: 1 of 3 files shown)


lib/tables.c
```diff
@@ -2023,20 +2025,37 @@ table_sorter_sort_sites(table_sorter_t *self)
...
-            sorted_sites[j].ancestral_state,
-            sorted_sites[j].ancestral_state_length);
-        ancestral_state_offset += sorted_sites[j].ancestral_state_length;
-        self->sites->metadata_offset[j] = metadata_offset;
-        memcpy(self->sites->metadata + metadata_offset,
-            sorted_sites[j].metadata,
-            sorted_sites[j].metadata_length);
-        metadata_offset += sorted_sites[j].metadata_length;
-    }
+        self->site_id_map[sorted_sites[j].id] = jj;
```
petrelharp: If I wrote this correctly, this means that there's a single "output" site ID for all "input" sites having the same position; so that all mutations referring to the same position will (afterwards) refer to the same site, even if they did not beforehand.

petrelharp:
> you're not suggesting that we allow duplicate sites in a tree sequence, just in a set of tables?

right.

> How do we deal with mutations that are referring to the two different site IDs? (I don't think this is handled in your PR?)

I think it is handled - see the comment in the code (but, I need to test that it actually does what I say =).

> What if the mutations are contradictory?

I think the only way mutations can be *contradictory* is if two sites have different ancestral states. Well, ok - I guess we could worry that (1) a mutation appears, going from `A -> T`; and then (2) another one appears on this background, going from `T -> G`; and then we mistakenly record a new site for the second one, resulting in two sites, with ancestral states `A` and `T` respectively; these are not actually contradictory, but we'd have to build the trees to tell.  Is there any way the mutations themselves can be contradictory?

> Perhaps it might be cleaner to have a separate function merge_sites or something, which takes a sites and mutation table and drops duplicate sites?

I agree, but it's so natural to do it in the `sort_tables` function, since we're already rearranging indices in the mutation table there already. If we split it out into a new function, we'd need to do another pass through.

Hm, although since mutations will be sorted by site at this point, this is do-able without a lookup table. It'd just require a pass through the already-sorted table.  This is not sounding so bad.

petrelharp:
> Suppose one is running a nucleotide-based model in SLiM. To initialize the model, one might set each different individuals in the initial population to have somewhat different nucleotide sequences. So then do we have multiple segregating mutations at the same site with different ancestral states? 

I think the *best* way to do this is to initialize your simulation with prior history coming from a tree sequence. =) But, if you don't do that, then I'd say you'd (a) pick an ancestral state, and (b) add mutations for everyone in the initial population that differs from this state.

Is that the best way to do it?  This is effectively asking how we deal with different states at the roots of different trees, if there is more than one tree at a site. To use `Sites` for this we'd need to include a way of associating sites with roots (which are just nodes); but this is what `Mutations` are for.

bhaller:
OK, after our discussion here, since "empty" can be an ancestral state I think the issue I raised is not an issue.  :->

lib/tables.c
jeromekelleher: Aha, yes; clever!

jeromekelleher:
> If there are two mutations on the same branch, but at different sites, then we're in trouble, since we don't know which one came first. This means that we can't just keep around redundant Sites through simplify (which I was secretly hoping we could), because we may be throwing away information about which mutation came first.

This is difficult, but I'm not sure there's anything we can do about it. The simulator is just going to have to keep this state somehow, I think. Which leads me to wonder, is it really that big a deal keeping a lookup table of mutated site positions within the simulator? Although, if site positions are integers, then I agree some sort of "recycling" of site positions as things become fixed will be necessary. If site positions are doubles, then it's surely simpler to keep a dictionary of all positions that have ever existed and rejection sample when generating new positions.

> I think the right thing to do is to separate this out into a separate function: it means an extra pass through the tables; but it also means that people who don't need to look for duplicated sites don't have to.

Agreed.

petrelharp:
I'm currently implementing in slim with the lookup table; we will then compare to the other solution above. I'll keep this PR open for now.

petrelharp:
This is replaced by #457.

Issue #437, opened 2018-03-22
Add table.truncate(num_rows) to Tables API.


jeromekelleher:
This allows us to efficiently chop off the end of a table, keeping the first num_rows.

PR #451, opened 2018-04-08
record and reset positions in table collections

petrelharp:
There's two things here (in theory: untested! but running the idea by you):

1. Ability to record a position in a set of tables, which would be important for a number of things, including sorting only the last chunk of the tables or looking through the mutations recorded on a given individual.

2. Ability to "rewind" tables to a previously recorded position.  This is useful in use cases where a large number of possible offspring are generated, but are discarded before moving on to the next offspring.

If this looks good, I'll write some tests.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -1221,6 +1221,32 @@ out:
     return ret;
 }
 
+static PyObject *
+IndividualTable_truncate(IndividualTable *self, PyObject *args)
+{
+    PyObject *ret = NULL;
+    Py_ssize_t num_rows;
+    int err;
+
+    if (IndividualTable_check_state(self) != 0) {
+        goto out;
+    }
+    if (!PyArg_ParseTuple(args, "n", &num_rows)) {
+        goto out;
+    }
+    if (num_rows < 0 || num_rows > (Py_ssize_t) self->table->num_rows) {
+        PyErr_SetString(PyExc_ValueError, "num_rows out of bounds");
+        goto out;
+    }
+    err = individual_table_truncate(self->table, (size_t) num_rows);
+    if (err != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = Py_BuildValue("");
+out:
+    return ret;
+}
 
 static PyObject *
 IndividualTable_get_max_rows_increment(IndividualTable *self, void *closure)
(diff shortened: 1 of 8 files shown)


petrelharp:
Thinking more on this, I think this maybe should *not* be in tskit.  If this was a useful thing to do with canonical (sorted) tables, then sure.  But, it's only useful as part of a forwards scheme that works in a particular way. So, maybe I'll keep it as separate code in SLiM.

jeromekelleher:
What's your thinking on this at the moment @petrelharp? I thought I'd responded to this earlier, but I don't see any comments so I must not have.

Basically I'm good with this idea, but I think we should internalise it as state it in the ``table_collection_t`` class. So, we'd say ``table_collection_checkpoint`` to store a checkpoint (basically store the num rows for each table), and then say  ``table_collection_rewind`` (or similar) to set the state back to this checkpoint. Internally, the ``table_collection_rewind`` method could use the ``x_table_truncate`` method described in #437.

jeromekelleher:
Are you still thinking of using this @petrelharp, or should we close this PR?

petrelharp:
We are actually using it; it's on my list to make these very good suggested
changes.

petrelharp:
Finally coming back to this: I think it shouldn't be part of the `table_collection` class, because (a) it makes that class more complex; and (b) I think of this as a "bookmark" into a set of tables; and in some cases one wants to have more than one bookmark at once. (Like, we have real use cases.)  So we could allow table collections to have *lists* of bookmarks... but this seems overly complex. I vote to do this more-or-less as-is.

jeromekelleher:
OK, if you want to have several bookmarks then I agree keeping it out of the table collection class is the way to go.

petrelharp:
I think this is ready to go; I've removed some unnecessary things and written some tests.

petrelharp:
Oh wait, except perhaps this should invalidate the indexes.

petrelharp:
This partially addresses #437.

petrelharp:
The changes look good.

jeromekelleher:
Now closes #437.

I refactored this a bit because you had basically implemented truncate, but not as "methods" of the C objects. Also, the ``record_position`` and ``reset_position`` were really methods of the TableCollection class, so I brought them in to fit in a bit better. I mainly added the Python interface to truncate to get better test coverage on the low-level code.

I'll merge if you're happy with the changes and I've checked the test coverage report.

petrelharp:
yes, looks great. thanks!

jeromekelleher:
OK, merging. Thanks @petrelharp.

PR #455, opened 2018-04-12
single_sweep model introduced into algorithms.py

andrewkern:
Jerome-- as we discussed I've now implemented sweeps in algorithms.py as a separate model. So far in my testing this checks out and plays well with other, neutral events (e.g. population size changes). I've reached a point where I'd like to get going on the C side so I'm submitting this PR.

cheers!
Andy


algorithms.py
```diff
@@ -427,20 +520,228 @@ def hudson_simulate(self):
...
+                    self.migration_event(mig_source, mig_dest)
+        return self.t
+   
+    def single_sweep_simulate(self):
+        """
+        Simulates a single sweep model using
+        phases of coalescent and structured coalsct
+        """
+        
+        assert( True in (["single_sweep_event" in x for x in self.modifier_events]))
```
jeromekelleher: I think 
```python
assert any(["single_sweep_event" in x for x in self.modifier_events])
```
is probably a bit more idiomatic?

algorithms.py
```diff
@@ -947,14 +1331,21 @@ def run_simulate(args):
         population_growth_rates = args.population_growth_rates
     if args.population_sizes is not None:
         population_sizes = args.population_sizes
+    if args.sweep is not None:
+        label_number = 2
+        if num_populations > 1:
+            label_number = 3 ## extra label for other popns
```
jeromekelleher: I don't follow this --- I thought labels and populations were orthogonal?

algorithms.py
```diff
@@ -427,20 +520,228 @@ def hudson_simulate(self):
...
+   
+    def single_sweep_simulate(self):
+        """
+        Simulates a single sweep model using
+        phases of coalescent and structured coalsct
+        """
+        
+        assert( True in (["single_sweep_event" in x for x in self.modifier_events]))
+        #pull out sweep
+        idx = ["single_sweep_event" in x for x in self.modifier_events].index(True)
```
jeromekelleher: I think we'll probably want to return to this and add the ``ModelChangeEvent`` to this inferface later, but this is good for now.

jeromekelleher:
I think we're definitely ready to start working on the C side of things now. Usually I find it best to use the command line interface when developing this sort of stuff, and to fill in the Python-C interface only when this is finished. The C library [developer docs](http://msprime.readthedocs.io/en/stable/development.html#c-library) should come in handy.

The more I look at it, the more convinced I am that the trajectory should be provided as a parameter to the C layer, so I think the best option would be to add a ``single_sweep`` model, that holds a pair of vectors as parameters. See the ``dev.cfg`` file for how this is done with the current models. Obviously adding a list of allele frequencies to ``dev.cfg`` is rubbish for actual simulations, but this is really just for development and testing. Also, for now, I think we should just run the ``single-sweep`` model on its own, and not try to combine it with Hudson simulations before and after (this will happen naturally when we plumb everything together at the Python level). This should give us something that we can work with productively for development. It doesn't matter if it doesn't fully coalesce --- we'll still get perfectly valid tree sequences out of it.

I'm happy to pitch in now with some of the coding to get the infrastructure in place, and then hopefully you can port in the Python code @kern-lab?

In terms of how we proceed, I'd suggest that we both work from this PR? We can merge this much first if you prefer, but for a large change like this merging in little bits as you go along isn't really an option, and we're going to end up with a very large PR in any case, so we might as well keep it all in one place.

Also, as there's other changes happening in the codebase at the moment, we're going to end up doing a lot of rebasing. To make this as easy as it can be, it helps to have fewer commits. So, would you mind rebasing this branch and squashing the current commits down to one please? (Ping me offline if you're not sure how to do this; it's something of a dark art!)

Anyway, this is very exciting! I think you have the most of the real work done here. I'm confident we can get this all merged and working in a couple of weeks.

algorithms.py
kern-lab: I'll remove the notebook for sure. just there for my sandboxing. sorry to have committed it. and yes I think using 'any' is more idiomatic-- i'll change that.

algorithms.py
kern-lab: right. this is needed because during a sweep we want to be able to choose recomb events that occur in segments that are population specific. Because of the way that hudson_recombination_event() works, I accomplish this by moving all ancestors in non-sweeping populations (i.e. all but popn 0) into an extra label, and thus I can select segments using the Fenwick tree convention as you had originally done. 

At the beginning of sweep_phase_simulate_until() this happens

https://github.com/kern-lab/msprime/blob/bdd7e2728f2caa0b13821c8a775b195b9d59a1c6/algorithms.py#L629-L645

then this is used for the recombination routines:

https://github.com/kern-lab/msprime/blob/bdd7e2728f2caa0b13821c8a775b195b9d59a1c6/algorithms.py#L816-L821

kern-lab:
Okay I rebased and squashed! I think this worked fine. I had created an upstream remote when I first forked so I had been keeping it synced with your original master. I'll go read the C library developer docs and take a peek through that dev.cfg file. Let me know how I can point you in the right direction to start doing the infrastructure work and then I'll port code.

jeromekelleher:
I'll make a start on the infrastructure as soon as I can---I'll hopefully get a bit done tomorrow and we can see if it's going in the right direction.

jeromekelleher:
I've made pretty decent progress getting the infrastructure in place for this now @kern-lab, and I think it's ready for you to start porting in the real logic. Because there's a bunch of PRs editing the C code open at the moment, I rebased this on the most recent so that it'll be easy to merge. Basically, there's a bunch of stuff in this PR that'll disappear over the next week as those PRs are merged.

One practical upshot of this though is that I force-pushed over your master branch, so you'll have to pull this into your local clone. You just want to replace your local version of master with origin/master. Do this with ``git reset --hard origin/master``.

In terms of working with the code, the simplest thing to do is use the development inferface in ``main``. Use 
```
./main simulate dev.cfg
```
to run the code. You can tweak the simulation parameters by editing ``dev.cfg``. If you want to see some output, use ``./main simulate -v dev.cfg``.

The main entry point is ``msp_run_single_sweep``; you should find the stuff that you need close by here. All the parameters should be present and easily accessible. I haven't quite got it straight in my head how time works here, but I think this is a fairly minor detail that we can resolve once you've ported in the main simulation logic. Once that's done we can set up the Python-C interface and then run ``hudson`` models on either side of the sweep to complete the simulation.

Good luck!

andrewkern:
closing this down to reflect pulling in this model into the newest refactor

PR #457, opened 2018-04-12
"clean tables" by removing duplicate sites

petrelharp:
This is WIP - not tested - but I'm putting it here now for comments on the API, the structure, and whether we want it in tskit at all.

Note that it really should update the site table in place, as there's no reason it can't do that; but currently it makes a new table and copies it back over.

I'm calling it "clean tables" because (a) it's shorter than "remove_duplicate_sites_in_tables" or equivalent, and (b) we could potentially add other "cleaning" operations to this one thing (although maybe those should just all be separate).  Feel free to complain about this.

### DIFF
--- a/lib/tables.c
+++ b/lib/tables.c
@@ -4551,61 +4551,100 @@ out:
     return ret;
 }
 
-
-/*******************
- * Cleaning tables *
- ******************/
-
-typedef struct {
-    site_table_t *sites;
-    mutation_table_t *mutations;
-    /* Map of old site IDs to new site IDs. */
-    site_id_t *site_id_map;
-} table_cleaner_t;
-
-static int table_cleaner_run(table_cleaner_t *self)
+/*
+ * Remove any sites with duplicate positions, retaining only the *first*
+ * one. Assumes the tables have been sorted, throwing an error if not.
+ */
+int WARN_UNUSED
+table_collection_deduplicate_sites(table_collection_t *self, int flags)
 {
-    // Remove any sites with duplicate positions, retaining only the *first*
-    // one. Assumes the tables have been sorted, throwing an error if not.
-
     int ret = 0;
-    table_size_t j, k, site_j;
+    table_size_t j, site_j;
     table_size_t as_length, as_offset;
     table_size_t md_length, md_offset;
     table_size_t num_input_sites;
     double last_position, position;
+    site_id_t mutation_site;
+    /* Map of old site IDs to new site IDs. */
+    site_id_t *site_id_map = NULL;
 
-    num_input_sites = self->sites->num_rows;
-    site_j = 0; // the index of the next output row
-    // this will need to change if negative positions are allowed!
+    num_input_sites = self->sites.num_rows;
+    site_id_map = malloc(num_input_sites * sizeof(*site_id_map));
+    if (site_id_map == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    /* Check the input first. This avoids leaving the table in an indeterminate
+     * state after an error occurs, which could lead to nasty downstream bugs.
+     * The cost of the extra iterations is minimal. If the user is super-sure
+     * that their input is correct, then we could add a flag to skip these
+     * checks. */
     last_position = -1;
-    as_offset = 0;
-    md_offset = 0;
-
-    // self->site_id_map[0] = 0;
-    // last_position = self->sites->position[0];
-    // as_offset = self->sites->ancestral_state_offset[1];
-    // md_offset = self->sites->metadata_offset[1];
-    for (j = 0; j < self->sites->num_rows; j++) {
-        position = self->sites->position[j];
+    for (j = 0; j < self->sites.num_rows; j++) {
+        position = self->sites.position[j];
+        if (position < 0) {
+            ret = MSP_ERR_BAD_SITE_POSITION;
+            goto out;
+        }
         if (position < last_position) {
             ret = MSP_ERR_UNSORTED_SITES;
             goto out;
         }
+        /* Checking the offsets is arguably unnecessary, since these should
+         * be validated when calling add_row/or append_rows. However,
+         * we can't be sure that users won't edit tables directly and
+         * we'll end up with hard-to-debug memory access violations when
+         * doing the memcpy'ing below. */
+        if (self->sites.metadata_offset[j + 1] > self->sites.metadata_length) {
+            ret = MSP_ERR_BAD_OFFSET;
+            goto out;
+        }
+        if (self->sites.metadata_offset[j] > self->sites.metadata_offset[j + 1]) {
+            ret = MSP_ERR_BAD_OFFSET;
+            goto out;
+        }
+        if (self->sites.ancestral_state_offset[j + 1]
+                > self->sites.ancestral_state_length) {
+            ret = MSP_ERR_BAD_OFFSET;
+            goto out;
+        }
+        if (self->sites.ancestral_state_offset[j]
+                > self->sites.ancestral_state_offset[j + 1]) {
+            ret = MSP_ERR_BAD_OFFSET;
+            goto out;
+        }
+        last_position = position;
+    }
+    for (j = 0; j < self->mutations.num_rows; j++) {
+        mutation_site = self->mutations.site[j];
+        if (mutation_site < 0 || mutation_site >= (site_id_t) num_input_sites) {
+            ret = MSP_ERR_SITE_OUT_OF_BOUNDS;
+            goto out;
+        }
+    }
+
+    site_j = 0; // the index of the next output row
+    // NOTE: this will need to change if negative positions are allowed!
+    last_position = -1;
+    as_offset = 0;
+    md_offset = 0;
+
+    for (j = 0; j < self->sites.num_rows; j++) {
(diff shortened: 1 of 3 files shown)


petrelharp:
Ok - I've added a simple test that it's doing the right thing.  Also, I've rewritten this to modify the SiteTable in place.  However, I think this is probably slower, but haven't yet checked.

lib/tables.c
```diff
@@ -210,7 +210,7 @@ node_table_copy(node_table_t *self, node_table_t *dest)
 
 int WARN_UNUSED
 node_table_set_columns(node_table_t *self, size_t num_rows, uint32_t *flags, double *time,
-        population_id_t *population, char *metadata, uint32_t *metadata_offset)
+        population_id_t *population, const char *metadata, uint32_t *metadata_offset)
```
jeromekelleher: Good cleanup... definitely need to sort out the const'iness of these signatures before finalising. 'const' is my least favourite C feature... definitely causes more problems than it solves for me.

lib/tables.c
```diff
@@ -4675,3 +4675,146 @@ table_collection_simplify(table_collection_t *self,
...
+    // table.
+
+    int ret = 0;
+
+    if (self->sites->num_rows <= 1) {
+        // nothing to do
+        goto out;
+    }
+
+    table_size_t j, k, site_j;
```
jeromekelleher: Style-wise, I try to keep all variable declarations before any code. I know this isn't necessary in C99, but I think it's a good habit and encourages short functions. Also, it's important safety wise. Suppose you had
```c
int function(int x) 
{
    int ret = 0;
    if (x < 0) {
        ret = -1;   
        goto out;
     }
     int *p = malloc(x * sizeof(int));
     ..
out:
     if (p != NULL) {
         free(p);  
     }
     return ret;
}
```
In this case, the compiler will probably be smart enough to know that ``p`` is undefined in the error case, but there will be situations where it can't. So, to guarantee safety, I always make sure that every variable is declared and all pointers have been assigned some value **before** running any code.

lib/tables.c
```diff
@@ -4675,3 +4675,146 @@ table_collection_simplify(table_collection_t *self,
...
+    // Remove any sites with duplicate positions, retaining only the *first*
+    // one. Assumes the tables have been sorted, throwing an error if not.
+    // TODO: this could be done in place rather than by constructing a new
+    // table.
+
+    int ret = 0;
+
+    if (self->sites->num_rows <= 1) {
+        // nothing to do
+        goto out;
```
jeromekelleher: Slightly uncomfortable with this --- gotos should really only be used for error handling, and this feels like control flow. Do we really need to make this special case here? Usually if the logic is correct the empty-table case will just work without problems.

lib/tables.c
```diff
@@ -4675,3 +4675,146 @@ table_collection_simplify(table_collection_t *self,
...
+                self->sites->metadata_offset[site_j] = md_offset;
+            }
+            as_offset += as_length;
+            md_offset += md_length;
+            last_position = position;
+        }
+        self->site_id_map[j] = site_j;
+    }
+
+    if (site_j == j) {
```
jeromekelleher: OK, this is definitely ``goto`` as control flow! Change to
```c
if (site_j != j) { 
    // reset site table size
    self->sites->num_rows = site_j + 1;
```
Or, you could just leave it fall through? The fewer paths through the code the better IMO, I don't really mind if a it's a little less efficient in some corner cases because of this. (However, simplicity is more important --- if having fewer paths through the function makes the code more complicated, then I'd go with the simpler and easier to understand option.)

lib/tables.c
```diff
@@ -4675,3 +4675,146 @@ table_collection_simplify(table_collection_t *self,
...
+    return ret;
+}
+
+static int 
+table_cleaner_alloc(table_cleaner_t *self, 
+        site_table_t *sites,
+        mutation_table_t *mutations) {
+    int ret = 0;
+
+    memset(self, 0, sizeof(table_cleaner_t));
```
jeromekelleher: This is perfect. Just to clarify the reasoning for setting the struct to zero first. This is to make sure that if an error happens during initialisation, and pointers within the struct will be set to NULL, and so ``msp_safe_free`` will work properly. If we don't do this, then any errors during ``_alloc`` will lead to a crash during ``_free``.

lib/tables.c
petrelharp: not sure what you mean by "fall through" -- here we are avoiding looping through every entry in the mutation table and setting its' `site` equal to itself -- if the tables are big, this seems not insignificant, and easy to avoid?

lib/tables.c
petrelharp: That's exactly why I had that in there -- before I had it, I got a crash during `free`. =)

petrelharp:
I have tidied the flow control up, and a couple of things that @bhaller noticed.  It'd be great if you could get the python version in (if we indeed want it: it'd be quite easy to do with numpy, I think?).  I have not figured out more corner cases, though.

lib/tables.c
jeromekelleher: Sure, you're right. I was just thinking in a big-O sense there's no real difference between iterating over the tables once and twice. But I think this it's more readable the way you have it and it is an easily avoidable cost.

lib/tables.c
```diff
@@ -22,6 +22,7 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <math.h>
```
jeromekelleher: What's math.h for? I don't think we're linking against libm at the moment, it'd be nice to keep it that way unless we need it.

lib/tables.c
```diff
@@ -4675,3 +4676,142 @@ table_collection_simplify(table_collection_t *self,
...
+    int ret = 0;
+    table_size_t j, k, site_j;
+    table_size_t as_length, as_offset;
+    table_size_t md_length, md_offset;
+    table_size_t num_input_sites;
+    double last_position, position;
+
+    num_input_sites = self->sites->num_rows;
+    site_j = 0; // the index of the next output row
+    last_position = -INFINITY;
```
jeromekelleher: I usually use DBL_MIN, or -1 for this (presumably this is what the math.h is for?). -1 is the same thing really since we're not allowing negative positions.

jeromekelleher:
> I have tidied the flow control up, and a couple of things that @bhaller noticed. It'd be great if you could get the python version in (if we indeed want it: it'd be quite easy to do with numpy, I think?). I have not figured out more corner cases, though.

Yeah, it looks great to me. The Python version is really just to make it easier to test this code. If you're happy with it now, would you mind squashing the commits down please? I'll add in the Python interface then and a few tests and ping you back to see if your happy. We can merge then assuming all is good.

lib/tables.c
petrelharp: `-INFINITY` - the only way i could think to cleanly start at `j=0`

lib/tables.c
petrelharp: ah, thats better.

petrelharp: Actually, it's `-DBL_MAX` (since `DBL_MIN` is the smallest *positive* float), and we need to include `float.h` for that one...

petrelharp:
I have squashed them.

lib/tables.c
bhaller: I wonder whether -DBL_MAX is guaranteed to be representable in double.  That's a little tricky for signed integer types; INT8_MAX is 127 whereas INT8_MIN is -128.  So -INT8_MAX is correctly representable, but -INT8_MIN is not.  I'm not sure what the rules are for IEEE double, and whether -DBL_MAX is safe.  Surely a smaller negative value, like -1e1000 or something, would suffice?

petrelharp: According to [this post](https://stackoverflow.com/questions/1153548/minimum-double-value-in-c-c) which references [ANSI C](http://www.csse.uwa.edu.au/programming/ansic-library.html#float) it is...

petrelharp: although that link to ANSI C does not state it.

jeromekelleher: I think -1 is probably fine. It's what I usually use for this requirement: any value < 0 will do.

petrelharp:
I have changed the `-DBL_MAX` to `-1`, with a note to any future people who want to remove the assumption that positions are nonnegative.

jeromekelleher:
I've updated this a bit @petrelharp, and I think it's good to go if you agree. Changes:

1. Change the name to ``deduplicate_sites`` and made it a method of ``table_collection``. This seems a bit neater, and you're right, ``clean_tables`` is too generic.
2. Added a bunch of paranoia around error conditions, and added some tests to exercise them.

One might argue that the error checking is a bit heavy-handed and we should add a flag to skip it. However, I'd like to see some evidence that this makes a real difference in run-times for realistic inputs first. This seems like a pretty fast function to me, and I doubt that the error checking will add up to much extra time.

jeromekelleher:
Also, I meant to add: I didn't bother adding a Python interface here as we really need to make a native TableCollection type there to mirror the ``table_collection_t`` low-level class. We can add in an interface for this function when it's done.

petrelharp:
Looks great.  Much tidier and more (helpfully) paranoid.

jeromekelleher:
Great; I'll merge this after #453 (and possibly a couple of others), as it's easier to rebase.

Issue #459, opened 2018-04-18
How to fill in missing `parent` atttributes of mutations?


petrelharp:
As discussed in #122, we decided to add a `parent` property of mutations to make things more efficient. It is also necessary if there is more than one mutation on a given branch. Since this property is encoded by the trees themselves (except in the latter case), we wanted to provide a method to fill in this column for tables where it isn't provided already: #286 and #288.

However:

1. computing the parent column of the mutation table requires constructing the trees and iterating through them, but
2. constructing a proper tree sequence requires the parent column (at least, to get genotypes correctly; without this, mutations may be in the wrong order within each site).

The proposed fix was to add a flag to `load_tables` that would go through and correctly set the parents of each mutation. We have code for this that will appear in #286 soon thanks to @bhaller. Then, we could dump the tables, and read them back in again... but it seems like there should be a better way to do this?

By the way, the flag to `load_tables` (or, maybe a different flag?) can also *check* the parent column (basically, compute the parents; fill in if they are `MSP_NULL_NODE` and throw an error if they exist but are wrong).  This is a possibly important check we haven't got in yet.

jeromekelleher:
This seems like an excellent plan to me; I'm looking forward to seeing the code!

petrelharp:
Update:

1. Checking when there's more than one mutation per branch is a little tricky (but do-able).
2. The previous code (e.g., python) assumes the mutations come in the right order, which they don't necessarily; this needs to be fixed to produce (and, check) parents correctly even if they aren't ordered correctly.

Here, "ordered correctly" means ordered so that any mutation comes after its parent; an easy way to ensure this is to order by (-1) * node time.  Currently `sort_tables` sorts mutations by site, and then by ID, thus preserving ordering within each site.  If mutations are input in order by time, then sort_tables should thus preserve time ordering.  However, it turns out that `simplify_tables` does not always preserve this ordering, at least if mutation parents are not set (sorry, no MWE yet). 

This means that what we want to happen is:

1. Produce tables in forwards time, with mutations ordered by time of appearance, and mostly leaving out mutation.parent information (only providing it when there's more than one mutation on actually the same node).
2. `sort`, `clean`, and `simplify` tables. 
3. At the end, when we want a tree sequence, fill in the `mutation.parent` information, by building a tree sequence and working out whot the parents are.

However, what actually happens is:

[...]

Issue #424, opened 2018-03-15
Add TableCollection to low-level Python API


jeromekelleher:
Following on #403, need to add the ``TableCollection`` object to mirror the ``table_collection_t`` object. The low-level ``load_tables`` method should take this as an argument.

The high-level ``msprime.load_tables`` function should optionally take single ``TableCollection`` *or* the existing list of tables as named arguments. There may be some breakage caused if ``nodes`` and ``edges`` are used as positional arguments --- will need to make a decision about the best approach here.

jeromekelleher:
Closed in #472

PR #461, opened 2018-04-22
find parents method

petrelharp:
Thanks to @bhaller.  This PR also includes a bit of cleanup.

See discussion in #459 for background.  Main idea: if `mutation.parent` is not specified, but mutations appear in the MutationTable so that each mutation's parent occurs before it in the table (which is a requirement), then the `mutation.parent` information can be inferred.  This inserts into `load_tables` two flags that will (a) check if existing `mutation.parent` information is correct; and (b) fill it in if missing.

To-do yet, I believe:

- write tests for finding parents (note that this *has* been tested quite a bit in SLiM)
- write tests for bad mutation order input
- clarify in docs that `sort_tables` preserves relative ordering of mutations at the same site (an addition in this PR)


lib/tables.c
```diff
@@ -2941,7 +2950,7 @@ simplifier_overlapping_segments_init(simplifier_t *self)
     /* Sort the segments in the buffer by left coordinate */
     qsort(self->segment_queue, self->segment_queue_size, sizeof(simplify_segment_t),
             cmp_segment);
-    assert(self->segment_queue_size < self->max_segment_queue_size - 1);
+    assert(self->segment_queue_size < self->max_segment_queue_size);
```
petrelharp: This was an error in the new simplifier that we actually encountered. =)

lib/tables.h
```diff
@@ -71,19 +71,19 @@ typedef struct {
 } node_table_t;
 
 typedef struct {
-    size_t num_rows;
-    size_t max_rows;
-    size_t max_rows_increment;
+    table_size_t num_rows;
+    table_size_t max_rows;
+    table_size_t max_rows_increment;
```
petrelharp: these were errors before, we believe

petrelharp:
Currently `tree_sequence_compute_parents( )` just takes a flag that says whether to check parents only, or to fill in msising values.  And, if this flag is true, then it will:

- if the parent is `MSP_NULL_MUTATION`, fill it in
- if not, and the parent disagrees with the computed parent, throw an error

A side effect is that currently we can't check if parents are *correctly* `MSP_NULL_MUTATION` (i.e., we don't throw an error anywhere if they are but shouldn't be). I've decided that if `fix_missing_parents` is `false`, then it should check the null parents, also.

petrelharp:
I've made that change (testing NULL parents if `fix` is false); and added some unit tests for errors.

petrelharp:
I've changed the behavior so that parents are filled in on branch groups.  I've also added a test to see if mutation parents are being correctly assigned, but this isn't working; I don't understand why.

lib/tree_sequence.c
```diff
@@ -986,6 +1001,300 @@ tree_sequence_simplify(tree_sequence_t *self, node_id_t *samples, size_t num_sam
...
+            // repeatedly scootching our size up by small increments
+            if (node_map__capacity < site->mutations_length)
+            {
+                if (node_map__capacity == 0)
+                    node_map__capacity = 16;    // arbitrary start size
+
+                while (node_map__capacity < site->mutations_length)
+                    node_map__capacity <<= 1;
+
+                node_map = realloc(node_map, node_map__capacity * sizeof(node_map_t));
```
jeromekelleher: All mallocs must be checked.  See other ``realloc`` examples elsewhere for the local convention on dealing with reallocs.

lib/tests.c
```diff
@@ -4175,6 +4225,66 @@ test_single_tree_simplify(void)
...
+    CU_ASSERT_EQUAL(tree_sequence_get_num_sites(&ts), 3);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_mutations(&ts), 6);
+    tree_sequence_free(&ts);
+
+    for (int j=0; j < tables.mutations.num_rows; ++j) {
+        parent = tables.mutations.parent[j];
+        tables.mutations.parent[j] = MSP_NULL_MUTATION;
+        ret = tree_sequence_load_tables(&ts, &tables, MSP_FIX_PARENTS | MSP_COMPUTE_PARENTS);
+        CU_ASSERT_EQUAL(ret, 0);
+        printf("\n%d ---> %d = %d\n", ret, ts.mutations.parent[j], parent);
```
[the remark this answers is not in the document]
petrelharp: as written, `MSP_FIX_PARENTS` requires `MSP_COMPUTE_PARENTS`, but I was thinking the same thing: the former should imply the latter.

lib/tables.c
```diff
@@ -1870,7 +1870,16 @@ cmp_site(const void *a, const void *b) {
     const site_t *ia = (const site_t *) a;
     const site_t *ib = (const site_t *) b;
     /* Compare sites by position */
-    return (ia->position > ib->position) - (ia->position < ib->position);
+    int ret = (ia->position > ib->position) - (ia->position < ib->position);
+	if (ret == 0) {
```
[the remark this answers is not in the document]
petrelharp: on this topic, we wondered why this is the convention for comparisons rather than just
```
   ret = ia->position - ib->position;
```

lib/tests.c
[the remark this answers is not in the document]
petrelharp: oh and the printf is because that test isn't passing yet

lib/tree_sequence.c
```diff
@@ -986,6 +1001,300 @@ tree_sequence_simplify(tree_sequence_t *self, node_id_t *samples, size_t num_sam
...
+                    // check that the branch group forms a linear chain
+                    for (table_size_t group_index = first_in_group + 1;
+                            group_index <= last_in_group; ++group_index)
+                    {
+                        mutation_id_t putative_parent_mut_id = node_map[group_index - 1].mutation_id;
+                        table_size_t mut_site_index = node_map[group_index].site_index;
+                        mutation_id_t actual_parent_mut_id = site->mutations[mut_site_index].parent;
+
+                        if (fix_missing_parents && (actual_parent_mut_id == MSP_NULL_MUTATION)) {
+                            site->mutations[mut_site_index].parent = putative_parent_mut_id;
```
petrelharp: IIUC, this bit doesn't actually record the parents in the tree sequence, only in the (temporary) mutations that are kept by that site: it should be
```
self->mutations.parent[mutation_table_index] = putative_parent_mut_id;
```
where `mutation_table_index` also needs to be defined to keep track of which mutation in the overall tree sequence we're looking at.

Is this right, @jeromekelleher? And, if we dynamically update the `tree_sequence.mutations.parent` array, is there anything else in the tree sequence that needs updating afterwards?  I think not...

lib/tree_sequence.c
petrelharp: we're fine with that; the reasoning was that `malloc` never fails (or, if it does, worse things happen) on modern systems; so adding all the checks obfuscates things.

lib/tree_sequence.c
```diff
@@ -986,6 +1001,300 @@ tree_sequence_simplify(tree_sequence_t *self, node_id_t *samples, size_t num_sam
...
+    if (ret != 0) goto out;
+    ret = sparse_tree_first(&tree);
+    if (ret < 0) goto out;
+
+    do
+    {
+        // for this tree, loop through sites
+        for (size_t tree_site_index = 0; tree_site_index < tree.sites_length; ++tree_site_index)
+        {
+            site_t *site = &tree.sites[tree_site_index];
```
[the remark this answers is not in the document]
petrelharp: the argument here was that `site` is scoped only within the loop, thus avoiding errors.

lib/tree_sequence.c
jeromekelleher: Can't say I agree with that reasoning... out of memory conditions occur reasonably often, and you should always try to return a sensible error to the Python interpreter. It's a basic philosophy that I follow: every possible error, ever under any circumstances **must** be accounted for. This is the source of a lot of the weird conventions I've adopted (but I mostly learned them from the CPython code).

petrelharp:
Update: I believe this now to be correctly assigning the parents.

petrelharp:
I looked at the C algorithm and rewrote it in python and in so doing simplified it considerably. 

Note that this is more complicated than the python version in #288 because that function assumed that mutations appeared after their parents; we want to be able to check this assumption here.  Therefore, at each site we:

1. Sort mutations on node, and ID, to retain ordering within nodes.
2. Iterate over mutations; at each we are either 
   (a) on the same node as the previous mutation, so the previous should be our parent
   (b) or, not, in which case we traverse up the tree, at each node searching backwards through the list of mutations to see if there's any on that node

The "backwards" ensures we find the last one on that node.

And recall the purpose of the flag is to control what happens when the existing parent is NULL but should not be: either, error, or fill in the correct one.

It should be easy to translate this back over to the C side. Happy to let anyone else do that, or else I'll get to it.

jeromekelleher:
I've had a play with this @petrelharp and I think I've got a better algorithm. This only needs two integer arrays as state, and requires (1) a single forward pass through the mutations; and (2) a single tree traversal for each site. It could possibly be done more efficiently, but I think this is the simplest way.

I don't think we should complicate this with checking the correctness of an existing parent column. I don't see what this would be used for aside from testing, and there's no need to have it done at ``load`` time to achieve this.

If you agree and you like the algorithm I'll code it up in C. We can add some extra tests steps to the test framework we've set up here so that we also check that we get the correct result when we set the parent column to -1 and call ``load_tables`` with the COMPUTE_PARENTS flag set.

jeromekelleher:
Aha, no, I changed my mind. Rather than do a full tree traversal for each site, it's a lot more efficient if we just traverse upwards from the node for each mutation (under reasonable assumptions). This is what you were doing before @petrelharp?

Anyway, this is hard to beat now I think unless we start caching paths through the tree. I don't think it's worth this though, unless this turns out to be an expensive and regular operation.

petrelharp:
> I don't think we should complicate this with checking the correctness of an existing parent column. I don't see what this would be used for aside from testing, and there's no need to have it done at load time to achieve this.

Hm, I think it's easy to check, although to check correctness does require one extra pass through the mutations. And, it's probably an easy thing to get wrong, so worthwhile checking?  We should at least have code to do this *somewhere*.

But, maybe arguing in the other direction: in our examples and papers we often omit the `mutation.parent` column. We are also not going to the substantial trouble to output it in forwards simulations. This suggests to me that a common use case will be to omit the column entirely, and let it be inferred from order. So, I think that the default behavior for `load_tables` should be to infer the `mutation.parent` column if it does not exist.  And if this is the expected use case, maybe we don't need to check parent correctness every time.

petrelharp:
> Aha, no, I changed my mind. Rather than do a full tree traversal for each site, it's a lot more efficient if we just traverse upwards from the node for each mutation (under reasonable assumptions). This is what you were doing before @petrelharp?

That is what I was doing.

tests/tsutil.py
```diff
@@ -373,32 +373,44 @@ def jukes_cantor(ts, num_sites, mu, multiple_per_node=True, seed=None):
...
+            for mutation in site.mutations:
+                u = mutation.node
+                if top_mutation[u] != msprime.NULL_MUTATION:
+                    # Traverse upwards until we find a another mutation or root.
+                    v = tree.parent(u)
+                    while v != msprime.NULL_NODE \
+                            and top_mutation[v] == msprime.NULL_MUTATION:
+                        v = tree.parent(v)
+                    if v != msprime.NULL_NODE:
+                        mutation_parent[top_mutation[u]] = bottom_mutation[v]
```
petrelharp: Couldn't this bit be
```
                v = tree.parent(mutation.node)
                # Traverse upwards until we find a another mutation or root.
                while v != msprime.NULL_NODE \
                            and top_mutation[v] == msprime.NULL_MUTATION:
                    v = tree.parent(v)
                if v != msprime.NULL_NODE:
                        mutation_parent[top_mutation[mutation.node]] = bottom_mutation[v]
```

petrelharp:
Yes, that's very nice. You've replaced the sorting step with keeping track of the top and bottom mutations.  It might be a bit clearer to replace the `top_mutation` with an `already_checked` array of booleans; see comments.

petrelharp:
Er, there - I've made some suggested changes.

jeromekelleher:
> Yes, that's very nice. You've replaced the sorting step with keeping track of the top and bottom mutations. It might be a bit clearer to replace the top_mutation with an already_checked array of booleans; see comments.

Very nice. I've tweaked this a little bit to use some existing state to check for branch groups, which will avoid a bit of mallocing tedium on the C side. Also note that we can skip the first mutation, since that cannot have a parent. This allows us to also efficiently handle the common case of one mutation, without any inelegant special cases.

jeromekelleher:
> Hm, I think it's easy to check, although to check correctness does require one extra pass through the mutations. And, it's probably an easy thing to get wrong, so worthwhile checking? We should at least have code to do this somewhere.

Well it does incur a full tree generation pass, with potentially quite a bit of traversing up and down the trees.  This is more than I'm willing to accept as the default cost of a load, I think. What are we guarding against here? I don't think we can have memory violations as a result of bad mutation parent information, so I'm reasonably relaxed about it. 

> But, maybe arguing in the other direction: in our examples and papers we often omit the mutation.parent column. We are also not going to the substantial trouble to output it in forwards simulations. This suggests to me that a common use case will be to omit the column entirely, and let it be inferred from order. So, I think that the default behavior for load_tables should be to infer the mutation.parent column if it does not exist. And if this is the expected use case, maybe we don't need to check parent correctness every time.

How would we know it doesn't exist though? I think having a flag to compute the parent column is fine. We capture the use case in which people don't want to be bothered computing it, but only if we're asked. It's a substantial extra cost, which will make it impossible to have fast load times. Fast load times are something I'm keen on, which we'll be able to do very well with ``kastore``. Any checks beyond a single pass through the data will hurt this.

petrelharp:
Ok, you've convinced me. We can leave it as optional.

> ... what are we guarding against?

The most likely scenario is someone reading in a table with multiple mutations that does not have `mutation.parent` set (i.e., they are all `-1`); and then getting wrong results.

But, it would be nice if there was *some* indication that there could be a problem in this case, because I think that currently the code would behave as if everything was fine, but do incorrect things.  A simple but not very good check would be to, say, produce a warning if some sites had more than one mutation, but no mutations had `parents` specified.

> How would we know if it doesn't exist?

Well, when reading in from text (which is my mental model :) this is easy: see if there's a column labeled `parent`. In binary formats, i don't know.

jeromekelleher:
I'm just updating the C code here at the moment and I've hit loads of compiler errors (catching most of the problems discussed elsewhere). How did you compile the C code and run the tests?

jeromekelleher:
I've changed this from a load-time option to a method of the tree sequence object. This is because there were all sorts of tricky state-issues cropping up, which would become serious headaches down the road. For example, I'm hoping to use an mmaped file as the memory for the tables used in the tree sequence. Then, we won't be able to write the table memory. In general, anyway, we should see the tables as read-only as far as tree sequences are concerned. I think there would be all sorts of ugly consequences if we violated this principle.

I think this works fine for your immediate application, since you're not particularly interested in actually doing anything with the tree sequence; you just want to compute the mutation parents. (You can even cheat a bit, and pass ``mutation_table.parent`` in as the argument, if you want to avoid an extra memcpy.)  

Dealing with cases in which the  mutation parents provided to ``load`` are incorrect is a separate issue I think.

petrelharp:
I compiled with `make tests` in `lib/` and also `make ext3`?

petrelharp:
>  In general, anyway, we should see the tables as read-only as far as tree sequences are concerned.

OK; I wondered about this.

> Dealing with cases in which the mutation parents provided to load are incorrect is a separate issue I think.

So, is the correct procedure then `load_tables | tree_sequence_compute_mutation_parents | tree_sequence_dump_tables | mutation_table_set_column`?  I think this makes sense -- we should really add the `mutation.parent` column before the tables exit SLiM, really, although it's tempting to allow SLiM to never deal with tree sequences (just tables).  

Huh: I still think there should be an easy method to add the `mutation.parent` column to a mutation table not having one (somehow).  And, there is an argument to be made for *not* having `compute_mutation_parent` be a method of a tree sequence at the python level; because if you have a tree sequence without correct mutation parents, then it's *incorrect*.  Perhaps this should be a method of a table collection, which under the hood makes a tree sequence, etcetera?  I guess that since it really is something a tree sequence does, then it's fine to expose it in python (useful for testing), but making reference in its docstring to the table collection method.

jeromekelleher:
> Huh: I still think there should be an easy method to add the mutation.parent column to a mutation table not having one (somehow). And, there is an argument to be made for not having compute_mutation_parent be a method of a tree sequence at the python level; because if you have a tree sequence without correct mutation parents, then it's incorrect. Perhaps this should be a method of a table collection, which under the hood makes a tree sequence, etcetera? I guess that since it really is something a tree sequence does, then it's fine to expose it in python (useful for testing), but making reference in its docstring to the table collection method.

I'm pretty much in agreement with you here... All we need is a single pass over the trees, which we don't actually need the full ``sparse_tree_t`` stuff to do. How about this:

- make a method ``table_collection_compute_mutation_parents`` which requires that the indexes have been built (or builds them if they don't exist?). 
- Then, within this function we do a manual tree generation using the basic algorithm (easy, and *fast*). All we need is the ``parent`` array, so there's no faffing with roots, children etc. We use this instead of the ``sparse_tree``, and so then there's no need for the ``tree_sequence_t`` to be involved at all.
- In the future, if someone clever comes up with a way of doing this that doesn't require iterating over the trees, then users of the API will never know any better.

What do you reckon?

petrelharp:
That sounds perfect.

jeromekelleher:
See ``table_collection_build_indexes``. This builds the **I** and **O** indexes needed for [Algorithm T](http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1004842).

jeromekelleher:
> I compiled with make tests in lib/ and also make ext3?

Weird; it should have spat out a bunch of errors.

> That sounds perfect.

OK, I'll have a go.

petrelharp:
> OK, I'll have a go.

That'd be great; in principle, I'm willing, but in practice won't get to it for a bit.  It'd be most efficient for me to contribute to the docs and tests.

jeromekelleher:
Here you go @petrelharp, this should do it. It's going to be very hard to make something faster than this I think. (If there are lots of mutations deep in the trees it might be worth setting up a path cache, but I doubt it'd get us much in the vast majority of cases.)

Unfortunately I couldn't set up a Python interface easily because we don't have the TableCollection object exposed (#424), so we're going to have to do some more testing on the C side to make sure it's all OK. I haven't thought about malicious/stupid inputs here at all yet, so there will need to be some additional paranoia and some tests for it.

petrelharp:
I've added tests for the only paranoia I could think of; which isn't much, because we're making strong assumptions about the input.  However, note that I [have a test](https://github.com/tskit-dev/msprime/pull/461/files#diff-a324fa314273baf14b13a482aab8f602R4246) that messes up the mutation order, so that a mutation appears in the table before its parent; but it does *not* catch this.  This was by design, but a bit worrisome. It would be better if `compute_parents` computed the correct parent; and then `load_tables` would throw an error.

petrelharp:
Ah-ha.  The problem was the "don't bother checking the first mutation" optimization - my out of order mutation happened to be the first one, so it didn't get checked.  I've removed the `+ 1` and added another test or so.  There could be an `if (first_mutation + 1 == mutation) { continue; }` but don't know if that's worth it?

I'm not feeling any more paranoid than this.

jeromekelleher:
I've added some integrity checks to the start of the method @petrelharp (we could have invalid node ids, for example, leading to segfaults). I've moved all of the checks so that errors are detected *before* we call ``tree_sequence_load``. This seems like a good idea to me since plenty people won't be calling tree sequence load at all, and they should get errors  when creating the file rather than when their users try to load it into tskit. Generally, the earlier you catch an error the less pain it causes...

I think it's worth optimising for the common case of one mutation, since the cost of traversing up the tree will be substantial if we do it millions of times.

Nice catch on the parent ordering check! I've moved this into ``compute_parents``, which should help in tracking down what would become hard to find bugs.

I think this is ready to merge now; do you agree?

jeromekelleher:
Also, I meant to add: doing these integrity checks over and over is tedious. We should abstract it out a bit: see #465.

petrelharp:
Looks good to me.  Thanks!!

jeromekelleher:
This is a great addition, thanks @petrelharp!

PR #464, opened 2018-04-26
remove simplify code to remove certain invisible mutations

### LINKED ISSUE #463: revisit simplify removing 'invisible' mutations
[This piece of code](https://github.com/tskit-dev/msprime/blob/00009872d9990286c9e4b06c1761d26850f7428d/lib/tables.c#L3201-L3221) has just bit me. As advertised, what it does, in `simplify_tables`, is
>                     /* If there is no parent and the ancestral state is equal to the
>                      * derived state, then we remove this mutation. */

But: as discussed elsewhere, I'm trying to be lazy about setting mutation parents, since this information is implicit in the trees plus the mutation table ordering. And, here's why it be me: any mutation back to the ancestral state, with no parent specified, will be removed.

You could convince me that no, we just have to somehow properly record mutation parents.  But instead of this, I'd like to propose one of these things:

1. Add a flag that controls this behavior: `MSP_REMOVE_INVISIBLE_MUTATIONS`. Or,
2. Remove this bit. Under what situation does it actually come up? The only situation I can think of is when mutations are placed without reference to the existing state. That's a wierd case, but I guess it'd be nice to deal with it.  But if we're dealing with this case, shouldn't we also compare the derived state to the ancestral state of the parent, and remove it if equal?

I feel like I"m forgetting something; apologies.

petrelharp:
This would close #463.  Note that if we wanted to put this back in, in the future, we should also add the test for derived state == derived state of parent when the parent exists.

### DIFF
--- a/tests/simplify.py
+++ b/tests/simplify.py
@@ -289,13 +289,9 @@ class Simplifier(object):
                 if mut.parent != -1:
                     mapped_parent = mutation_id_map[mut.parent]
                 if mapped_node != -1:
-                    keep = True
-                    if mapped_parent == -1 and site.ancestral_state == mut.derived_state:
-                        keep = False
-                    if keep:
-                        mutation_id_map[mut.id] = num_output_mutations
-                        num_output_mutations += 1
-                        num_output_site_mutations += 1
+                    mutation_id_map[mut.id] = num_output_mutations
+                    num_output_mutations += 1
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Looks good to me. We should remove the corresponding logic from the Python implementation in tests/simplify too though.

PR #468, opened 2018-04-29
Initial kastore

jeromekelleher:


jeromekelleher:
This will need quite a bit more work before it's ready to merge (mainly providing upgrade paths from HDF5 to kastore, and also more testing), but this is basically working now. Bye bye HDF5, you've served us well!

cc @petrelharp and @bhaller

petrelharp:
Moved from the other thread:

There may be a memory leak in table_collection_load(): Beforehand, we need to do a table_collection_alloc(), passing in MSP_ALLOC_TABLES flag; or else e.g. the node_table itself won't be alloc'ed. This calls node_table_alloc(), which allocs self->time. However, node_table_load() calls read_table_cols() with the address to self->time, which calls kastore_gets(), which replaces the pointer in self->time with the pointer to the memory read in by kastore.

jeromekelleher:
Thanks @petrelharp, well spotted. I'm surprised this didn't show up in valgrind, but I may have forgotten to run it.

jeromekelleher:
> There may be a memory leak in table_collection_load(): Beforehand, we need to do a table_collection_alloc(), passing in MSP_ALLOC_TABLES flag; or else e.g. the node_table itself won't be alloc'ed. This calls node_table_alloc(), which allocs self->time. However, node_table_load() calls read_table_cols() with the address to self->time, which calls kastore_gets(), which replaces the pointer in self->time with the pointer to the memory read in by kastore.

I see the issue here @petrelharp --- when calling ``table_collection_load`` you **don't** call ``table_collection_alloc`` with the ``MSP_ALLOC_TABLES``. The tables are embedded within the table_collection_t struct, so we don't need to allocate memory for the table itself, just for the columns. So, ``MSP_ALLOC_TABLES`` really means, "allocate memory for table columns". I've the tests through valgrind, and it's all checking out fine.

jeromekelleher:
I've done a good bit more work here @petrelharp, and I think it's nearly ready to be merged. Since this is such a critical change, it would be great if you could have a look over it and see if you spot any problems.

TODO:
- Define high-level exceptions for file format errors, and have these output useful error messages.
- Add check for legacy HDF5 file formats at ``load`` time, and raise correct exception. (This is essential, because otherwise users won't know that they need to run msp upgrade.
- Propagate FormatTooNew and FormatTooOld exceptions up to the top level

Hopefully I'll get these done in the next day or two and we can merge.

petrelharp:
I've looked through everything; it looks great.  Thanks a bunch for this, I think it's a big improvement.

One thing I'm not clear on: if you `load_indexes` from a kastore, then `malloced_locally` is `false`. Why is that? Like, who is in charge of dealloc'ing in that case?

jeromekelleher:
Just for the record, here are some load benchmarks from before and after for a tree sequence with 10^6 samples:
```
jk@powderfinger$ python3 tmp.py 10000000.trees 
msprime =  0.5.1.dev77+g0340755.d20180508
loaded in  1.7452379460446537
jk@powderfinger$ python3 tmp.py ../10000000.uncompressed.hdf5 
msprime =  0.5.0
loaded in  3.889940964989364
```
Load time is quite a bit faster. Most of the remaining time is spent in tree_sequence_load. I don't think we need to make it any faster for now though.

Here are the sizes:
```
jk@powderfinger$ ls -lh ../10000000.*
-rw-r--r-- 1 jk jk 1.1G May  8 12:15 ../10000000.trees
-rw-r--r-- 1 jk jk 785M Apr 12 10:27 ../10000000.uncompressed.hdf5
```
The kastore is a bit bigger all right. Here's where the space is going:
```
jk@powderfinger$ kastore ls 10000000.trees  -lH
int32   22482605  85.8M edges/child
float64 22482605 171.5M edges/left
int32   22482605  85.8M edges/parent
float64 22482605 171.5M edges/right
int8          11    11B format/name
uint32         2     8B format/version
int32   22482605  85.8M indexes/edge_insertion_order
int32   22482605  85.8M indexes/edge_removal_order
int32          0     0B migrations/dest
float64        0     0B migrations/left
int32          0     0B migrations/node
float64        0     0B migrations/right
int32          0     0B migrations/source
float64        0     0B migrations/time
uint8     669526 653.8K mutations/derived_state
uint32    669527   2.6M mutations/derived_state_offset
uint8          0     0B mutations/metadata
uint32    669527   2.6M mutations/metadata_offset
int32     669526   2.6M mutations/node
int32     669526   2.6M mutations/parent
int32     669526   2.6M mutations/site
uint32  20479399  78.1M nodes/flags
uint8          0     0B nodes/metadata
uint32  20479400  78.1M nodes/metadata_offset
int32   20479399  78.1M nodes/population
float64 20479399 156.2M nodes/time
uint8        839   839B provenances/record
uint32         3    12B provenances/record_offset
uint8         52    52B provenances/timestamp
uint32         3    12B provenances/timestamp_offset
float64        1     8B sequence_length
uint8     669526 653.8K sites/ancestral_state
uint32    669527   2.6M sites/ancestral_state_offset
uint8          0     0B sites/metadata
uint32    669527   2.6M sites/metadata_offset
float64   669526   5.1M sites/position
```

petrelharp:
>  the memory is really in the kastore, which is freed when the store is closed. 

Ah-ha.  That was the part I was missing.  I noticed the memory was in the kastore, but hadn't figured out that the correct workflow was to just close it when you're all done (and not before). See, I was thinking of the store like a file: read the data in from it, then close it, then do stuff with the data. But the store is also the container.

petrelharp:
> `$ kastore ls 10000000.trees  -lH`

wow, pretty sweet.

bhaller:
Hi guys.  I'm late to the party.  @jeromekelleher when you write:

> I see the issue here @petrelharp --- when calling table_collection_load you don't call table_collection_alloc with the MSP_ALLOC_TABLES. The tables are embedded within the table_collection_t struct, so we don't need to allocate memory for the table itself, just for the columns. So, MSP_ALLOC_TABLES really means, "allocate memory for table columns".

I'm not sure exactly what that really means; I can't tell whether you're saying that we do need MSP_ALLOC_TABLES, or that we don't, and in which cases.  We have several places where we call table_collection_alloc() in SLiM, and I don't know which ones should pass MSP_ALLOC_TABLES and which shouldn't.  One is in SLiMSim::StartTreeRecording(); that one presumably should, since it is starting recording into new tables.  But then there is the new SLiMSim::TreeSequenceDataFromAscii() method from @petrelharp which does:

    int ret = table_collection_alloc(p_tables, MSP_ALLOC_TABLES);
    if (ret < 0) handle_error("read_from_ascii", ret);

    ret = table_collection_load_text(...

I suppose whether MSP_ALLOC_TABLES is right there depends on what Peter defines table_collection_load_text() to do, so this is just something to be careful on.  Then we have SLiMSim::TreeSequenceDataToAscii(), which does:

	table_collection_t tables_copy;
	int ret = table_collection_alloc(&tables_copy, MSP_ALLOC_TABLES);
	if (ret < 0) handle_error("convert_to_ascii", ret);
	
	ret = table_collection_copy(p_tables, &tables_copy);
	if (ret < 0) handle_error("convert_to_ascii", ret);

Is that correct?  Similarly, SLiMSim::WriteTreeSequence() does:

	table_collection_t output_tables;
	ret = table_collection_alloc(&output_tables, MSP_ALLOC_TABLES);
	if (ret < 0) handle_error("table_collection_alloc", ret);
	
	ret = table_collection_copy(&tables, &output_tables);
	if (ret < 0) handle_error("table_collection_copy", ret);

And SLiMSim::CrosscheckTreeSeqIntegrity() does:

		table_collection_t *tables_copy;
		
		tables_copy = (table_collection_t *)malloc(sizeof(table_collection_t));
		ret = table_collection_alloc(tables_copy, MSP_ALLOC_TABLES);
		if (ret != 0) handle_error("CrosscheckTreeSeqIntegrity table_collection_alloc()", ret);
		
		ret = table_collection_copy(&tables, tables_copy);
		if (ret != 0) handle_error("CrosscheckTreeSeqIntegrity table_collection_copy()", ret);

So for all of those, I guess the question is: does table_collection_copy() expect tables to already be allocated, or does it do that itself (which might make more sense, since it knows how big the allocs for the columns need to be)?

Finally, in SLiMSim::_InitializePopulationFromMSPrimeBinaryFile() I have:

	// free the existing table collection
	int ret = table_collection_free(&tables);
	if (ret != 0) handle_error("table_collection_free", ret);
	
	// read the file from disk
	ret = table_collection_alloc(&tables, MSP_ALLOC_TABLES);
	if (ret != 0) handle_error("table_collection_alloc", ret);
	
	ret = table_collection_load(&tables, p_file, 0);
	if (ret != 0) handle_error("table_collection_load", ret);

Does this code make sense?

As an aside: why have this complication over whether or not to pass MSP_ALLOC_TABLES, in the first place?  Wouldn't it be possible to just have a policy of always deferring allocation of the columns until they are actually needed, and then always allocating them silently?

petrelharp:
Good questions. I think I have fixed this in SLiM, I believe: https://github.com/jgallowa07/SLiM/commit/547b234a55accb55251f6eed170f92db619dc801
(thanks for the reminder)

bhaller:
OK, so table_collection_load() does not want MSP_ALLOC_TABLES, but table_collection_copy() does?

jeromekelleher:
I think it *is* inconsistent and confusing at the moment @bhaller; I've made an issue to track the problem (#480). This is one of these things that has grown over time and been co-opted into new situations and needs to be rethought for the final API.

In the mean time, the best I can suggest is to try leaving it out first. If you segfault, then you needed to put it in! I know this isn't satisfactory, but it's the best I can offer at the moment I'm afraid.

bhaller:
OK.  :->

PR #481, opened 2018-05-17
Text input method

petrelharp:
This would be a handy thing to have.  But, maybe it should be kept separate?  And, it should really be against #472.

This does not do population tables yet.

I believe it is working; but, it fails on the "gappy" test tree sequence, because that one puts a site to the right of all the edges; and this method has no way to record sequence_length other than inferring it from the edges.  I can just skip that tests, unless there's a compelling reason not to. 

This is not supposed to be a completely robust input method: it assumes that all columns are present, and that columns are separated by exactly one tab.

Also, there were (predictably) issues with floating point precision. I've changed all the precisions in `_dump_text` to be larger to fix this problem.  How do you feel about this, @jeromekelleher?

### DIFF
--- a/lib/tables.c
+++ b/lib/tables.c
@@ -184,6 +184,69 @@ out:
     return ret;
 }
 
+/* ****
+ * Tab-separated parsing:
+ * These will be used on null-terminated strings, which are included,
+ * so `end`, if not NULL, will always be at least one before the end
+ * of the string.
+ *
+ * These return: 
+ *   1 if `sep` is found and delimits a nonzero-length token; 
+ *   0 if `sep` is found as the first character;
+ *   and -1 otherwise.
+ * ***/
+
+int
+get_sep_atoi(char **start, int *out, int sep)
+{
+    int ret;
+    char *next;
+    next = strchr(*start, sep);
+    if (next == NULL) {
+        ret = -1;
+    } else {
+        ret = (int) (next != *start);
+        *next = '\0';
+    }
+    *out = atoi(*start);
+    *start = (next == NULL) ? NULL : next + 1;
+    return ret;
+}
+
+int
+get_sep_atof(char **start, double *out, int sep)
+{
+    int ret;
+    char *next;
+    next = strchr(*start, sep);
+    if (next == NULL) {
+        ret = -1;
+    } else {
+        ret = (int) (next != *start);
+        *next = '\0';
+    }
+    *out = atof(*start);
+    *start = (next == NULL) ? NULL : next + 1;
+    return ret;
+}
+
+int
+get_sep_atoa(char **start, char **out, int sep)
+{
+    int ret;
+    char *next;
+    next = strchr(*start, sep);
+    if (next == NULL) {
+        ret = -1;
+    } else {
+        ret = (int) (next != *start);
+        *next = '\0';
+    }
+    *out = *start;
+    *start = (next == NULL) ? NULL : next + 1;
+    return ret;
+}
+
 
 /*************************
  * node table
@@ -458,7 +521,7 @@ node_table_print_state(node_table_t *self, FILE *out)
 int
 node_table_dump_text(node_table_t *self, FILE *out)
 {
-    int ret = 0;
+    int ret = MSP_ERR_IO;
     size_t j;
     table_size_t metadata_len;
     int err;
@@ -469,7 +532,7 @@ node_table_dump_text(node_table_t *self, FILE *out)
     }
     for (j = 0; j < self->num_rows; j++) {
         metadata_len = self->metadata_offset[j + 1] - self->metadata_offset[j];
-        err = fprintf(out, "%d\t%d\t%f\t%d\t%d\t%.*s\n", (int) j,
+        err = fprintf(out, "%d\t%d\t%.17g\t%d\t%d\t%.*s\n", (int) j,
                 (int) (self->flags[j] & MSP_NODE_IS_SAMPLE),
                 self->time[j], self->population[j], self->individual[j],
                 metadata_len, self->metadata + self->metadata_offset[j]);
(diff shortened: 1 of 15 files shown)


lib/tables.c
```diff
@@ -482,6 +545,75 @@ node_table_dump_text(node_table_t *self, FILE *out)
...
+    err = getline(&line, &k, file);
+    if (err < 0) {
+        goto out;
+    }
+    err = strcmp(line, header);
+    if (err != 0) {
+        goto out;
+    }
+
+    while ((err = getline(&line, &k, file)) != -1) {
```
jeromekelleher: Unfortunately getline is a POSIX function, so this won't work on windows.

lib/tables.c
```diff
@@ -482,6 +545,75 @@ node_table_dump_text(node_table_t *self, FILE *out)
...
+    size_t k;
+    size_t MAX_LINE = 1024;
+    char *line = NULL;
+    double time;
+    int flags, population, individual, id, is_sample;
+    char *name;
+    const char *header = "id\tis_sample\ttime\tpopulation\tindividual\tmetadata\n";
+    char *start;
+
+    line = malloc(MAX_LINE);
```
jeromekelleher: Need to check return value here.

lib/tables.c
```diff
@@ -482,6 +545,75 @@ node_table_dump_text(node_table_t *self, FILE *out)
     return ret;
 }
 
+int
+node_table_load_text(node_table_t *node_table, FILE *file)
+{
+    int ret = MSP_ERR_FILE_FORMAT;
+    int err;
+    size_t k;
+    size_t MAX_LINE = 1024;
```
jeromekelleher: Since we're accepting arbitrary metadata here, I don't think a fixed alloc is going to work. We're going to need someway to read in an arbitrary line length.

lib/tables.c
```diff
@@ -482,6 +545,75 @@ node_table_dump_text(node_table_t *self, FILE *out)
...
+        if (err <= 0) {
+            goto out;
+        }
+        err = get_sep_atoa(&start, &name, '\n');
+        if (err < 0 || *start != '\0') {
+            goto out;
+        }
+        ret = node_table_add_row(node_table, flags, time, population, individual,
+                name, strlen(name));
+        assert(ret >= 0);
```
jeromekelleher: Check and goto out. Nothing to be gained here from not doing it, as ``add_row`` will check its parameters and potentially return errors.

lib/tables.c
jeromekelleher: Scratch that, readline reallocs the buffer as necessary.

petrelharp:
@bhaller had the use case: what do you think?

petrelharp:
I've tidied those things up. There's still an issue with the "gappy" example: even after settting the sequenc length to be the max of the largest `right` position in  an `edge` *and* the largest `position` in a `site`, it fails because now there's a site at the right endpoint of the sequence.

Once we decide where this is going we can figure out how to address this.

bhaller:
"@bhaller had the use case: what do you say: how useful/important is this? only for debugging, or for general purpose? I am not motivated to make this cross-platform, personally."  I suppose there's probably some way for me to reply in place up where that comment occurred, but I can't figure out how.  :->  Well, at present the SLiM Eidos API supports writing out text tables, so it seemed natural to allow them to be read in, too; that's about as far as my thinking on it went :->.  I actually thought this was a standard msprime file format, but now I am getting the impression from you guys that it's more of an internal/debug format, not something end users are expected to use?  Well, anyway, if reading them in works on all POSIX platforms, that is good enough; SLiM makes no attempt to be Windows-compatible, and is probably already limited to POSIX-compliant systems, so that's fine.  It does seem potentially useful for end users, since they might conceivably want to read/modify the text files (harder to do with binary files), but if it makes sense to remove the text-output option for trees files, I'm not opposed.

petrelharp:
How about this - I'll move this code over to SLiM; we can use it there: it may not be 100% safe, but I do trust it to round-trip tables (that have ASCII metadata not including tabs). still want it, @bhaller?

jeromekelleher:
> How about this - I'll move this code over to SLiM; we can use it there: it may not be 100% safe, but I do trust it to round-trip tables (that have ASCII metadata not including tabs). still want it, @bhaller?

Just to stand back a bit from this --- why bother with outputting text from SLiM in the first place? It's vastly less efficient and you have the problem of dealing around 6 (I've lost count) different files rather than one. Provenance is much more problematic, because you've lost the way that a particular node, edge, site, mutation, etc tables are bound together.

If it's so that a user can 'look' at the data, then the commands:
```
$ cat somets.edges.txt
```
and 
$ msp edges somets.trees
```
give exactly the same results. The ``msp nodes``, ``msp edges`` etc commands all take a trees file as input and output the text form. (These commands will be moved into the ``tskit`` program when tskit is done.) This is very useful for debugging, or just getting a sense of what the data looks like.

Basically, IMO text-based interchange is something left-over from a bygone age. Nobody moves real data around as text anymore, and I want to discourage the use of text in tskit where ever possible.

bhaller:
I like the idea of moving it to SLiM, if it doesn't belong in tskit.  It's very useful to be able to get text output files for debugging, and having an extra step of using msp to view that (which depends upon my having a working msp build, etc.) is just a hassle.  And being able to read in text files will similarly be very useful for debugging in SLiM, since the steps after the tables are read in are all the same; it's a useful way to see and control what's going in, and to isolate whether a problem is in the table-read stage or a later stage.  I am strongly in favor of having this functionality, even if it is only in SLiM.  However, it does seem like we probably want to remove the option for it from the user-visible API; there is, perhaps, no reason why an end user would ever want to save out text trees info.

My battery is at 11% and I appear to have left my charger brick/cable in the U.S. (aargh!), so I will be going dark until I manage to find a new one here in Korea.  And I'm not even in Seoul, so that may not be easy.  :-<  Peace out.

jeromekelleher:
Sound good to me! I think it would be a good idea to remove text from the user-visible API in, but keeping this around in SLiM for debugging sounds like an excellent idea.

bhaller:
OK, the "binary" parameter to treeSeqOutput() is now named "_binary", which has the effect of making it invisible to the user; but it can still be supplied by us for debugging purposes (if we remember that it's there :->).

jeromekelleher:
> OK, the "binary" parameter to treeSeqOutput() is now named "_binary", which has the effect of making it invisible to the user; but it can still be supplied by us for debugging purposes (if we remember that it's there :->).

Sounds perfect to me. Glad you found your charger!

@petrelharp, shall we close this PR then?

petrelharp:
I've moved this code over.  Note that as attractive as text input seems, there are unresolved issues around the few things not stored in tables (well, currently just one thing: `sequence_length`).

PR #483, opened 2018-05-24
avl syntax fixes

petrelharp:
Fixes thanks to @bhaller that I think have been discussed already with @jeromekelleher.

### DIFF
--- a/lib/avl.c
+++ b/lib/avl.c
@@ -142,7 +142,10 @@ int avl_search_closest(const avl_tree_t *avltree, const void *item, avl_node_t *
 	node = avltree->top;
 
 	if(!node)
-		return *avlnode = NULL, 0;
+	{
+		*avlnode = NULL;
+		return 0;
+	}
 
 	cmp = avltree->cmp;
 
@@ -153,14 +156,21 @@ int avl_search_closest(const avl_tree_t *avltree, const void *item, avl_node_t *
 			if(node->left)
 				node = node->left;
 			else
-				return *avlnode = node, -1;
+			{
+				*avlnode = node;
+				return -1;
+			}
 		} else if(c > 0) {
 			if(node->right)
 				node = node->right;
 			else
-				return *avlnode = node, 1;
+			{
+				*avlnode = node;
+				return 1;
+			}
 		} else {
-			return *avlnode = node, 0;
+			*avlnode = node;
+			return 0;
 		}
 	}
 }
--- a/lib/avl.c
+++ b/lib/avl.c
@@ -25,6 +25,8 @@
     the sourcecode, change the interface and squash a few bugs.
     Mail him if you find new bugs.
 
+    Minor modifications to remove compiler warnings by Ben Haller on 6/1/2018.
+
 *****************************************************************************/
 
 #include <stdio.h>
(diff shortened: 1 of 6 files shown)


petrelharp:
Also a minor fix in `trees.h` that caused errors if it was included more than once.

jeromekelleher:
Looks good, thanks @petrelharp. Since this is someone else's code that we're modifying, I feel like we should record this fact somewhere in the header. Just the date and something like "Minor modifications to remove compiler warnings"?

This file won't be part of ``tskit`` BTW, so technically we don't really need this change if we're not worried about these compiler warnings in tskit-less msprime.

bhaller:
The warnings that remain on my side are all to do with unused parameters (in tree_sequence.c, tables.c, and newick.c).  I saw that you defined KAS_UNUSED() in kastore.h, and I've been wondering whether it would be helpful for me to define essentially the same thing in tskit somewhere and add it where needed to get rid of those warnings.  Would that go in util.h?  I see that everything there still uses an "MSP_" naming convention; do you intend to rename stuff to TSK_ or some such?  Anyway, if you give me a bit of guidance I'm happy to push a fix for the remaining warnings that @petrelharp can then add to this pull request...

jeromekelleher:
> The warnings that remain on my side are all to do with unused parameters (in tree_sequence.c, tables.c, and newick.c). I saw that you defined KAS_UNUSED() in kastore.h, and I've been wondering whether it would be helpful for me to define essentially the same thing in tskit somewhere and add it where needed to get rid of those warnings. Would that go in util.h? I see that everything there still uses an "MSP_" naming convention; do you intend to rename stuff to TSK_ or some such? Anyway, if you give me a bit of guidance I'm happy to push a fix for the remaining warnings that @petrelharp can then add to this pull request...

Sure, go for it! Yes, we'll change all the "MSP_" prefixes to "TSK_" when tskit finally gets split out (but lets keep them as MSP_ for now, for simplicity). ``util.h`` is a good place to put this.

bhaller:
So if you've got a moment to explain, what does `#define MSP_UNUSED(x) MSP_UNUSED_ ## x __attribute__((__unused__))` actually do?  :->

jeromekelleher:
The short answer is "it does what we want it to do", the longer answer is "I don't really know the details, I just copied it from somewhere when I googled how to do this sort of stuff"!

The idea of having a macro rather than just putting in the ``__attribute__((__unused__)`` values though is so that it can be used on compilers that don't support this sort of syntax (i.e., MSVC).

bhaller:
> The short answer is "it does what we want it to do", the longer answer is "I don't really know the details, I just copied it from somewhere when I googled how to do this sort of stuff"!

lol.  The mysteries of the C preprocessor.

> The idea of having a macro rather than just putting in the __attribute__((__unused__) values though is so that it can be used on compilers that don't support this sort of syntax (i.e., MSVC).

Yep, that part I get.  :->

petrelharp:
OK. I merged those WARN UNUSEDs in. Hopefully this doesn't make merging in `individual_tables` difficult.

jeromekelleher:
Looks good, thanks. I'll merge one CI is done. Should be easy to rebase #472 on the result.

PR #505, opened 2018-06-03
Add UUID to file format.

### LINKED ISSUE #479: Add UUID to file format.
Add a UUID to each tree sequence file so that it is uniquely identifiable. This is to facilitate provenance information, where we want to track the history of a given file through other files.

A randomly generated uuid4 is most appropriate.

jeromekelleher:
Closes #479 . 

This should do fine for the unix implementation. Will need to do something else to get random bytes for Windows, but it shouldn't be too hard.

The protocol is to just generate a new UUID for every time we dump. I think this is probably fine --- we can worry about the how to store the provenance of UUIDs we're generating from later. Basically, the UUID is a property of the **file** not the tree sequence/tables.


petrelharp:
Random bytes; why do anything else? Like it.

PR #507, opened 2018-06-04
Initial version of mutate().

### LINKED ISSUE #473: Add function msprime.mutate(ts)
We need a nice simple way to dump neutral mutations on a tree sequence. What about:
```python
def mutate(tree_sequence, mutation_rate,  model=None):
     """
     Returns a copy of the specified tree sequence with mutations thrown down under the specified model.
     If model is None, the default is simple infinite sites mutations.
     """
     # Do stuff
     return new_tree_sequence
```
  
Any thoughts @petrelharp or  @molpopgen? I'm trying to keep it simple but leave the door open for more complex mutation in the future with the model argument. Maybe the rate parameter is a mistake though, and all the parameters should be subsumed into the MutationModel?

We can also do a version that takes a table collection as an argument I guess. I would imagine that most mutation models will need the trees though...

jeromekelleher:
Closes #473.

Pinging @petrelharp, @bhaller and @molpopgen for review.

We can add complexity from there, but I think this is what the basic interface and default behaviour should be --- simple infinite sites 0/1 mutations, overwriting any existing mutations.

The main thing we need for forward simulations I think is to add a ``keep`` (default=False) argument that tells ``mutate`` to keep any existing mutations and discard any new mutations that are generated with the same site position.

Someone mentioned a 'since' argument to limit the time slice over which mutations are applied. This is a good idea, but I think we can add this a planned feature rather than something essential in the first version. If we agree we can open an issue to track this and schedule it in for some future release.

In the future then we'll add a ``model=msprime.JukesCantor()`` argument (e.g.) which will allow the user to generate mutations under some more complex model than infinite sites. (See #473 for more discussion here.) I might add a ``msprime.InfiniteSites`` model here now just to make things concrete and make it clear how we envisage the API growing over time.


bhaller:
This plan sounds reasonable to me, and I think this implementation will suffice for my needs at the present moment (i.e. will let me release SLiM 3, etc.).  The `keep` parameter sounds like a good idea for the future, and with infinite sites the probability of a site collision will be very small, so it will be able to effectively do the same thing as SLiM with respect to "stacking" of mutations at the same base position, I guess.

We will eventually want to be able to open a .trees file with overlaid mutations back up in SLiM, to run further simulation on it, I think; to do that the overlaid mutations will need to get the appropriate metadata added to them, which I believe @petrelharp has a plan for...?  That SLiMification process would also involve rounding off site positions to integers and merging mutations at the same site into stacked mutations with the appropriate stacked derived state, so it will be a bit of an adventure to do right.  But the ability to do that does not need to make it into 0.6.0 or SLiM 3, I think.  So for 0.6.0 this looks great, and I think the future plans should work.

molpopgen:
This looks like a good  start.  I think the `since` idea was mine, and seconded by @petrelharp.  It is handy when processing forward simulations "on the go" rather than after they're done.  We can definitely table it for now, though.

jeromekelleher:
> This plan sounds reasonable to me, and I think this implementation will suffice for my needs at the present moment (i.e. will let me release SLiM 3, etc.). The keep parameter sounds like a good idea for the future, and with infinite sites the probability of a site collision will be very small, so it will be able to effectively do the same thing as SLiM with respect to "stacking" of mutations at the same base position, I guess.

I think we probably do need this for the initial version as putting neutral mutations onto something with selected sites seems like a fairly important thing to do. Let's see what @petrelharp thinks though.

> We will eventually want to be able to open a .trees file with overlaid mutations back up in SLiM, to run further simulation on it, I think; to do that the overlaid mutations will need to get the appropriate metadata added to them, which I believe @petrelharp has a plan for...? That SLiMification process would also involve rounding off site positions to integers and merging mutations at the same site into stacked mutations with the appropriate stacked derived state, so it will be a bit of an adventure to do right. But the ability to do that does not need to make it into 0.6.0 or SLiM 3, I think. So for 0.6.0 this looks great, and I think the future plans should work.

Sounds good. I've opened issue #509 to track this.

jeromekelleher:
> This looks like a good start. I think the since idea was mine, and seconded by @petrelharp. It is handy when processing forward simulations "on the go" rather than after they're done. We can definitely table it for now, though.

Great, thanks. I've opened issue #508 to track this.

petrelharp:
Looks good for now. Anyone sufficiently motivated could use infinite sites as fodder for a more complicated scheme.

So, would the current `MutationGenerator` be renamed `InfiniteSitesMutationGenerator`?

jeromekelleher:
> So, would the current MutationGenerator be renamed InfiniteSitesMutationGenerator?

This would be hidden away I think --- the interface is too low-level to expose to users. I think if we did end up writing different mutation models the whole low-level infrastructure would need to change, so this class would probably disappear.

It might not even be necessary do some models in C for example. I'm sure we could rewrite the current infinite sites generator in numpy and it would be exactly as fast as the C method we have. The reasons I haven't done this so far is (a) I can't be bothered as it all works pretty well at the moment, and such a change would need to be carefully tested statistically; and (b) it's handy having mutations for testing in C.

petrelharp:
Wow, very impressive. Looking now.

> There's bunch of debugging in there at the moment as I've hit a problem with deduplicate_sites. [...]  I also realise that I don't understand what it does with mutations. Shouldn't we remove any mutations for sites we've removed? 

Not sure I follow, but: the idea with `deduplicate_sites` is that we don't want to bother looking up if a site has already been mutated at some point or not, so we just throw in a new site for every mutation. So, if sites `a` and `b` both have the same `position`, then these will get collapsed into one (with label `c`, say), and all mutations whose `site` is `a` or `b` will now have label `c`.

petrelharp:
Other notes about `deduplicate_sites`:

- Suppose in the example above that the mutation at site `a` is above the mutation at site `b` in the tree. The scheme above might reasonably check if the ancestral state at `b` matches the derived state at the mutation at site `a`. We do **not** check this, however, because getting this right would require the sort of looking things up that we want to avoid.

- After deduplicating sites, you also need to `compute_mutation_parents`.  This is a separate function because we don't need to do it as we go along in the simulation, but maybe it should also be done by the python `deduplicate_sites`, since if you don't do it then you don't have a valid tree sequence.

petrelharp:
Other than the issue with `mutation_parents`, the method looks good.

jeromekelleher:
Thanks @petrelharp. I'm not sure I agree about the logic on ``deduplicate_sites`` to be honest, it seems pretty complicated...

I think there's a simpler way though. What if we just rejection sample any mutations that happen to fall on existing site positions in the MutationGenerator? It's simple to do, and I think it's more logically consistent. We're generating infinite sites mutations, so by definition these should not fall on any site that already exists. That way we don't need ``deduplicate_sites`` at all, and we don't need to call ``compute_mutation_parents`` either.

What do you think?

molpopgen:
> I think there's a simpler way though. What if we just rejection sample any mutations that happen to fall on existing site positions in the MutationGenerator?

This is how I do it, and it is super efficient given a good hash table implementation.

jeromekelleher:
> This is how I do it, and it is super efficient given a good hash table implementation.

It's already done within the existing mutation generator (using an AVL tree so not that fast, but good enough) to ensure that there are no duplicate site positions produced. So all we need to do here is to populate the lookup table with the site positions that are already in the table and it should all just work.

petrelharp:
I think that would be the most consistent thing to do, because new infinite sites mutations should really not hit existing sites...

petrelharp:
A bit more on the logic of `deduplicate_sites` - I think you mean the logic of why we need it at all? - it seems natural that a forwards simulation ought to be able to know which positions are mutated without too much hassle.  That's true - maybe with a bit of bookkeeping - for the question of "what positions are polymorphic in the current population".  But that's not the question we need to ask when inserting into the Site table.  For that, we need to know if there have been any mutations at a given position in the past that are still present in the tables, i.e., are polymorphic in any individuals that haven't been simplified away yet (including ancestral samples recorded in the tables).

OK, so the list of positions to check against isn't already available in the simulator; but it is certainly available in the Site table; so all we'd have to do is to check if a position is already present in `tables.sites.position`, and add it if not.  But, this should really be the job of `tskit`, to preserve the one-way principle.  So, maybe we could add internal state to the site table that allows easy lookup for existing positions, and then have `site_table_add_row()` check first to see if the position exists.  But, that seems a lot more complicated, and less efficient, than our current solution.

jeromekelleher:
OK, thanks for the rationale @petrelharp, that's helpful. I think the way forward here so is:

1. Change the mutation generator to look at the existing sites and rejection sample. This means that the ``mutate`` function is nice and simple and doesn't need to call deduplicate sites, etc.

2. File an issue noting the bug in deduplicate sites, and leave it undocumented for now.

I'm still not crazy on the ``deduplicate_sites`` thing, as I tend to think it's the job of the code that generates the mutations to decide how the state transitions occur. It seems very messy to just bung mutations for one site onto another and hope for the best that they make sense. 

One thing a simulator could do to look up the sites is to use binary search on the sorted positions in ``tables.sites.position``. This will be fast enough for most purposes...

petrelharp:
> 1.    Change the mutation generator to look at the existing sites and rejection sample. This means that the mutate function is nice and simple and doesn't need to call deduplicate sites, etc.

Yes, agree.

>    File an issue noting the bug in deduplicate sites, and leave it undocumented for now.

What is the bug? But I guess I will see when you file the bug.

> I'm still not crazy on the deduplicate_sites thing, as I tend to think it's the job of the code that generates the mutations to decide how the state transitions occur. It seems very messy to just bung mutations for one site onto another and hope for the best that they make sense.

True, but if this is happening in a consistent forwards simulation, then it is guarenteed to make sense.

It might make sense to leave `deduplicate_sites` only at the C level, without a python interface.

jeromekelleher:
> What is the bug? But I guess I will see when you file the bug.

There's something screwy with site metadata and ancestral states. I'll leave some test cases which provoke it.

> True, but if this is happening in a consistent forwards simulation, then it is guarenteed to make sense.

Yes that's true. Fair enough.

jeromekelleher:
OK, this should do it. I've updated the low-level code to generate infinite sites mutations, conditioned on an existing set of positions. The ``keep`` parameter is now propagated through.

I've left a few skipped tests for the issue with ``deduplicate_sites``, and created an issue for it (#511).

@petrelharp would you mind taking one last look at this before I merge?

lib/mutgen.c
```diff
@@ -164,13 +279,23 @@ mutgen_populate_tables(mutgen_t *self, site_table_t *sites, mutation_table_t *mu
         }
         for (j = 0; j < site->mutations_length; j++) {
             mutation = site->mutations + j;
+            parent = mutation->parent;
+            if (parent != MSP_NULL_MUTATION) {
+                parent += new_mutations;
```
petrelharp: very nice

msprime/mutations.py
```diff
@@ -80,9 +80,9 @@ def mutate(tree_sequence, rate=None, random_seed=None, model=None, keep=False):
 
     By default, sites and mutations in the parameter tree sequence are discarded.
     If the ``keep`` parameter is true, however, *additional* mutations are
-    simulated. To ensure that the returned tree sequence is valid, and
-    mutations generated by this function that have precisely the same position
-    are discarded.
+    simulated. Under the infinite sites mutation model, any sites that are
+    generated with precisely the same position as an existing position will be
+    rejection sampled.
```
petrelharp: How about "Under the infinite sites mutation model, all new mutations generated will occur at distinct positions from each other and from any existing mutations (by rejection sampling)."

jeromekelleher:
> Looks good. Nice implementation, should make future mutation models quite easy.

Thanks! It was due a spring clean, I had been hacking it just enough to work for a long time. Hopefully this clean implementation will inspire someone to make lots of useful mutation models.

PR #510, opened 2018-06-08
Added root argument to Tree.newick().

### LINKED ISSUE #487: newick() output should (optionally) allow root to be specified, if multiple roots exist
In the case of a ts.tree with multiple roots, we should probably output multiple trees, each terminated by a semicolon.

jeromekelleher:
Closes #487. Removes redundant time_scale argument.

Would you mind having a quick look over this please @hyanwong? 

In particular, do you think it's correct that a newick tree with a single node should be rendered as "1;"?

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -6748,17 +6748,16 @@ static PyObject *
 SparseTree_get_newick(SparseTree *self, PyObject *args, PyObject *kwds)
 {
     PyObject *ret = NULL;
-    static char *kwlist[] = {"precision", "time_scale", NULL};
+    static char *kwlist[] = {"root", "precision", NULL};
     int precision = 14;
-    double time_scale = 1.0;
-    int err;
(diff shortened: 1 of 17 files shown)


hyanwong:
I think there is no other way to represent a tree with one tip and no internal nodes. The only other possibility might be `(1);`, but if these are all rooted trees, I think that is a tree with a tip and one internal node between the tip and the root.

jeromekelleher:
Also adds support for labels in newick output. @hyanwong, would you mind taking a quick look at the changes in ``msprime/trees.py`` in the diff and seeing if you agree with the interface and implementation? The idea is for it to be more-or-less the same as for tree.draw. I wimped out and implemented it in Python, as I figured that performance really doesn't matter here. This can be improved if needed.

Adding newick escapes should be an easy enough thing to do if needed (please feel free to open an issue / PR to do this).

PR #512, opened 2018-06-11
New interface for load_tables

### LINKED ISSUE #499: load_tables should accept TableCollection as parameter
I suggest changing ``load_tables`` as follows:

```python
def load_tables(tables=None, nodes=None, edges=None, ..., sequence_length):
    # stuff
```
The default way to use ``load_tables`` will be 
```python
ts = msprime.load_tables(tables)
```
which seems like what we'd want. We can also support existing code by looking at the keyword arguments and accepting **either** a TableCollection in ``tables``, or a bunch of other tables in the old keywords. This will be **mostly** backwards compatible with old code. Where it will break old code is if people we using the pattern

```python
ts = msprime.load_tables(nodes, edges)
```
rather than
```python
ts = msprime.load_tables(nodes=nodes, edges=edges)
```
It's unfortunate, but I don't see any other way of doing it without making the API clumsier in the long term. Since it's a pretty recent API with a fairly niche usage, I think it's OK.

@petrelharp, @molpopgen, what do you think?

jeromekelleher:
Closes #499.

This PR depends on #507; it'll get much smaller once that is merged.

Rather than complicate the interface for ``load_tables`` to make it take a ``TableCollection`` as a parameter, I decided to add a ``TableCollection.tree_sequence()`` method instead. This returns a new tree sequence instance based on the tables, and does the same thing as ``load_tables``, essentially (but much more simply). I suggest we undocument ``load_tables`` then and keep it around as a deprecated interface.

So, the forward simulation now goes:

```python
tables = msprime.TableCollection()
tables.nodes.add_row(...)
tables.edges.add_row(...)
# Etc, etc. filling in some stuff in your tables
tables.sort()
tables.simplify(samples)
tree_sequence = tables.tree_sequence()
```

What do you think @petrelharp, @molpopgen?

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -5495,143 +5495,24 @@ TreeSequence_load_tables(TreeSequence *self, PyObject *args, PyObject *kwds)
 {
     int err;
     PyObject *ret = NULL;
-    NodeTable *py_nodes = NULL;
-    EdgeTable *py_edges = NULL;
-    MigrationTable *py_migrations = NULL;
-    SiteTable *py_sites = NULL;
-    MutationTable *py_mutations = NULL;
-    ProvenanceTable *py_provenances = NULL;
-    IndividualTable *py_individuals = NULL;
-    PopulationTable *py_populations = NULL;
-    table_collection_t tables;
-    double sequence_length = 0.0;
-    static char *kwlist[] = {"nodes", "edges", "migrations",
-        "sites", "mutations", "provenances", "individuals",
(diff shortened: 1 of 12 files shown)


molpopgen:
This sounds reasonable to me.

jeromekelleher:
Aha, just realised that this is PR number 512. An auspicious number.

petrelharp:
Great idea, yes.  Much more intuitive.

jeromekelleher:
OK, merging. Will document in later in one pass.

PR #513, opened 2018-06-12
Fix site dedup bug

### LINKED ISSUE #511: Deduplicate sites not working with multichar strings
To reproduce:
```
$ python3 -Wignore -m nose -vs tests/test_tables.py:TestDeduplicateSites.test_multichar_ancestral_states
```
gives
```
 id  position        ancestral_state  metadata
0       0.27492011
1       0.27492011      0
2       0.30636063      A
3       0.30636063      0
4       0.40236064      AA
5       0.40236064      0
6       0.76582637      AAA
7       0.76582637      0
id      position        ancestral_state metadata
0       0.27492011
1       0.30636063      A
2       0.40236064      AA
3       0.76582637
```
These are the before and afters. 0.76582637  should have  "AAA" but it has the empty string instead.

cc @petrelharp

jeromekelleher:
Closes #511.

Depends on #510.

cc @petrelharp. The existing algorithm was missing the last offsets, so wasn't getting the sizes of metadata and ancestral state correctly. More seriously though, the algorithm required memcpys on overlapping memory areas within these buffers. Rather than use memmove (which might have had other subtle consequences) I decided to use a simpler algorithm based on taking a copy of the site table rather than doing it in-place.

### DIFF
--- a/lib/tables.c
+++ b/lib/tables.c
@@ -5061,21 +5061,24 @@ int WARN_UNUSED
 table_collection_deduplicate_sites(table_collection_t *self, int MSP_UNUSED(flags))
 {
     int ret = 0;
-    table_size_t j, site_j;
-    table_size_t as_length, as_offset;
-    table_size_t md_length, md_offset;
-    table_size_t num_input_sites;
+    table_size_t j;
     double last_position, position;
     site_id_t mutation_site;
     /* Map of old site IDs to new site IDs. */
     site_id_t *site_id_map = NULL;
+    site_table_t copy;
 
-    num_input_sites = self->sites->num_rows;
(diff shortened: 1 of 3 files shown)


msprime/cli.py
```diff
@@ -138,6 +135,7 @@ def __init__(
         if population_configurations is not None:
             sample_size = None
         self._simulator = msprime.simulator_factory(
+            Ne=0.25,
```
petrelharp: I think this warrants an explanatory note, perhaps.

petrelharp:
>  the algorithm required memcpys on overlapping memory areas within these buffers

Darn it, you're right.

Looks good. Thanks a ton for fixing this.

PR #519, opened 2018-06-13
Rough outline of Python simplify individuals.

### LINKED ISSUE #501: Update simplify to remove unreferenced individuals
Simplify should have an option of removing individuals that are not referenced by any nodes. I think this should be the default behaviour, and keeping these individuals should require a boolean flag be provided. This should be analogous to the existing ``filter_zero_mutation_sites``. Call it ``filter_zero_node_individuals``?

What do you think @petrelharp

jeromekelleher:
Closes #501

I it's not clear to me how to do this and **guarantee** that the output nodes for a given individual are always adjacent. Under the implementation I have here we output individuals as we encounter them forwards in time, which seems like a nice property. But, there may be some way we can have multiple nodes that have the same time value referencing different individuals, and then somehow end up with these nodes in the table with non-adjacent individuals. I think that if the input has the property that all nodes for a given individual are adjacent then all output nodes for the same individual must also be adjacent, but I'm not 100% confident in that.

What do you think @petrelharp? Do you think the algorithm I've outlined here will work? If we're not totally confident that this approach will be correct I think we should take the functionality out and just raise an error if someone tries to simplify a set of tables in which there are 1 or more individuals. That way at least we're sure we're not shipping something that's incorrect and that we'll have to backtrack on afterwards.

I do want to release 0.6.0 on Friday with the functionality needed for SLiM 3, so if simplifying with individuals isn't essential (ISTR that you add the individuals at the end?) then cutting it out might be the best short term solution.

### DIFF
--- a/msprime/tables.py
+++ b/msprime/tables.py
@@ -1344,7 +1344,7 @@ class TableCollection(object):
     def __str__(self):
         s = self.__banner("Individuals")
         s += str(self.individuals) + "\n"
-        s = self.__banner("Nodes")
+        s += self.__banner("Nodes")
         s += str(self.nodes) + "\n"
         s += self.__banner("Edges")
         s += str(self.edges) + "\n"
--- a/msprime/trees.py
+++ b/msprime/trees.py
@@ -2373,7 +2373,9 @@ class TreeSequence(object):
         for record in converter:
             output.write(record)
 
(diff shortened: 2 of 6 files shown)


petrelharp:
It is not necessary for SLiM, yes. And I can't deal with this before Friday, so I think that's the thing to do.

jeromekelleher:
OK, lets do that then. Shifting #501 to the next point release and closing this PR for now.

PR #520, opened 2018-06-14
Docs updates for 0.6.0

jeromekelleher:


jeromekelleher:
Some slight changes to the definitions of ``node`` and ``individual`` here @petrelharp --- it'd be good to get your thoughts on what I've done.

docs/interchange.rst
```diff
@@ -38,16 +38,24 @@ tree sequence
...
-    unique ``time``, thought of as her birth time, which determines the height
-    of any branching points she is associated with.
+    Each branching point in each tree is associated with a particular chromosome
+    in a particular ancestor, called "nodes".  Since each node represents a
+    specific gamete it has a unique ``time``, thought of as its birth time,
+    which determines the height of any branching points it is associated with.
+
+individual
+    In certain situations we are interested in how nodes (representing individual
+    homologous chromosomes) are grouped together into polyploid individuals.
```
petrelharp: often "polyploid" connotes more than diploid, maybe just "grouped together into individuals (e.g., two nodes per diploid individual)."

docs/cli.rst
```diff
@@ -22,7 +22,7 @@ msp
 
 The ``msp`` program provides a convenient interface to the :ref:`msprime API
 <sec_api>`. It is based on subcommands that either generate or consume a
-:ref:`history file <sec_hdf5_file_format>`. The ``simulate`` subcommand runs a
+:ref:`ancestry file <sec_ancestry_file_format>`. The ``simulate`` subcommand runs a
```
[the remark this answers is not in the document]
jeromekelleher: Sure, I'll change this.

docs/interchange.rst
```diff
@@ -89,16 +97,20 @@ Sequence length
...
 :ref:`Provenance <sec_provenance_table_definition>`.
-The first two store the genealogical
-relationships that define the trees; the next two describe where mutations fall
-on those trees; the Migration table describes how lineages move across space;
+The Node and Edge tables store the genealogical
+relationships that define the trees, and the Individual table
+describes how multiple chromosomes are grouped within individuals;
+the Site and Mutation tables describe where mutations fall
+on the trees; the Migration table describes how lineages move across space;
 and the Provenance table contains information on where the data came from.
```
petrelharp: insert "Only Node and Edge tables are necessary to encode the genealogical trees; Sites and Mutations are optional but necessary to encode polymorphism (sequence) data; the remainder are optional."

docs/interchange.rst
```diff
@@ -108,16 +120,16 @@ more detail.
...
 
-
 .. _sec_node_table_definition:
 
 Node Table
 ----------
 
-A **node** defines a specific ancestor that was born at some time in
+A **node** defines single chromosome in a specific ancestor
+that was born at some time in
```
petrelharp: A node might be a set of non-homologous chromosomes; but it should be the set that are inherited together. So, maybe "A **node** defines the haploid set of chromosomes of a specific individual that was born at some time in the past: the set of chromosomes inherited from a particular one of the individual's parents."

docs/interchange.rst
```diff
@@ -126,14 +138,20 @@ Column              Type                Description
...
+individual          int32               The individual the node belongs to.
 metadata            binary              Node :ref:`sec_metadata_definition`
 ================    ==============      ===========
 
 The ``time`` column records the birth time of the individual in question,
 and is a floating point value. Similarly,
 the ``population`` column records the ID of the population where this
 individual was born. If not provided, ``population`` defaults to the
-null ID (-1).
+null ID (-1). The population ID must refer to a row in the
```
petrelharp: Should say "Otherwise, the population ID must..."

docs/interchange.rst
```diff
@@ -126,14 +138,20 @@ Column              Type                Description
 flags               uint32              Bitwise flags.
 time                double              Birth time of node
 population          int32               Birth population of node.
+individual          int32               The individual the node belongs to.
```
petrelharp: perhaps these tables should have another column labeled "Optional?" for those that can be omitted.

docs/interchange.rst
```diff
@@ -162,6 +180,47 @@ more details on how metadata columns should be used.
...
 
+.. _sec_individual_table_definition:
+
+
+Individual Table
+----------------
+
+An **individual** defines how nodes (which can be seen
+as representing single chromosomes) group together in a polyploid individual.
+The individual table contains three columns of which ``flags`` is mandatory.
```
petrelharp: comma before "of which"

docs/interchange.rst
```diff
@@ -371,32 +440,50 @@ Valid tree sequence requirements
...
 Node requirements
 -----------------
 
-Nodes are the most basic type in a tree sequence, and are not defined with
-respect to any other tables. Therefore, the requirements for nodes are
-trivial.
+Given a valid set of individuals and populations, the requirements for
+each node are:
+
+- Node times must be non-negative;
```
petrelharp: I don't think tihs is true.

docs/interchange.rst
```diff
@@ -371,32 +440,50 @@ Valid tree sequence requirements
...
 
-- Node times must be non-negative.
+For simplicity and algorithmic efficiency, all nodes referring to the same
+(non-null) individual must be contiguous.
 
-There are no requirements regarding the ordering of nodes with respect to time
-or any other field. Sorting a set of tables using :func:`.sort_tables` has
-no effect on the nodes.
+Sorting a set of tables using :meth:`.TableCollection.sort`
+has no effect on nodes.
```
petrelharp: I forget: what about adjacency of nodes in the same individual?  That's up to the user to guarentee? (which is ok for now...)

petrelharp:
Wow, this is great. See comments.

docs/cli.rst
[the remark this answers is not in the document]
petrelharp: "ancestry file" could mean a lot of other things in other contexts, so I think it's good to keep the name of the format tied to "tree sequence".

docs/interchange.rst
jeromekelleher: Yes, that's up to the user for now. I think we should update TableCollection.sort() to sort nodes too so we can guarantee this, but for now it's fine.

(To address the concerns about extra time spent sorting nodes pointlessly, we can add a flag SKIP_NODES to sort for those who know their node tables are already OK.)

docs/cli.rst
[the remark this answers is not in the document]
jeromekelleher: Done.

docs/interchange.rst
jeromekelleher: Too much work to do properly now; but feel free to open an issue to track it if you think it'd be good to do.

PR #540, opened 2018-06-21
Change to use kastore's dynamic C API.

jeromekelleher:
Avoids the need to embed the kastore C code in the msprime distribution by using the kastore dynamic C API. This loads the kastore code at runtime from the shared library distributed with the ``kastore`` module.

This depends on https://github.com/tskit-dev/kastore/pull/45 being merged and released. We need to get some code in to rigorously check API version matches, but other than that it should be good. 

We can remove the git submodule also, and instead advise developers to put their own clone in the lib directory.


jeromekelleher:
This won't be merged as we'll split off tskit first. Leaving open for now though as the methods will be the same in tskit when pulling in the kastore API.

jeromekelleher:
Various debugging here for mysterious issue on windows python 3.7. For some reason it suddenly can't find the '_msprime' module I think.

PR #541, opened 2018-06-21
Simulate from

jeromekelleher:
WIP


petrelharp:
The amount of replumbing is a bit daunting.  Thanks!

jeromekelleher:
Pinging @DomNelson. 

This PR is all about starting a backwards-in-time simulation from an arbitrary tree sequence, and is mostly working now. In the process of doing this, I realise that it'd be quite a bit simpler now if we got rid of the ``ModelChangedEvent``s and instead implemented the hybrid coalescent-DTWF simulations like this (this should work fine now in this branch):
```
initial_ts = msprime.simulate(100, Ne=1000, max_time=100, model="dtwf")
full_ts = msprime.simulate(from_ts=initial_ts, Ne=1000)
```
I guess there is a bit more overhead here in creating an intermediate tree sequence and so on, but it's probably not worth worrying about. Arguably it's more useful to have the two tree sequences lying around anyway, in case you want to do some analysis on the separate bits.

What do you think? If you agree, I'll chop out the ModelChange stuff which should simplify things a bit.

Also, it'd be great if you could take a skim through the C code to see if you can spot anything!

jeromekelleher:
@petrelharp and @bhaller: This is basically working now. I have a lot of work still to do on testing with demographic events, ancient samples and so on, but for simple single-population simulations this should work fine:
```python
slim_ts = msprime.load("slim.trees")
full_ts = msprime.simulate(from_ts=slim_ts, Ne=1000)
```

It's not ready for use yet or merging as there's a whole bunch of stuff I haven't tested like mutations and replication, and I haven't done any statistical tests yet either. 

Any thoughts on the API? I would have preferred to make the argument ``from``, but this is a keyword in Python, so I kludged it to ``from_ts``...

jeromekelleher:
TODO:

- [ ] Tests with mutations (throwing mutations over the entire tree sequence, but keeping any existing mutations seems like the right behaviour for the ``mutation_rate`` argument.
- [ ] Tests for replication. Generate many ancient histories for a given recent past.
- [ ] Specifying the correct number of populations
- [ ] Demographic events: do we correctly skip events that happen before the first node in ``from_ts`` and correctly apply those that are older?
- [ ] Do we deal with historical samples correctly? (Should the semantics of the initial time be changed? Currently we take the oldest node time as the start, but it should probably be the oldest non-sample time or zero. We would then have to deal with ancient sampling events, but could just disallow these initially.)
- [ ] Statistical tests
- [ ] Documentation.

bhaller:
> Any thoughts on the API? I would have preferred to make the argument from, but this is a keyword in Python, so I kludged it to from_ts...

Seems fine.  Perhaps there's some other term, rather than "from", that would not be reserved and would feel a bit nicer?  "begin", "initial", "basis", "start", etc...?

As far as usage with SLiM is concerned: would it work just as well to call `pyslim.load` instead of `msprime.load`?  And would the SLiM-specific metadata, etc., then be preserved?  That's where we want to end up, I think.

How does it tell which genomes, exactly, are to be simulated backward from?  All genomes with no ancestor?  Or the genomes of all defined individuals, tracing backward through the population's history?  Or what?  I'm thinking here of the way that SLiM lets you add genomes to the sample permanently, as "remembered genomes", which – by the time a `.trees` file is written out – may no longer represent any extant individual.  I'm not sure whether one would want such "remembered genomes" to be included in a recapitation operation or not.  @petrelharp what do you think?  Possibly there should be an (optional) way to specify exactly which genomes should be recapitated?  Or do you think the standard behavior can always be what the user wants?

Great to hear that it's coming together.  Thanks for all the work.  I'll be dropping off the radar for a while, starting in two days; I'm heading off to a place in remote Colorado with no internet.  After that break I'll be back in Eugene, and I imagine I'll then get up to speed with this stuff and put it to use!  :->

DomNelson:
Looks like a good strategy to me. I agree the overhead is manageable, especially since hybrid sims will tend to be pretty large-scale. I've skimmed through the changes in `msprime.c` and all looks good as far as I can tell - this is a lot of work! I'll let you know if I spot anything odd as I go through the rest

jeromekelleher:
Thanks @DomNelson, good to know. Most of the change is an overdue refactoring to use updated APIs; it seemed as well to do this as I was rummaging around in there anyway.

petrelharp:
Great! I've not got the big picture of how the simulator gets set up, but:

> Currently we take the oldest node time as the start, but it should probably be the oldest non-sample time or zero. We would then have to deal with ancient sampling events, but could just disallow these initially.)

I think we need to have the time to simulate above as a parameter that's passed in (which is not currently a parameter, right?). Otherwise: suppose you forwards simulate for 10 generations, and all of the roots happen to be at 5 generations; then when you run the coalescent simulation to fill in prior history you might get a coalescence at 7 generations before present, which is inconsistent with the forwards simulation.  And, i'd vote for making this a required parameter, as guessing the right thing seems complicated (maybe impossible in general).

petrelharp:
>  I'm not sure how this would interact with the "remembered genomes" stuff in SLiM though.

The remembered genomes are just samples at some previous time, nothing special, so there's nothing to worry about there.

petrelharp:
> Throwing mutations over the entire tree sequence, but keeping any existing mutations seems like the right behaviour for the mutation_rate argument.

I guess this is the natural use case for fowards-simulating with only selected mutations. We should definately document this well, though, as it'd be easy for people to end up with twice as many mutations as they want on the lower bit of the trees.

jeromekelleher:
> I think we need to have the time to simulate above as a parameter that's passed in (which is not currently a parameter, right?). Otherwise: suppose you forwards simulate for 10 generations, and all of the roots happen to be at 5 generations; then when you run the coalescent simulation to fill in prior history you might get a coalescence at 7 generations before present, which is inconsistent with the forwards simulation. And, i'd vote for making this a required parameter, as guessing the right thing seems complicated (maybe impossible in general).

This is a subtle point @petrelharp...  Thinking it through I think you're right. We need to specify a time from which the coalescent model starts... For forward sims in general then this would be the number of generations that you've simulated, right?

This is fine, and can be used in the standard case as well (for what, I don't know, but it easier to just make it a full parameter). I'll add it in.

I'm also adding an ``end_time`` argument here (currently called ``max_time``), mainly to make it easy to generate tree sequences that need to be completed. I'm planning on making the semantics of this parameter "simulate until the first event happens whose time would exceed ``end_time``, and terminate without actually applying this event". This would guarantee that the time of any node is strictly less than ``end_time``. This seems more useful than the alternative, which would mean that any number of nodes could have times greater than ``end_time``. It also seems nice to keep the general "half closed" pattern going so that ``start_time`` is inclusive and ``end_time`` is exclusive. What do you think?

jeromekelleher:
> Hmm. But typically you want to reach coalescence, not just go backwards to whatever generation 0 happens to be, right? Would it continue into "negative time" until coalescence was reached? If not, how would one achieve the result that one typically wants to achieve – coalescence, guaranteed

jeromekelleher:
Agh! Slip of the "close" button here. I was responding to @bhaller here:

> Hmm. But typically you want to reach coalescence, not just go backwards to whatever generation 0 happens to be, right? Would it continue into "negative time" until coalescence was reached? If not, how would one achieve the result that one typically wants to achieve – coalescence, guaranteed

This is at the other end of time @bhaller, specifying what time the coalescent model starts (in backward time). We'll always simulate to full coalescence by default.

petrelharp:
> We need to specify a time from which the coalescent model starts... For forward sims in general then this would be the number of generations that you've simulated, right?

Yes, as long as the tree sequence has had times shifted so that time in the tree sequence is measured in units before the end of the simulation. (Recall that it's easiest to measure time in units before the *start* of the simulation; in which case we'd want to start at `0.0`.)

I think your proposal for `end_time` is the right one.

It'd be nice to think of something to call these things that doesn't connote a directionality to time. Like instead of `end_time`, how about `oldest_time`?  `longest_ago_time`? Or maybe I even like `max_time` better.  The corresponding time-ago from which to run coalescence could be called `youngest_time` or `most_recent_time_ago`? Or `start_time_ago` and `end_time_ago`?  Hmph.

petrelharp:
Just to add a bit:

> >   Hmm. But typically you want to reach coalescence, not just go backwards to whatever generation 0 happens to be, right? Would it continue into "negative time" until coalescence was reached? If not, how would one achieve the result that one typically wants to achieve – coalescence, guaranteed
>
> This is at the other end of time @bhaller, specifying what time the coalescent model starts (in backward time). We'll always simulate to full coalescence by default.

"Time" here is "time ago", i.e., units of time before the zero-time.  (I think "time ago" is better than "backwards time"?)

bhaller:
Well, this time-related stuff just makes my head spin.  :->  It sounds like you guys have a good sense of what needs to be done, and I'm clearly just confused, so I will leave it to you.  I would add just one suggestion: Peter, perhaps you could throw together a proof-of-concept of Example 4 from our TreeSeq paper, recapitating a SLiM run, and make sure that it works with what Jerome is doing?  "Works" in both the technical sense and in the semantic sense of the APIs being sensical and easy to use for that usage pattern.  If you could put that together (hopefully should be completely trivial), that would be great; I can then leverage what you have done when I get around to doing the example.

petrelharp:
Well, this should work:
```
decap = pyslim.load("decapitated.trees", slim_format=True)
recap = msprime.simulate(
            from_ts=decap,
            # start_time_ago=decap.slim_generation,
            Ne=1000)
```
but I'm getting a `MSP_ERR_INCOMPATIBLE_FROM_TS` error, possibly because the number of populations hasn't been set. (setting `length=decap.sequence_length` doesn't fix the problem, but shouldn't be necessary?)

jeromekelleher:
> but I'm getting a MSP_ERR_INCOMPATIBLE_FROM_TS error, possibly because the number of populations hasn't been set. (setting length=decap.sequence_length doesn't fix the problem, but shouldn't be necessary?)

Yes, you have to set the number of populations in the simulation to whatever is in the tree sequence. We could set this up automatically, but them we wouldn't know the migration rates so the most likely thing to happen is that the simulation will run forever. In the interest of "fail early, fail loudly", I think requiring the user to explicitly set the populations is the most user-friendly thing to do.

There should be better errors thrown though; I'll put in some checks in the Python code.

petrelharp:
> Yes, you have to set the number of populations in the simulation to whatever is in the tree sequence.

Even in the one-population case?

jeromekelleher:
> Even in the one-population case?

No, the default model should be one population. Must be something else then...

petrelharp:
Oh, right: SLiM starts with population ID 1, and so has *two* entries in the population table, but the first is not used.

petrelharp:
Something is not working out still.  Here's the MWE. First, the SLiM recipe:
```
initialize() {
    initializeTreeSeq(); 
    initializeMutationRate(0.0);
    initializeMutationType("m1", 0.5, "f", -0.1);
    initializeGenomicElementType("g1", m1, 1.0);
    initializeGenomicElement(g1, 0, 9);
    initializeRecombinationRate(1e-1);
}
1 {
    sim.addSubpop("p1", 5);
}
3 {
    sim.treeSeqOutput("decapitated.trees");
    sim.simulationFinished();
}
```
and the python:
```
import msprime
import pyslim

decap = pyslim.load("decapitated.trees", slim_format=True)

sample_pops = decap.tables.nodes.population[decap.samples()]
pops = [msprime.PopulationConfiguration(sample_size=sum(sample_pops == k))
        for k in range(decap.num_populations)]

recap = msprime.simulate(
            from_ts=decap, length=decap.sequence_length,
            #start_time_ago=decap.slim_generation,
            population_configurations=pops,
            Ne=1000)

recap.dump("recapitated.trees")
```
produces
```
ValueError: Cannot specify samples with from_ts
```
But, changing to
```
pops = [msprime.PopulationConfiguration()
        for k in range(decap.num_populations)]
```
doesn't fix it. I haven't been able to figure out what's going on here in the python/C.
doesn't fix this, though.

jeromekelleher:
OK, thanks @petrelharp. I haven't gone through the multiple populations stuff yet in tests, so that's probably broken in general.

That vestigial population 0 is going to end up being confusing though, isn't it? Any transition from 1 based indexes to 0 based is going to be confusing though, it's just a choice of where the confusion happens.

petrelharp:
Ah, finally found it. The code [here](https://github.com/mcveanlab/msprime/blob/9cf33bd46b4683678fa0c78f1256a4940e64d59d/msprime/simulations.py#L140) is setting `samples` to be `[]`; putting a `if from_ts is None:` before that line would avoid this.

petrelharp:
> That vestigial population 0 is going to end up being confusing though, isn't it? Any transition from 1 based indexes to 0 based is going to be confusing though, it's just a choice of where the confusion happens.

True; maybe we should switch this over at load time.  But population `0` is legal in SLiM, just not default.  We should provide a `pyslim.recapitate` function anyhow, to make these things easy: migration rates and everything can be pulled out of the metadata.

petrelharp:
After that fix, I'm getting an error related to the recombination map when the recomb rate is 0.  Somehow, `num_loci` is set to -1, resulting in negative positions returned by `recomb_map_phys_to_genetic`; strangely, printing `num_loci` [here](https://github.com/mcveanlab/msprime/blob/9cf33bd46b4683678fa0c78f1256a4940e64d59d/lib/msprime.c#L359) gives me `-1`, while printing it [here](https://github.com/mcveanlab/msprime/blob/9cf33bd46b4683678fa0c78f1256a4940e64d59d/msprime/simulations.py#L169) gives me `4294967295`.

It looks like most places, `num_loci` is `uint32_t`, except [here](https://github.com/mcveanlab/msprime/blob/simulate_from/lib/msprime.c#L2615) and [here](https://github.com/mcveanlab/msprime/blob/simulate_from/lib/recomb_map.c#L192).

jeromekelleher:
Thanks for looking into this @petrelharp, looks like there's work to be done. I'll try and finish it up this weekend.

jeromekelleher:
> It'd be nice to think of something to call these things that doesn't connote a directionality to time. Like instead of end_time, how about oldest_time? longest_ago_time? Or maybe I even like max_time better. The corresponding time-ago from which to run coalescence could be called youngest_time or most_recent_time_ago? Or start_time_ago and end_time_ago? Hmph.

I don't like using ``start_time_ago`` @petrelharp as it's inconsistent with all the other documentation and APIs. I agree it would be less confusing if everything was called ``time_ago`` but looking through the code and docs, I'm afraid that boat has sailed (without a huge effort, at least). I'd rather everything was at least consistently confusing rather than confusingly inconsistent!

I'm going to go with ``start_time`` as this would have a nice symmetry with ``end_time``.

jeromekelleher:
Good progress here @petrelharp. There was a lot of corner cases to worry about with ``start_time`` with demographic events, ancient samples and so on, but I think I've nailed them all down now. There seems to be two remaining problems:

1. The issue you pointed out with recombination maps when we have a single locus. I'll look at that next and see if I can sort it out.
2. There's an issue with SLiM tree sequences that I don't understand. If you try the simplest possible SLiM examples (see the end of ``tests/test_simulate_from.py``, and the example I've added to ``tests/data/SLiM``), we end up with a 'contradictory children over interval' error when trying to iterate over the trees. This is possibly due to a loss in precision when mapping coordinates back into genetic space, but I'm not sure. I've only just seen it, so not dug in to see what's happening.

Other than statistical tests, documentation and these issues, I think it's basically done though.

jeromekelleher:
FYI @petrelharp, I've figured out the root cause of the weirdness here. It's all down to the recombination map: we need a method to map continuous physical coordinates to discrete genetic loci. The whole thing is pretty clunky, so I'm going to have a bit of a cleanup while I'm having to rummage around in there. It should be pretty simple, but there's a few annoying corner cases to be dealt with.

Anyway, I did successfully complete a SLiM tree sequence by tacking on an msprime simulation, so it all works apart from this coordinate mapping issue. Yay for interoperability!

petrelharp:
Yay! Let me know if I can help with tests or with sorting out the right way to map coordinates.

petrelharp:
Thinking out loud: If there's a region with zero recomb rate, the phys->genetic map is not one-to-one. This makes the inverse map not well-defined; to implement it, you have to pick something from the inverse image. This means that the genetic coordinate of a zero recomb rate region can be mapped to anywhere in that region and we'll still be able to round trip genetic->phys->genetic.  Clearly we want something consistent, though. One could define what should happen for zero recomb regions by taking the limit of maps with all positive recombination; however, the result depends on how you take that limit, so this doesn't help either. The discretization occurs in genetic coordinates, so if a discretization point falls exactly on a region of zero recomb, there's no guidance for where map it to in physical coordinates.

But, what's the left or right coordinate of an edge mean? Since intervals are closed on the left and open on the right, that means the crossovers occurred epsilon *before* the given coordinates. So, if a edge boundary falls in genetic space right on a zero recomb region, then that means that the whole zero recomb region should be *after* that crossover, which happens if the genetic position gets mapped to the *left* endpoint of the region.

But but, what if the zero recombination region occurs at the *end* of the chromosome? Then, in genetic coordinates the whole region corresponds to the very endpoint, which since intervals are right-open, is never actually included in any edges. This seems wrong, and suggests mapping the genetic endpoint to the *right* end of that final zero recombination region; and thus the entire region to its right endpoint under the phys -> genetic -> phys map. Is this too messy? Let's see.

But but but, what about when the *whole* chromosome is zero recomb? This case is wierd, because you've got more than one discretization point, but the genetic length is zero. In this case, all physical edges should be [0, sequence_length), which should clearly map to [0, num_loci) under the phys->genetic. 

Both of the wierdnesses above would be fixed by mapping (a) a genetic position corresponding to zero recombination regions to the left endpoint (b) UNLESS that genetic position is the largest genetic position (num_loci), in which case it gets mapped to sequence_length. This preserves coverage of edges at the end of a chromosome with zero recomb at the end; the remaining wierdness is that for a zero recomb chromosome, the phys->genetic->phys map takes sequence_length to 0; in all other cases the phys -> genetic -> phys map is the identity for every point in the image of the genetic->phys map. 

I think this is exactly what you do already. Going to check now.

ps. I think a suggested workaround for this wierdness should be to not have zero recombination regions; just regions with very small recombination rates.

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def _physical_to_genetic_zero_rate(self, x):
+        """
+        If we have a zero recombination rate throughout, then we only have
+        two possible values. Any value < L maps to 0, as this is the start
+        of the interval. If x = L, then we map to num_loci because the interval
+        is half-open. This is consistent with the behaviour we get when we
+        have an interval of zero recombination within the map: any value
+        within the interval is mapped down to the start coordinate, and the
+        endpoint signifies the beginning of the next interval.
```
petrelharp: I think this is the right behavior but I don't understand what is meant by "the endpoint signifies the beginning of the next interval" - what interval? I think that an entire physical region of zero recombination, within the map, gets mapped to a single genetic point, including both endpoints.

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        ret = 0
+        if x >= self._sequence_length:
+            ret = self._num_loci
+        return ret
+
+    def _genetic_to_physical_zero_rate(self, v):
+        """
+        If we have a zero recombination rate throughout then 0 maps to the
+        start of the interval and everything else maps to L.
```
petrelharp: I think this should map everything but `num_loci` to zero.

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        return 0 if v == 0 else self._sequence_length
+
+    def physical_to_genetic(self, x):
+        if self.get_total_recombination_rate() == 0:
+            # TODO this check at zero shouldn't be necessary as should be
+            # able to use the same code to deal with an interval within
+            # the map.
```
petrelharp: I don't know about that - the totally zero case is really a separate case.

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+            phys_x = self._positions[j]
+            rate = self._rates[j - 1]
+            s += (phys_x - last_phys_x) * rate
+            j += 1
+            last_phys_x = phys_x
+        if rate == 0:
+            # If the rate for the current interval is zero, map back to
+            # first position unless we're at the end of the interval
+            assert False, "this hasn't been tested"
+            y = last_phys_x if s < u else self._positions[j]
```
petrelharp: This isn't the first position. I think we'd want `y = self._positions[j-1]` if we knew there was only one region of zero recombination; but in the presence of more than one consecutive one this won't be right.

petrelharp: also note that as written, `last_phys_x` and `self._positions[j]` are the same.

petrelharp: I think this is (tricky but) right, except it should be only
```
y = last_phys_x
```
i.e., without the subsequent `if s < u` stuff.

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+    @unittest.skip("Mappings for genetic to physical not clear")
+    def test_zero_rate_start(self):
+        positions = [0, 50, 100]
+        rates = [0, 1, 0]
+        num_loci = 50
+        maps = [
+            # msprime.RecombinationMap(positions, rates, num_loci),
+            PythonRecombinationMap(positions, rates, num_loci)]
+        for rm in maps:
+            # Anything <= 50 maps to 0
```
petrelharp: yes

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+            for x in [0, 10, 49, 50]:
+                self.assertEqual(0, rm.physical_to_genetic(x))
+            self.assertEqual(0, rm.genetic_to_physical(0))
+            # values > 50 should map to x - 50
```
petrelharp: yes again, agree

tests/test_recombination_map.py
```diff
@@ -0,0 +1,437 @@
...
+            for x in [50, 51, 55, 99, 100]:
+                genetic_x = x - 50
+                self.assertEqual(genetic_x, rm.physical_to_genetic(x))
+                self.assertEqual(rm.genetic_to_physical(genetic_x), x)
+
+    @unittest.skip("Uncertain about what happens at the end of an interval")
```
petrelharp: I agree with this test also

petrelharp:
Comments left; minor changes. Whew, that's tricky.

jeromekelleher:
> ps. I think a suggested workaround for this wierdness should be to not have zero recombination regions; just regions with very small recombination rates.

Do you mean, we suggest to users to avoid using zero recombination rates as they result in weirdness, or we don't accept zero recombination rates? Either makes sense to me. One thing to bear in mind is that most of the time people are using this will be with HapMap (or similar) genetic maps, which probably will contain zeros, and the user won't be aware of this. Seems annoying to force users to edit their recombination map files to change zero with 1e-14 (or whatever).

> Comments left; minor changes. Whew, that's tricky.

Thanks! And, phew. I'm glad it's not just me. I definitely regret letting Gil talk me into doing variable recombination rates this way...

petrelharp:
> Do you mean, we suggest to users to avoid using zero recombination rates as they result in weirdness, or we don't accept zero recombination rates?

I think this is just an issue when recaptitating and thing like that? So, how about we just add to the documentation an explanation of the recombination model that includes saying what's wierd about zero recombination regions, and link to it from the docs about recaptiation? No need to exclude them.

jeromekelleher:
> I think this is just an issue when recaptitating and thing like that? So, how about we just add to the documentation an explanation of the recombination model that includes saying what's wierd about zero recombination regions, and link to it from the docs about recaptiation? No need to exclude them.

Makes sense, will do.

jeromekelleher:
I've written some statistical tests here @petrelharp, and hit a bit of head-scratcher. The idea is that we should obtain identical tree sequences in distribution if we simulate a model the whole way through, or if we stop it some time in the middle and complete the simulation using ``simulate_from``.

However, with a reasonable amount of recombination, this isn't what I'm seeing:
```
original        mean trees =  89.668    mean nodes =  85.828    mean edges =  292.912
t =  0.5        mean trees =  80.279    mean nodes =  87.882    mean edges =  289.993
t =  1.0        mean trees =  81.806    mean nodes =  91.515    mean edges =  302.945
t =  1.5        mean trees =  82.028    mean nodes =  92.129    mean edges =  299.409
t =  5.0        mean trees =  88.08     mean nodes =  88.437    mean edges =  293.974
```
The first line is what we get for the mean number of trees, nodes and edges if we simulate all the way until coalescence. The remaining lines are what we get when we stop at different time points. 

It looks like we get too many nodes and edges, and too few trees, and I'm puzzled about how this would happen.

Do you think the basic test makes sense? There must either be some weird bug which didn't show up with unit tests, or some subtle effect regarding the initial conditions I didn't think of. Any thoughts?

jeromekelleher:
Aha! I see what the problem is. But, ouch, this complicates things.

I've been assuming that we can completely restore the state of a simulation just by looking at the edges and finding the roots of the trees. This is only partially true, unfortunately. Here is the state of a simulation which has been stopped early:

```
        [0][(1084-10000) 0] 
        [0][(2033-10000) 2] 
        [0][(0-6840) 12] [(6840-10000) 1] 
        [0][(6840-10000) 11] 
        [0][(0-3066) 13] [(3066-10000) 7] 
        [0][(5804-10000) 4] 
        [0][(0-5804) 4] 
        [0][(0-10000) 3] 
        [0][(0-2033) 2] 
        [0][(0-1084) 0] [(3066-10000) 5] 
```
Each line here is a list of (left, right, node) segments representing a single ancestor. Here's what we reconstruct:
```
        [0][(6840-10000) 1] 
        [0][(6840-10000) 11] 
        [0][(3066-10000) 5] 
        [0][(3066-10000) 7] 
        [0][(0-3066) 13] 
        [0][(0-10000) 4] 
        [0][(0-10000) 3] 
        [0][(0-10000) 2] 
        [0][(0-6840) 12] 
        [0][(0-10000) 0] 
```

So, we get all of the interval->node mappings correct, but we lose all the information about which segments go together within individuals. This is rather a poor approximation I think, which is why we're seeing these statistical departures from the coalescent.

Hmm. As this PR contains a lot of refactoring which will be useful in any case, I'm going to merge this much, and move discussion of how we fix the problem to #572.

PR #543, opened 2018-06-25
Do something to allow binary derived/ancestral states.

petrelharp:
See discussion in #544.


petrelharp:
Never mind: see #544.

Issue #498, opened 2018-06-02
Add text input for Individuals, Populations, Migrations, Provenances


jeromekelleher:
We can't round trip a table collection through text without them, and we'll definitely need at least individuals and Populations  for doing decent testing.

petrelharp:
Also: Individuals and Populations are not printed when printing a TableCollection.

jeromekelleher:
Closed in #555. Migrations are handled in #389.

gtsambos:
I'm currently trying to use this functionality to set up some tests, but am getting some errors. Do you know what might be going wrong for me here?

https://gist.github.com/gtsambos/d20f85954fd9325b94af548e7eda7051

PR #546, opened 2018-06-27
binary population metadata

petrelharp:
Other sorts of metadata are allowed to be binary, and are printed out using base64 encoding. The Population metadata was not. This would, I believe, fix it to allow binary metadata.

Pros:

-  to store information about population here, it's more convenient to store things as binary, and avoid floating point text <-> binary conversion issues.
- it is consistent with what we do in other tables.

Cons:

- you can't read it.

I say we make this change, and add this to the impetus to come up with a "metadata decoder" option somehow, that we've referred to elsewhere.


jeromekelleher:
Yep, agreed. I'm surprised this didn't show up anywhere else --- is there some way we can test this? I.e., checking that str(table) doesn't bork if we have unprintable characters stored in the metadata?

petrelharp:
Yes, that'd do it.  Related: it might get caught if `str(table_collection)` were to print individuals and populations (#498).  I'll add a test, though.

petrelharp:
This throws an error in `master`:
```
class TestPopulationTable(unittest.TestCase, CommonTestsMixin):
    columns = []
    ragged_list_columns = [
        (CharColumn("metadata"), UInt32Column("metadata_offset"))]
    equal_len_columns = [[]]
    string_colnames = []
    binary_colnames = ["metadata"]
    input_parameters = [("max_rows_increment", 1024)]
    table_class = msprime.PopulationTable

    def test_simple_example(self):
        t = msprime.PopulationTable()
        t.add_row(metadata=b"\xf0")
        t.add_row(b"1")
        str(t)
        self.assertEqual(len(t), 2)
        self.assertEqual(t[0], (b"\xf0",))
        self.assertEqual(t[0].metadata, b"\xf0")
        self.assertEqual(t[1], (b"1",))
        self.assertRaises(IndexError, t.__getitem__, -3)
```
but not with this PR.

I've added this, and a similar thing to the other Table tests.

jeromekelleher:
Merging; thanks @petrelharp.

PR #551, opened 2018-07-03
Properly randomizes single-locus inheritance. Fixes #439

### LINKED ISSUE #439: Fix statistical issues with DTWF single locus
Seems like the TMRCA is out by a factor of 2. Perhaps this is because we're basically defining N as 2N in the coalescent, but N is "1N" in the DTWF? (I really, really hate this mad 2N popgen thing!)

DomNelson:


jeromekelleher:
Looks good, thanks @DomNelson. Merging.

PR #567, opened 2018-08-09
Generalise simplify

jeromekelleher:
This PR is to add support for individuals and populations to simplify.

The first step here is to remove the restriction that nodes for individuals must be contiguous. I realised that this wouldn't be possible to guarantee with simplify, unless we change the property that the nodes in the output tree sequence are 0 to n. I think I'm basically the only person who thought this restriction was worthwhile anyway, so I think it's fine to ditch it.

The second step is to update simplify so that it can filter unreferenced individuals and populations from the output tables. I'm thinking of changing the API so that it looks like:

```python
def simplify(..., filter_sites=True, filter_individuals=True, filter_populations=True):

```

Each of these meaning, if there's no objects referring to these entities, we remove them from the tables. The ``filter_sites`` parameter would replace the current ``filter_zero_mutation_sites``. I thought about making them ``filter_unreferenced_sites``, ``filter_unreferenced_individuals``, etc, but I think ``filter_sites`` is unambiguous enough here in this context.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,31 @@
 ********************
+[0.6.1] - 2018-xx-xx
+********************
+
+**Breaking changes**:
+
+- Change in the semantics of how populations are treated by simplify. By
+  default, populations that are not refenced will now be removed from the
+  data model. This can be avoided by setting ``filter_populations=False``.
+
+- Simplify now raises an error if called on a set of tables that contain
+  one or more migrations.
+
+**New features**:
+
+- Simplify generalised to support individuals, and the ``filter_populations``,
+  ``filter_individuals`` and ``filter_sites`` parameters added to allow
+  filtering of unreferenced objects from the data model. (#567).
+
+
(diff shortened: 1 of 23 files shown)


petrelharp:
That's an excellent plan. Sounds just right.

jeromekelleher:
OK, this is pretty much done except for some more tests and documentation. I got side tracked with some refactoring, but I think it's made a big difference to the code and was well worth it. Adding ``x_table_get_row`` has made dealing with offsets and so on much easier to deal with. Also adding the centralised ``table_collection_check_integrity`` should make it much easier to write safe code. 

It'd be good to get your thoughts on ``table_collection_check_integrity`` and on the various restrictions that are now imposed on things like ``sort_tables`` when you get a chance @petrelharp, as I think these will affect you a little in SLiM. There's no hurry though.

jeromekelleher:
I'm going to merge this as I want to truck ahead with #569 and so some perf work there.

PR #571, opened 2018-08-14
Detect when simulations produce bad node times.

### LINKED ISSUE #87: Multiple bottlenecks may occur at the same time, resulting in illegal trees
We should insist that bottlenecks happen at different times.

### LINKED ISSUE #570: very small populations can cause children at the same time as parents due to floating point errors
Here's a MWE:
```
import msprime
import _msprime

n = 3

population_configurations = [
    msprime.PopulationConfiguration(sample_size=10000,
        initial_size = 10000,
        growth_rate=0)
    for k in range(n) ]


demographic_events = [
    msprime.PopulationParametersChange(
        time=0.01, initial_size=1e-10,
        population_id=2, growth_rate=0),
    msprime.MassMigration(
        time=0.02, source=1, destination=2, proportion=1.0)]

M = [[0.0, 1.0, 1.0], [1.0, 0.0, 1.0], [1.0, 1.0, 0.0]]

reps = msprime.simulate(
    population_configurations=population_configurations,
    demographic_events=demographic_events,
    migration_matrix=M,
    num_replicates=100,
    recombination_rate=0.0,
    mutation_rate=0.0)

for j, ts in enumerate(reps):
    print(" ts  ", j, ":", ts.num_mutations)
```

which produces
```
_msprime.LibraryError: time[parent] must be greater than time[child]
```

This is obviously not something that would be very common.

jeromekelleher:
Closes #87
Closes #570 

cc @petrelharp. The message now is:
```
  File "/home/jk/work/github/msprime/tests/test_demography.py", line 44, in test_multiple_bottlenecks
    random_seed=1)
  File "/home/jk/work/github/msprime/msprime/simulations.py", line 304, in simulate
    return next(_replicate_generator(sim, mutation_generator, 1, provenance_dict))
  File "/home/jk/work/github/msprime/msprime/simulations.py", line 89, in _replicate_generator
    sim.run()
  File "/home/jk/work/github/msprime/msprime/simulations.py", line 561, in run
    self.ll_sim.run()
_msprime.LibraryError: The simulation model supplied resulted in a parent node having a time value >= to its child. This can occur either as a result of multiple bottlenecks happening at the same time or because of numerical imprecision with very small population sizes.
```

Is this more helpful? It's not great if you want to catch this stuff programmatically though, as it's just raising the generic ``_msprime.LibraryError``. We could create another exception class here if you want to (e.g.) catch this error and realise that your population sizes have gotten too small in your MCMC or whatever.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -992,8 +992,10 @@ msp_store_edge(msp_t *self, double left, double right, node_id_t parent, node_id
 {
     int ret = 0;
     edge_t *edge;
+    const double *node_time = self->tables.nodes->time;
 
     assert(parent > child);
+    assert(parent < (node_id_t) self->tables.nodes->num_rows);
     if (self->num_buffered_edges == self->max_buffered_edges - 1) {
         /* Grow the array */
         self->max_buffered_edges *= 2;
@@ -1003,6 +1005,10 @@ msp_store_edge(msp_t *self, double left, double right, node_id_t parent, node_id
             goto out;
         }
         self->buffered_edges = edge;
+    }
+    if (node_time[child] >= node_time[parent]) {
+        ret = MSP_ERR_TIME_TRAVEL;
+        goto out;
     }
     edge = self->buffered_edges + self->num_buffered_edges;
     edge->left = left;
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@
 
 **Bug fixes**:
 
+- Throw a more intelligle error during simulation if a topology is produced
+  where the time of a parent is equal to the time of the child. (#570, #87).
+
 **Deprecated**:
 
 - The ``filter_zero_mutation_sites`` parameter for simplify has been deprecated
--- a/lib/simulation_tests.c
+++ b/lib/simulation_tests.c
@@ -773,6 +773,40 @@ test_demographic_events_start_time(void)
     recomb_map_free(&recomb_map);
 }
 
+static void
+test_time_travel_error(void)
+{
+    int ret;
+    uint32_t n = 100;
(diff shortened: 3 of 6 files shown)


petrelharp:
No need to be more fancy than this, I think. About the error message: perfect, except that it should be `<=`, not `>=`.

petrelharp:
Although, maybe the error message should be more generic, because one could get this from loading in tables produced not by msprime simulation. How about
```
Illegal node times: parent node time >= child node time. This can occur in a simulation if multiple bottlenecks happen at the same time, or because of numerical imprecision with very small population sizes.
```

jeromekelleher:
> No need to be more fancy than this, I think. About the error message: perfect, except that it should be <=, not >=.

Ahem, yes, well spotted!

> Although, maybe the error message should be more generic, because one could get this from loading in tables produced not by msprime simulation. How about

This is just about msprime (i.e., not tskit) though: we catch the error within the simulation and exit early if it happens. This prevents a malformed tree sequence from ever being made, which is definitely better.

lib/msprime.c
```diff
@@ -1152,6 +1153,10 @@ msp_store_edge(msp_t *self, double left, double right, node_id_t parent, node_id
         self->edges = edge;
         self->num_edge_blocks++;
     }
+    if (self->nodes[child].time >= self->nodes[parent].time) {
+        ret = MSP_ERR_TIME_TRAVEL;
```
petrelharp: :+1:

petrelharp:
> This is just about msprime (i.e., not tskit) though

Ah, right. =) I should have actually looked at the code.  :+1:

jeromekelleher:
OK, cool. I'll merge this after #541 is done (as it's easier to rebase this PR).

PR #580, opened 2018-08-17
Genotypes perf

jeromekelleher:
WIP.

### DIFF
--- a/tests/tsutil.py
+++ b/tests/tsutil.py
@@ -457,3 +457,111 @@ def algorithm_T(ts):
             right = min(right, edges[out_order[k]].right)
         yield (left, right), parent
         left = right
+
+
+
+class LinkedTree(object):
+    def __init__(self, tree_sequence):
+        self.tree_sequence = tree_sequence
+        num_nodes = tree_sequence.num_nodes
+        # Quintuply linked tree.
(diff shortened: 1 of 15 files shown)


tests/tsutil.py
```diff
@@ -457,3 +457,137 @@ def algorithm_T(ts):
...
+        # Quintuply linked tree.
+        self.parent = [-1 for _ in range(num_nodes)]
+        self.left_sib = [-1 for _ in range(num_nodes)]
+        self.right_sib = [-1 for _ in range(num_nodes)]
+        self.left_child = [-1 for _ in range(num_nodes)]
+        self.right_child = [-1 for _ in range(num_nodes)]
+        self.sample_head = [-1 for _ in range(num_nodes)]
+        self.sample_tail = [-1 for _ in range(num_nodes)]
+        # This is too long, but it's convenient for printing.
+        self.sample_next = [-1 for _ in range(num_nodes)]
```
molpopgen: Isn't this an ok length for the case where the sample list contains both currently-alive and ancestral samples?

jeromekelleher: The length of the list should be equal to the number of samples, because we're using sample indexes. It doesn't make any difference if the samples are from the present day or ancient, they're treated the same.

PR #581, opened 2018-08-19
Finish simulate from

jeromekelleher:
This has been another long journey, but I think it's almost done now. I'm pretty sure it's statistically correct now, and it's also fast. Here's a benchmark I did with 50 megabases of human-ish sequence:

```
====================
n =  1000
====================
Full sim required 13.25 sec
Initial sim required 2.19 sec
         87237 trees, mean roots =  13.927931955477607
Final sim required 11.21 sec
====================
n =  10000
====================
Full sim required 14.94 sec
Initial sim required 5.10 sec
         130141 trees, mean roots =  14.901230204163177
Final sim required 12.39 sec
====================
n =  100000
====================
Full sim required 34.59 sec
Initial sim required 20.16 sec
         178233 trees, mean roots =  14.033069072506214
Final sim required 12.31 sec
```

Basically, there's no cost to doing this.

The trick is to stick on extra edges to represent the ancient ancestor that we must recreate. This is a bit fiddly, but I think it basically works in the coalescent simulation case. 

The requirements for forward simulations are actually quite straightforward. Basically, we need to mark the nodes from the initial population (i.e., generation 0) in the simulation as samples, and always keep these around. The trick then is to *unmark* these as samples before you output the final tree sequence, keeping only the extant population as samples. Consider the following WF simulation example (these trees probably won't render properly for everyone; curse the lack of monospace fonts in the world!):
```
   1        3   
   ┃        ┃   
   15       ┃   
┏━┳┻━━┓     ┃   
┃ ┃   13    12  
┃ ┃  ┏┻┓   ┏┻┓  
6 11 7 10  8 9  

  3       1     
  ┃       ┃     
  14      15    
┏━┻━┓   ┏━┻┓    
┃   12  ┃  13   
┃  ┏┻┓  ┃ ┏┻┓   
11 8 9  6 7 10  

 1    0    3     
 ┃    ┃    ┃     
 ┃    ┃    14    
 ┃    ┃  ┏━┻━┓   
 13   ┃  ┃   12  
┏┻┓   ┃  ┃  ┏┻┓  
7 10  6  11 8 9  
```
These are three adjacent trees, where the samples are 6-11. The nodes for the initial population were marked as samples, and these are showing up here as 0, 1 and 3. Note that they are no longer samples though, as 0 does not show up as a root in the first two trees. We can then finish this simulation using simulate_from, and the first tree we get is:
```
    15       
 ┏━━┻━┓      
 ┃    9      
 ┃  ┏━╋━━┓   
 6  ┃ ┃  7   
┏┻┓ ┃ ┃ ┏┻┓  
2 3 0 5 1 4  
```
The nodes have been renumbered, but you can see where the topology is preserved.

The way I ended up doing it was by using the root segments of the trees. I spent most of today on a different algorithm in which we just look at the root segments, but there proved to be very tricky corner cases.

One nasty thing about the current algorithm is that we have to run simplify() at the end to get rid of the unary nodes (we need to do this as (a) it's hard to compare things statistically if we don't; and (b) it'll confuse users if we don't). I'm sure there is an algorithm in which we remove the unary edges from the table and the nodes corresponding to ancient individuals, but it seems quite fiddly to me.

A good thing about this approach also is that I'm sure that in 99% of the cases there's no need to specify ``start_time``. If we follow the recipe above and keep track of our generation-zero nodes, then we know exactly how long the simulation was run for.

There's still quite a few test-cases to fill out, but I think it's mostly working now.


jeromekelleher:
I think this is basically done now @petrelharp, other than documentation.  Would you mind taking a look through this please, and maybe trying it out with SLiM? It should work with any tree sequence as input, but will only produce statistically correct output if the initial population is retained, as discussed above.

I'll do the documentation ASAP if you agree with what's been done here.

(Note to self: revert the implementation in algorithms.py, as the current version doesn't match the real one.)

petrelharp:
(trying it out now; in the meantime:) I think this is a good solution; but, for clarity - did you come across a problem with the proposal to have simplify record uncoalesced root segments, or was that option ditched for expediency?

jeromekelleher:
> (trying it out now; in the meantime:) I think this is a good solution; but, for clarity - did you come across a problem with the proposal to have simplify record uncoalesced root segments, or was that option ditched for expediency?

I had a good hard look at this, and I can't see how it could be done with the current version of simplify. In an older version we were taking the trouble to maintain ancestry, so that at the end of the process we should have the root segments. But in the current version we deliberately don't bother with this which makes things much faster. Reverting back to this older version of simplify just didn't seem worth it to me (and, going by recent experience, it wouldn't have worked out that easily either!).

The current approach seems more elegant to me in any case, as it also solves the start_time problem too. Should be an easy update for SLiM and something we can document for forward-simulator developers?

petrelharp:
Ok, I've messed around with this a good bit with SLiM, and it all checks out. I've also added a `.recapitate()` method to `pyslim` that does this unmarking of the first generation. Happily, recapitating my terrible messy tree sequence that previously caused my computer to thrash now works with very little memory. =)  I'll take a better pass through the code now, but it's probably good to go.

lib/msprime.c
```diff
@@ -1763,9 +1759,365 @@ msp_insert_sample(msp_t *self, node_id_t sample, population_id_t population)
...
+    ret = sparse_tree_alloc(&t, self->from_ts, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    if (root_segments_head == NULL || root_segments_tail == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    /* Reset the tables to their correct position for replication */
+    ret = table_collection_reset_position(&self->tables, &self->from_position);
```
petrelharp: This is a bit confusing. As far as I can tell, the tables can't have been modified since this position was set over [here](https://github.com/tskit-dev/msprime/pull/581/files#diff-b1775af7fb1c1f04ebe0aa2492b2953bR2141). Is this just paranoia?  Or could the tables have been actually modified?

lib/msprime.c
```diff
@@ -1763,9 +1759,365 @@ msp_insert_sample(msp_t *self, node_id_t sample, population_id_t population)
...
+            if (seg == NULL) {
+                ret = MSP_ERR_NO_MEMORY;
+                goto out;
+            }
+            root_segments_head[root] = seg;
+            root_segments_tail[root] = seg;
+        } else {
+            tail = root_segments_tail[root];
+            if (tail->right == left) {
+                tail->right = right;
```
petrelharp: :+1:

lib/msprime.c
```diff
@@ -2530,10 +2953,92 @@ msp_run(msp_t *self, double max_time, unsigned long max_events)
...
 {
-    return table_collection_copy(&self->tables, tables);
+    int ret = 0;
+
+    ret = table_collection_copy(&self->tables, tables);
+    if (ret != 0) {
+        goto out;
+    }
+    if (!msp_is_completed(self)) {
+        ret = msp_insert_uncoalesced_edges(self, tables);
```
petrelharp: :+1:

lib/msprime.c
jeromekelleher: I think the confusion is because of the commented out code I've left lying around. But it's being reset at the start of every replicate back to the where the input tree sequence started.

jeromekelleher:
> One question: I think that if you remove this simplify then the tables from the input from_ts will remain, unchanged, up at the top of the resulting tree sequence. Is that right? If so, should we document that fact? (It could be useful, but then we couldn't change it...)

Getting rid of it now --- it's much nicer having this property.

lib/msprime.c
petrelharp: ah, re-use across replicates .I see. that's what "for replication" means. =)

jeromekelleher: Yep -- there's a lot of effort in there to make replication as efficient as possible.

jeromekelleher:
OK, the ``simplify`` is gone and the code tidied up. I think it's basically done except for documentation.

I've made a first pass at a tutorial for this @petrelharp --- would you mind taking a look over it and see what you think?

docs/examples.py
```diff
@@ -338,17 +339,98 @@ def historical_samples_example():
...
+
+def simulate_from_example():
+
+    num_loci = 2
+    wf_ts = wright_fisher(10, 5, L=num_loci, random_seed=3)
+    for tree in wf_ts.trees():
+        print("interval = ", tree.interval)
+        print(tree.draw(format="unicode"))
+
+    recomb_map = msprime.RecombinationMap.uniform_map(num_loci, 1, num_loci)
```
petrelharp: shouldn't that be
```
     recomb_map = msprime.RecombinationMap.uniform_map(num_loci, 0.01, num_loci) 
```
, since the `rate` argument is per unit of length, and the sequence length is `L=100`? 
(Or, to be more clear, it could be `rate = 1/wf_ts.sequence_length`?)

petrelharp:
That tutorial looks great.

docs/tutorial.rst
```diff
@@ -1123,48 +1123,214 @@ We can then finally obtain the tree sequence::
...
 
+Because our Wright Fisher simulation ran for only 5 generations, there has not
+been enough time for the trees to fully coalesce. Therefore, instead of having
+one root, the trees have several --- the first tree has 2 and the second 4.
 
+**TODO FINISH THIS BIT**
 
+.. code-block:: python
+
+    recomb_map = msprime.RecombinationMap.uniform_map(num_loci, 1, num_loci)
```
petrelharp: same comment as above

jeromekelleher:
I've updated this with more docs @petrelharp. I'm hoping to wrap this up tomorrow and release, so it'd be great if you could have a scan through the docs to see if you can spot any problems.

docs/tutorial.rst
```diff
@@ -1233,21 +1234,53 @@ We get::
...
+(Please see the :ref:`API <sec_api_simulate_from>` section for more details on the
+required properties of input tree sequences.)
+
+The process of completing this tree sequence using a coalescent simulation
+begins by first examining the root segments on the input trees. We get the
+following segments::
+
+    [(0, 2, 0), (0, 2, 7), (1, 2, 8), (1, 2, 4)]
+
+where each segment is a ``(left, right, node)`` tuple. As nodes 0 and 7 are present in both
```
petrelharp: how about add
```
For instance, root ``8`` has not coalesced over the interval from ``1.0`` to ``2.0``, while root ``0`` has not coalesced over the entire segment from ``0.0`` to ``2.0``.
```
```

File: docs/tutorial.rst [hunk 7]
```diff
@@ -1123,48 +1123,246 @@ We can then finally obtain the tree sequence::
...
+    ┃      ┃      ┃ ┏┻┓
+    ┃      11     ┃ ┃ ┃
+    ┃ ┏━┳━┳┻┳━┳━┓ ┃ ┃ ┃
+    2 4 9 0 3 6 8 1 5 7
+
+This final tree sequence is topologically identical to the original tree sequence,
+but has the redundant nodes and edges removed. Note also that he node IDs have been
+reassigned so that the samples are 0 to 9 --- if you need the IDs from the original
+tree sequence, please set ``map_nodes=True`` when calling ``simplify`` to get a
+mapping between the two sets of IDs.
```
petrelharp: Perhaps add here:
```
Why recording the initial generation?
=============================

We can now see why it is essential that the forwards simulator records the *initial* generation in a tree sequence
that will later be used as a ``from_ts`` argument to ``msprime.simulate( )``. In the example above, if node ``7`` was not in the tree sequence, we would not know that the segment that node ``20`` inherits from on ``[0.0, 1.0)`` and the segment that node ``12`` inherits from on ``[1.0, 2.0)`` both exist in the same node (here, node ``7``).

However, note that although the intial generation (above, nodes ``0``, ``4``, ``7``, and ``8``) must be in the tree
sequence, they do *not* have to be samples. The easiest way to do this (as is what is done in ``pyslim``) is to
(a) retain the initial generation as samples throughout the forwards simulation (so they persist through ``simplify()``), but then (b) before recapitation, removing the flags that mark them as samples, so that 
``simulate()`` does not simulate their entire history as well.
```

docs/tutorial.rst
```diff
@@ -1123,48 +1123,246 @@ We can then finally obtain the tree sequence::
...
+
+The ``msprime`` simulator generates tree sequences using the backwards in
+time coalescent model. But it is also possible to output tree sequences
+from `forwards-time <https://www.biorxiv.org/content/early/2018/01/16/248500>`_
+simulators such as `SLiM <https://messerlab.org/slim/>`_.
+There are many advantages to using forward-time simulators, but they
+are usually quite slow compared to similar coalescent simulations. In this
+section we show how to combine the best of both approaches by simulating
+the recent past using a forwards-time simulator and then complete the
+simulation of the ancient past using ``msprime``.
```
petrelharp: Add
```
We refer to this process as "recapitation".
```

docs/tutorial.rst
```diff
@@ -1123,48 +1123,246 @@ We can then finally obtain the tree sequence::
...
+            ┃         ┃  ┏┻━┓  ┃
+            21        ┃  ┃  ┃  ┃
+    ┏━━┳━━┳━┻┳━━┳━━┓  ┃  ┃  ┃  ┃
+    14 19 10 13 16 18 12 15 17 11
+
+The trees have fully coalesced and we've successfully combined a forwards-time
+Wright-Fisher simulation with a coalescent simulation: hooray! However, these
+trees have some slightly odd properties which are important to be aware of.
+In both trees we can see that the old roots are still present in the trees,
+even through they have only one child and are clearly redundant. While this
```
petrelharp: Insert here, maybe:
```
This is because the tables of ``from_ts`` have been retained, without modification, at the top of the tables of the output tree sequence. 
```

petrelharp:
I've had a good pass through. I kinda wanted to insert text about how the `from_ts` tables are retained in the docs for `simulate`, but it seemed like it'd clutter that up too much.  LGTM!

jeromekelleher:
Thanks @petrelharp! One issue though: I mildly object to the term 'recapitation', simply because it implies that the thing had a head in the first place. These are headless tree sequences that we're grafting a head on (or, possibly many different heads!). "Encapitate" or just "capitate" would be less inaccurate, but they're not very catchy. 

I'll admit that what I've gone with doesn't roll off the tongue as well as 'recaptitation'. I just think it's going to confuse people...

petrelharp:
>  I mildly object to the term 'recapitation'

No problem. it's just so much shorter than "run simulate with `from_ts` specified".

jeromekelleher:
Ahh, interesting point here:

> (a) retain the initial generation as samples throughout the forwards simulation (so they persist through ``simplify()``), but then (b) before recapitation, removing the flags that mark them as samples, so that 
``simulate()`` does not simulate their entire history as well.

I would have said the right thing to do here is to (a) retain the initial generation as samples throughout the forwards simulation; and (b) just before final output **unmark** these ancient individuals as samples. 

Why output the initial generation as samples? This seems quite confusing to me, and not what I would expect when I load a forward-time tree sequence.

I'm changing the text for now so I can merge this, but we can change it back if I'm wrong.

jeromekelleher:
Merging this so the docs go live on RTD. Happy to make changes though.

petrelharp:
> I would have said the right thing to do here is to (a) retain the initial generation as samples throughout the forwards simulation; and (b) just before final output unmark these ancient individuals as samples.

Ah, good point. Er, I figured that output tree sequences should be simplified, so they need to be marked as samples. And, I've added the unmarking to the `pyslim.SlimTreeSequence.recapitate` method.  I guess I think this is confusing, but stil maybe the right way to go?  Currently it is the user's responsibility to mark the first generation; but you're right, it probably should not be. I will discuss with @bhaller

PR #583, opened 2018-08-21
Initial version of provenance schema.

### LINKED ISSUE #556: Provenance schema
We need some sort of structure for the provenance data that we are recording. I suggest we create a simple JSON schema for this, so that we can document the format and validate it. There is a basic outline at the moment, which covers things like the software environment quite well, but is poor for capturing the actual command/function that was run. Currently we have:
```json
{
  "version": "0.6.0b1",
  "environment": {
    "os": {
      "version": "#50~16.04.1-Ubuntu SMP Wed May 30 11:18:27 UTC 2018",
      "node": "powderfinger",
      "release": "4.13.0-45-generic",
      "system": "Linux",
      "machine": "x86_64"
    },
    "python": {
      "implementation": "CPython",
      "version": [
        "3",
        "5",
        "2"
      ]
    },
    "libraries": {
      "kastore": {
        "version": [
          0,
          1,
          0
        ]
      },
      "gsl": {
        "version": [
          2,
          1
        ]
      }
    }
  },
  "command": "simulate",
  "parameters": {
    "TODO": "encode simulation parameters"
  },
  "software": "msprime"
}
```

The  ``command`` and ``parameters`` bit is weak I think, and doesn't generalise very well between CLI and API invocations.

With SLiM now outputting tree sequence files I think it's important we nail this down reasonably well so that we can document the expected structure.

Some principles:

1. The schema should document a **top-level** structure that specifies just enough to be useful. Users are entirely free to put whatever extra stuff they like into the provenance documents.  Hopefully the top-level structure will make it obvious where the stuff that they want to add should go.
2. The schema should provide structure sufficient to make it possible to fully reproduce a particular simulation/tree sequence inference. The schema is not responsible for this though, it's up to the client programs to ensure that they are providing enough information.
3. The schema structure for specifying commands and parameters used should generalise across API and CLI invocations (i.e., having a ``command_line`` attribute somewhere is no good, as this doesn't work for API invocations).

Pinging @molpopgen, @bhaller, @petrelharp, @ashander, @DomNelson, @hyanwong for input.

See also https://github.com/tskit-dev/tsinfer/issues/29

jeromekelleher:
First pass at defining the provenance schema using [JSON schema](http://json-schema.org/).   Closes #556 

The purpose of provenance is **reproducability**: given a tree sequence that has this provenance information associated with it, you should be able to reproduce it. This won't always be possible for various reasons (parameters too big to encode as JSON, referencing files that don't exist, numerical precision of parameters, ...), but this is the basic spirit of the idea.

**UPDATED to include schema_version**

Here's the current schema:
```json
{
  "schema": "http://json-schema.org/draft-07/schema#",
  "version": "1.0.0",
  "title": "tskit provenance",
  "description": "The combination of software, parameters and environment that produced a tree sequence",
  "type": "object",
  "required": ["schema_version", "software", "parameters", "environment"],
  "properties": {
    "schema_version": {
      "description": "The version of this schema used.",
      "type": "string",
      "minLength": 1
    },
    "software": {
      "description": "The primary software used to produce the tree sequence.",
      "type": "object",
      "required": ["name", "version"],
      "properties": {
        "name": {
          "description": "The name of the primary software.",
          "type": "string",
          "minLength": 1
        },
        "version": {
          "description": "The version of primary software.",
          "type": "string",
          "minLength": 1
        }
      }
    },
    "parameters": {
      "description": "The parameters used to produce the tree sequence.",
      "type": "object"
    },
    "environment": {
      "description": "The computational environment within which the primary software ran.",
      "type": "object",
      "properties": {
        "os": {
          "description": "Operating system.",
          "type": "object"
        },
        "libraries": {
          "description": "Details of libraries the primary software linked against.",
          "type": "object"
        }
      }
    }
  }
} 
```

This is what we end with from msprime:
```json
{
  "schema_version": "1.0.0",
  "environment": {
    "libraries": {
      "gsl": {
        "version": "2.1"
      },
      "kastore": {
        "version": "0.1.0"
      }
    },
    "python": {
      "version": "3.5.2",
      "implementation": "CPython"
    },
    "os": {
      "system": "Linux",
      "node": "powderfinger",
      "release": "4.15.0-29-generic",
      "version": "#31~16.04.1-Ubuntu SMP Wed Jul 18 08:54:04 UTC 2018",
      "machine": "x86_64"
    }
  },
  "parameters": {
    "sample_size": 5,
    "command": "simulate",
    "TODO": "add other simulation parameters"
  },
  "software": {
    "name": "msprime",
    "version": "0.6.1.dev123+ga252341.d20180820"
  }
}
```

This is slightly different to what we had before, but in the same spirit. The basic idea is that we have three things we need to document for provenance: the **software** used, the **parameters** provided to this software and the **environment** within which this software was run. 

The schema is pretty light-touch, as there's no point in over-specifying this stuff. The minimal compliant instance is:
```json
    {
            "schema_version": "1",
            "software": {
                "name": "x",
                "version": "x",
            },
            "environment": {},
            "parameters": {}
     }
```
Basically, you have to have these three keys, and you have to give non-empty strings for the **name** and **version** keys, and specify the **schema_version** You can put anything extra in there that you want.

**TODO**

- High-level documentation explaining what this is for.

Pinging @petrelharp, @molpopgen and @bhaller for comments!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include lib/*.h
 include lib/kastore/c/kastore.h
+include msprime/provenance.schema.json
 include README.txt
 include ez_setup.py
 include LICENSE
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -635,3 +635,21 @@ Table functions
 .. autofunction:: msprime.pack_bytes
 
 .. autofunction:: msprime.unpack_bytes
+
+
+.. _sec_provenance_api:
+
+**************
+Provenance API
+**************
+
+We provide some preliminary support for validating JSON documents against the
+:ref:`provenance schema <sec_provenance>`. Programmatic access to provenance
+information is planned for future versions.
+
+.. autofunction:: msprime.validate_provenance
+
+.. autoexception:: msprime.ProvenanceValidationError
+
+
+
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,6 +27,7 @@ Contents:
    api
    cli
    interchange
+   provenance
    development
    CITATION
 
--- a/docs/provenance.rst
+++ b/docs/provenance.rst
@@ -1,0 +1,230 @@
+.. _sec_provenance:
+
+==========
+Provenance
+==========
+
+Every tree sequence has provenance information associated with it. The purpose of this
+information is to improve `reproducibility <https://en.wikipedia.org/wiki/Reproducibility>`_:
+given the provenance associated with a given tree sequence, it should be possible to
+reproduce it. Provenance is split into three sections: the primary **software** used to
+produce a tree sequence; the **parameters** provided to this software; and the computational
+**environment** where the software was run.
+
+This documentation serves two distinct purposes:
+
+1. For developers using ``tskit`` in their own applications, it provides normative documentation
+   for how provenance information should be stored.
+2. For end-users of ``tskit``, it provides documentation to allows them to inspect and interpret
+   the provenance information stored in ``.trees`` files.
+
+Provenance information is encoded using `JSON <https://www.json.org/>`_.
+To standardise the provenance information produced by different software and improve
+interoperability we define a formal specification using `JSON Schema <http://json-schema.org/>`_.
+The full schema is provided :ref:`below <sec_provenance_schema>`, which may be used to
+automatically validate input. In the following we describe the intention of the various
+sections in more detail.
+
+This document defines specification version 1.0.0. Specification version numbers follow
+`SemVer <https://semver.org/>`_ semantics.
+
+.. _sec_provenance_example:
+
+*******
+Example
+*******
+
+To make things more concrete, let's consider an example:
+
+.. code-block:: json
+
+    {
+      "schema_version": "1.0.0",
+      "software": {
+        "name": "msprime",
+        "version": "0.6.1.dev123+ga252341.d20180820"
+      },
+      "parameters": {
+        "sample_size": 5,
+        "random_seed": 12345,
+        "command": "simulate"
+      },
+      "environment": {
+        "libraries": {
+          "gsl": {
+            "version": "2.1"
+          },
+          "kastore": {
+            "version": "0.1.0"
+          }
+        },
+        "python": {
+          "version": "3.5.2",
+          "implementation": "CPython"
+        },
+        "os": {
+          "system": "Linux",
+          "node": "powderfinger",
+          "release": "4.15.0-29-generic",
+          "version": "#31~16.04.1-Ubuntu SMP Wed Jul 18 08:54:04 UTC 2018",
+          "machine": "x86_64"
+        }
+      }
+    }
+
+This information records the provenance for a very simple msprime simulation. The record is a JSON
+object with three mandatory fields ("software", "parameters" and "environment")
+which we discuss separately in the following sections.
+
+.. _sec_provenance_software:
+
+********
+Software
+********
+
+Every tree sequence is produced by some piece of software. For example, this may be a
+coalescent simulation produced by ``msprime``, a forwards-time simulation from ``SLiM``
+or tree sequence inferred from data by ``tsinfer``. The software provenance is
+intended to capture the details about this primary software.
+
+================    ==============      ===========
+Field               Type                Description
+================    ==============      ===========
+name                string              The name of the software.
+version             string              The software version.
+================    ==============      ===========
+
+
+Note that libraries that the primary software links against are considered part of the
+:ref:`sec_provenance_environment` and should be recorded there.
+
+.. _sec_provenance_parameters:
+
+**********
+Parameters
+**********
+
+The parameters section of a provenance document records the input that was used to
+produce a particular tree sequence. There are no requirements on what may be stored
+within it, but we make some recommendations here on how to encode such information.
+
+As a general principle, sufficient information should be recorded in the parameters
+section to allow the output tree sequence to be reproduced exactly. There will be instances,
+however, where this is not possible due to missing files, issues with numerical precision
+and so on.
+
++++++++++++++++
+API invocations
++++++++++++++++
+
+Consider an API call like the following simple msprime simulation:
+
+.. code-block:: python
+
+    ts = msprime.simulate(sample_size=10, recombination_rate=2)
+
+We recommend encoding the parameters provenance as follows (other fields omitted
+for clarity):
+
+.. code-block:: json
+
+    {
+      "parameters": {
+        "command": "simulate",
+        "sample_size": 10,
+        "recombination_rate": 2,
+        "random_seed": 123456789,
+      }
+    }
+
+Specifically, we encode the name of the function using the ``command`` key and
+the function parameters in the obvious way. Note that we include the ``random_seed``
+here even though it was automatically generated.
+
+
++++++++++++++++
+CLI invocations
++++++++++++++++
+
+Consider the following invocation of a hypothetical command line program:
+
+.. code-block:: bash
+
+    $ supersim --sample-size=10 --do-some-stuff -O out.trees
+
+We recommend encoding the parameters provenance as follows (other fields omitted
+for clarity):
+
+.. code-block:: json
+
+    {
+      "parameters": {
+        "command": "supersim",
+        "args": ["--sample-size=10", "--do-some-stuff", "-O", "out.trees"],
+        "random_seed": 56789
+      }
+    }
+
+Here we encode the name of the program using the ``command`` key
+and its command line arguments as a list of strings in the ``args`` key. We
+also include the automatically generated random seed in the parameters list.
+
+If parameters that affect the output tree sequence are derived from environment
+variables these should also be recorded.
+
+.. _sec_provenance_environment:
+
+***********
+Environment
+***********
+
+The environment section captures details about the computational environment in
+which the software was executed. Two optional fields are defined: ``os``
+and ``libraries``. We recommend including any additional relevant platform
+information here; for example, if using Python store the interpreter information
+as shown in the example above.
+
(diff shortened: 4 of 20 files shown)


bhaller:
If "environment" and "parameters" were made optional, and if "software" were split out into a "software" string and a "version" string (both at the top level), it would match what SLiM already produces, which would be nice.  The point is not so much saving work on my end (changing this stuff is trivial) as allowing SLiM 3.0 files to be schema-compliant, and not changing the format between versions unnecessarily (which just causes confusion all around).  If there's a compelling reason driving those changes, then that's fine and we can just say SLiM 3.0 files are non-compliant, and so it goes; but offhand I don't see any particularly compelling reasons for those alterations, so maybe we can preserve backward compatibility?

The other thing I would note is that requiring "parameters" to be a key-value dictionary (which I take it is what is meant in JSON by "object"?) is perhaps unnecessarily restrictive and undesirable.  Right now SLiM generates an array, instead, for "parameters", composed of all of the command-line parameters that were supplied (straight from argv/argc), and that feels rather natural and simple.  Some command-line parameters take no argument (so they would have a key with no value), others might allow more than one argument (multiple values per key).  It seems like it is more flexible and natural to just let the generating program represent its parameters as it sees fit.  Is there a need to impose the key-value structure as a requirement here?

jeromekelleher:
Thanks for the feedback @bhaller, comments below.

> If "environment" and "parameters" were made optional, and if "software" were split out into a "software" string and a "version" string (both at the top level), it would match what SLiM already produces, which would be nice. The point is not so much saving work on my end (changing this stuff is trivial) as allowing SLiM 3.0 files to be schema-compliant, and not changing the format between versions unnecessarily (which just causes confusion all around). If there's a compelling reason driving those changes, then that's fine and we can just say SLiM 3.0 files are non-compliant, and so it goes; but offhand I don't see any particularly compelling reasons for those alterations, so maybe we can preserve backward compatibility?

I don't think we should worry too much about backward compatibility here to be honest. Tskit isn't going to be checking these provenance strings for compliance, and I have no plans for adding APIs to do anything with the values at the moment. Users never see these provenance values the right now, so there's nothing to be confused by there. When we do add APIs, they'll have to be reasonably robust to provenance information that doesn't conform (i.e., maybe just print out the string as-is rather than nicely formatting, say). To me this exercise is about avoid these issue **in the future** when we do start using the values. So, I would rather get this right and try and get a nice, clean and extensible schema.

Old msprime files will also have provenance information that isn't compliant, btw. 

> The other thing I would note is that requiring "parameters" to be a key-value dictionary (which I take it is what is meant in JSON by "object"?) is perhaps unnecessarily restrictive and undesirable. Right now SLiM generates an array, instead, for "parameters", composed of all of the command-line parameters that were supplied (straight from argv/argc), and that feels rather natural and simple. Some command-line parameters take no argument (so they would have a key with no value), others might allow more than one argument (multiple values per key). It seems like it is more flexible and natural to just let the generating program represent its parameters as it sees fit. Is there a need to impose the key-value structure as a requirement here?

That's a good point. I think for recording CLI parameters the right way to do it would be something like:

```json
...
   "parameters": {"command": ["-X", "1", "--something"]}
...
```
Having parameters as a mapping gives the flexibility to cover both CLI invocations as well as API calls. If we use a list, then it's quite limiting, as API calls would have to be either a list of (parameter, value) pairs or just a straight list of positional parameters.

bhaller:
OK, that all makes sense.  I'm on board with this proposal.  :->

jeromekelleher:
I've added some [documentation](https://github.com/tskit-dev/msprime/pull/583/files#diff-b375af127779c2dfcad4ae1c40691ea2) and I think this is ready to go now. I'll merge it in a couple of days if I don't hear any objections.

bhaller:
I notice you're writing out a "schema_version" entry but no "schema" entry, whereas previous versions of this that I saw from you had a "schema" tag with value "tskit-provenance".  Did this get removed deliberately, or should it be added back in?

jeromekelleher:
> I notice you're writing out a "schema_version" entry but no "schema" entry, whereas previous versions of this that I saw from you had a "schema" tag with value "tskit-provenance". Did this get removed deliberately, or should it be added back in?

I took it out because it doesn't seem to be the done-thing in JSON schema. Best practices indicate that you should include a [$id](http://json-schema.org/latest/json-schema-core.html#rfc.section.8.2) URI to uniquely identify the schema (in the definition), but there's no guidance on making the actual documents self-identifying. Since JSON schema isn't worrying about it and this is such a limited and specialised application, I figured it was overboard to include the schema details in the actual documents.

I left out the $id field in the schema definition because I'm not sure what it should be and can't be bothered setting up a stable URL where it could be hosted. This can always be added in later if we change our minds.

Adding a version number did seem prudent though, since this schema is likely to evolve over time and it'll be good to have SemVer semantics so that clients can deal with this.

I'd be good to hear your thoughts on this (if you have any!)...

bhaller:
OK, if this is standard practice then no worries.  Re: "stable URL where it could be hosted", I have a stable domain (benhaller.com) where I'm happy to host (not-huge) stuff if you ever want that.  We already host the downloads for SLiM there.

bhaller:
Here is a sample SLiM provenance table now (@petrelharp):

```
{
    "environment": {
        "os": {
            "machine": "x86_64",
            "node": "darwin-447.local",
            "release": "17.6.0",
            "system": "Darwin",
            "version": "Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64"
        }
    },
    "parameters": {
        "command": []
    },
    "schema_version": "1.0.0",
    "slim": {
        "file_version": "0.2",
        "generation": 2000,
        "model": "initialize() {\n\tinitializeTreeSeq();\n\tinitializeMutationRate(1e-7);\n\tinitializeMutationType(\"m1\", 0.5, \"f\", 0.0);\n\tinitializeGenomicElementType(\"g1\", m1, 1.0);\n\tinitializeGenomicElement(g1, 0, 99999);\n\tinitializeRecombinationRate(1e-8);\n}\n1 {\n\tsim.addSubpop(\"p1\", 500);\n}\n2000 late() { sim.treeSeqOutput(\"~/Desktop/junk.trees\"); }\n",
        "model_type": "WF",
        "remembered_node_count": 0,
        "seed": 1722162964723
    },
    "software": {
        "name": "SLiM",
        "version": "3.0"
    }
}
```

  The parameters under "command" are just an empty array because this was run in SLiMgui.  I believe it follows the schema.  SLiM-specific cruft has been put under a "slim" top-level tag.  This will require an update to pyslim, presumably.  Note that file_version was 0.1 for SLiM 3.0 (and was a top-level tag), and now it is 0.2 (and is under "slim").  The previous post-3.0 file format also called itself 0.2, but I think we should just pretend that it never existed, as it was only a thing briefly in GitHub, never in release.

jeromekelleher:
> Re: "stable URL where it could be hosted", I have a stable domain (benhaller.com) where I'm happy to host (not-huge) stuff if you ever want that. We already host the downloads for SLiM there.

Thanks for the offer @bhaller --- it's not so much the web hosting, it just feels like the long-term stable URI for a schema like this should be something associated with the project. Something like ``tskit.org/provenance.schema.json``. So, I think the simplest thing is just leave it out for now, as it's easy to add in later without breaking anything.

Having said that, tskit.org is free. Maybe I should just register it...

jeromekelleher:
The SLiM provenance looks great @bhaller. I have a minor suggestion but feel free to ignore:
```
{
    "environment": {
        "os": {
            "machine": "x86_64",
            "node": "darwin-447.local",
            "release": "17.6.0",
            "system": "Darwin",
            "version": "Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64"
        }
    },
    "parameters": {
        "command": [],
        "model": "initialize() {\n\tinitializeTreeSeq();\n\tinitializeMutationRate(1e-7);\n\tinitializeMutationType(\"m1\", 0.5, \"f\", 0.0);\n\tinitializeGenomicElementType(\"g1\", m1, 1.0);\n\tinitializeGenomicElement(g1, 0, 99999);\n\tinitializeRecombinationRate(1e-8);\n}\n1 {\n\tsim.addSubpop(\"p1\", 500);\n}\n2000 late() { sim.treeSeqOutput(\"~/Desktop/junk.trees\"); }\n",
        "model_type": "WF",        
        "seed": 1722162964723
    },
    "schema_version": "1.0.0",
    "slim": {
        "file_version": "0.2",
        "remembered_node_count": 0,
        "generation": 2000     
    },
    "software": {
        "name": "SLiM",
        "version": "3.0"
    }
}
```

The idea is that ``parameters`` should contain all information that you would need to supply to SLiM to recreate the tree sequence, so I've moved the model, model_type and seed in there. The stuff left in the ``slim`` tag then is the miscellaneous information that pyslim uses to better interpret the tree sequence.

bhaller:
I like your proposed shift, I'll do that this morning.  Thanks for the feedback.

jeromekelleher:
The documentation is up [here](https://msprime.readthedocs.io/en/latest/provenance.html) --- should explain the rationale for things.

bhaller:
Great, I'll link to that in my documentation, thanks.  @petrelharp, SLiM's file_version 0.2 provenance format is now as Jerome proposed immediately above.  Done and committed.

jeromekelleher:
> Great, I'll link to that in my documentation, thanks. @petrelharp, SLiM's file_version 0.2 provenance format is now as Jerome proposed immediately above. Done and committed.

Excellent. Probably best to link to the stable version: https://msprime.readthedocs.io/en/stable/provenance.html

Doesn't exist right now, but will later today hopefully.

PR #584, opened 2018-08-21
clarified error

petrelharp:
Computing pairwise diversity on a tree sequence with more than one mutation per site was producing the crpytic error
```
_msprime.LibraryError: Operation cannot be performed in current configuration
```

This changes that error to
```
_msprime.LibraryError: Only infinite sites mutations are supported for this operation.
```

Also in the LD code in a few places.


jeromekelleher:
LGTM, thanks @petrelharp!

PR #591, opened 2018-08-24
Document flags reserved and unreserved status.

jeromekelleher:
cc @bhaller and @petrelharp --- making it official that the top 16 bits are free for application to do whatever they want with.


jeromekelleher:
OK, looks like we're all happy with this. Merging.

PR #599, opened 2018-09-10
Fixes to DTWF demographic models

DomNelson:
Ok it's looking good! Final piece of the puzzle was the way migrations were done sequentially - lineages were migrated as they were drawn, which meant they could be drawn again and re-migrated from their new population. Also explains why it got worse with high migration rates. Not sure if my fix is the most efficient - I stored migrating nodes in new AVL trees so they could only be drawn once - but for now it works. Two pops with migration rates `[[0, 0.5], [0.7, 0]]` now looks like

num_trees:
num_trees

tmrca_mean:
tmrca_mean

Also, SLiM uses discrete loci only, so scenarios where we have lots of breakpoints show small departures. This can be fixed by increasing the number of loci until both models are essentially continuous, but I do find it odd that this can happen even when I explicitly specify `num_loci` in the recombination map. Maybe this is handled differently now?

num_trees with 1 Morgan = 1e8 loci:
num_trees

num_trees with 1 Morgan = 1e14 loci:
num_trees

Speaking of which, I'm getting this error when running DTWF checks in `verification.py` after syncing my fork:

```
~/project/msprime/msprime/simulations.py in create_ll_instance(self)
    654             segment_block_size=self.segment_block_size,
    655             avl_node_block_size=self.avl_node_block_size,
--> 656             node_mapping_block_size=self.node_mapping_block_size)
    657         return ll_sim
    658

TypeError: 'recombination_map' is an invalid keyword argument for this function
```

I'll look into how these checks are run now, but thought you might know how to fix this off the top of your head. All these new checks also take quite a while to run if I'm honest...

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2486,12 +2486,8 @@ msp_dtwf_generation(msp_t *self)
         parents = NULL;
     }
 out:
-    if (parents != NULL) {
-        free(parents);
-    }
-    if (segment_mem != NULL){
-        free(segment_mem);
-    }
+    msp_safe_free(parents);
+    msp_safe_free(segment_mem);
     return ret;
 }
 
@@ -2583,9 +2579,9 @@ msp_run_dtwf(msp_t *self, double max_time, unsigned long max_events)
             // For proper sampling, we need to calculate the proportion
             // of non-migrants as well
             sum = 0;
-            for (uint32_t z = 0; z < self->num_populations; z++) {
-                mig_tmp[z] = self->migration_matrix[j * self->num_populations + z];
-                sum += mig_tmp[z];
+            for (k = 0; k < self->num_populations; k++) {
+                mig_tmp[k] = self->migration_matrix[j * self->num_populations + k];
+                sum += mig_tmp[k];
             }
             assert(mig_tmp[j] == 0);
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2405,13 +2405,15 @@ msp_dtwf_generation(msp_t *self)
             continue;
         }
         /* For the DTWF, N for each population is the reference population size
-         * from the model multiplied by the current population size, rounded
-         * to the nearest integer. Thus, the population's size is always relative
+         * from the model multiplied by the current population size, rounded to
+         * the nearest integer. Thus, the population's size is always relative
          * to the reference model population size (which is also true for the
          * coalescent models. */
-        N = (uint32_t) round(
+        // This may not be ideal, but with low migration / high growth rates it
+        // is possible for populations to reach zero individuals before all
+        // lineages coalesce. We round up to avoid this.
+        N = (uint32_t) ceil(
             msp_get_population_size(self, pop) * self->model.population_size);
-        /* printf("%u parents to choose from %f \n", N, msp_get_population_size(self, pop)); */
         if (N == 0) {
             ret = MSP_ERR_INFINITE_WAITING_TIME;
             goto out;
@@ -2448,7 +2450,6 @@ msp_dtwf_generation(msp_t *self)
                 msp_free_avl_node(self, node);
 
                 // Recombine ancestor
-
                 if (self->recombination_rate > 0) {
                     ret = msp_dtwf_recombine(self, x, &u[0], &u[1]);
                     if (ret != 0) {
@@ -2494,6 +2495,52 @@ out:
     return ret;
 }
 
+static int WARN_UNUSED
+msp_store_simultaneous_migration_events(msp_t *self, avl_tree_t *nodes,
+       population_id_t source_pop) {
+    int ret = 0;
+    uint32_t j;
+    avl_node_t *node;
+    avl_tree_t *source;
+
+    source = &self->populations[source_pop].ancestors;
+
+    // Choose node to migrate
+    j = (uint32_t) gsl_rng_uniform_int(self->rng, avl_count(source));
+    node = avl_at(source, j);
+    assert(node != NULL);
+
+    // Move to temp storage for actual migration later
+    avl_unlink_node(source, node);
+    node = avl_insert_node(nodes, node);
+    assert(node != NULL);
+
+    return ret;
+}
+
+static int WARN_UNUSED
+msp_simultaneous_migration_event(msp_t *self, avl_tree_t *nodes,
+        population_id_t source_pop, population_id_t dest_pop) {
+    int ret = 0;
+    avl_node_t *node;
+    size_t index;
+
+    index = ((size_t) source_pop) * self->num_populations + (size_t) dest_pop;
+    self->num_migration_events[index]++;
+
+    // Iterate through nodes in tree and move to new pop
+    // -- removed from "nodes" in msp_move_individual()
+    while (avl_count(nodes) > 0) {
+        node = nodes->head;
+        ret = msp_move_individual(self, node, nodes, dest_pop);
+        if (ret != 0) {
+            goto out;
+        }
+    }
+out:
+    return ret;
+}
+
 /* The main event loop for the Wright Fisher model. */
 static int WARN_UNUSED
 msp_run_dtwf(msp_t *self, double max_time, unsigned long max_events)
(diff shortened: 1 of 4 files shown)


jeromekelleher:
The differences with numbers of loci is odd and suggests to me that there's some slight differences in the recombination model between the two. Given that it converges for large numbers of loci, I think we shouldn't worry too much about it right now though, but we should keep it in mind and hopefully resolve at some point.

DomNelson:
Great, thanks for the comments! I'll clean up the things you suggested, and in the mean .ime here's the full back trace in `verification.py`:

```
  File "verification.py", line 1925, in <module>
    main()
  File "verification.py", line 1922, in main
    verifier.run(keys)
  File "verification.py", line 229, in run
    runner()
  File "verification.py", line 1280, in f
    recombination_map=recombination_map)
  File "verification.py", line 1034, in run_dtwf_coalescent_comparison
    for ts in replicates:
  File "/home/dnelson/project/msprime/msprime/simulations.py", line 118, in _replicate_gen
erator
    sim.run(max_time)
  File "/home/dnelson/project/msprime/msprime/simulations.py", line 665, in run
    self.ll_sim = self.create_ll_instance()
  File "/home/dnelson/project/msprime/msprime/simulations.py", line 656, in create_ll_inst
ance
    node_mapping_block_size=self.node_mapping_block_size)
TypeError: 'recombination_map' is an invalid keyword argument for this function
```

Testing it a bit more though the original msprime-ms tests also raise this error, so something fishy may have happened in my merge...?

jeromekelleher:
Your merge looks fine, but there's definitely something fishy here. Have you recompiled your ``_msprimemodule.cpython...so`` file? (``make clean; make`` should do the trick). This is the only thing I can think of that would cause it (but nothing would work if that's not updated, so I don't know).

DomNelson:
Arrgh, I really should have thought to try that before posting... anyway, it worked!

verification.py
```diff
@@ -1024,6 +1075,348 @@ def f():
...
+            ts = subsample_simplify_slim_treesequence(ts, slim_args['sample_sizes'])
+
+            t_mrca = np.zeros(ts.num_trees)
+            try:
+                for tree in ts.trees():
+                    t_mrca[tree.index] = tree.time(tree.root) + slim_args['NGENS']
+            except ValueError:
+                ## Assuming this is due to multiple roots in the tree - 
+                ## increase number of generations and run again
+                slim_args['NGENS'] = slim_args['NGENS'] * 2
```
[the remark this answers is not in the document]
DomNelson: This has been tricky. Implementing coalescent checking broke the validation by some means I haven't found yet... both tmrca and num_trees were distorted.

I'm a bit hesitant to put too much work into understand this at the moment since SLiM tree sequence recording seems to be under active development. At least versions 3.0 and 3.1 required different handling of node times to give a proper comparison to DTWF - in 3.0 times were (coalescence_time - num_gens_simulated), which would always be negative.

What do you think? It's definitely slower than it needs to be, but functional

lib/msprime.c
```diff
@@ -2482,6 +2483,49 @@ msp_dtwf_generation(msp_t *self)
...
+    size_t index;
+
+    index = ((size_t) source_pop) * self->num_populations + (size_t) dest_pop;
+    self->num_migration_events[index]++;
+
+    // Iterate through nodes in tree and move to new pop
+    // -- removed from "nodes" in msp_move_individual()
+    while (avl_count(nodes) > 0) {
+        node = nodes->head;
+        ret = msp_move_individual(self, node, nodes, dest_pop);
```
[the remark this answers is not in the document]
DomNelson: Still working on this habit!

DomNelson:
Ok coalescence-checking strategy is updated, so now we run SLiM until coalescence, run again for an equal number of generations, and then return the simplified tree sequence.

In the meantime I've been trying to track down the cause of some small departures vs. SLiM when we have 3 or 4 populations:

num_trees:
num_trees

tmrca_mean:
tmrca_mean

Talking with Simon I realized that the inefficient but "functional" strategy of iterating over all lineages to draw migrants was actually incorrect! What we want is to draw a number of migrants for all populations together from a multinomial distribution - that's getting debugged now, and with any luck might be what was behind these departures

petrelharp:
> Ok coalescence-checking strategy is updated, so now we run SLiM until
coalescence, run again for an equal number of generations, and then return
the simplified tree sequence.


It may be worth checking whether "run again for 10x as many generations"
fixes your discrepancies: an equal number should be pretty good, but might
be the cause of small discrepancies.

jeromekelleher:
> Talking with Simon I realized that the inefficient but "functional" strategy of iterating over all lineages to draw migrants was actually incorrect! What we want is to draw a number of migrants for all populations together from a multinomial distribution - that's getting debugged now, and with any luck might be what was behind these departures

Aha, sneaky one. Let me know if I can help!

lib/msprime.c
```diff
@@ -2537,8 +2537,10 @@ msp_run_dtwf(msp_t *self, double max_time, unsigned long max_events)
     unsigned long events = 0;
     int mig_source_pop, mig_dest_pop;
     sampling_event_t *se;
-    double mu;
-    uint32_t j, k, i, n;
+    uint32_t j, k, i, N, num_migs, max_migs;
+    unsigned int n[self->num_populations];
```
jeromekelleher: This won't work on a lot of compilers I'm afraid (probably why the windows build is failing), and is a bit dodgy in any case. If you had a lot of populations you could overflow the stack frame, which would be a bad time for all. Best to just malloc/free the memory for now.

DomNelson:
This took me down the rabbit hole a bit, but I think the remaining differences are due to the different models being implemented, namely that SLiM simulates diploid individuals and msprime single chromosomes (albeit with diploid-based recombination). @petrelharp running for more multiple of coalescence time didn't make a difference I was able to see, although I didn't do any systemic checking of that.

The remaining discrepancies all occur when simulating multiple populations, so migration is a prime suspect. To approximate diploid migrants, I set up msprime to draw half as many migration events, but move two lineages each time instead of one - the mean number of migrant lineages stays the same, but the variance is affected. This changed this:

num_trees
num_trees

tmrca_mean
tmrca_mean

to this:

num_trees
num_trees

tmrca_mean
tmrca_mean

So far so good! However, I put a special case when a population contains only a single lineage so that it could migrate alone. This seems to cause differences for smaller-scale simulations, I assume since the special case has greater relative effect, and the approximation of individuals as diploid migrants may break down - for one thing, "selfing" is still possible and may distort things relative to SLiM in small populations. An example with 3 pops, 10 inds / 5 samples each, 0.1 Morgans, symmetrical migration rate 0.01: 

num_trees
num_trees

tmrca_mean
tmrca_mean

Single-locus simulations interestingly lined up well with the original single-lineage migration, but not with double-lineage migration. I haven't precisely pinned down the explanation for this yet, but it seems likely to be similar.

I'm currently seeing if I can show more explicitly what's going on. It seems as though some of this will simply come down to our choice of model - implementing full diploid individuals in msprime is something I've been planning to do (through genealogies) but I wasn't going to start that until the base DTWF model is complete. 

Any ideas how to explore what's happening here? Or convince ourselves that these are model differences and not bugs?

jeromekelleher:
Ouch; man, this looks hard! So, I'm happy to believe that the differences we're seeing are down to the model. It seems inevitable now that small choices in how we implement things will lead to measurable differences, and there's not a lot we can do about it. Probably the only way to be absolutely certain we're simulating exactly what we want to simulate would be to implement our own forwards-time simulator (using tree sequence recording) and compare with that. That seems like a lot of work, especially if we're going to be dealing with migration etc.

Would you mind summarising where we're at here? If you think we're basically correct and we have at least some tests against SLiM that verify things are working, then I'm happy to merge whenever you are.

lib/msprime.c
```diff
@@ -2490,49 +2537,104 @@ msp_run_dtwf(msp_t *self, double max_time, unsigned long max_events)
     unsigned long events = 0;
     int mig_source_pop, mig_dest_pop;
     sampling_event_t *se;
-    double mu;
-    uint32_t j, k, i, num_migrations, n;
+    uint32_t j, k, i, N, num_migs, max_migs;
+    unsigned int *n;
```
jeromekelleher: Minor thing here, but need to make sure that the pointers for stuff that's malloced are initialised to NULL and that the return value is checked:
```c
     unsigned int *n = malloc(self->num_populations * sizeof(int));
￼     double *mig_tmp = malloc(self->num_populations * sizeof(double));￼
￼     double sum;
￼     avl_tree_t *node_trees = NULL;
￼     avl_tree_t *nodes;
￼
     if (n == NULL || mig_tmp == NULL) {
        ret  = MSP_ERR_NO_MEMORY;
        goto out;
    }
```
The reason for making sure that pointers (or at least pointers to stuff we eventually malloc/free) are always initialised to NULL is that if an error occurs before malloc is called and we ``goto out`` then we'll try to free some arbitrary memory, and things will explode. This couldn't actually have happened here, but it's better to get into the habit of doing this stuff automatically.


￼

DomNelson: Ahh yes, thanks!

DomNelson:
Ok great, it sounds like we're on the same page then - I was also starting to feel as though the only "exact" validation would be to do something other than what we set out to do. I do think we'll line up a lot more closely with SLiM when we can follow genealogies, which will handle these haploid/diploid issues fairly automatically. That's high on my list of things to tackle next, but until then I'm comfortable with what we have now.

If this is sounding good to you too then, I'll double-check a few things and get set for the merge. Noob git question: Can I squash the commits and push to this branch, or will that mess with all the references in this PR? I could just open a new one too if that's easier.

petrelharp:
Damn, this does look hard. I'm glad to hear running it for longer past coalescence wasn't the culprit - I was surprised that made such a big difference.

DomNelson:
All set!

I implemented a crude lineage-sampling algorithm for diploid migration to see if that improves things vs SLiM and it looks quite good. Not included in the merge though - it's O(N^2) and not properly debugged, but roughly the idea is to draw a number of migrants from the number of diploid individuals ( == Ne / 2) and then draw lineages for each migrant from *all* lineages in the population (== Ne, since we include those not carrying ancestral material)

```
for (i = 0; i < num_migrants; i++) {
    // Sample two active/inactive lineages for this migrant
    for (l = 0; l < 2; l++) {
        prop_active_lineages = ((double) num_active_lineages) / ((double) num_total_lineages);

        if (gsl_rng_uniform(self->rng) < prop_active_lineages) {
            // Migrate an active lineage
            ret = msp_store_simultaneous_migration_events(
                    self, nodes, mig_source_pop);
            if (ret != 0) {
                goto out;
            }
            num_active_lineages--;
            num_total_lineages--;
        } else {
            // Migrate an inactive lineage
            num_total_lineages--;
        }
    }
}
```

Then for the 3-pop scenario above we now get:
tmrca_mean
screen shot 2018-10-26 at 2 03 53 pm

And excellent agreement in other multi-pop simulations. Still better to solve this through genealogies I think, but I'm happy that this seems to be behind the majority of the differences we were seeing.

petrelharp:
Just to check, you got this good agreement also with running longer than coalescence? Or just with the migration fix?

jeromekelleher:
Looks good, thanks @DomNelson. I've made a few changes to make it easier to run SLiM and sorted out some style issues that had crept into the ``verification.py`` script. If it all looks good to you I'll merge this, and we can work on the other issues that popped up separately.

DomNelson:
> Just to check, you got this good agreement also with running longer than coalescence? Or just with the migration fix?

@petrelharp Besides the migration fix, I reverted to running until coalescence, then running for an equal number of generations before returning the tree sequence. Running for 10x as many generations didn't make any difference I could see, at least in the statistics we're looking at. 

@jeromekelleher Yep sounds like a plan, all looks good to me.

PR #603, opened 2018-09-11
Xi beta

shajoezhu:


jeromekelleher:
I've tidied this up and refactored a bit @shajoezhu, would you mind taking a quick look to see if it's OK please?

The main thing I've done is factored out the integration parameters so that they are set at model init time.  This is to (a) save the effort of allocating the workspace over and over again; (b) make it so we can handle errors without aborting; and (c) set it up so we can set these parameters more easily from Python. 

Ultimately, I think we should probably set up the workspace size, epsrel and epsabs as full parameters of the model which can be set from Python. There's no one-size-fits-all for these intregration tuning parameters, and it's nearly always easier to allow them to be set by the user. We don't have to do this now, but we should do it before release.

shajoezhu:
Hi @jeromekelleher , it looks really good! Thanks a lot for tidying things up. Should have done a better job myself. I really like what you did with factoring out the error term with msprime exception. I agree, I think setting up workspace size, epsrel and epsabs as full parameters from Python is a good idea. I think I will do a bit of investigation on these gsl integration solvers.

jeromekelleher:
OK great, will merge once CI passes.

shajoezhu:
That's brilliant! Thanks a lot @jeromekelleher ! I should remember to squash next time

PR #605, opened 2018-09-15
Updated test to work on newer glibc.

jeromekelleher:
For some reason we seem to have got some extra precision on newer versions of glibc. Made the test more stringent so these are caught.


jeromekelleher:
But also, seemed to be failing in the same way on Windows. Weird...

PR #634, opened 2018-11-21
Added inferface for managing GSL error handlers.

### LINKED ISSUE #632: GSL error handler clobbering
In lib/msprime.c, msp_alloc sets the GSL error handler to "off" with every call.  The effect of this behavior is to undo the error handler of any other package running in the same memory space.  I noticed this when unit tests started suddenly failing in molpopgen/fwdpy11#142.  The cause was having msprime and fwdpy11 both listed as top-level imports.

Not sure what the best approach to fixing this is.

jeromekelleher:
Closes #632.

@molpopgen, this will hopefully fix the issues that you've seen. If you do the following:

```python
import msprime
import _msprime

_msprime.restore_gsl_error_handler()
```
It should restore the GSL error handler to it's value before ``msprime`` was imported. I've left the function in the low-level module as this seems like a fairly obscure case that we don't need to complicate the high-level interface for. I'm happy to provide a function in the top level so we can avoid importing the low-level package though, if you think this would be easier to work with. Also happy to provide a different interface, if anything would work better for you.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -28,6 +28,7 @@
 
 #include <gsl/gsl_version.h>
 #include <gsl/gsl_math.h>
+#include <gsl/gsl_errno.h>
 
 #include "msprime.h"
 
@@ -40,6 +41,9 @@
 
 #define SET_COLS 0
 #define APPEND_COLS 1
+
+/* We keep a reference to the gsl_error_handler so it can be restored if needed */
+static gsl_error_handler_t *old_gsl_error_handler;
 
 static PyObject *MsprimeInputError;
 static PyObject *MsprimeLibraryError;
@@ -10050,6 +10054,22 @@ msprime_get_gsl_version(PyObject *self)
 }
 
 static PyObject *
+msprime_restore_gsl_error_handler(PyObject *self)
+{
+    gsl_set_error_handler(old_gsl_error_handler);
+    return Py_BuildValue("");
+}
+
+static PyObject *
+msprime_unset_gsl_error_handler(PyObject *self)
+{
+    /* turn off GSL error handler so we don't abort on errors. Can be restored
+     * by calling restore_gsl_error_handler() */
+    old_gsl_error_handler = gsl_set_error_handler_off();
+    return Py_BuildValue("");
+}
+
+static PyObject *
 msprime_get_library_version_str(PyObject *self)
 {
     return Py_BuildValue("s", MSP_LIBRARY_VERSION_STR);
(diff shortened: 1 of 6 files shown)


molpopgen:
I think this works for many cases.  Fwdpy11 unit tests will still break, I think, but I'll have to check next week.

jeromekelleher:
> I think this works for many cases. Fwdpy11 unit tests will still break, I think, but I'll have to check next week.

Imports for unit tests are pretty opaque all right... Any other ideas for how to structure this? I could set it up so the error handler is only set on the first call to simulate(), *if* some module level variable hasn't been set. Something like:

```python

# Set this to False if you don't want msprime to touch the GSL error handler
set_gsl_error_handler = True
_gsl_error_handler_set = False

def _check_gsl_error_handler():
    if set_gsl_error_handler and not _gsl_error_handler_set:
        _msprime.set_gsl_error_handler()
        _gsl_error_handler_set = True

def simulate(...):
     _check_gsl_error_handler()
    # Do everything else
```

This way, you know that the GSL error handler hasn't been touched before you call simulate for the first time, and you can easily prevent it from mucking with it. This should work, right?

molpopgen:
I'll have to ponder it some more, but I won't have a machine until Monday.  I'm a bit concerned that the solution is to make all imports local when writing unit tests.  Or, there is an option to pass to "unittest discover" that I'm not aware of...

jeromekelleher:
> I'll have to ponder it some more, but I won't have a machine until Monday. I'm a bit concerned that the solution is to make all imports local when writing unit tests. Or, there is an option to pass to "unittest discover" that I'm not aware of...

Well the proposed update above should take care of that --- nothing at all is done until msprime.simulate is called, so you could just have:
```python
def run_msprime_sim(*args, **kwargs):
    msprime.set_gsl_error_handler = False
    return msprime.simulate(*args, **kwargs)
```
or something similar. This has to work, no matter how convoluted the import structure is.

jeromekelleher:
Or, I guess we could even add a ``set_gsl_error_hander`` argument to msprime.simulate() which would have the same effect, and avoid the messing with module variables. The only thing is in principle we should be setting this also when making a the RandomGenerator instance, since allocating the rng can fail (not really in practise though).

molpopgen:
I'll take a look today.

molpopgen:
I modified the fwdpy11 unit tests to trigger the failure and then implement the fix described above.  All tests pass once `_msprime.restore_gsl_error_handler()` is called.  I think this can be merged.

PR #636, opened 2018-11-22
Separating into tskit and msprime

jeromekelleher:
This is the biggie, and it's a WIP.

~~I have the C code more-or-less separated out, but the Python is all broken now. The plan is to make a another package in this repo for now, which will exist in parallel. Once that's all working, I think we should commit that much, so that msprime development isn't impacted too much. We can then work on the initial tskit code drop, and work away there until it's ready to take over from the version embedded in msprime. Once this is working, we delete the version here in msprime.~~

I've separated out the C code and everything seems to be working OK. The plan now is to refine the C API until we're happy with it, and then start moving things over to the tskit repo.


jeromekelleher:
Note to self: add in missing test_genotypes.c file.

jeromekelleher:
@petrelharp FYI, I'm merging this in as a step towards pulling tskit out. Basically we have the outline of the C API on its own. There's a bit of refinement of the semantics still to be done, but 95% of the C level changes should be done now. I'll be plugging away on this where I can for a while --- the aim is for the next release of msprime to depend on tskit.

petrelharp:
awesome!

PR #638, opened 2018-11-26
added single_sweep model into algorithms.py

andrewkern:
I've implemented a single_sweep model in algorithms.py. this relies on the implementation of 'labels' for segments which leads to a bunch of lower level changes, but nothing too drastic. @jeromekelleher please take a look when you get a chance and then we can start moving this into the C side of things.

PR #642, opened 2018-12-04
Basic label infrastructure in C.

jeromekelleher:
@andrewkern, this is the basic infrastructure needed for structured population labels in C.Would you mind taking a look through this and seeing what you think please?

If we're happy with this much, I vote we merge and then continue  the actual development of the sweep algorithm in #638 (after a rebase).


andrewkern:
oops maybe i was supposed to tag you @jeromekelleher... still no good with github

PR #648, opened 2018-12-07
Split up low-level C modules for msprime and tskit

jeromekelleher:
WIP, depends on #540.


jeromekelleher:
This is another big step towards pulling out tskit, so I'm pinging @tskitdev/all to keep you all in the loop.

This PR splits the Python code into two modules, ``msprime`` and ``tskit`` which are both currently held in the same repo and managed by a single setup.py. From msprime's perspective, there's not a lot to be done, so the simulation code should be pretty stable from here. So, if you're working on the simulator then you should be able to rebase onto master and everything should go fairly smoothly from this point.

There's a couple of technical points on the low-level interaction between msprime and tskit and msprime that I want to nail down before pulling the tskit code out entirely, so there might be a little bit more churn at the C level before we're clear to go there, and tskit development moves to it's own repo. I'm hoping that this can be completed by mid January (I have other stuff I need to do before the holidays).

Issue #564, opened 2018-07-28
Implement max_time for simulations


jeromekelleher:
#541 partially implemented the idea of max_time for simulations. ~Since this gives us a neat way of running hybrid simulations within msprime (using the simulate from feature), we would not longer need ModelChange events to implement this. Remove ModelChange events from Python code. It may also be worth refactoring C code to remove some complexity associated with changing models, but it's probably not worth the effort.~

The ``max_time`` should guarantee that all times in the returned tree sequence are strictly **less than** max_time. The current implementation doesn't provide this guarantee.

jeromekelleher:
Update: this parameter should be called ``end_time`` to mirror the ``start_time`` parameter.

jeromekelleher:
Updated: The ModelChange events are still useful, as it avoids the messiness that we have in remapping coordinates in the process recapitation process.

PR #650, opened 2018-12-08
algorithms.py updated with labels and single_sweep mode-- take 2

andrewkern:
okay here is a new PR from a brand new fork. overkill i think and i apologize for opening a second PR but this seemed easiest?

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1017,7 +1017,6 @@ msp_store_edge(msp_t *self, double left, double right, node_id_t parent, node_id
         self->buffered_edges = edge;
     }
     if (node_time[child] >= node_time[parent]) {
-        printf("Storing %d %d time = %f %f\n", child, parent, node_time[child], node_time[parent]);
         ret = MSP_ERR_TIME_TRAVEL;
         goto out;
     }
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -1115,12 +1115,12 @@ class TestSimulator(LowLevelTestCase):
                 self.assertRaises(ValueError, f, [d, d])
         self.assertRaises(ValueError, f, [{"start_time": 1, "type": -1000}])
         for bad_number in ["", None, [], {}]:
-                self.assertRaises(
-                    TypeError, f,
-                    [get_population_configuration(initial_size=bad_number)])
-                self.assertRaises(
-                    TypeError, f,
-                    [get_population_configuration(growth_rate=bad_number)])
+            self.assertRaises(
+                TypeError, f,
+                [get_population_configuration(initial_size=bad_number)])
+            self.assertRaises(
+                TypeError, f,
+                [get_population_configuration(growth_rate=bad_number)])
         # Cannot have negative for initial_size
         for bad_size in [-1, 0]:
             self.assertRaises(
--- a/tests/test_simulate_from.py
+++ b/tests/test_simulate_from.py
@@ -846,6 +846,7 @@ class TestErrors(unittest.TestCase):
             with self.assertRaises(ValueError):
                 msprime.simulate(from_ts=base_ts, start_time=start_time)
 
+    @unittest.skip("FIXME C_0.99.1")
     def test_malformed_tree_sequence(self):
         tables = tskit.TableCollection(1)
         tables.populations.add_row()


andrewkern:
huh not sure why these checks are failing??

petrelharp:
The circleci test looks to do with the upgrade to kastore - maybe either you need to submodule update or we need to ask circleci to do it?

andrewkern:
Will try submodule update. I tried to git pull locally within kastore but it had no effect

andrewkern:
no dice on `git submodule update`. didn't seem to update anything. i went ahead and did `git fetch origin; git reset --hard origin/master` in `lib/kastore/` and now I can't compile my local version due to an error with `kastore`:

`lib/tskit/tsk_tables.c:5499:52: error: use of undeclared identifier 'KAS_NO_MMAP'
    ret = kastore_open(self->store, filename, "r", KAS_NO_MMAP);`

any other ideas?

petrelharp:
OK, right. That's one of the changes [made recently](https://github.com/tskit-dev/kastore/pull/57) by @jeromekelleher; but those changes haven't been propagated to msprime yet, I think.  So updating `kastore` was the *wrong* thing... but by doing it you got the same error that circleci did. So, it looks like circleci is maybe using the most recent kastore version instead of 0.2.0?

petrelharp:
Yep, so that's the problem,then: circleCI (like you) is using the most recent version of `kastore`:
```
#!/bin/bash -eo pipefail
git submodule update --init --recursive --remote

Submodule 'lib/kastore' (https://github.com/tskit-dev/kastore) registered for path 'lib/kastore'
Cloning into '/home/circleci/msprime/lib/kastore'...
Warning: Permanently added the RSA host key for IP address '192.30.253.113' to the list of known hosts.

Submodule path 'lib/kastore': checked out 'f5254264698b355a7e3e1ecd2f7680f0f7beb894'
```
but it should be using `0.2.0` still.  The `KAS_NO_MMAP` thing @jeromekelleher removed [recently](https://github.com/tskit-dev/kastore/pull/57) in kastore, but hasn't propagated the changes to msprime yet.

petrelharp:
You should move your kastore submodule back to 0.2.0; I'm not sure how to fix circleci.

andrewkern:
that fixed it right up! circleCI now is still broken i guess but the changes I push here have nothing to do with kastore or the C side of things at all.

jeromekelleher:
Sorry guys, this is my fault. The current master isn't compatible with kastore < 0.2.0, which I forgot about. I'll sort it out and merge this PR either later today or tomorrow.

jeromekelleher:
Thanks for this @andrewkern. I think it's looking good, but we should figure out some basic structural issues before going further.

At the moment,  it seems like the ``single_sweep`` is a mix of an event and a simulation model. I think we should resolve this so that the ``single_sweep`` is a simulation model which we think about in isolation. In the main code we have the idea of a [SimulationModelChange](https://github.com/tskit-dev/msprime/blob/master/msprime/simulations.py#L1035) event which then runs until either some maximum time or full coalescence. I think we should fit the ``single_sweep`` model into this framework.

In the low-level code, we also have a [max_time](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L2899) parameter built in, so that we can ensure that simulations stop at a certain time point (although the semantics of the operation aren't currently quite right; see #564). I think this removes the need for the ``simulate_until_x`` variants we have here?

What I'd like to see here I think is just running the sweep model from the standard initial conditions until it finishes, and then we output the tree sequence and terminate. So, there's no initial or final Hudson phases; we start in a sweep and we end in the sweep. It doesn't matter if the trees aren't completed. If this is in place, I can implement the model change stuff easily enough which will allow us to plug these phases in cleanly (or it may not be worth the bother for the prototype Python version). Does this sound OK, or am I missing something fundamental here?

## Style issues

Coding style is personal, and  it really doesn't matter at the end of the day, but I think it's extremely important that style is reasonably uniform within a project. To minimise the time wasted on style discussions, I've implemented automated checks using [flake8](http://flake8.pycqa.org/en/latest/). I've made the changes here to make the code [PEP8](https://www.python.org/dev/peps/pep-0008/)  compliant. It's a good idea to run it locally on the code before pushing changes though, as it'll save a lot of time chasing down CI failures.

I've also changed the variables from camelCase to snake_case.

jeromekelleher:
ps. Don't worry about the CodeCov failure. This is just because the project-wide C test coverage has slipped below 90%. We'll fix this as part of adding the C implementation of the sweeps.

jeromekelleher:
Oh yeah, I also rebased this branch on upstream to pull in some changes. To use it, you need to do:
```
$ git fetch origin
$ git checkout master
$ git reset --hard origin/master
```

andrewkern:
Cool. So it seems like using the SimulationModelChange paradigm would do the right thing. Should I implement that paradigm in algorithms.py? Basically for a single sweep we have up to 3 phases of the coalescent simulation-- a neutral phase until the sweep event, a sweep phase, and then another neutral phase (not all are needed of course). The way I have written things in algorithms.py accomplishes this by using those `*_simulate_until` variants, as you noticed. 

One wrinkle we should be thinking about now is that the trajectory simulation which we will condition on is only compatible with piecewise constant population size changes. I'm currently dealing with this in the inelegant loop [here](https://github.com/andrewkern/msprime/blob/35ec6818d25cfda53240ceaa6174a770d81810f6/algorithms.py#L680).

jeromekelleher:
> Cool. So it seems like using the SimulationModelChange paradigm would do the right thing. Should I implement that paradigm in algorithms.py? Basically for a single sweep we have up to 3 phases of the coalescent simulation-- a neutral phase until the sweep event, a sweep phase, and then another neutral phase (not all are needed of course). The way I have written things in algorithms.py accomplishes this by using those *_simulate_until variants, as you noticed.

I wouldn't bother implementing the `SimulationModelChange` yet in algorithms.py. If we need it, I can splice it in later pretty quickly. For now though, all we need is just the implementation of the ``single_sweep`` model. We can just run it from the initial conditions and finish the simulation when it's done. It doesn't matter that it's unrealistic, we just need to get the important code in there in the simplest way possible.

So, would you mind pulling out the ``simulate_until`` stuff and setting up ``single_sweep_simulate`` to run from the start and exit when it's done?

> One wrinkle we should be thinking about now is that the trajectory simulation which we will condition on is only compatible with piecewise constant population size changes. I'm currently dealing with this in the inelegant loop here.

I'm imagining that we'll be able to do all the trajectory simulations in Python (they don't need to be fast/big, right?), so I'm not worrying too much about how we go about this at the moment. Maybe I'm missing something though?

andrewkern:
okay so i believe i made all of the requested changes. along the way i found some bugs in the basic label plumbing too, but i *think* everything is now working.

what i've done here is get rid of the `*_simulate_until()` functions, and pulled all the `single_sweep` code under a `model`. this model runs a sweep back in time from the current sample-- there is no option for the structured coalescent phase to have begun at some point in the past other than time zero. it's quite basic but i think it achieves what you wanted. 

i've also added some code that might annoy you, like this [helper](https://github.com/andrewkern/msprime/blob/217bdc18f6a746a53776dbea12d70b5982c3c46d/algorithms.py#L241) which is used to remove a segment from a Population's ancestors array [here](https://github.com/andrewkern/msprime/blob/217bdc18f6a746a53776dbea12d70b5982c3c46d/algorithms.py#L902). It's a bit ugly, so maybe there is a cleaner way to do this?

jeromekelleher:
Thanks @andrewkern, it's looking good and I think we're nearly there! The main problem I realise is that I don't really understand what's happening here, so I'm struggling to isolate the simplest instance that I can run and work with. Here's some points where I'm confused:

- I don't really understand interaction between the sweep and populations here. Can we not have a sweep happening in a single population? If so, does it make sense to get this simpler case working first?
- I also don't understand the interaction between the sweep and population size changes. At the moment, it seems that if we don't have any population size changes then the whole thing is pretty trivial and very little happens. Is this true? Can we have a meaningful sweep happening in a single constant size population? If so, does it make sense to get this simple case working first?

I've refactored the code a bit to try and avoid repetition and to decouple the trajectory and sweep simulation phases. I've put some comments into the code, and made some inline comments here on GitHub too to explain. I've probably broken your code with this, so apologies for that. I hope I'm not annoying you too much with the stupid questions and nitpicking also!

algorithms.py
```diff
@@ -228,6 +232,90 @@ def add(self, individual):
...
+        find the index of an ancestor in population
+        """
+        return self._ancestors.index(indv)
+
+
+class TrajectorySimulator(object):
+    """
+    Class to simulate an allele frequency trajectory on which to condition
+    the coalescent simulation.
+    """
```
jeromekelleher: I've changed the interface of this a bit so that it's now reponsible for simulating trajectories.

algorithms.py
```diff
@@ -228,6 +232,90 @@ def add(self, individual):
...
+        # because it seems like an unnecessary complexity for this initial prototype and
+        # I don't understand what happens in the single population size regime. What's
+        # the simplest possible instance of this trajectory that we can use as input
+        # to the coalescent simulation phase?
+        # I've added the _modifier_events variable here so that the code runs.
+        self._modifier_events = []
+
+        # Go through each event looking for population size changes in popn0
+        # not dealing with sweeps in other popns or growth
+        for event in self._modifier_events:
```
jeromekelleher: Rather than passing in events, it would be cleaner here to pass in the actual sequence of population sizes. That way this class won't need to know anything about the events and it'll be easier to test. Does this make sense?

algorithms.py
```diff
@@ -254,19 +342,20 @@ def __init__(
...
         self.max_segments = max_segments
         self.segment_stack = []
         self.segments = [None for j in range(self.max_segments + 1)]
         for j in range(self.max_segments):
             s = Segment(j + 1)
             self.segments[j + 1] = s
             self.segment_stack.append(s)
-        self.P = [Population(id_) for id_ in range(N)]
-        self.L = FenwickTree(self.max_segments)
+        self.P = [[Population(id_) for id_ in range(N)] for x in range(num_labels)]
```
jeromekelleher: Ah, I've just realised what some of my confusion is. In the C code I've put population as the first index and label as the second. So, each population contains a bunch of labels. 

Do you mind if I change the code here to follow the same convention? It'll be very confusing if they're different.

andrewkern: Oh yes totally fine to change which order of the indexes!

jeromekelleher: OK, I'll do the old swaparoony now. Shouldn't take long.

jeromekelleher:
I've swapped the order of the indexes here, so that labels are stored within population objects. The ``add`` and ``remove`` methods take an optional parameter which should abstract things a little bit. I fixed up the code for Hudson and DTWF, but I'm afraid I've broken it for ``single_sweep``: I ran out of time, sorry!

Hopefully this doesn't muck things up too badly anyway, and it'll make the code a bit easier to follow.

jeromekelleher:
I've cleaned up the TrajectorySimulator and added some notes on how I see it working for more complex scenarios in the real code @andrewkern; how does this look? Basically, I think we can abstract out the population size calculation here which will simplify things a lot.

Looking forward to seeing the results!

andrewkern:
Hey-- i've carved out some time to push this ahead and have a question. Right now with the switch to labeling that you have made there is a bunch of broken code. The most fundamental might be the behavior of the `Population` class upon iteration. What should we do in this regard? In particular this breaks the current implementation of `single_sweep_simulate()` right now, but I'm hoping that you can make a call as to what we should do and I'll work around that. 

The trajectory stuff should be all set as well.

andrewkern:
fuck looks like i've introduced some conflicts that can't get auto-merged. christ...

andrewkern:
at any rate please advise @jeromekelleher and i'll will hop to it.

jeromekelleher:
Sorry, I should have warned you that we'd updated algorithms.py in the meantime @andrewkern --- things move quickly here!

In general, I'd advise against merging changes back into topic branches. It's better to **rebase** so that the tail of your topic branch starts at the tip of upstream/master. See the [network illustration](https://github.com/tskit-dev/msprime/network) --- it's after getting messy here. See the advice [here](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow). 

The CI failures are due to problems that occured elsewhere, which I've fixed. These should go away when you rebase on master and force push.

jeromekelleher:
> Hey-- i've carved out some time to push this ahead and have a question. Right now with the switch to labeling that you have made there is a bunch of broken code. The most fundamental might be the behavior of the Population class upon iteration. What should we do in this regard? In particular this breaks the current implementation of single_sweep_simulate() right now, but I'm hoping that you can make a call as to what we should do and I'll work around that.

I've forgotten where we were at with this, to be honest. I remember refactoring things around a bit to make them map into the low-level code better and leaving some stuff a bit broken (hoping you'd fix it!). Might be best to arrange a quick call to talk it through?

andrewkern:
Yeah sounds good— I’m free all morning (PST) what time works best for you? Shouldn’t take long, we just need to make a quick decision about what to do with the iterator.

jeromekelleher:
> Yeah sounds good— I’m free all morning (PST) what time works best for you? Shouldn’t take long, we just need to make a quick decision about what to do with the iterator

9am PST/5pm GMT?

andrewkern:
how about 9:15/5:15? google chat?

jeromekelleher:
SGTM; see you then.

PR #659, opened 2018-12-14
Using dictionaries of arrays as the long-term interchange format.

jeromekelleher:


jeromekelleher:
The goal here is twofold:

1) Sort out a long-term stable API between tskit and msprime so that they can interchange tree sequences reliably, robustly to API and ABI changes in the tskit C library and with minimal dependencies between the two at an ABI level;
2) Sort out the lifecycle of the ``_tskit.TableCollection`` and the low level semantics of the ``tsk_tbl_collection_alloc``, which are very complicated at the moment.

Up to this point I was trying to set things up so that ``msprime`` could dynamically link against the ``tskit`` C API exported as part of the ``_tskit`` Python module. This avoids having several different versions of the ``tskit`` C API living at the same time, simplifying some issues allowing us to pass a pointer to a ``tsk_tbl_collection_t`` between the two modules. While this works, it places very hard restrictions on the versions of the tskit C API that both are compiled against and we'd have to be very careful indeed about versioning so that ABI requirements are met. We could easily get into a situation where updating ``tskit`` leads to breaking an existing msprime installation in a very messy and user-unfriendly way.

So, I've come to the conclusion that it's actually better to compile in two independent versions of the tskit C API into the msprime and tskit Python modules. Using the dictionary encoding of the tables we can interchange the data reliably with no ABI worries at all. There's no cost efficiency-wise because we're using numpy arrays as lightweight wrappers around the underlying table data, so we can make a copy of the simulator's tables without an intermediate copies of the columns.

The lifecycle stuff is something is a hangover from earlier ways of doing things, which I hope I can simplify a lot if we have this canonical dictionary-based way of creating a new TableCollection. This stuff gets quite tricky though, so we'll have to see how it works out.

jeromekelleher:
Notes to self:

1. Change the low-level ``columns`` methods to take a dictionary as an argument instead of kargs for each of the columns.
2. Change the constructor for `_tskit.TableCollection` to take a dictionary describing the full table collection as parameter. Ideally the ``tables.nodes`` etc references will then return a new reference to a node table which depends on the the TableCollection. (Perhaps the different tables could *optionally* have a reference to a TableCollection -- if they do, the table is backed by the memory from the collection, if not it's backed by a local object?)
3. Once this is all tested and the dictionary encoding used to create a low-level ``tbl_collection_t`` is finished and well tested, copy this into ``_msprimemodule`` and use this code for getting the tables in the ``from_ts`` argument.
4. Once this is all working, also update the low-level C API to change the ``increment_sizes`` from something set at ``alloc`` time to something we specify afterwards using a setter function. This should simplify things. Perhaps we should remove the ``node_table_alloc`` etc entirely, as these should only be used in the context of a table_collection anyway? Maybe not.

jeromekelleher:
OK, this is starting to look good and we're nearly ready to split tskit off I think. Things remaining to be done:

1. Sort of out provenance recording. Msprime is using a private tskit API at the moment, which isn't good.
2. Final semantics updates on the C API, which should be much simpler now that the Python API has been streamlined.

jeromekelleher:
> Any particular area you want extra eyes on?

Not right now, thanks. If you want to make a pass through and comments you have would be great though.

I'll be calling in some serious review soon once the tskit C API is ready to go!

jeromekelleher:
This is in good shape mostly, so I'm going to merge so we don't have such a massive patch ready to drop. I'll document the remaining issues required for full tskit extraction.

Issue #616, opened 2018-09-28
Tools for ancestry/inheritance calculation


petrelharp:
Here's some closely related use cases that we're seeing (e.g. #612). I'd like to collect them in one place so we can settle on a minimal number of useful methods and an API.  So far:

1. (ancestry painting) Given a list of nonoverlapping sets of nodes, the `ancestors`, and a collection of samples, return somehow an assignment for each sample of which set of nodes it inherits from on each part of the genome (if any).

2. (ancestry proportions) Instead of the whole ancestry painting, return summaries of it in windows: what proportion of the samples's genomes inherit from each set of ancestors, in each window.

3. (number of descendants) Given a list of samples and a list of nonoverlapping sets "ancestral" nodes, return for each set of ancestors what proportion of the samples' genomes that inherit from that set of nodes. (This should also be done in windows.)

The last two sound very similar, but one is indexed by ancestors; the other by descendants.

jeromekelleher:
Totally. 

For future reference, see also the functions ``node_span`` and ``mean_sample_ancestry`` [here]( https://github.com/tskit-dev/tsinfer/blob/master/tsinfer/eval_util.py#L721)

gtsambos:
Hi @petrelharp and @jeromekelleher , I’ve been thinking about (1) myself too, and have thought of a solution that will suit my purposes. The basic idea is to make a table that lists all sample:ancestral node relationships in my TreeSequence. These are pretty similar to EdgeTables, except that they contain information about branches of the tree rather than edges.

A descriptive notebook about this is [here](https://github.com/gtsambos/AdmixtureSim/blob/master/6.true.local.ancestry/1.finding-true-local-ancestry.ipynb). Feel free to have a look and let me know what you think!

(2) and (3) have not been as high on my radar, but I think it should be pretty easy to aggregate this kind of information from the more detailed output that I’m producing.

I’m going to keep chugging on with this by myself as it’s something I’ll need for my own analyses in the near future, but I’d be happy to submit a pull request if you find any of it useful!

Btw, I'm copying my supervisor @dvukcevic into this so he can see what I’m up to. (I’ll be away for a few days on holiday, again, so I won’t respond very quickly)

jeromekelleher:
> A descriptive notebook about this is here. Feel free to have a look and let me know what you think!

This notebook is really cool! I don't quite get it on my first quick reading through, but I'll think about it again early next week and see if I can get my head around it.

jeromekelleher:
[...]

PR #675, opened 2018-12-21
document mean_descendants

petrelharp:
I've written documentation for mean_descendants... but, I think we ought to change the behavior.  Currently, here's what it does:
```
        Computes for every node the mean number of samples in each of the `reference_sets`
        that descend from that node, averaged over the portions of the genome for which
        the node is ancestral to *any* sample.  The output is an array, `C[node, j]`,
        which reports the total length of all genomes in `reference_sets[j]` that
        inherit from `node`, divided by the total length of the genome on which `node`
        is an ancestor to any sample in the tree sequence.

        .. note:: This interface *may change*, particularly the normalization by proportion
            of the genome that `node` is an ancestor to anyone.
```
I think that this normalization should either:

- be optional, e.g., by a flag called `normalize=False`. The normalization is a nice idea, but it makes the statistic depend on who's marked as a sample in the tree sequence, which isn't so great.

- or, we provide another function that computes the mean proportion of the genome for which each node is ancestral to *anyone* in `reference_sets`; say, `mean_is_ancestor()`, so then the current behavior of `mean_descendants` would be the un-normalized version divided by `mean_is_ancestor`.

But, this is part of a bigger API discussion around #616. I am bringing this up as a concrete example of what we'll need to think about.


petrelharp:
Darn it, I don't know how to get rid of the submodule changes - `kastore` is at a different version for some reason. WTF, submodules?

andrewkern:
do you have to do the submodule init / submodule update thing?

jeromekelleher:
Yeah, submodules are a PITA. Not sure there's a better way to do things though, unfortunately. Try doing something like ``git checkout lib/kastore`` maybe? The idea is that kastore should be checked out at a tagged release. If you ``cd`` into the kastore directory, check out a different version at then ``git add`` the kastore directory, this will commit that particular version into the main repo. (Not sure how helpful this description is though!).

Re the docs, I think we should decouple this from the idea of 'samples' entirely. I thought we'd already done this, but if not we should. The nodes in the reference sets should be arbitrary. The normalisation then is with respect to the union of all the reference sets, which should address your issue here?

jeromekelleher:
Any thoughts here @petrelharp? I'm going to be moving the code to tskit in the next couple of days, so we should either fix this up here or open an issue over in tskit for it.

petrelharp:
There. That should do it.

jeromekelleher:
> OK, thanks. What about the normalisation stuff though? I don't think it's correct as it is.

Scratch that, I was mixing this up with genealogical_nearest_neighbours. I think the normalisation is wrong, and we should probably  change this function to normalise wrt to the union of the reference sets.

petrelharp:
Did you mean a code bug or a conceptual bug?  But yes, I think we need to have a bigger-picture think about what to compute, with @gtsambos.

PR #677, opened 2019-01-08
Doc Fix for SparseTree.num_nodes

jgallowa07:
Simply clarifies (in the docs) what SparseTree.num_nodes returns. This created a bug in my code as I expected it to be equivalent to `len([n for n in st.nodes()])`

### DIFF
--- a/tskit/trees.py
+++ b/tskit/trees.py
@@ -604,7 +604,9 @@ class Tree(object):
     @property
     def num_nodes(self):
         """
-        Returns the number of nodes in the sparse tree.
+        Returns the number of nodes in the :class:`.TreeSequence` this tree is in.
+        Equivalent to ``tree.tree_sequence.num_nodes``. To find the number of
+        nodes that are reachable from all roots use ``len(list(tree.nodes()))``.
 
         :rtype: int
         """


tskit/trees.py
```diff
@@ -615,7 +615,12 @@ def is_sample(self, u):
     @property
     def num_nodes(self):
         """
-        Returns the number of nodes in the sparse tree.
+        Returns the number of nodes in the treeSequence that encapsulates this tree.
+        Equivalent to: ``tree.tree_sequence.num_nodes``.
+
+        note that a simplified SparseTree is a proper binary tree. meaning the
```
jeromekelleher: Looks good, thanks @jgallowa07.

This second bit isn't true though --- we can have trees that aren't binary, even if they're simplified.

jgallowa07: Ah, then the best way to know all nodes associated with a single tree is simply `len([n for n in st.nodes()])`?

jeromekelleher: Yep, ``len(list(tree.nodes()))`` is probably the best way to do it. We could add a method to compute it more efficiently if it's a common operation. How are you using it?

jeromekelleher:
Merged; thanks @jgallowa07!

PR #680, opened 2019-01-10
Remove tskit

jeromekelleher:
Final removal of tskit code. Currently depends on a pre-release version of tskit.


jeromekelleher:
Windows builds are failing because tskit doesn't have the right link options set. Will need to fix before we can go further.

petrelharp:
yay! go, jerome!! :fireworks: :champagne:

jeromekelleher:
Thanks Peter, we're getting there!!!

PR #699, opened 2019-02-06
Full arg migration test

JereKoskela:
I've made three changes here:

1) The most immediate is that I've added a test which runs through the list of stored migration events, verifies that each corresponds to exactly one full-arg edge, and vice versa. I haven't worked out how to actually run these tests from the `test_demography.py` file, but have run a functionally identical code as a stand-alone Python script (just replacing self.assertEquals with assert, for example), and that seemed to work fine.

2) I've added the `MSP_NODE_IS_MIG_EVENT` flag to nodes created by the `msp_move_individual` function under `store_full_arg`. Previously the flag was set to 0. The test in the Python file uses the hard-coded, numerical value of this flag - is that an ok thing to do?

3) I've changed the condition for running `defrag_required` when merging ancestors to take into account the fact that all segments will point to a common node under `store_full_arg`. This wasn't captured by the earlier implementation since the values of the segments were compared to determine `defrag_required` before running `msp_store_arg_edges`.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -790,8 +790,11 @@ class Simulator(object):
                     pop.add(alpha)
                     self.L.set_value(alpha.index, alpha.right - alpha.left - 1)
                 else:
-                    defrag_required |= (
-                        z.right == alpha.left and z.node == alpha.node)
+                    if self.full_arg:
+                        defrag_required |= z.right == alpha.left
+                    else:
+                        defrag_required |= (
+                            z.right == alpha.left and z.node == alpha.node)
                     z.next = alpha
                     self.L.set_value(alpha.index, alpha.right - z.right)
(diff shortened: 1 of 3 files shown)


JereKoskela:
Sorry to do things in two parts, but I realised I needed to update `defrag_required` in `algorithms.py` as well, and answered my own question from point 2 above in the process by working out how to depict migration flags in Python properly. I've also fixed the niggling whitespaces that CircleCI pointed out.

jeromekelleher:
This is excellent, thanks @JereKoskela!  Merging.

>  I've also fixed the niggling whitespaces that CircleCI pointed out.

Apologies for the level of pedantry here, but it's the only way to make sure that code is consistent...

PR #700, opened 2019-02-06
single_sweep model take 3

andrewkern:
okay Jerome I think I'm there. I've cleaned up the broken labels stuff, changed the paradigm to the reverse ordering you had wanted, and implemented a very minimal case of single sweeps. Note that the `single_sweep_simulate()` does not guarantee that the entire sample has coalesced, for that we would need to add a round of the vanilla coalescent after the sweep has completed. 

take a look and let me know what i should move to next.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -123,6 +123,7 @@ class Segment(object):
         self.prev = None
         self.next = None
         self.population = None
+        self.label = 0
         self.index = index
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


algorithms.py
```diff
@@ -482,20 +576,123 @@ def hudson_simulate(self):
...
+        allele_freqs, times = self.sweep_trajectory
+        sweep_traj_step = 0
+        x = allele_freqs[sweep_traj_step]
+
+        assert self.num_populations == 1
+
+        # go through segments and assign labels
+        # a bit ugly with the two loops because
+        # of dealing with the pops
+        indices = []
```
jeromekelleher: We must be missing something here I think. The ``indices`` array is always empty, so the second loop never runs.

algorithms.py
```diff
@@ -604,21 +802,67 @@ def hudson_recombination_event(self):
...
+        self.L[label].set_value(z.index, z.right - z.left - 1)
+        self.P[z.population].add(z, label)
         if self.full_arg:
             self.store_node(lhs_tail.population, flags=msprime.NODE_IS_RE_EVENT)
             self.store_arg_edges(lhs_tail)
             self.store_node(z.population, flags=msprime.NODE_IS_RE_EVENT)
             self.store_arg_edges(z)
+        ret = None
+        if return_heads:
+            if x is None:
```
jeromekelleher: We can simplify this a bit now:
```suggestion
                    x = lhs_tail
                    # Seek back to the head of the x chain
￼	            while x.prev is not None:
￼	                x = x.prev
￼	            ret = x, z
```

jeromekelleher: Not sure how these suggested changes things work, so that won't be valid code. Basically, we don't need the ``if x is None`` check any more.

algorithms.py
andrewkern: this was vestigal... forgot to remove it from some checks i was performing

algorithms.py
andrewkern: so i'm probably missing something here, but i think we still need the `if x is None:` check. From the bit of testing I've done the code is working...

jeromekelleher: No worries, this wasn't very clear. I'll sort it out and merge.

I'll be force pushing to your master branch though, so beware it'll be mucked up. (It would be simpler if you used a topic branch rather than master, as you can just delete these after they're merged)

algorithms.py
jeromekelleher: I got a message on email where you had replied to this I think, but it's not showing up on github. Can I just delete these two loops then?

andrewkern: no we need these-- the first loop goes through and switches the labels of segments. the second loop pops them out and moves them over. could probably condense into one loop but i wasn't sure of a clean pythonic way to do this, so left it as two loops for clarity

algorithms.py
andrewkern: force push away my friend. will use topic branches from now on-- getting used to them with the stdpopsim repo!

jeromekelleher:
OK, merging. Thanks @andrewkern!

PR #711, opened 2019-02-19
Mutation map

### LINKED ISSUE #710: add genomic interval argument to mutate()
It is useful to be able to mutate different regions of the genome at different rates. To allow this, we could either:

1. add a `left` and `right` parameter to `msprime.mutate()`, or
2. add something like a `MutationRateMap` class, with breakpoints and rates

I guess I'm leaning towards the second option.

This is maybe orthogonal to other efforts to better mutations in #553.

jeromekelleher:
Closes #710.

@petrelharp, I started thinking about this and I realise I was over complicating things. Let's just do it the simplest reasonable way first and see how it performs after.

### DIFF
--- a/lib/interval_map.c
+++ b/lib/interval_map.c
@@ -23,6 +23,18 @@
 #include "util.h"
 #include "msprime.h"
 
+void
+interval_map_print_state(interval_map_t *self, FILE *out)
+{
+    size_t j;
+
+    fprintf(out, "interval_map (%p):: size = %d\n", (void *) self, (int) self->size);
+    fprintf(out, "\tsequence_length = %f\n", interval_map_get_sequence_length(self));
+    fprintf(out, "\tindex\tposition\tvalue\n");
+    for (j = 0; j < self->size; j++) {
+        fprintf(out, "\t%d\t%f\t%f\n", (int) j, self->position[j], self->value[j]);
+    }
+}
 
 int MSP_WARN_UNUSED
 interval_map_alloc(interval_map_t *self, size_t size, double *position, double *value)
@@ -65,6 +77,13 @@ out:
     return ret;
 }
 
+int MSP_WARN_UNUSED
+interval_map_alloc_single(interval_map_t *self, double sequence_length, double value)
+{
+    double position[2] = {0, sequence_length};
+    return interval_map_alloc(self, 2, position, &value);
+}
+
 int
 interval_map_free(interval_map_t *self)
 {
@@ -91,15 +110,13 @@ interval_map_get_num_intervals(interval_map_t *self)
     return self->size - 1;
 }
 
-void
-interval_map_print_state(interval_map_t *self, FILE *out)
+size_t
+interval_map_get_index(interval_map_t *self, double x)
 {
-    size_t j;
+    size_t index = tsk_search_sorted(self->position, self->size, x);
 
-    fprintf(out, "interval_map (%p):: size = %d\n", (void *) self, (int) self->size);
-    fprintf(out, "\tsequence_length = %f\n", interval_map_get_sequence_length(self));
-    fprintf(out, "\tindex\tposition\tvalue\n");
-    for (j = 0; j < self->size; j++) {
-        fprintf(out, "\t%d\t%f\t%f\n", (int) j, self->position[j], self->value[j]);
+    if (self->position[index] > x) {
+        index--;
     }
+    return index;
 }
--- a/lib/interval_map.c
+++ b/lib/interval_map.c
@@ -1,0 +1,105 @@
+/*
+** Copyright (C) 2020 University of Oxford
+**
+** This file is part of msprime.
+**
+** msprime is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** msprime is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with msprime.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+
+#include "util.h"
+#include "msprime.h"
+
+
+int MSP_WARN_UNUSED
+interval_map_alloc(interval_map_t *self, size_t size, double *position, double *value)
+{
+    int ret = 0;
+    size_t j;
+
+    memset(self, 0, sizeof(interval_map_t));
+    if (size < 2) {
+        ret = MSP_ERR_INSUFFICIENT_INTERVALS;
+        goto out;
+    }
+    /* Check the framing position */
+    if (position[0] != 0.0) {
+        ret = MSP_ERR_INTERVAL_MAP_START_NON_ZERO;
+        goto out;
+    }
+    self->position = malloc(size * sizeof(*self->position));
+    self->value = malloc(size * sizeof(*self->value));
+    if (self->position == NULL || self->value == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    self->size = size;
+    for (j = 0; j < size; j++) {
+        if (position[j] < 0) {
+            ret = MSP_ERR_NEGATIVE_INTERVAL_POSITION;
+            goto out;
+        }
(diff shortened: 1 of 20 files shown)


tests/test_mutations.py
```diff
@@ -470,3 +470,57 @@ def test_keep_mutation_parent_zero_rate(self):
...
+def mutate_map(tables, mutation_map, seed):
+    """
+    Parallel implementation of the C code which should produce precisely the same
+    results.
+    """
+    rng = msprime.RandomGenerator(seed)
+    time = tables.nodes.time
+    for edge in tables.edges:
+        branch_length = time[edge.parent] - time[edge.child]
+        index = np.searchsorted(mutation_map.position, edge.left)
```
jeromekelleher: This is my suggested implementation @petrelharp, does it look OK to you? Basically we can replace ``searchsorted`` with bsearch and it'll be basically identical after that.

petrelharp:
Gee, looks just right.  Good thinking.

jeromekelleher:
OK, that's most of it done. Just need to update the high-level code and test now.

The only change here is that the map is defined by two equal length arrays now. There's no point in specifying the end value, as it just makes things tricky.

jeromekelleher:
@daniel-goldstein, I'd appreciate a quick look over this please. It's a WIP where I'm refactoring the recombination map class to make use of a ``interval_map_t``. We want the interval_map because there are other values we want to associate with intervals (here, mutation rates; the idea here is to add variable mutation rates along the genome).

What do you think of the basic setup? I think the plan now is to make a high-level Pythonic interface for the IntervalMap class and to give it some useful methods. We then refactor the RecombinationMap class to use this (not quite sure how yet), but maintaining the old interface for compatibility.

I'm also changing my mind about insisting the arrays are the same length in Python - really, the ``values`` array should be one shorter, and the C code should never look at this value.

cc also @grahamgower - I know you were thinking about #902. I think this PR should also close that issue (or at least plan to with a concrete follow up).

lib/util.h
```diff
@@ -78,6 +78,14 @@
 #define MSP_ERR_BAD_PEDIGREE_ID                                     -40
 #define MSP_ERR_BAD_ALPHA                                           -41
 #define MSP_ERR_BAD_TRUNCATION_POINT                                -42
+#define MSP_ERR_BAD_MUTATION_MAP_SIZE                               -43
+#define MSP_ERR_BAD_MUTATION_MAP_POSITION                           -44
+#define MSP_ERR_BAD_MUTATION_MAP_RATE                               -45
```
jeromekelleher: FIXME: some of these are unused now.

grahamgower:
This is very cool. One thing that I had in mind for #902 is that it would be nice for the python interface to do zero memory allocations/copies when getting the `rate` and `position` properties of the RecombinationMap---just return a numpy array view of the underlying memory. I didn't think it through fully due to lack of familiarity with the code, but it probably only needs small changes to memory ownership semantics in the C interface.

jeromekelleher:
> This is very cool.

Thanks!

>  One thing that I had in mind for #902 is that it would be nice for the python interface to do zero memory allocations/copies when getting the rate and position properties of the RecombinationMap---just return a numpy array view of the underlying memory. I didn't think it through fully due to lack of familiarity with the code, but it probably only needs small changes to memory ownership semantics in the C interface.

Ah, yes, now is this tricky. I've thought about doing it before (for table columns in tskit) and ended up giving up and just returning copies. In principle, we should be able to use [this method](https://docs.scipy.org/doc/numpy/reference/c-api.array.html#c.PyArray_SimpleNewFromData) and set the base member to the IntervalMap instance so that ref counting works properly. There are some worries about then changing the shape of the array (or something) and then breaking underlying C code, but maybe it can be set to a read-only-ish mode. It would  be handy to directly access the underlying arrays when manipulating these maps...

I might have another go at it, it's good idea.

jeromekelleher:
Ah yes, now I remember why I didn't do this for Table columns. This is because they are dynamic, and the column memory can be realloced, therefore making the memory pointed to by the numpy array invalid. We don't have this problem here though.

benjeffery:
> The column memory can be realloced, therefore making the memory pointed to by the numpy array invalid.
Could you get round this by supporting a callback that the C API calls with the new memory address when ever it re-allocates? Would clearly add complexity - do you have a handle on how much copies are a perf issue?

jeromekelleher:
> Could you get round this by supporting a callback that the C API calls with the new memory address when ever it re-allocates? Would clearly add complexity - do you have a handle on how much copies are a perf issue?

Copies are totally not a perf issue, which is mainly why I decided to not worry about zero-copy semantics for this stuff. It's more about nice APIs really than perf. We'd like to be able to change the data in place sometimes, and modifying the array is much nicer.

The realloc stuff isn't an issue here though as the size is fixed, that's a tskit Tables API concern.

jeromekelleher:
I started messing around with zero copy stuff there and realised there's actually not much point in this case. The gains are pretty minimal and we'd have to do a lot of work to make sure that (e.g.) the user couldn't modify the position array while it was in use by the mutation generator. Taking copies is safe and simple. Having a good high-level API is more important here than a super-efficient low-level one.

jeromekelleher:
I think the thing to do here is to merge this PR as soon as possible and to follow up with more work on improving the high-level API and testing out the mutation mapping code. There's quite a big diff here, so it would be good to get it merged now so we can keep other PRs going through.

I've added #920 as a starting follow-up where we add the create the high-level interval map tools and change the semantics to something more sensible.

daniel-goldstein:
> I'd appreciate a quick look over this please. It's a WIP where I'm refactoring the recombination map class to make use of a interval_map_t. We want the interval_map because there are other values we want to associate with intervals (here, mutation rates; the idea here is to add variable mutation rates along the genome).

@jeromekelleher Sounds good to me. I wonder though if `interval_map` is a bit too broad to benefit fully from the abstraction for recombination/mutation maps.

It makes sense the way `recomb_map` is currently refactored because there's no promise that `interval_map.value` is a *rate*. But a mutation map would also treat it as such and I think would then duplicate some core functionality like expected # of mutations/recombinations within interval `[start, end)` or up to position `x`. I don't know much about how mutation maps are used differently than recombination maps, but to me they seem like they could be thin skins around an `interval_rate_map`, with recombination map doing a little extra work surrounding loci/discrete space etc. Something like an `interval_rate_map` could then manage position, rate, and cumulative_value, and could give the useful method of position-at-value.

> I'm also changing my mind about insisting the arrays are the same length in Python - really, the values array should be one shorter, and the C code should never look at this value.

I am also a fan of nixing the last entry on the values array. I think it's no more mental overhead to make sure it's one shorter than make sure they're the same length *and* ends in zero.

> I started messing around with zero copy stuff there and realised there's actually not much point in this case. The gains are pretty minimal and we'd have to do a lot of work to make sure that (e.g.) the user couldn't modify the position array while it was in use by the mutation generator. Taking copies is safe and simple. Having a good high-level API is more important here than a super-efficient low-level one.

Worth noting that the C simulator copies its `recomb_map` handed to it from python, and does some mutation translating model rates, so we'd have to be careful with views into the arrays. I haven't seen any perf reasons not to copy, so I think this sounds like the best move right now.

jeromekelleher:
Thanks @daniel-goldstein, I was wondering myself about an ``interval_rate_map``. Let's follow this up in #920, once this is merged.

petrelharp:
Wow, this is great.  And: Yes!! Please make values one shorter!!! That eliminates uncertainty about which number goes with which interval, a serious source of bugs.

lib/interval_map.c
```diff
@@ -0,0 +1,122 @@
...
+#include "msprime.h"
+
+void
+interval_map_print_state(interval_map_t *self, FILE *out)
+{
+    size_t j;
+
+    fprintf(out, "interval_map (%p):: size = %d\n", (void *) self, (int) self->size);
+    fprintf(out, "\tsequence_length = %f\n", interval_map_get_sequence_length(self));
+    fprintf(out, "\tindex\tposition\tvalue\n");
```
petrelharp: Shouldn't this be left, right, value?

lib/interval_map.c
```diff
@@ -0,0 +1,122 @@
...
+}
+
+double
+interval_map_get_sequence_length(interval_map_t *self)
+{
+    return self->position[self->size - 1];
+}
+
+size_t
+interval_map_get_size(interval_map_t *self)
```
petrelharp: Maybe `num_breakpoints` instead of size?

jeromekelleher: Good idea. Will follow up in #920

jeromekelleher:
OK, thanks @petrelharp. I'm going to merge this and then pick it up again in #920 to tidy up the rough edges on the interval map stuff.

PR #716, opened 2019-03-03
trailing zero recombination rate fix

### LINKED ISSUE #715: nothing happens in terminal zero-recomb segments
It looks like if the last segments of chromosome have zero recombination rate, then the chromosome *functionally* acts as if it ended before those segments, although the `sequence_length` attribute is still correct.

Here is a MWE (modified from @jradrion's in #704):
```
import msprime
recomb_map = msprime.RecombinationMap(positions = [0, 100, 200, 300], 
             rates = [0.01, 0.00, 0.0, 0.0])

ts=msprime.simulate(
        sample_size=50,
        Ne=100,
        mutation_rate=0.01,
        recombination_map=recomb_map)

print("Recombination_map.get_positions():", recomb_map.get_positions())
print("Sequence length:", ts.sequence_length)
print("Rightmost edge end:", max(ts.tables.edges.right))
print("Rightmost SNP:", ts.site(ts.num_sites - 1).position)
```
Even though `sequence_length` is (correctly) 300, the end of the rightmost edge and the position of the rightmost SNP are both 100.

And, I think I know what the problem is.
1. There aren't any mutations on the righthand end because no edges extend that far, and mutations are added to the edges.
2. The edges don't actually extend to the end because (I'm guessing) when we map physical position -> discrete genetic map -> back to physical positions it's taking the end of the chromosome (300) to 100.

Indeed, that's what it's doing:
```
>>> recomb_map.genetic_to_physical(recomb_map.physical_to_genetic(300))
100.0
```

petrelharp:
This hopefully fixes #715 (and also better documents `read_hapmap()`).

I've done it by just inserting a check into `recomb_map_genetic_to_phys()`, which seems the cleanest way to do it.  I've inserted a new test, also.  

But, I'm having trouble testing this right now, due to some tskit conflict.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -758,19 +758,29 @@ class RecombinationMap(object):
     @classmethod
     def read_hapmap(cls, filename):
         """
-        Parses the specified file in HapMap format. These files must contain
-        a single header line (which is ignored), and then each subsequent
-        line denotes a position/rate pair. Positions are in units of bases,
-        and recombination rates in centimorgans/Megabase. The first column
-        in this file is ignored, as are subsequence columns after the
-        Position and Rate. A sample of this format is as follows::
+        Parses the specified file in HapMap format. These files must be
+        white-space-delimited, and contain a single header line (which is
+        ignored), and then each subsequent line contains the starting position
+        and recombination rate for the segment from that position (inclusive)
+        to the starting position on the next line (exclusive). Starting
+        positions of each segment are given in units of bases, and
+        recombination rates in centimorgans/Megabase. The first column in this
+        file is ignored, as are additional columns after the third (Position is
+        assumed to be the second column, and Rate is assumed to be the third).
+        If the first starting position is not equal to zero, then a
+        zero-recombination region is inserted at the start of the chromosome.
+
+        A sample of this format is as follows::
 
             Chromosome	Position(bp)	Rate(cM/Mb)	Map(cM)
-            chr1	55550	2.981822	0.000000
-            chr1	82571	2.082414	0.080572
-            chr1	88169	2.081358	0.092229
-            chr1	254996	3.354927	0.439456
-            chr1	564598	2.887498	1.478148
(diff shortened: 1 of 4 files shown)


petrelharp:
I'll fix up those tests when I fix up my testing set-up.

petrelharp:
Well, if you all agree with my fix, then this is good to go, closing #716.

jeromekelleher:
Superb, thanks @petrelharp!

msprime/simulations.py
```diff
@@ -758,19 +758,29 @@ def uniform_map(cls, length, rate, num_loci=None):
...
-            chr1	254996	3.354927	0.439456
-            chr1	564598	2.887498	1.478148
+            chr1	55550	        2.981822	0.000000
+            chr1	82571	        2.082414	0.080572
+            chr1	88169	        2.081358	0.092229
+            chr1	254996	        3.354927	0.439456
+            chr1	564598	        2.887498	1.478148
+            ...
+            chr1	182973428	1.036438	122.832331
+            chr1	183030013	0.000000	124.482178
```
andrewkern: @petrelharp i don't understand where the Map (cM) position of the last entry (124.48) is coming from. Shouldn't it be more like ((183.03-182.97)*1.036) + 122.83 = 122.8922

petrelharp: whoops; I forgot to fix that. =)

jeromekelleher: > @petrelharp i don't understand where the Map (cM) position of the last entry (124.48) is coming from. Shouldn't it be more like ((183.03-182.97)*1.036) + 122.83 = 122.8922

Well spotted @andrewkern --- who wrote this garbage?! (Oh, wait...)

petrelharp: that was me, actually - I added those last few lines. =)

jeromekelleher: Doh! Well that's what happens when I try to be funny.

PR #726, opened 2019-03-16
added two helpful functions to the DemographyDebugger class

andrewkern:
I wrote a function that will be helpful for tracing the population size histories of samples drawn from individual populations moving back in time. The idea here is that we may want to know the population size experienced by a given sample as it goes through both population size changes and population mergers (back in time). 

To go along with this a wrote a simple helper function that simply outputs the epoch start times for a given demographic model. This may be deemed superfluous.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -26,9 +26,9 @@ import math
 import random
 import sys
 import os
-import numpy
 
 import tskit
+import numpy as np
 
 from . import provenance
 import _msprime
@@ -1251,6 +1251,7 @@ class DemographyDebugger(object):
 
     def _make_epochs(self, simulator, demographic_events):
         self.epochs = []
+        self.num_populations = simulator.num_populations
         ll_sim = simulator.create_ll_instance()
         N = simulator.num_populations
         start_time = 0
@@ -1348,12 +1349,11 @@ class DemographyDebugger(object):
     @property
     def population_size_history(self):
         """
-        Returns a matrix of population sizes seen
-        by samples from each population at each
-        time defined by the epochs.
+        Returns a (num_pops, num_epochs) numpy array giving the starting population size
+        for each population in each epoch.
         """
         num_pops = len(self.epochs[0].populations)
-        pop_size = numpy.zeros((num_pops, len(self.epochs)))
+        pop_size = np.zeros((num_pops, len(self.epochs)))
         for j, epoch in enumerate(self.epochs):
             for k, pop in enumerate(epoch.populations):
                 pop_size[k, j] = epoch.populations[k].start_size
@@ -1362,7 +1362,13 @@ class DemographyDebugger(object):
     @property
     def epoch_times(self):
         """
-        Returns array of epoch times defined by
-        the demographic model
+        Returns array of epoch times defined by the demographic model
         """
-        return numpy.array([x.start_time for x in self.epochs])
+        return np.array([x.start_time for x in self.epochs])
+
+    @property
+    def num_epochs(self):
+        """
+        Returns the number of epochs defined by the demographic model.
+        """
+        return len(self.epochs)
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -26,6 +26,7 @@ import math
 import random
 import sys
 import os
+import numpy
 
 import tskit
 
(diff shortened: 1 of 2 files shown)


andrewkern:
codecov is bitching about C but I only changed a single python file... what gives?

msprime/simulations.py
```diff
@@ -1343,3 +1344,33 @@ def print_history(self, output=sys.stdout):
...
+        time defined by the epochs.
+        """
+        num_pops = len(self.epochs[0].populations)
+        pop_size = numpy.zeros((num_pops, len(self.epochs)))
+        popn_set = [[x] for x in range(num_pops)]
+        for j, epoch in enumerate(self.epochs):
+            for k, pop in enumerate(epoch.populations):
+                idx = [k in x for x in popn_set].index(True)
+                pop_size[k, j] = epoch.populations[idx].start_size
+            for event in epoch.demographic_events:
```
jeromekelleher: Hmmm, I'm not convinced about this. The population size isn't actually zero, it's just that there aren't any samples in there. This seems like an unnecessary complication to me. (If you want to set the population size of zero you can do this in the model explicitly with an extra event)

msprime/simulations.py
```diff
@@ -1343,3 +1344,33 @@ def print_history(self, output=sys.stdout):
             print("=" * len(s), file=output)
             self._print_populations(epoch, output)
             print(file=output)
+
+    def popn_size_history(self):
```
jeromekelleher: We're not using the 'popn' abbreviation anywhere else, so it'll be confusing to have it here. Can you change to ``population_size_history`` please?

msprime/simulations.py
```diff
@@ -1343,3 +1344,33 @@ def print_history(self, output=sys.stdout):
...
+                    print(f"epoch {j} source: {event.source} dest: \
+                          {event.dest}")
+                    popn_set[event.dest] = popn_set[event.dest] + \
+                        popn_set[event.source]
+                    popn_set[event.source] = []
+        return pop_size
+
+    def epoch_times(self):
+        """
+        Helper fxn. Returns appay of epoch times defined by
```
jeromekelleher: The first sentence here is unnecessary (plus, why the abbreviation?). Spelling errorr also: "appay"

jeromekelleher:
> codecov is bitching about C but I only changed a single python file... what gives?

Codecov is complaining because the project level C coverage is lower than the target (basically because the selection model stuff hasn't been full implemented and tested). I've dropped the coverage target in #727 to avoid this confusion.

msprime/simulations.py
jeromekelleher: Also, probably both of these functions would be better as properties (unless we can imagine passing parameters?)

msprime/simulations.py
andrewkern: not sure i follow here- are you referring to the fact that i'm initializing the population sizes to zero in the pop_size array? Or to when i'm iterating in the loop on 1358 over populations present in the given epoch?

msprime/simulations.py
andrewkern: that's what i get for coding too late at night...

andrewkern:
well I pushed some asserts of the new functions (now written as class properties) and codecov still is borking. not enough? check out my changes in `tests/test_demography.py`

msprime/simulations.py
jeromekelleher: No, I'm objecting to looping over the demographic events and setting the population size to zero if there was a full MassMigration. This seems like an unnecessary complication to me, and probably isn't safe in any case --- what if there was still migration into this deme?

jeromekelleher:
> well I pushed some asserts of the new functions (now written as class properties) and codecov still is borking. not enough? check out my changes in tests/test_demography.py

Asserts look good, you're just not touching the code where you're special-casing for MassMigrations, which I'm pretty sure needs to be deleted in any case. You'll need to rebase to sort out the C-level stuff.

msprime/simulations.py
andrewkern: ah I see. No population sizes in the history object are actually getting set to zero except at initialization as I'm setting by `epoch.populations[idx].start_size` which should always be positive. I am however ignoring the case where there has been a Mass Migration with proportion = 1 (i.e. a back in time pop merger) followed by subsequent migration back in to the source population. This case would make sense biologically, but I can see that the code would not be safe.

Thoughts about a better way to do this? The goal is slightly strange-- I want to know the population size history experienced by samples tracing back in time through population merger events.

andrewkern:
I should be rebasing against upstream/master or a different branch? i can write more tests to hit that bit of code

jeromekelleher:
> I should be rebasing against upstream/master or a different branch? i can write more tests to hit that bit of code

upstream/master.

msprime/simulations.py
jeromekelleher: If we adopt the convention that any 'closed' population should have its population size set to zero by the user, then there's no need to have any special cases in here. I'm probably misinterpreting the goal here though. My understanding here is that the population size of the deme in question is effectively zero because there was a population merger, and we're trying to show this. Is this right or am I misunderstanding?

andrewkern:
should be all set here. I pulled out those conditionals about Mass Migration events and will pull that into the stdpopsim/analysis side of things

jeromekelleher:
Looks good, thanks @andrewkern. I've added a couple of small bits to the API and some more tests. Will merge once CI passes.

jeromekelleher:
Merging --- sorry, totally forgot about this @andrewkern!

PR #729, opened 2019-03-18
added mean recombination rate

jgallowa07:
I could probably make better tests, any thoughts?

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -26,6 +26,7 @@ import math
 import random
 import sys
 import os
+import numpy as np
 
 import tskit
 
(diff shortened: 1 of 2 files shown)


msprime/simulations.py
```diff
@@ -26,6 +26,7 @@
 import random
 import sys
 import os
+import numpy
```
jeromekelleher: We're using ``numpy as np`` elsewhere, so may as well follow this convention.

msprime/simulations.py
```diff
@@ -804,6 +805,24 @@ def read_hapmap(cls, filename):
             f.close()
         return cls(positions, rates)
```
jeromekelleher: We can make this a property I think, ``mean_recombination_rate``.

msprime/simulations.py
```diff
@@ -804,6 +805,24 @@ def read_hapmap(cls, filename):
...
 
+    def get_mean_recombination_rate(self):
+        """
+        Use numpy to compute the weighted mean recombination rate
```
jeromekelleher: The documentation for a method should tell you *what* it does, not *how* it does it. "Returns the..."

jgallowa07:
@jeromekelleher Awesome, should be all cleaned up now.

jeromekelleher:
Very nice, thanks @jgallowa07!

jgallowa07:
np @jeromekelleher, closing and deleting the branch.

PR #735, opened 2019-04-16
Update tutorial.rst

mufernando:
What msprime calls a genotype matrix is actually a HaplotypeArray in scikit-allel, so I just changed the tutorial to make that clear.


jeromekelleher:
Great, thanks @mufernando!

PR #737, opened 2019-04-26
Add demographic tests in C for DTWF - fixes #626

### LINKED ISSUE #626: No C test coverage for DTWF demographic events
Add simple test to simulation_tests.c

DomNelson:
Added DTWF to existing demographic tests, but still not sure about some things

Some fiddling is due to a check in DTWF that events occur exactly on (integer-valued) time, which is done to avoid `self->time = self->next_demographic_event->time` in `msp_apply_demographic_events()` getting generations out of sync.

If I have any sample times before demographic events I get  `Assertion "overlaps[k] == count" failed.` which I'm currently trying to understand - could be related to the way historical samples are handled in DTWF?

### DIFF
--- a/lib/tests/tests.c
+++ b/lib/tests/tests.c
@@ -619,7 +619,7 @@ static void
 test_demographic_events(void)
 {
     int ret;
-    uint32_t j, k;
+    uint32_t j, k, model;
     uint32_t n = 10;
     uint32_t m = 10;
     sample_t *samples = malloc(n * sizeof(sample_t));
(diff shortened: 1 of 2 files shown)


lib/tests/tests.c
```diff
@@ -816,15 +856,31 @@ test_demographic_events_start_time(void)
...
+            CU_ASSERT_EQUAL(ret, 0);
+            ret = msp_set_simulation_model_dtwf(msp, 1);
+            CU_ASSERT_EQUAL(ret, 0);
+        }
+
+        ret = msp_initialise(msp);
+        CU_ASSERT_EQUAL(ret, MSP_ERR_BAD_DEMOGRAPHIC_EVENT_TIME);
+    }
 
     ret = msp_free(msp);
```
jeromekelleher: Memory leak here, need to have this within the loop

lib/tests/tests.c
```diff
@@ -638,156 +638,195 @@ test_demographic_events(void)
...
-    /* Zero or negative population sizes are not allowed */
-    ret = msp_set_population_configuration(&msp, 0, -1, 0);
-    CU_ASSERT_EQUAL(ret, MSP_ERR_BAD_PARAM_VALUE);
-    ret = msp_set_population_configuration(&msp, 0, 0, 0);
-    CU_ASSERT_EQUAL(ret, MSP_ERR_BAD_PARAM_VALUE);
+    for (model = 0; model < 2; model++) {
+        for (j = 0; j < n; j++) {
+            samples[j].time = j;
+            /* Ensures DTWF sample times are after demographic events */
+            if (model == 1) {
```
jeromekelleher: I think it's a bug if we need to do this --- there's nothing stopping these happening at the same time, right? This must be a slight issue with the logic in the main loop. Are these events mutually exclusive or something?

DomNelson:
Ok there seem to have been two issues:

- Historical samples weren't being added in `msp_debug_demography()`, which led to an overlap assertion error if demographic events occurred after any historical samples (in both models)

- Historical samples in DTWF sims were being added before parent selection, which meant they could choose parents of their own generation, leading to `MSP_ERR_TIME_TRAVEL`

But now it looks like CI is failing! Didn't see those errors locally - I'll investigate.

jeromekelleher:
Good work @DomNelson, I think these are the root of some of the problems I was hitting as well. Let me know if I can help.

DomNelson:
Whew! I'd made a mess of the rebase I think... looks better now

jeromekelleher:
Thanks @DomNelson, I'll pick this up first thing tomorrow and get it merged.

jeromekelleher:
I've updated this a little to fix some memory leaks in the tests (it's good to run these through valgrind every now and again). I removed the assertion requiring event times to be integers, as (a) having events that fall in between generations seems harmless to me; and (b) if we did want to prevent this, we'd need to throw an error rather than bomb out.

Is there a good reason for requiring events to happen at integer time?

PR #742, opened 2019-05-01
Add pop size and coalescent rate trajectories to demography debugger

jgallowa07:
These changes add 3 new methods to `msprime.DemographyDebugger` 

1: `pop_size_and_migration_at_t(self, t)` - Which returns all population sizes and the migration matrix at any time (ago), t

2: `population_size_trajectory(self, end, num_steps=10)` - This function allows the user to specify an end time and a number of steps to take between (and including) - [0,end]. The function will return the steps, as well as each population's respective population size at each step.

3: `coalescence_rate_trajectory(self, end, num_samples, num_steps=10, min_pop_size=1)` - The kicker (brought to light and solved by @petrelharp), This function follows the same regime as `population_size_trajectory` except it instead computes the ground truth coalescent rates for populations with multiple locations and migration. `num_samples` should be a list the same length as the number of populations, and `min_pop_size` is there for math reasons (read the doc string for more). Return the steps, respective coalescent rates, along with with the sum of a matrix `P` which represents the total probability that two lineages have not yet coalesced. 

There are currently only two unit tests for this, and I am working on writing more. In the meantime, feel free to make suggestions. 

Below is what we are computing. 
[2.HEIC.pdf](https://github.com/tskit-dev/msprime/files/3136185/2.HEIC.pdf)
[3.HEIC.pdf](https://github.com/tskit-dev/msprime/files/3136186/3.HEIC.pdf)
[1.HEIC.pdf](https://github.com/tskit-dev/msprime/files/3136187/1.HEIC.pdf)

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -26,6 +26,7 @@ import math
 import random
 import sys
 import os
+import warnings
 
 import tskit
 import numpy as np
@@ -1345,7 +1346,6 @@ def _matrix_exponential(A):
     Note: this is not a general purpose method and is only intended for use within
     msprime.
     """
-
     d, Y = np.linalg.eig(A)
     Yinv = np.linalg.pinv(Y)
     D = np.diag(np.exp(d))
@@ -1492,13 +1492,12 @@ class DemographyDebugger(object):
     def population_size_trajectory(self, steps):
         """
         This function returns an array of per-population effective population sizes,
-        as defined by the demographic model. These are just the `population_size`
+        as defined by the demographic model. These are the `initial_size`
         parameters of the model, modified by any population growth rates.
         The sizes are computed at the time points given by `steps`.
 
         :param list steps: List of times ago at which the population
             size will be computed.
-
         :return: Returns a numpy array of population sizes, with one column per
             population, whose [i,j]th entry is the size of population
             j at time steps[i] ago.
@@ -1506,63 +1505,58 @@ class DemographyDebugger(object):
         num_pops = self.num_populations
         N_t = np.zeros([len(steps), num_pops])
         for j, t in enumerate(steps):
-            N, _ = self.pop_size_and_migration_at_t(t)
+            N, _ = self._pop_size_and_migration_at_t(t)
             N_t[j] = N
         return N_t
 
-    def coalescence_rate_trajectory(self, steps, num_samples, min_pop_size=1,
-                                    double_step_validation=True):
+    def coalescence_rate_trajectory(
+            self, steps, num_samples, min_pop_size=1, double_step_validation=True):
         """
-        This function will calculate the mean coalescence rates and
-        proportions of uncoalesced lineages between the
-        lineages of the sample configuration provided in `num_samples`,
-        at each of the times ago listed by steps, in this demographic model.
-        The coalescence rate at time t in the past is the average rate
-        of coalescence of as-yet-uncoalesed lineages, computed as follows:
-        let p(t) be the probability that the lineages of a randomly
-        chosen pair of samples has not yet coalesced by time t,
-        let p(z,t) be the probability that the lineages of a randomly
-        chosen pair of samples has not yet coalesced by time t *and* are
-        both in population z, and let N(z,t) be the diploid effective
-        population size of population z at time t. Then the mean coalescence
-        rate at time t is r(t) = (sum_z p(z,t) / (2 * N(z,t) ) / p(t).
-        The computation is done approximating population size trajetories
-        with piecewise constant trajectories between each of the steps.
-        For this to be accurate, the distance between the steps
-        must be small enough so that (a) short epochs
-        (e.g., bottlenecks) are not missed, and (b) populations do not
-        change in size too much over that time, if they are growing or shrinking.
-        "This function optionally provides a simple check of this approximation
-        by recomputing the coalescence rates on a grid of steps twice as fine and
-        throwing a warning if the resulting values do not match to a relative
-        tolerance of 0.001.
+        This function will calculate the mean coalescence rates and proportions
+        of uncoalesced lineages between the lineages of the sample
+        configuration provided in `num_samples`, at each of the times ago
+        listed by steps, in this demographic model. The coalescence rate at
+        time t in the past is the average rate of coalescence of
+        as-yet-uncoalesed lineages, computed as follows: let :math:`p(t)` be
+        the probability that the lineages of a randomly chosen pair of samples
+        has not yet coalesced by time :math:`t`, let :math:`p(z,t)` be the
+        probability that the lineages of a randomly chosen pair of samples has
+        not yet coalesced by time :math:`t` *and* are both in population
+        :math:`z`, and let :math:`N(z,t)` be the diploid effective population
+        size of population :math:`z` at time :math:`t`. Then the mean
+        coalescence rate at time :math:`t` is :math:`r(t) = (\\sum_z p(z,t) /
+        (2 * N(z,t)) / p(t)`.
+
+        The computation is done by approximating population size trajectories
+        with piecewise constant trajectories between each of the steps. For
+        this to be accurate, the distance between the steps must be small
+        enough so that (a) short epochs (e.g., bottlenecks) are not missed, and
+        (b) populations do not change in size too much over that time, if they
+        are growing or shrinking. This function optionally provides a simple
+        check of this approximation by recomputing the coalescence rates on a
+        grid of steps twice as fine and throwing a warning if the resulting
+        values do not match to a relative tolerance of 0.001.
 
         :param list steps: The times ago at which coalescence rates will be computed.
-
-        :param list num_samples: is a list of the same length as the number
+        :param list num_samples: A list of the same length as the number
             of populations, so that `num_samples[j]` is the number of sampled
             chromosomes in subpopulation `j`.
-
-        :param int min_pop_size: is the smallest allowed population size during
+        :param int min_pop_size: The smallest allowed population size during
             computation of coalescent rates (i.e., coalescence rates are actually
             1 / (2 * max(min_pop_size, N(z,t))). Spurious very small population sizes
             can occur in models where populations grow exponentially but are unused
             before some time in the past, and lead to floating point error.
             This should be set to a value smaller than the smallest
             desired population size in the model.
-
         :param bool double_step_validation: Whether to perform the check that
             step sizes are sufficiently small, as described above. This is highly
             recommended, and will take at most four times the computation.
-
-        :return: a tuple of arrays whose jth elements, respectively, are the
+        :return: A tuple of arrays whose jth elements, respectively, are the
             coalescence rate at the jth time point (denoted r(t[j]) above),
             and the probablility that a randomly chosen pair of lineages has
             not yet coalesced (denoted p(t[j]) above).
-
-        :rtype: class:'tuple'
+        :rtype: (numpy.array, numpy.array)
         """
-
         num_pops = self.num_populations
         if not len(num_samples) == num_pops:
             raise ValueError(
@@ -1585,18 +1579,13 @@ class DemographyDebugger(object):
             r_prediction_close = np.allclose(r[1:], rd[::2], rtol=1e-3)
             p_prediction_close = np.allclose(p_t[1:], p_td[::2], rtol=1e-3)
             if not (r_prediction_close and p_prediction_close):
-                # should we use the `warning` package?
-                raise UserWarning(
+                warnings.warn(
                     "Doubling the number of steps has resulted in different"
                     " predictions, please re-run with smaller step sizes to ensure"
                     " numerical accuracy.")
         return r, p_t
 
     def _calculate_coalescence_rate_trajectory(self, steps, num_samples, min_pop_size):
-        """
-        Private function is used for doing all the calculation for
-        coalescence_rate_trajectory.
-        """
         num_pops = self.num_populations
         P = np.zeros([num_pops**2, num_pops**2])
         IA = np.array(range(num_pops**2)).reshape([num_pops, num_pops])
@@ -1622,7 +1611,7 @@ class DemographyDebugger(object):
         for j, time in enumerate(steps_b):
             dt = time - prev
             prev = time
-            N, M = self.pop_size_and_migration_at_t(time)
+            N, M = self._pop_size_and_migration_at_t(time)
             C = np.zeros([num_pops**2, num_pops**2])
             for idx in range(num_pops):
                 C[IA[idx, idx], IA[idx, idx]] = 1 / (2 * max(min_pop_size, N[idx]))
@@ -1656,13 +1645,16 @@ class DemographyDebugger(object):
         p_t = np.delete(p_t, diff_mask)
         return r, p_t
 
-    def pop_size_and_migration_at_t(self, t):
+    def _pop_size_and_migration_at_t(self, t):
         """
         Returns a tuple (N, M) of population sizes (N) and migration rates (M) at
         time t ago.
 
-        :param float t: The time ago.
+        Note: this isn't part of the external API as it is be better to provide
+        separate methods to access the population size and migration rates, and
+        needing both together is specialised for internal calculations.
 
+        :param float t: The time ago.
         :return: A tuple of arrays, of the same form as the population sizes and
             migration rate arrays of the demographic model.
         """
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -1338,6 +1338,21 @@ class Epoch(object):
         return repr(self.__dict__)
 
 
+def _matrix_exponential(A):
+    """
+    Returns the matrix exponential of A.
+    https://en.wikipedia.org/wiki/Matrix_exponential
+    Note: this is not a general purpose method and is only intended for use within
+    msprime.
+    """
+
+    d, Y = np.linalg.eig(A)
+    Yinv = np.linalg.pinv(Y)
+    D = np.diag(np.exp(d))
+    B = np.matmul(Y, np.matmul(D, Yinv))
+    return B
+
+
 class DemographyDebugger(object):
     """
     A class to facilitate debugging of population parameters and migration
@@ -1345,7 +1360,10 @@ class DemographyDebugger(object):
     """
     def __init__(
             self, Ne=1, population_configurations=None, migration_matrix=None,
-            demographic_events=[], model="hudson"):
+            demographic_events=None, model='hudson'):
+        if demographic_events is None:
+            demographic_events = []
+        self.demographic_events = demographic_events
         self._precision = 3
         # Make sure that we have a sample size of at least 2 so that we can
         # initialise the simulator.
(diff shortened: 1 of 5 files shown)


msprime/simulations.py
```diff
@@ -26,6 +26,8 @@
 import random
 import sys
 import os
+import scipy.linalg
```
petrelharp: Note that this brings in scipy as a prerequisite. We could write our own expm function (not too bad for this nicely behaved case) to avoid this.

msprime/simulations.py
```diff
@@ -1365,6 +1368,139 @@ def print_history(self, output=sys.stdout):
             self._print_populations(epoch, output)
             print(file=output)
 
+    def population_size_trajectory(self, end, num_steps=10):
+        """
+        This function computes the defined effective population size
```
petrelharp: To disambiguate this from the effective population size that is 1/(coal rate), how about:
```
        This function returns an array of per-population effective population sizes,
        as defined by the demographic model. These are just the `population_size`
        parameters of the model, modified by any population growth rates.
        The sizes are computed at a regular grid of time steps, between time 0 and `end`
        units of time ago.

        Returns a tuple where the first value
        is an array giving the times at which population sizes have been computed,
        the second is a num_steps-by-num_pops ndarray containing the effective
        population sizes for each subpopulation individually, at the corresponding step.
```

msprime/simulations.py
```diff
@@ -1365,6 +1368,139 @@ def print_history(self, output=sys.stdout):
...
+        steps = np.linspace(0, end, num_steps)
+        for j, t in enumerate(steps):
+            N = self.pop_size_and_migration_at_t(t)["population_sizes"]
+            N_t[j] = N
+        return steps, N_t
+
+    def coalescence_rate_trajectory(self, end, num_samples,
+                                    num_steps=10, min_pop_size=1):
+        """
+        This function will calculate the ground truth
```
petrelharp: Edits:
```
        This function will calculate the mean coalescence rates between the lineages of the 
        sample configuration provided in `num_samples`, at a regularly spaced set of times,
        in this demographic model. The coalescence rate at time t in the past is the average rate of coalescence of 
        as-yet-uncoalesed lineages, computed as follows: let p(t) be the probability that the
        lineages of a randomly chosen pair of samples has not yet coalesced by time t, 
        let p(z,t) be the probability that the lineages of a randomly chosen pair of samples 
        has not yet coalesced by time t *and* are both in population z, and let N(z,t) be the
        diploid effective population size of population z at time t. Then the mean coalescence
        rate at time t is r(t) = (sum_z p(z,t) / (2 * N(z,t) ) / p(t) .

        The computation is done approximating population size trajetories with piecewise constant
        trajectories, in steps of size dt = end / num_steps. For this to be accurate, dt must be small 
        enough so that (a) short epochs (e.g., bottlenecks) are not missed, and (b) populations do not
        change in size too much over that time, if they are growing or shrinking. If in doubt, check that
        results are simliar after doubling `num_steps`.

        `num_samples` is a list of the same length as the number of populations, so that `num_samples[j]` is the number of sampled chromosomes in subpopulation `j`.
        `num_steps` determines at how many time points the coalescence rate is computed.
        `min_pop_size` is the smallest allowed population size during computation of coalescent rates (i.e.,
             coalescence rates are actually 1 / 2 * max(min_pop_size, N(z,t))). Spurious very small population sizes
             can occurIn models where populations grow exponentially but are unused before some time in the past,
             and lead to floating point error. This should be set to a value smaller than the smallest desired population size in the model.

        Returns a tuple where the first value
        is an array of time points, the second is an array whose jth element is the coalescence rate at the jth time
        point (denoted r(t[j]) above), and the third is the probablility that a randomly pair of lineages
        has not yet coalesced (denoted p(t[j]) above).
```

msprime/simulations.py
```diff
@@ -1365,6 +1368,139 @@ def print_history(self, output=sys.stdout):
...
+        for j in range(num_steps):
+            N_M = self.pop_size_and_migration_at_t(steps[j])
+            N = N_M["population_sizes"]
+            M = copy.deepcopy(N_M["migration_matrix"])
+            C = np.zeros([num_pops**2, num_pops**2])
+            for idx in range(num_pops):
+                C[IA[idx, idx], IA[idx, idx]] = 1 / (2 * max(min_pop_size, N[idx]))
+            for idx, row in enumerate(M):
+                M[idx][idx] = -1 * sum(row)
+            G = (np.kron(M, Identity) + np.kron(Identity, M)) - C
```
petrelharp: we could skip the deepcopy (trying to avoid extra dependencies here) by doing the following thing instead of lines 1457-1459:
```
             dM = np.diag(sum(M))
             G = (np.kron(M - dM, Identity) + np.kron(Identity, M - dM)) - C
```

petrelharp:
I think we just need a test using a population growth rate: maybe a single population, two epochs with different growth rates, check that pop size and 1/coal rate match.

msprime/simulations.py
andrewkern: super cool!!!

quick comment. the phrase "...in models where populations grow exponentially but are unused..."-- do you mean unused in the sense that neither of the lineages has been able to migrate into the focal population?

petrelharp: I mean unused like there was a mass migration event that moved everyone out of that population already, and if you drew the population diagram it wouldn't even be there. We should probably put in an example. Like if in forwards time, population 0 split into populations 0 and 1, a thousand years ago; then since msprime always has the same number of pops around, older than 1000 years ago pop 1 would be there, but be 'unused'.

msprime/simulations.py
jgallowa07: oh, right, much easier

msprime/simulations.py
```diff
@@ -1365,6 +1366,169 @@ def print_history(self, output=sys.stdout):
...
+            j += 1
+        N = self.population_size_history[:, j]
+        for i, pop in enumerate(self.epochs[j].populations):
+            s = t - self.epochs[j].start_time
+            g = pop.growth_rate
+            N[i] *= np.exp(-1 * g * s)
+        return {"population_sizes": N,
+                "migration_matrix": self.epochs[j].migration_matrix}
+
+    def _expm(self, A):
```
petrelharp: nice.

jgallowa07:
I can't seem to get CI to pass, ha. Seems like it's a problem with the docstring of coalescent_rate_trajectories(), Once someone helps with this I'm happy to squash the commits and we should be good to go.

petrelharp:
Those blank lines weren't the problem, it was a lack of indent. Here's a diff that fixes it. 
[docs.txt](https://github.com/tskit-dev/msprime/files/3139206/docs.txt)

msprime/simulations.py
```diff
@@ -1365,6 +1366,167 @@ def print_history(self, output=sys.stdout):
...
+                    if x == a:
+                        S[IA[a, a], IA[a, b]] = S[IA[a, a], IA[b, a]] = p * (1 - p)
+                        S[IA[a, a], IA[b, b]] = p ** 2
+                        S[IA[a, a], IA[a, a]] = (1 - p) ** 2
+                    else:
+                        S[IA[x, a], IA[x, b]] = S[IA[a, x], IA[b, x]] = p
+                        S[IA[x, a], IA[x, a]] = S[IA[a, x], IA[a, x]] = 1 - p
+                P = np.matmul(P, self._expm(ds * G))
+                P = np.matmul(P, S)
+                P = np.matmul(P, self._expm((dt - ds) * G))
```
petrelharp: This is a bug: we're still using the old `G` for the first (dt - ds) of the new epoch. In the Gutenkunst et al model, at time 848, there's a mass migration from 2 -> 1 *and* concurrently all the migration rates to 2 get set to zero, so no lineages can go back there. But, since we don't update `G`, some lineages sneak back into 2.

petrelharp:
Proposal, to fix that bug: drop the requirement that `steps` be regularly spaced, and ensure that every epoch change is one of the entries of `steps` (exactly), like:
```
steps = sorted(list(set(steps + epoch_breaks)))
```

jgallowa07:
Got it, good catch. Should I do this to pop_size_trajectory as well? Otherwise, the equality test doesn't work out so well ...

petrelharp:
Yes, totally.

petrelharp:
Oh, and we should probably remove those extra `steps` from the output?

jgallowa07:
> Oh, and we should probably remove those extra `steps` from the output?

So that `num_steps` stays true to the user? meaning the user will never actually see the calculations we added at `epoch_splits` unless the linear space overlaps?

jgallowa07:
> Got it, good catch. Should I do this to pop_size_trajectory as well? Otherwise, the equality test doesn't work out so well ...

Actually, we should leave `pop_size_trajectory` alone if we're only returning the original linspace

msprime/simulations.py
```diff
@@ -1488,13 +1494,19 @@ def coalescence_rate_trajectory(self, end, num_samples,
                     else:
                         S[IA[x, a], IA[x, b]] = S[IA[a, x], IA[b, x]] = p
                         S[IA[x, a], IA[x, a]] = S[IA[a, x], IA[a, x]] = 1 - p
-                P = np.matmul(P, self._expm(ds * G))
+                P = np.matmul(P, self._expm(G))
```
petrelharp: this line should not be there

msprime/simulations.py
```diff
@@ -1488,13 +1494,19 @@ def coalescence_rate_trajectory(self, end, num_samples,
...
                 P = np.matmul(P, S)
-                P = np.matmul(P, self._expm((dt - ds) * G))
-            else:
-                P = np.matmul(P, self._expm(dt * G))
+            P = np.matmul(P, self._expm(dt * G))
```
petrelharp: it happens here

msprime/simulations.py
```diff
@@ -1365,6 +1366,176 @@ def print_history(self, output=sys.stdout):
...
+        for j, time in enumerate(steps_b):
+            dt = time - prev
+            prev = time
+            N_M = self.pop_size_and_migration_at_t(time)
+            N = N_M["population_sizes"]
+            M = N_M["migration_matrix"]
+            C = np.zeros([num_pops**2, num_pops**2])
+            for idx in range(num_pops):
+                C[IA[idx, idx], IA[idx, idx]] = 1 / (2 * max(min_pop_size, N[idx]))
+            dM = np.diag([sum(s) for s in M])
```
jgallowa07: should this be `np.array(M) + np.diag([sum(s) for s in M])` @petrelharp ?

jgallowa07: Oops, nvm.

jeromekelleher:
This looks excellent and a really useful piece of functionality, thanks @jgallowa07 and @petrelharp! I'll need to spend a bit of time playing with it, and might make a few tweaks before merging (after checking with you guys to see if you agree, of course). Can you squash the commits down in prep for this please @jgallowa07?

jeromekelleher:
One quick UI question: why not make the parameter to the trajectory functions an array of times? It's much more general, and not too taxing on the user. I.e., we change from ``end, num_steps=10`` to ``t``, and call it like:
```
t = np.arange(end, num_steps) # Or whatever the actual syntax is
N = db.population_size_trajectory(t)
# To stuff with N and t.
```
In practise, the user is going to end up making the ``t`` vector anyway in order to plot the values in N, and this way is far more general.

jeromekelleher:
Quick note here: if it's useful we can wrap some linear algebra functions from GSL. I can't find a matrix exponential function quickly, but I'd be surprised if we couldn't do it using the primitives. (But, this may be no better than using numpy primitives)

jgallowa07:
Added user defined steps and squashed @jeromekelleher. I'm going to hand it off to you to play with for a bit. Let me know If there's anything I can do to help. - Note that if your the steps you're taking are too large, the matrix exponential will spike; I generally keep the size of steps to about one order of magnitude in terms of generations.

jeromekelleher:
Thanks @jgallowa07, playing now. Just FYI, "Squashing - just added user defined steps" isn't a very helpful commit message: you want to describe what that commit *does*, not how you got to it. Think about skimming back through commits in a years time and how useful the message would be in understanding what the commit does...

jeromekelleher:
Also, using ``master`` as your topic branch isn't a great idea...

jgallowa07:
> Also, using `master` as your topic branch isn't a great idea...

Yea this was a mistake, I created a new branch and forgot to switch before making some major changes; Is there a way to move this to a new branch, then reset my master to upstream/master?

jeromekelleher:
> Yea this was a mistake, I created a new branch and forgot to switch before making some major changes; Is there a way to move this to a new branch, then reset my master to upstream/master?

It's not worth changing now, it doesn't make much difference. In general it's  simpler to just forget about the master branch in your fork. You never need to use it.

jeromekelleher:
I made some updates here:

1. Split out matrix_exponential and added some tests. It would be good to document when we think this is a good approx and when we think it'll fail. Maybe a few more tests also? (I see the existing tests have failed on Appveyor --- we really should be reasonably clear about when this function is appropriate to use.)
2. Removed docstring comments from tests. These make it harder to see which test is being run when using nose.
3. Changed the UI  to stop returning the ``steps`` array. There's no point in returning a parameter unchanged, and it'll just confuse people if we do.

I think the main thing to do from here is to bring the documentation up to date and fix up the tests for the matrix exp function. Can you take this please @jgallowa07?

jgallowa07:
>     1. Split out matrix_exponential and added some tests. It would be good to document when we think this is a good approx and when we think it'll fail. Maybe a few more tests also? (I see the existing tests have failed on Appveyor --- we really should be reasonably clear about when this function is appropriate to use.)

I'm not actually sure where this is going to fail (differ from other implementations) exactly. might be slightly out of my realm of understanding of the matrix exponential; I know I was computing `eigendecomposition` vs. `Padé approximation` (which is implemented by scipy.linalg).

any thoughts @petrelharp?

petrelharp:
We could work out when it fails and when it doesn't, but I'd like to avoid re-implementing our own fully-robust `expm` function, since that's a [whole can of worms](https://epubs.siam.org/doi/10.1137/S00361445024180?mobileUi=0&). So, I'd vote for making `matrix_exponential` "private" again. As for checking if it's OK, I think my advice in the docs is a good way to do it:
```
        For this to be accurate, dt must be small enough so that (a) short epochs
        (e.g., bottlenecks) are not missed, and (b) populations do not
        change in size too much over that time, if they are growing or shrinking.
        If in doubt, check that results are simliar after doubling `num_steps`.
```
We could even do this automatically: we could do the same computation with all steps divided in two, and if the answer we get is substantially different, throw an error or warning. I want to just rely on the user doing this rather than doing it ourselves, but it *would* be pretty easy.

petrelharp:
There *was* a problem that *could* be hit in practice, I think. We think we have a fix, but if not, let me record this alternative method here (which doesn't quite work yet):
```
def expm(A, n=10):
    # only works for stochastic matrices
    m = A.shape[0]
    dA = (-1) * np.diag(A)
    dmax = max(dA)
    P = A / dmax
    for j in range(m):
        P[j, j] = (-1) * A[j, j] / dmax
    nscales = max(0, np.ceil(np.math.log2(dmax) - np.math.log2(.2)))
    print(nscales)
    t = dmax / 2 ** nscales
    Pk = np.exp(-t) * t * P
    expA = Pk
    for j in range(m):
        expA[j, j] = np.exp(-t) + expA[j, j]  # identity
    for k in range(2, n + 1):
        Pk = (t/k) * np.matmul(P, Pk)
        expA = expA + Pk
    while nscales > 0:
        expA = np.matmul(expA, expA)
        nscales = nscales - 1
    return expA

```

petrelharp:
This is a cleaned-up version of what we had, and passes the tests, for me:
```
def matrix_exponential(A):
    """
    Returns the matrix exponential of A.
    https://en.wikipedia.org/wiki/Matrix_exponential
    Note: this is not a general purpose method and is only intended for use within
    msprime.
    """
    d, Y = np.linalg.eig(A)
    Yinv = np.linalg.pinv(Y)
    D = np.diag(np.exp(d))
    B = np.matmul(Y, np.matmul(D, Yinv))
    return B
```

petrelharp:
... but, the original function seems to work equivalently. This *should* be fine. @jgallowa07, can you confirm if this works?

jgallowa07:
> ... but, the original function seems to work equivalently. This _should_ be fine. @jgallowa07, can you confirm if this works?

So it doesn't _for me_. I'm positive that this floating point error is sensitive the processor and machine, remember how we got different answers for this function given `np.ones((4,4))` yesterday @petrelharp ? I've tried it both on `sesame` and `deep`, and neither pass.

jeromekelleher:
All we want here I think is just some test coverage on some matrices that are typical of our usage. The example I put in we're what I could think of in the 5 minutes I had. Feel free to change them! I agree the function should be private, but we need to describe for future-us what the target inputs are and get some test coverage on some typical instances. Totally agree that we don't want to make a general purpose expm.

msprime/simulations.py
```diff
@@ -1236,19 +1236,19 @@ def __repr__(self):
 def matrix_exponential(A):
     """
     Returns the matrix exponential of A.
-
     https://en.wikipedia.org/wiki/Matrix_exponential
-
     Note: this is not a general purpose method and is only intended for use within
     msprime.
     """
```
petrelharp: Could we use this version? I don't think the rounding or `Q` are actually doing anything.
```
    d, Y = np.linalg.eig(A)
    Yinv = np.linalg.pinv(Y)
    D = np.diag(np.exp(d))
    B = np.matmul(Y, np.matmul(D, Yinv))
    return B
```

petrelharp: and the `asmatrix( )` calls are just making `*` do `matmul`, but less obviously.

tests/test_demography.py
```diff
@@ -570,6 +571,208 @@ def test_two_pop_change_growth_rates(self):
...
+        for j in range(10):
+            A = np.array([[j]])
+            self.verify(A)
+
+    def test_zeros(self):
+        for j in range(1, 10):
+            A = np.zeros((j, j))
+            self.verify(A)
+
+    def test_ones(self):
```
jeromekelleher: This is a stupid test (and is causing problems) - can we replace it with a few example matrices where we expect to do well?

jeromekelleher:
@jgallowa07, please don't merge upstream/master into the topic branch, use rebase and force push instead. Merging in master only confuses things. We'll need to rebase and squash these commits before it can be merged.

jgallowa07:
good to know @jeromekelleher , it was the first time I've fixed up a merge conflict on the github UI, and the suggested the merge; I'll rebase now. 

re tests: I'm unclear by what tests we expect to do well on. I've taken one iteration from `test_ones` and it now passes on all machines: I've only seen the problem on j > 3. Should I write some tests with more realistic user input migration matrices?

jeromekelleher:
> re tests: I'm unclear by what tests we expect to do well on. I've taken one iteration from test_ones and it now passes on all machines: I've only seen the problem on j > 3. Should I write some tests with more realistic user input migration matrices?

Yes, the np.ones example is clearly not a realistic input, so we can get rid of it. We don't need anything elaborate, just one or two non-trivial example matrices that we would regard as typical inputs for our version of expm.

tests/test_demography.py
petrelharp: These are almost good tests; if these matrices are modified to be valid generators then we're good.

petrelharp:
Examples: 
 - (k x k) matrices of ones, but with (-k) on the diagonal, for k >= 2. 
 - a 1 x 1 matrix consisting of just 0 (compared to exp(0) = 1)
 - a 1 x 1 matrix consisting of just -1 (compared to exp(-1))
 - (-1) * np.eye(k), compared to exp(-1) * np.eye(k)

I think that should be fine, really.

jeromekelleher:
LGTM - can you squash please @jgallowa07? (It would be good if you could do this by default...)

jgallowa07:
How do you do squash by default? with `--amend`?

petrelharp:
He means by default in your head - there is not a git default for this.

msprime/simulations.py
```diff
@@ -1338,14 +1338,30 @@ def __repr__(self):
         return repr(self.__dict__)
 
 
+def matrix_exponential(A):
```
petrelharp: I still think this should be `_matrix_exponential(A)`?

msprime/simulations.py
```diff
@@ -1338,14 +1338,30 @@ def __repr__(self):
...
+
 class DemographyDebugger(object):
     """
     A class to facilitate debugging of population parameters and migration
     rates in the past.
     """
     def __init__(
             self, Ne=1, population_configurations=None, migration_matrix=None,
-            demographic_events=[], model="hudson"):
+            demographic_events=[], model='hudson'):
```
petrelharp: I think this is not how this default value should be specified: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

perhaps: `demographic_events=None` and then
```
if demographic_events is None:
    demographic_events = []
```

msprime/simulations.py
```diff
@@ -1471,6 +1489,156 @@ def print_history(self, output=sys.stdout):
...
+
+        :return: Returns a tuple where the first value is an array whose jth
+            element is the coalescence rate at the jth time point
+            (denoted r(t[j]) above), and the second is the probablility that a
+            randomly pair of lineages has not yet coalesced (denoted p(t[j]) above).
+
+        :rtype: class:'tuple'
+        """
+        num_pops = self.num_populations
+        # TODO throw a useful error if this fails.
```
petrelharp: TODO

jgallowa07: I'm was thinking on this, @jeromekelleher do you have a preference/suggestion about what type of error to throw?

jgallowa07:
> We could even do this automatically: we could do the same computation with all steps divided in two, and if the answer we get is substantially different, throw an error or warning. I want to just rely on the user doing this rather than doing it ourselves, but it _would_ be pretty easy.

oops, forgot about this, I'll get it done by tonight.

msprime/simulations.py
```diff
@@ -1471,6 +1489,178 @@ def print_history(self, output=sys.stdout):
...
+            This should be set to a value smaller than the smallest
+            desired population size in the model.
+
+        :return: Returns a tuple where the first value is an array whose jth
+            element is the coalescence rate at the jth time point
+            (denoted r(t[j]) above), and the second is the probablility that a
+            randomly pair of lineages has not yet coalesced (denoted p(t[j]) above).
+
+        :rtype: class:'tuple'
+        """
```
jgallowa07: We should probably do more user parameter input checking here, so we can throw useful errors/warnings. 
check that:
1: num_samples has len equal to num_pops
2: The number of samples doesn't exceed the number being sampled for the demographic model (maybe not?)
3: min_pop_size is sufficient? (maybe not?)
4: the steps are sequential?
5: Other requirements?

jgallowa07: If I knew which specific `msprime` errors/ warnings I should throw I could get this done real quick.

petrelharp: 1: num_samples has len equal to num_pops

`raise ValueError("num_samples must have the same length as the number of populations")`

2: The number of samples doesn't exceed the number being sampled for the demographic model (maybe not?)

No need for this check.

3: min_pop_size is sufficient? (maybe not?)

I think not: we don't have a good way of doing this beforehand.

4: the steps are sequential?

`raise ValueError("steps must be a sequence of increasing times.")`

5: Other requirements?

No zero-size steps.

We need documentation for `double_step_validation` also (or just *always* do it: this is my vote actually).

msprime/simulations.py
```diff
@@ -1471,6 +1489,178 @@ def print_history(self, output=sys.stdout):
...
+        if double_step_validatation:
+            offset = np.append(steps[1:], steps[-1]+1)
+            double_steps = [j for i in zip(steps, (steps+offset)/2) for j in i]
```
petrelharp: This assumes that the last `+1` is smaller than the stepsize elsewhere. I'd just intercalculate one extra step in each spot, like so:
```
double_steps = sorted(np.concatenate([x[1:], x[1:] + np.diff(x)/2]))
```

msprime/simulations.py
```diff
@@ -1471,6 +1489,178 @@ def print_history(self, output=sys.stdout):
...
+            rd, p_td = self._calculate_coalescence_rate_trajectory(
+                steps=double_steps,
+                num_samples=num_samples,
+                min_pop_size=min_pop_size)
+        r, p_t = self._calculate_coalescence_rate_trajectory(
+                steps=steps,
+                num_samples=num_samples,
+                min_pop_size=min_pop_size)
+        if not (np.allclose(r[-1], rd[-2]) and np.allclose(p_t[-1], p_td[-2])):
```
petrelharp: need to check only every other return value

msprime/simulations.py
petrelharp: `ValueError( )`, I tihnk.

msprime/simulations.py
jgallowa07: Sounds good

> We need documentation for double_step_validation also (or just always do it: this is my vote actually).

Happy to do this, my two cents: I think that If I were to, for some reason put this in a pipeline where I could guarantee my step sizes correct (by testing the domain once) and wanted to do this in a loop, I would want the option to cut the time down.

jeromekelleher: Errors looks good to me --- we need some tests to provoke them though. See elsewhere for examples of testing error inputs using ``self.assertRaises``

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
             self._print_populations(epoch, output)
             print(file=output)
 
+    def population_size_trajectory(self, steps):
+        """
+        This function returns an array of per-population effective population sizes,
+        as defined by the demographic model. These are just the `population_size`
+        parameters of the model, modified by any population growth rates.
+        The sizes are computed at a regular grid of time steps, between time 0 and `end`
+        units of time ago.
```
petrelharp: "The sizes are computed at the time points given by `steps`.

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+
+        :param int end: Time (ago) at which to stop trajectory inference.
```
petrelharp: delete

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+
+        :param int num_steps: determines at how many time points the
+            coalescence rate is computed.
```
petrelharp: delete; add
"
:param list steps: List of times ago at which the population size will be computed.

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+
+        :return: Returns a tuple where the first value
+            is an array giving the times at which population sizes have been computed,
+            the second is a num_steps-by-num_pops ndarray containing the effective
+            population sizes for each subpopulation individually,
+            at the corresponding step
+
+        :rtype: :class:'tuple'
```
petrelharp: Remove tuple stuff

petrelharp: :return: Returns a numpy array of population sizes, with one column per population, whose [i,j]th entry is the size of population j at time steps[i] ago.

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+        of coalescence of as-yet-uncoalesed lineages, computed as follows:
+        let p(t) be the probability that the lineages of a randomly
+        chosen pair of samples has not yet coalesced by time t,
+        let p(z,t) be the probability that the lineages of a randomly
+        chosen pair of samples has not yet coalesced by time t *and* are
+        both in population z, and let N(z,t) be the diploid effective
+        population size of population z at time t. Then the mean coalescence
+        rate at time t is r(t) = (sum_z p(z,t) / (2 * N(z,t) ) / p(t).
+        The computation is done approximating population size trajetories
+        with piecewise constant trajectories, in steps of size dt = end / num_steps.
```
petrelharp: "with piecewise constant trajectories between each of the ``steps``.

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+        For this to be accurate, dt must be small enough so that (a) short epochs
```
petrelharp: "For this to be accurate, the distance between the steps..."

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+        (e.g., bottlenecks) are not missed, and (b) populations do not
+        change in size too much over that time, if they are growing or shrinking.
+        If in doubt, check that results are simliar after doubling `num_steps`.
```
petrelharp: Instead of "If in doubt..", 
"This function optionally provides a simple check of this approximation by recomputing the coalescence rates on a grid of steps twice as fine and throwing a warning if the resulting values do not match to a relative tolerance of 0.001.

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+
+        :param list steps: The respective time steps (in generations) for which the
```
petrelharp: How about just
:param list steps: The times ago at which coalescence rates will be computed."

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+        N_t = np.zeros([len(steps), num_pops])
+        for j, t in enumerate(steps):
+            N, _ = self.pop_size_and_migration_at_t(t)
+            N_t[j] = N
+        return N_t
+
+    def coalescence_rate_trajectory(self, steps, num_samples, min_pop_size=1,
+                                    double_step_validatation=True):
+        """
+        This function will calculate the mean coalescence rates between the
```
petrelharp: "This function will calculate the mean coalescence rates and proportions of uncoalesced lineages ..."

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+            has not yet coalesced, is computed and returned. The list given should be
+            sorted in ascending order.
+
+        :param list num_samples: is a list of the same length as the number
+            of populations, so that `num_samples[j]` is the number of sampled
+            chromosomes in subpopulation `j`.
+
+        :param int min_pop_size: is the smallest allowed population size during
+            computation of coalescent rates (i.e., coalescence rates are actually
+            1 / 2 * max(min_pop_size, N(z,t))). Spurious very small population sizes
```
petrelharp: need extra parentheses in the denominator

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+            can occurIn models where populations grow exponentially but are unused
```
petrelharp: "occurIn"

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+            before some time in the past, and lead to floating point error.
+            This should be set to a value smaller than the smallest
+            desired population size in the model.
+
+        :param bool double_step_validatation: provides the option to ensure your step
```
petrelharp: delete first sentence starting "provides the option" (since this is explained above)

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+            sizes are small enough to ensure the correctness of a model's coalescent
+            rate prediction. If `True` this will, at most, take four times
+            the computation. Note: This test is sesitive to small effective
+            population sizes. Currently, this check with throw a warning
+            if the relative difference, for all values of coalecence rate or
+            sum of P statistic exceed 1e-3.
```
petrelharp: How about
:param bool double_step_validation: Whether to perform the check that step sizes are sufficiently small, as described above. This is highly recommended, and will take at most four times the computation.

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+
+        :return: Returns a tuple where the first value is an array whose jth
+            element is the coalescence rate at the jth time point
+            (denoted r(t[j]) above), and the second is the probablility that a
+            randomly pair of lineages has not yet coalesced (denoted p(t[j]) above).
```
petrelharp: how about
"
Returns a tuple of arrays  whose jth elements, respectively, are the coalescence rate at the jth time point
            (denoted r(t[j]) above), and the probablility that a
            randomly pair of lineages has not yet coalesced (denoted p(t[j]) above).

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+        r, p_t = self._calculate_coalescence_rate_trajectory(
+            steps=steps,
+            num_samples=num_samples,
+            min_pop_size=min_pop_size)
+        r_prediction_close = np.allclose(r[1:], rd[::2], rtol=1e-3)
+        p_prediction_close = np.allclose(p_t[1:], p_td[::2], rtol=1e-3)
+        if not (r_prediction_close and p_prediction_close):
+            # should we use the `warning` package?
+            raise UserWarning(
+                "doubling the number of steps have resulted in different"
```
petrelharp: "Doubling the number of steps has resulted in different"

msprime/simulations.py
```diff
@@ -1471,6 +1489,198 @@ def print_history(self, output=sys.stdout):
...
+            if(steps[i] != steps_b[i+idx_counter]):
+                diff_mask.append(i+idx_counter)
+                idx_counter += 1
+        r = np.delete(r, diff_mask)
+        p_t = np.delete(p_t, diff_mask)
+        return r, p_t
+
+    def pop_size_and_migration_at_t(self, t):
+        """
+        Given a time, t (in generations), return a tuple (N, M) where
```
petrelharp: How about
"
Returns a tuple (N, M) of population sizes (N) and migration rates (M) at time ``t`` ago.

:param float t: The time ago.

:return: A tuple of arrays, of the same form as the population sizes and migration rate arrays of the demographic model.

jgallowa07:
This says it's failing, but I don't think it has anything to do with my latest commit ... (I could be wrong?)

jeromekelleher:
> This says it's failing, but I don't think it has anything to do with my latest commit. Not to mention all nose tests pass for me locally... (I could be wrong?)

You're right, this is from other changes recently merged (but annoyingly only happening on this background...). I'll push a fix; if you rebase and push then it should be fixed up.

jeromekelleher:
Should be fixed in #760.

jgallowa07:
Awesome. Unless there's anything else, this should be good to merge.

petrelharp:
LGTM!  :yay:

petrelharp:
I will wait for someone else to approve this before merging; probably this should be @jeromekelleher as he was looking at it already.

jeromekelleher:
Thanks @jgallowa07 and @petrelharp, this is a great addition! I've made some minor style changes and fixed a problem with the warning (warnings should be thrown in warnings.warn, not raised as an exception),

PR #744, opened 2019-05-04
single_sweep prototype implemented in C

andrewkern:
hey @jeromekelleher - so here is a work in progress PR. I've implemented most of the guts of the sweep routines in C and it's time for a few design decisions. In particular we need to decide how we will generate the trajectories that we will feed to `msp_run_single_sweep`. I've tried to implement this so that it can run in the mixed model run setting that I see in lib/tests/tests.c, so that constrains things a bit-- the times that we generate for the trajectory should probably be on the same scale as times outside of these loops (i.e. not relative to the start of the simulation). 

So take a look and tell me what I should do next! I think I've got my ahead well around the moving parts of the C side so would like to keep working on it while its fresh in my mind.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2365,25 +2365,9 @@ out:
     return ret;
 }
 
+/* Returns the size of the specified population at the specified time */
 static double
-msp_get_population_size(msp_t *self, population_t *pop)
-{
-    double ret = 0;
-    double alpha = pop->growth_rate;
-    double t = self->time;
-    double dt;
-
-    if (alpha == 0.0) {
-        ret = pop->initial_size;
-    } else {
-        dt = t - pop->start_time;
-        ret = pop->initial_size * exp(-alpha * dt);
-    }
-    return ret;
-}
-
-static double
-msp_get_population_size_time(population_t *pop, double t)
+get_population_size(population_t *pop, double t)
 {
     double ret = 0;
     double alpha = pop->growth_rate;
@@ -2397,6 +2381,7 @@ msp_get_population_size_time(population_t *pop, double t)
     }
     return ret;
 }
+
 /* Given the specified rate, return the waiting time until the next common ancestor
  * event for the specified population */
 static double
@@ -2619,7 +2604,7 @@ msp_dtwf_generation(msp_t *self)
         // is possible for populations to reach zero individuals before all
         // lineages coalesce. We round up to avoid this.
         N = (uint32_t) ceil(
-            msp_get_population_size(self, pop) * self->model.population_size);
+            get_population_size(pop, self->time) * self->model.population_size);
         if (N == 0) {
             ret = MSP_ERR_INFINITE_WAITING_TIME;
             goto out;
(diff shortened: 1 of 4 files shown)


andrewkern:
codecov is pissed cause I can't make tests that hit the bad `ret` values in the `msp_single_sweep_recombination_event` function. any advice guys?

jeromekelleher:
Hooray! This is fantastic, thanks for getting this together @andrewkern. I'd say we merge it ASAP. Can you squash the commits please? I'll make a few minor updates then and merge when it's ready (shouldn't take long).

> codecov is pissed cause I can't make tests that hit the bad ret values in the 
 msp_single_sweep_recombination_event function. any advice guys?

I'd say not to worry about it.

andrewkern:
okay squashed!

lib/msprime.h
```diff
@@ -129,6 +129,7 @@ typedef struct {
         double *allele_frequency;
         double *time;
         size_t num_steps;
+        double sweep_dt;
```
jeromekelleher: Is this needed as a separate parameter (i.e., it's not the difference between adjacent values in the time array)? If so, it would be good to add a comment here explaining what it's for, as it's not obvious.

andrewkern: probably not needed. i put this in to give us options design-wise for the trajectory times. we can definitely do it from the difference between adjacent times.

jeromekelleher: OK, let's change it over to computing the dts online. I want to keep things simple as possible at the C level; we can add user friendly stuff in Python. Would you mind updating to remove this? I think we can merge then, and I'll setup the Python-C plumbing. I see what you mean about the absolute times too, BTW, we should probably change these to relative to the start of the sweep.

andrewkern: happy to change over the dt parameter today. there is a big obstacle to changing over times to be relative to the beginning of the sweep: we need to account for population size changes when we construct the trajectories themselves. 

i think we should keep them absolute for the time being and start to nail down how we will implement the trajectory generation. For instance should I port over C code to do this? Should we instead do the trajectories using python?

jeromekelleher: OK, understood and agreed on the times. I want to do as little as possible in C - I'm sure we can use the DemographyDebugger to give us the population sizes that we need, and numpy arrays used well should be fast enough. If you can get rid of the dts I'll make the low-level C-python interface and we'll be able to see where to go from there.

andrewkern: okay took at the `sweep_dt` thing and moved to calculating those on the fly.

lib/msprime.c
```diff
@@ -2905,74 +2910,156 @@ msp_change_label(msp_t *self, segment_t *ind, label_id_t label)
...
+        while (event_prob > event_rand && curr_step < num_steps) {
+            sweep_dt = time[curr_step] - time[curr_step - 1];
+            /* using pop sizes grabbed from get_population_size */
+            p_coal_B = ((sweep_pop_sizes[1] * (sweep_pop_sizes[1] - 1) ) * 0.5)
+                / allele_frequency[curr_step]
+                * sweep_dt / get_population_size(&self->populations[0], time[curr_step]);
+            p_coal_b = ((sweep_pop_sizes[0] * (sweep_pop_sizes[0] - 1) ) * 0.5)
+                / (1.0 - allele_frequency[curr_step])
+                * sweep_dt / get_population_size(&self->populations[0], time[curr_step]);
+            p_rec_b = rec_rates[0] * sweep_dt;
```
andrewkern: in addition I added `p_rec_b` and `p_rec_B` to make the calculation more obvious at the expense of writing a bit more code

jeromekelleher:
Looks good to me @andrewkern, I'm going to do a quick squash and then merge unless you object.

andrewkern:
question: should I continue to develop on this branch and request PRs as I go, or should I delete the branch and start a new one for my next PR?

jeromekelleher:
It's a simpler/cleaner to delete the branch and create new topic branches from upstream master, but it doesn't really matter once you rebase before opening the PR.

jeromekelleher:
Working on Python/C interface for this now...

andrewkern:
sweet! let me know where I can starting writing the trajectory code when you get to a good point for me to add.

PR #745, opened 2019-05-08
Sweeps python c

jeromekelleher:
WIP; I'll get a basic high-level interface working here before merging.

cc @andrewkern

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4909,15 +4909,17 @@ single_sweep_model_free(simulation_model_t *model)
 
 int
 msp_set_simulation_model_single_sweep(msp_t *self, double population_size,
-        uint32_t locus, size_t num_steps, double *time, double *allele_frequency)
+        double position, size_t num_steps, double *time, double *allele_frequency)
 {
     int ret = 0;
     size_t j;
+    uint32_t locus;
     simulation_model_t *model = &self->model;
+    double L = self->recomb_map->sequence_length;
 
     /* Check the inputs to make sure they make sense */
-    if (locus >= self->num_loci) {
-        ret = MSP_ERR_BAD_SWEEP_LOCUS;
+    if (position < 0 || position >= L) {
+        ret = MSP_ERR_BAD_SWEEP_POSITION;
         goto out;
     }
     if (num_steps == 0) {
@@ -4941,6 +4943,10 @@ msp_set_simulation_model_single_sweep(msp_t *self, double population_size,
         }
     }
 
+    ret = recomb_map_phys_to_discrete_genetic(self->recomb_map, position, &locus);
+    if (ret != 0) {
+        goto out;
+    }
     ret = msp_set_simulation_model(self, MSP_MODEL_SINGLE_SWEEP, population_size);
     if (ret != 0) {
         goto out;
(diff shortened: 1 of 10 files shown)


andrewkern:
i'll hold off until you tell me to have at it

jeromekelleher:
OK, I've made some good progress here @andrewkern, and it's at least possible to make things run now from Python. It's quite brittle at the moment though, so there's definitely work to be done hardening the interface. In particular, we'll need to nail down what we mean by 'time' within the sweep, and what it means to run the sweep outside of the times in the trajectory (it would be by far simplest if this was a no-op). We also need to rescale the input times into model time, to make sure everything works.

Hopefully we're not far away from a flexible and user-friendly interface though. We can figure out ways of generating the trajectory nicely within Python, but I think it's really vital to decouple the trajectory and coalescent simulations --- the code is complicated enough as it is! 

Can you take a look through and see if you agree with what's been done here @andrewkern? If so I'll merge and we can have a chat about how to proceed from here.

andrewkern:
so this all looks good to me. the one design issue is that bit about removing the max time from the main single_sweep loop, but it is fine for now and we can tweak it moving forward, as you note in the comment.

msprime/simulations.py
```diff
@@ -1204,6 +1217,95 @@ def __init__(self, population_size=1, psi=0.5, c=10.0):
...
+        self.time = []
+        self.allele_frequency = []
+
+    def __genic_selection_stochastic_forwards(self, dt, freq, alpha):
+        ux = (alpha * freq * (1 - freq)) / np.tanh(alpha * freq)
+        sign = 1 if self.rng.random() < 0.5 else -1
+        freq += (ux * dt) + sign * np.sqrt(freq * (1.0 - freq) * dt)
+        return freq
+
+    def __simulate(self):
```
andrewkern: we are going to want to have the population size history available within this routine. i guess it could come in through a DemographyDebugger object?

lib/msprime.c
```diff
@@ -2967,17 +2994,31 @@ msp_run_single_sweep(msp_t *self, double max_time, unsigned long max_events)
...
+    p_coal_B = 0;
+
+    /* JK: I've removed the time and event limits on this function to simplify
+     * things as function is currently 'non-rentrant'; we can't stop in the middle
+     * of the sweep and pick it up again from where we left off later. This is
+     * different to all the other model runners. Probably the simplest thing to
+     * do here is to make the trajectory times absolute wrt to the simulation, and
+     * to then increment curr_step accordingly. We can then probably reason about
+     * when the call msp_single_sweep_initialise and msp_single_sweep_finalise
+     * depending on the value of curr_step, and hopefully reintroduce the max_time
```
andrewkern: yes we would want to go back to the hudson coalescent after a given sweep to make guarantee the MRCA is reached

jeromekelleher: > yes we would want to go back to the hudson coalescent after a given sweep to make guarantee the MRCA is reached

We'll do this with a SimulationModelChange at the top level --- we might not always want to change to Hudson, maybe we would want to finish the sim with DTWF, or SMC, or might just want to stop entirely.

jeromekelleher:
OK, merging. We should organise a quick video chat to figure out the remaining details @andrewkern, but I think we're super-close to having it all working now!

andrewkern:
super cool @jeromekelleher! talk tomorrow? i'll shoot you an email

PR #749, opened 2019-05-09
Round DTWF pop size, and set model earlier in demographic tests. Fixes #739

### LINKED ISSUE #739: Ceil or round in DTWF pop size calculation?
When computing the population size in the DTWF we take the ceiling rather than round to the nearest int [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L2606).

This seems like it might cause more problems that it would solve @DomNelson. Would you object to changing this to ``round``? I realise this will lead to more models that end up with a population size of 0, but I think the user is better off knowing that their simulation has entered this weird parameter regime rather than continuing on indefinitely.

DomNelson:
Previously used ceil to avoid zero population size in DTWF, but this only let
badly-specified models run without errors. A more descriptive error is now
thrown. Appropriately resizing populations in tests then led to time travel
errors in bottleneck events. These should not have been possible to specify in
DTWF (they are only supported in the Hudson model), so model is now set earlier
to catch this.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -230,17 +230,12 @@ msp_set_population_configuration(msp_t *self, int population_id, double initial_
     int ret = MSP_ERR_BAD_POPULATION_CONFIGURATION;
     simulation_model_t *model = &self->model;
 
-
     if (population_id < 0 || population_id > (int) self->num_populations) {
         ret = MSP_ERR_POPULATION_OUT_OF_BOUNDS;
         goto out;
     }
     if (initial_size <= 0) {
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Looks good, thanks. One minor thing: we should add in a test to provoke the zero population size error. Might be simplest to make a new test that just does this, rather than complicating the exsting tests. (test_dtwf_zero_pop_size, e.g.)

DomNelson:
Forgot to run valgrind first, have a leak to find still...

PR #752, opened 2019-05-12
Add error code for unsupported DTWF bottlenecks - Fixes #747

### LINKED ISSUE #747: Bottlenecks in DTWF
Running into time travel errors with bottlenecks in DTWF - looks like this was hidden by small pop sizes in the tests. Basically any event which calls `msp_merge_ancestors()` can potentially create a child node with the same time as the parent node (if called before parent selection) or a parent node with the same time as the child nodes (if called after).

Are bottleneck events strictly necessary in the DTWF, or would they be equivalent to population size changes? The simple bottleneck maybe couldn't be done this way, but they might have to be treated as a special case in Wright-Fisher sims and be done within the DTWF parent selection step. The only other thing I can think of at the moment is jiggling node times a bit, but I'm not sure that's a great solution.

DomNelson:
Also fix possible infinite loop in demography debugger

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -3946,10 +3946,6 @@ msp_add_simple_bottleneck(msp_t *self, double time, int population_id, double pr
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
     }
-    if (self->model.type == MSP_MODEL_DTWF) {
-        ret = MSP_ERR_DTWF_UNSUPPORTED_BOTTLENECK;
-        goto out;
-    }
     ret = msp_add_demographic_event(self, time, &de);
     if (ret != 0) {
         goto out;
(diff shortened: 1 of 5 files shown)


DomNelson:
Oops, wasn't running high-level tests...

lib/msprime.c
```diff
@@ -3945,6 +3949,10 @@ msp_add_simple_bottleneck(msp_t *self, double time, int population_id, double pr
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
     }
+    if (self->model.type == MSP_MODEL_DTWF) {
```
jeromekelleher: I don't think there's much point in catching these at 'add' time, as often we'll be using a different model at the start and end of the simulation. Might as well just throw the error at run time so we have less code to test. (Same for the other bottleneck below).

jeromekelleher:
OK, merging this, thanks @DomNelson.

PR #754, opened 2019-05-13
Model docs

jeromekelleher:
WIP

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -81,7 +81,7 @@ Running simulations
 The :func:`.simulate` function provides the primary interface to running
 coalescent simulations in msprime.
 
-.. autofunction:: msprime.simulate
+.. autofunction:: msprime.simulate()
 
 ********************
 Population structure
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -156,55 +156,63 @@ model. We also support a number of different models, which are documented in thi
 
 Simulations models are specified using the ``model`` parameter to
 :func:`.simulate`. This parameter can either take the form of a
-string describing the model (e.g. ``model="smc"``) or an instance of a
-model definition class (e.g ``model=msprime.SmcApproxCoalescent(1)``).
+string describing the model (e.g. ``model="dtwf"``) or an instance of a
+model definition class (e.g ``model=msprime.DiscreteTimeWrightFisher(1000)``).
 The available models are documented in the following subsections.
 
 A key element of simulation models in ``msprime`` is the concept
-of a "reference population size". Simulation models such as
-the coalescent are defined
-in terms of "scaled time". Time in the standard diploid coalescent
-is measured in units of :math:`4 N_e` generations, and
-one of the ways in which msprime tries to make life easier for
-users is to automatically convert times and rates to and
-from units of generations. Thus, a key responsibility for a simulation model is to
-convert times specified by users in generations into "model time"
-(in which the simulation is performed) and to tranlate the
-simulated model times back into generations. This is what the
-reference population size associated with a model is used for
-and fundamentally what it means. The ``Ne`` argument to
+of a "reference population size". The ``Ne`` argument to
 :func:`simulate` can also be used to define this parameter
 when combined with a string shorthand for a model.
-Thus
 
 .. code-block:: python
 
-    msprime.simulate(10, Ne=1000, model="smc")
+    msprime.simulate(10, Ne=1000, model="dtwf")
 
 and
 
 .. code-block:: python
 
-    msprime.simulate(10, model=msprime.SmcApproxCoalescent(1000))
+    msprime.simulate(10, model=msprime.DiscreteTimeWrightFisher(1000))
 
 define the same simulation.
 
-The situation is somewhat confused by the sizes associated
-with specific populations using, e.g., the
-:class:`.PopulationConfiguration` class. In coalescent models, these
-sizes are used to compute rates of coalescence, and have no direct
-relationship to the model's reference population size. Thus, we may
-have several populations, all of which have different sizes and none
-equal to the model's reference population size.
-The model's reference population size is used to scale all times
-and rates, irrespective of the sizes of the various individual
-populations.
+.. todo:: Add a discussion of population sizes here, describing
+    what Ne/model.reference_size really means, and how it interacts
+    with the individual population sizes.
 
-The situation for the :class:`.DiscreteTimeWrightFisher` model
-is different. In this case, there is no concept of rescaling time
-as time is always measured in generations. Therefore, in this case the
-reference population size has no function and is just used as a
-shortcut for specifying individual population sizes.
+.. JK: Commented this discussion out as it seemed likely to confuse. We
+.. definitely need to give a thorough description of what's actually
+.. going on somewhere though.
+
+.. Simulation models such as the coalescent are defined
+.. in terms of "scaled time". Time in the standard diploid coalescent
+.. is measured in units of :math:`N_e` generations, and
+.. one of the ways in which msprime tries to make life easier for
+.. users is to automatically convert times and rates to and
+.. from units of generations. Thus, a key responsibility for a simulation model is to
+.. convert times specified by users in generations into "model time"
+.. (in which the simulation is performed) and to tranlate the
+.. simulated model times back into generations. This is what the
+.. reference population size associated with a model is used for
+.. and fundamentally what it means.
+
+.. The situation is somewhat confused by the sizes associated
+.. with specific populations using, e.g., the
+.. :class:`.PopulationConfiguration` class. In coalescent models, these
+.. sizes are used to compute rates of coalescence, and have no direct
+.. relationship to the model's reference population size. Thus, we may
+.. have several populations, all of which have different sizes and none
+.. equal to the model's reference population size.
+.. The model's reference population size is used to scale all times
+.. and rates, irrespective of the sizes of the various individual
+.. populations.
+
+.. The situation for the :class:`.DiscreteTimeWrightFisher` model
+.. is different. In this case, there is no concept of rescaling time
+.. as time is always measured in generations. Therefore, in this case the
+.. reference population size has no function and is just used as a
+.. shortcut for specifying individual population sizes.
 
 We are often interested in simulating mixtures of models: for example,
 using the :class:`.DiscreteTimeWrightFisher` model to simulate the
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -126,6 +126,7 @@ and all rates are per-generation.
 .. autoclass:: msprime.PopulationParametersChange
 .. autoclass:: msprime.MigrationRateChange
 .. autoclass:: msprime.MassMigration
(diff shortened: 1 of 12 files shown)


jeromekelleher:
FYI @DomNelson, I'm making progress here. I think I just need to clear up the confusion about population sizes and the different ways they can be set and we're good to merge. That's basically it for DTWF and we should be ready to cut a release. 

Can you run the DTWF statistical tests and just make sure they're all looking OK please? This is the list of [outstanding issues](https://github.com/tskit-dev/msprime/milestone/6) for 0.7.1, and hopefully they'll get knocked off fairly quickly.

DomNelson:
Great! Ok I'll get the tests running

jeromekelleher:
This is nearly complete, and is adding some important changes so I'm tagging Dom, Peter and Andy for review. I've tried to clear up the confusion about "population_size" by changing it to ``reference_size`` for model objects and hopefully clarified what models *do* in the docs. The neat thing now is that we can do things like:
```
ts = msprime.simulate(
    100, Ne=1000, model="dtwf",
    demographic_events=[msprime.SimulationModelChange(500, "hudson")])
```
which will run 500 generations of DTWF before switching to the standard coalescent model, all with 'population size' of 1000. This seems like it's what people would want in most cases. We can also do things in a more long-handed way to specify different Ne's for the different models:
```
ts = msprime.simulate(
    100, model=msprime.DiscreteTimeWrightFisher(1000),
    demographic_events=[
         msprime.SimulationModelChange(500, msprime.StandardCoalescent(2000)])
```
which will run DTWF with Ne=1000 and Hudson with Ne=2000.

Does this look like a sensible framework to you? I'm pinging you Andy as the ``GenicSelectionSweep`` is going to fit into exactly the same framework (but with a slightly different notation for specifying times, since we don't know when the sweep will end, before the sim).

docs/api.rst
```diff
@@ -144,6 +145,105 @@ Variable recombination rates
...
+have several populations, all of which have different sizes and none
+equal to the model's reference population size.
+The model's reference population size is used to scale all times
+and rates, irrespective of the sizes of the various individual
+populations.
+
+The situation for the :class:`.DiscreteTimeWrightFisher` model
+is different. In this case, there is no concept of rescaling time
+as time is always measured in generations. Therefore, in this case the
+reference population size has no function and is just used as a
```
jeromekelleher: This isn't quite true as it's used in the population growth calculations, I need to fix this

msprime/simulations.py
```diff
@@ -1063,18 +1082,30 @@ def __str__(self):
...
+    :param float time: The time at which the simulation model changes
+        to the new model, in generations.
+    :param model: The new simulation model to use.
+        This can either be a string (e.g., ``"smc_prime"``) or an instance of
+        a simulation model class (e.g, ``msprime.DiscreteTimeWrightFisher(100)``.
+        Please see the :ref:`sec_api_simulation_models` section for more details
+        on specifying simulations models. If the argument is a string, the
+        reference population size is set from the top level ``Ne`` parameter
+        to :func:`.simulate`. If this is None (the default) we revert to
+        the standard coalescent with the reference population size set by ``Ne``.
```
DomNelson: Could mention that default reference pop size is also set by Ne if model is specified but with default reference population size (if that's actually the case)?

tests/test_models.py
```diff
@@ -30,6 +30,75 @@
...
+        simulation_models = [
+            ("hudson", msprime.StandardCoalescent),
+            ("smc", msprime.SmcApproxCoalescent),
+            ("smc_prime", msprime.SmcPrimeApproxCoalescent),
+            ("dtwf", msprime.DiscreteTimeWrightFisher)
+        ]
+        for name, model_class in simulation_models:
+            model = msprime.model_factory(model=name.upper())
+            self.assertIsInstance(model, model_class)
+            model = msprime.model_factory(model=name.upper())
```
DomNelson: Should one of these be `name.lower()`?

docs/tutorial.rst
```diff
@@ -736,39 +720,31 @@ Also note that recombinations will still occur in the gaps between chromosomes,
...
+This is done through a :class:`.SimulationModelChange` event, which is a special type of
+demographic event.
 
 For example, here we switch from the discrete-time Wright-Fisher model to the
 standard Hudson coalescent 100 generations in the past:
 
-.. code-block:: python
-
-    num_wf_generations = 100
+.. todo:: FIXME this doesn't work yet.
```
DomNelson: This will work once this PR is merged though right?

jeromekelleher: Yes, this was put in earlier so I didn't forget about it. I'll clear it up for the final version.

jeromekelleher:
@andrewkern, @petrelharp, it would be great if you could take a look at this. It's a big change and I want to be sure we're all on board with the model change infrastructure before going any further with it.

docs/api.rst
```diff
@@ -126,6 +126,7 @@ and all rates are per-generation.
 .. autoclass:: msprime.PopulationParametersChange
 .. autoclass:: msprime.MigrationRateChange
 .. autoclass:: msprime.MassMigration
+.. autoclass:: msprime.SimulationModelChange
```
jeromekelleher: TODO: need to clarify the exact semantics of this WRT to the time argument. Currently it's guaranteed to stop before a given time for DTWF, but for other models it will stop immediately after first event > this. These semantics will be updated at some point to have the more natural <= semantics.

andrewkern: can you just change the DTWF behavior to match that of the other models? i.e. the last time it returns is `MAXTIME` (or whatever the variable is called) for that epoch?

docs/api.rst
```diff
@@ -144,6 +145,105 @@ Variable recombination rates
...
+of a "reference population size". Simulation models such as
+the coalescent are defined
+in terms of "scaled time". Time in the standard diploid coalescent
+is measured in units of :math:`4 N_e` generations, and
+one of the ways in which msprime tries to make life easier for
+users is to automatically convert times and rates to and
+from units of generations. Thus, a key responsibility for a simulation model is to
+convert times specified by users in generations into "model time"
+(in which the simulation is performed) and to tranlate the
+simulated model times back into generations. This is what the
```
andrewkern: this is awesome!

docs/api.rst
andrewkern: also the discussion of the model reference size being divorced from the actual population sizes might throw users. perhaps better to suggest that the user choose a "sensible reference", like the population size of pop0 at time zero?

msprime/simulations.py
```diff
@@ -98,6 +99,43 @@ def almost_equal(a, b, rel_tol=1e-9, abs_tol=0.0):
...
+    - If model is an instance of SimulationModel, return a copy of it.
+      In this case, if the model's reference_size is None, set it to the
+      parameter reference size.
+    - Otherwise return a type error.
+    """
+    model_map = {
+        "hudson": StandardCoalescent(reference_size),
+        "smc": SmcApproxCoalescent(reference_size),
+        "smc_prime": SmcPrimeApproxCoalescent(reference_size),
+        "dtwf": DiscreteTimeWrightFisher(reference_size)
```
andrewkern: want to add the simple sweep case here now?

docs/api.rst
```diff
@@ -144,6 +145,105 @@ Variable recombination rates
...
+:func:`.simulate`. This parameter can either take the form of a
+string describing the model (e.g. ``model="smc"``) or an instance of a
+model definition class (e.g ``model=msprime.SmcApproxCoalescent(1)``).
+The available models are documented in the following subsections.
+
+A key element of simulation models in ``msprime`` is the concept
+of a "reference population size". Simulation models such as
+the coalescent are defined
+in terms of "scaled time". Time in the standard diploid coalescent
+is measured in units of :math:`4 N_e` generations, and
```
petrelharp: Isn't time in units of `N_e` generations? This phrase seems to contradict [these docs](https://msprime.readthedocs.io/en/stable/api.html?highlight=diploid#simulation-model).

docs/api.rst
```diff
@@ -144,6 +145,105 @@ Variable recombination rates
...
+*****************
+
+The default simulation model in ``msprime`` is the standard coalescent with recombination
+model. We also support a number of different models, which are documented in this section.
+
+Simulations models are specified using the ``model`` parameter to
+:func:`.simulate`. This parameter can either take the form of a
+string describing the model (e.g. ``model="smc"``) or an instance of a
+model definition class (e.g ``model=msprime.SmcApproxCoalescent(1)``).
+The available models are documented in the following subsections.
```
petrelharp: maybe don't use `smc` as the example, since it's nice to have but it's not really a popluation model and we don't want to encourage people to use it (unless they know what they are doing) - `dtwf`?

docs/api.rst
petrelharp: I think this point - that under the hood, many models just treat Ne as a time scaling - is a side note, and is kinda confusing to put right up the top of this. All the simulation models have a *population size*, and they all mean the same thing; and maybe by the way in some models this is just a time scaling.

docs/api.rst
petrelharp: Isn't it true for the coalescent also that the reference size is a shortcut for specifying individual population sizes? Like, I think that `Ne` is *not* an overall time scaling, because setting population sizes in population configurations overrides it?

docs/examples.py
```diff
@@ -421,6 +421,14 @@ def full_arg_example():
         print("interval:", tree.interval)
         print(tree.draw(format="unicode"))
 
+def hybrid_sim_example():
+    ts = msprime.simulate(
+        sample_size=6, Ne=1000, model="dtwf", random_seed=2,
+        demographic_events=[
+            msprime.SimulationModelChange(time=500, model="hudson")])
+    print(ts.tables.nodes)
```
petrelharp: This is **great**!

msprime/simulations.py
jeromekelleher: It's not clear to me how we'd do this with the single sweep as it's parametric (we need s, etc). Happy to discuss options though.

petrelharp:
This is great! I had some comments on how Ne is talked about (I think it means the same thing in all models?). Also: is there a list of possible models somewhere? I didn't see it.

docs/api.rst
jeromekelleher: Good points, thanks. I'll clear this up.

docs/api.rst
jeromekelleher: I've updated this so fix the semantics all round. See #765 if you're interested.

docs/api.rst
jeromekelleher: Ne is *definitely* used for scaling and unscaling times @petrelharp -- see [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L4219)

jeromekelleher:
Thanks for the feedback all. I've dealt with the "what is Ne issue" by sweeping it back under the carpet for now. I'll open an issue to track actually documenting this properly.

PR #756, opened 2019-05-13
adding discoal checks to verifaction.py

andrewkern:
work in progress here, but i'm getting the following error with existing code and haven't debugged yet. here is the error:

```
$ python verification.py
admixture-1-pop1
admixture-1-pop1 1000 1000 -t 2.0 -es 0.1 1 0.5 -em 0.1 1 2 1
	 msprime: 1000 1000 -t 2.0 -es 0.1 1 0.5 -em 0.1 1 2 1
  File "verification.py", line 2209, in <module>
    main()
  File "verification.py", line 2205, in main
    verifier.run(keys)
  File "verification.py", line 273, in run
    runner()
  File "verification.py", line 281, in f
    self._run_coalescent_stats(key, command_line)
  File "verification.py", line 257, in _run_coalescent_stats
    df_msp = self._run_msprime_coalescent_stats(args)
  File "verification.py", line 224, in _run_msprime_coalescent_stats
    sim.run()
  File "/home/adkern/my_msprime/msprime/simulations.py", line 674, in run
    self.ll_sim = self.create_ll_instance()
  File "/home/adkern/my_msprime/msprime/simulations.py", line 666, in create_ll_instance
    node_mapping_block_size=self.node_mapping_block_size)
TypeError: function takes at most 14 keyword arguments (15 given)
```
so it looks like something is going wrong with the admixture example check of `ms` vs `msp`. Am I simply using `verification.py` wrong?

In addition I patched a bug in the `check_slim_version()` function that was choking on a slim version that looked like `3.2.1`

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -133,6 +133,7 @@ class SimulationVerifier(object):
         self._ms_executable = ["./data/ms"]
         self._scrm_executable = ["./data/scrm"]
         self._slim_executable = ["./data/slim"]
+        self._discoal_executable = ["./data/discoal"]
         self._mspms_executable = [sys.executable, "mspms_dev.py"]
 
     def check_slim_version(self):
@@ -144,13 +145,18 @@ class SimulationVerifier(object):
             if version_list[i].lower() == 'version':
                 version_str = version_list[i+1]
                 break
-        version = float(version_str.strip(' ,'))
+        version = float(version_str.strip(' ,')[0:3])
         assert version >= min_version, "Require SLiM >= 3.1!"
 
     def get_ms_seeds(self):
         max_seed = 2**16
         seeds = [random.randint(1, max_seed) for j in range(3)]
         return ["-seed"] + list(map(str, seeds))
+
+    def get_discoal_seeds(self):
+        max_seed = 2**16
+        seeds = [random.randint(1, max_seed) for j in range(3)]
+        return ["-d"] + list(map(str, seeds))
 
     def _run_sample_stats(self, args):
         print("\t", " ".join(args))
@@ -167,6 +173,10 @@ class SimulationVerifier(object):
             f.seek(0)
             df = pd.read_table(f)
         return df
+
+    def _run_discoal_mutation_stats(self, args):
+        return self._run_sample_stats(
+            self._discoal_executable + args.split() + self.get_discoal_seeds())
 
     def _run_ms_mutation_stats(self, args):
         return self._run_sample_stats(
@@ -260,6 +270,23 @@ class SimulationVerifier(object):
             print(key, command_line)
             self._run_coalescent_stats(key, command_line)
             self._run_mutation_stats(key, command_line)
+        self._instances[key] = f
+
+    def _run_mutation_discoal_stats(self, key, args):
+        # convert discoal string to msprime string
+        tokens = args.split(" ")
+        msp_str = " ".join(tokens[0:2]+tokens[3:])
+        df_msp = self._run_msprime_mutation_stats(msp_str)
+        df_d = self._run_discoal_mutation_stats(args)
+        self._plot_stats(key, "mutation", df_d, df_msp)
+
+    def add_discoal_instance(self, key, command_line):
+        """
+        Adds a test instance with the specified discoal command line.
+        """
+        def f():
+            print(key, command_line)
+            self._run_mutation_discoal_stats(key, command_line)
         self._instances[key] = f
 
     def get_pairwise_coalescence_time(self, cmd, R):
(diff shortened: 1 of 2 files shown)


verification.py
```diff
@@ -186,6 +196,16 @@ def _run_ms_coalescent_stats(self, args):
             df = pd.read_table(f)
         return df
 
+    def _run_discoal_coalescent_stats(self, args):
+        executable = ["./data/ms_summary_stats"]
```
jeromekelleher: This won't work for discoal I think. The coalescent stats here are basically internal counts from ms --- the number of RE and CA events. The ms_summary_stats executable here is a patched version of ms that spits these out. Unless you want to do the same for discoal, we won't be able to do these comparisons.

verification.py
```diff
@@ -144,14 +145,19 @@ def check_slim_version(self):
...
-        version = float(version_str.strip(' ,'))
+        version = float(version_str.strip(' ,')[0:3])
         assert version >= min_version, "Require SLiM >= 3.1!"
 
     def get_ms_seeds(self):
         max_seed = 2**16
         seeds = [random.randint(1, max_seed) for j in range(3)]
         return ["-seed"] + list(map(str, seeds))
 
+    def get_discoal_seeds(self):
```
jeromekelleher: Might as well just call ``get_ms_seeds``? If they're the same form then there's no point in this extra method.

verification.py
```diff
@@ -262,6 +282,26 @@ def f():
             self._run_mutation_stats(key, command_line)
         self._instances[key] = f
 
+    def _run_coalescent_discoal_stats(self, key, args):
+        df_msp = self._run_msprime_coalescent_stats(args)
```
jeromekelleher: See point above about coalescent stats.

verification.py
andrewkern: difference is in the flag we provide at the command line. `discoal` uses `-d` vs `ms`'s `-seed`

verification.py
andrewkern: ah ok. will jettison these then.

andrewkern:
so why isn't the `verification.py` script working for me?  any idea what that error is?

jeromekelleher:
Oh yeah, sorry. You need to rebuild your _msprime.so module. Run ``make`` in the root dir.

andrewkern:
now getting this error
```
$ python verification.py
admixture-1-pop1
admixture-1-pop1 1000 1000 -t 2.0 -es 0.1 1 0.5 -em 0.1 1 2 1
	 msprime: 1000 1000 -t 2.0 -es 0.1 1 0.5 -em 0.1 1 2 1
	 ./data/ms_summary_stats 1000 1000 -t 2.0 -es 0.1 1 0.5 -em 0.1 1 2 1 -seed 10748 49462 9333
verification.py:196: FutureWarning: read_table is deprecated, use read_csv instead, passing sep='\t'.
  df = pd.read_table(f)
  File "verification.py", line 2209, in <module>
    main()
  File "verification.py", line 2205, in main
    verifier.run(keys)
  File "verification.py", line 273, in run
    runner()
  File "verification.py", line 281, in f
    self._run_coalescent_stats(key, command_line)
  File "verification.py", line 259, in _run_coalescent_stats
    self._plot_stats(key, "coalescent", df_ms, df_msp)
  File "verification.py", line 252, in _plot_stats
    f = self._build_filename(key, stats_type, stat)
  File "verification.py", line 242, in _build_filename
    os.mkdir(output_dir)
FileNotFoundError: [Errno 2] No such file or directory: 'tmp__NOBACKUP__/admixture-1-pop1'
```

should i not be running verification.py from the root level?

jeromekelleher:
> should i not be running verification.py from the root level? creating that tmp__NOBACKUP__ dir at the root level of msprime of course fixed things

Sorry, yeah, this is just shit programming from me. We assume ``tmp__NOBACKUP__`` exists for this script.

verification.py
jeromekelleher: Ah, didn't spot that! Ignore the above

jeromekelleher:
When we ditch the coalescent stats we're only left with the sequence based statistics, which is inherently unpleasing... 

We could use the [tsconvert.from_ms](https://github.com/tskit-dev/tsconvert/blob/master/tsconvert/newick.py#L28) function to parse the discoal output, and then compare the tree sequences (num edges, num nodes, max time, etc). This would make things a lot easier. The only reason for not using this method before was I hadn't realised you could go from ms format to tree sequence so straightforwardly.

The only issue is that tsconvert isn't available on PyPI, but we could fix that pretty easily.

jeromekelleher:
Looks good - shall we merge this much now (as it's the basic infrastructure) and add more tests later? (if so, do the old squasharoo)

andrewkern:
yeah sounds good. will squash and force push.

andrewkern:
ready for merge. i'll add more tests today as time allows. meanwhile how is bringing the DemographyDebugger into the trajectory simulator going?

jeromekelleher:
Sweet, merging.

> meanwhile how is bringing the DemographyDebugger into the trajectory simulator going?

Will try to make some progress this evening...

PR #757, opened 2019-05-14
Add DTWF support for demographic events between generations - fixes #750

### LINKED ISSUE #750: Demographic events in DTWF with non-integer times
@jeromekelleher I don't think we can have DTWF demographic events at non-integer times without a bit more work, since the demographic events set the internal simulation time, and right now we're only set up to do one event per generation.

So if we have events at `time=0.5` and `time=0.6`, the first will take effect, followed by a generation of recombination and parent-chosing, and set the internal time to `0.5`, half a generation behind the 'true' simulation state. Similarly after the next event another full generation will have passed but simulation time will only show `0.6`.

```
if (self->next_demographic_event != NULL) {                
    if (self->next_demographic_event->time <= self->time) {
        ret = msp_apply_demographic_events(self);          
        if (ret != 0) {                                    
            goto out;                                      
        }                                                  
    }                                                      
}                                                          
free(node_trees);               
node_trees = NULL;              
                                
ret = msp_dtwf_generation(self);
if (ret != 0) {                 
    goto out;                   
}                               

```

If we want to allow non-integer times, we could have something like

```
while (self->next_demographic_event->time <= self->time) {
    ret = msp_apply_demographic_events(self);          
    if (ret != 0) {                                    
        goto out;                                      
    }
}
```
then maybe
```
self->time = ceil(self->time)
```
so that nodes at least will always have integer times. What do you think?

DomNelson:


lib/msprime.c
```diff
@@ -2818,17 +2818,22 @@ msp_run_dtwf(msp_t *self, double max_time, unsigned long max_events)
...
+         * generation, and throw off the time between generations. We avoid
+         * this by saving the current time and returning to it. */
+        cur_time = self->time;
+        while (self->next_demographic_event != NULL &&
+                self->next_demographic_event->time <= self->time) {
+            ret = msp_apply_demographic_events(self);
+            if (ret != 0) {
+                goto out;
             }
+            self->time = cur_time;
```
jeromekelleher: Should't resetting time happen *outside* the loop? It's a bit confusing to see it being done in there.

DomNelson: True, but if there were two events between generations, say at times 0.5 and 0.6, then the first would set the time to 0.5 and the next wouldn't execute. This makes sure we get them all.

jeromekelleher: This is pretty confusing ... so self->time will be 1.0 when we enter this loop, then when we apply the event it'll get reset back to 0.5, and then we don't run it because self->time < 0.6.

What if we did it like:
```C
                cur_time = self->time;
￼	        while (self->next_demographic_event != NULL &&
￼	                self->next_demographic_event->time <= cur_time) {
￼	            ret = msp_apply_demographic_events(self);
￼	            if (ret != 0) {
￼	                goto out;
￼	            }
            }
￼	    self->time = cur_time;
```
That seems less confusing to me?

DomNelson: Yep that looks better!

jeromekelleher:
Thanks @DomNelson, I made a small tweak to the Python tests (need to use ``self.assertEquals`` as we're using nose). Will merge once CI passes.

PR #758, opened 2019-05-15
Simulate trajectories in C

jeromekelleher:
Outline for discussion - see comments below for details @andrewkern

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2931,7 +2931,7 @@ msp_change_label(msp_t *self, segment_t *ind, label_id_t label)
     return ret;
 }
 
-int
+static int
 msp_sweep_recombination_event(msp_t *self, label_id_t label,
         uint32_t sweep_locus, double population_frequency)
 {
@@ -2989,6 +2989,7 @@ msp_run_sweep(msp_t *self)
     double event_prob, event_rand, tmp_rand, e_sum, pop_size;
     double p_coal_b, p_coal_B, total_rate, sweep_pop_tot_rate;
     double p_rec_b, p_rec_B;
+    bool sweep_over;
 
     /* Keep the compiler happy */
     sweep_pop_tot_rate = 0;
@@ -3017,22 +3018,23 @@ msp_run_sweep(msp_t *self)
 
     while (msp_get_num_ancestors(self) > 0 && curr_step < num_steps) {
         events++;
-        /* quick sanity check; also set pop sizes & rec_rates */
+        /* Set pop sizes & rec_rates */
         for (j = 0; j < self->num_labels; j++) {
             label = (label_id_t) j;
             num_links = fenwick_get_total(&self->links[label]);
             ret = msp_sanity_check(self, num_links);
-            sweep_pop_sizes[j] = (double) avl_count(&self->populations[0].ancestors[label]);
-            rec_rates[j] = (double) num_links * self->recombination_rate;
             if (ret != 0) {
                 goto out;
             }
+            sweep_pop_sizes[j] = avl_count(&self->populations[0].ancestors[label]);
+            rec_rates[j] = (double) num_links * self->recombination_rate;
         }
 
         curr_step++;
         event_prob = 1.0;
         event_rand = gsl_rng_uniform(self->rng);
-        while (event_prob > event_rand && curr_step < num_steps) {
+        sweep_over = false;
+        while (event_prob > event_rand && curr_step < num_steps && !sweep_over) {
             sweep_dt = time[curr_step] - time[curr_step - 1];
             /* using pop sizes grabbed from get_population_size */
             pop_size = get_population_size(&self->populations[0], time[curr_step]);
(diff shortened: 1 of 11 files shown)


lib/msprime.c
```diff
@@ -3010,6 +3010,11 @@ msp_run_single_sweep(msp_t *self)
      * depending on the value of curr_step, and hopefully reintroduce the max_time
      * and max_steps parameters. */
 
+    ret = model->params.single_sweep.generate_trajectory(
+            &model->params.single_sweep, self, &num_steps, &time, &allele_frequency);
+    if (ret != 0) {
```
jeromekelleher: The idea here is that we decouple the trajectory simulation here by having a function pointer, which is pass 
- a reference to the ``single_sweep`` object (and should therefore know the trajectory parameters; more on this in a bit)
- a reference to the simulator so that (a) it can ask it to compute population sizes at a given time; and (b) can use its RNG instance
- And pointers to num_steps, time and allele_frequency. The generate_trajectory function will allocate and fill these out, and the calling function is responsible for freeing them.

lib/msprime.h
```diff
@@ -123,13 +123,31 @@ typedef struct {
...
+/* Forward declaration */
+struct _msp_t;
+
 typedef struct {
-    uint32_t locus;
+    double start_frequency;
+    double end_frequency;
+    double alpha;
+    double dt;
+} genic_selection_trajectory_t;
```
jeromekelleher: Different types of trajectories with different parameters are defined here. Question: do we expect there to be different ways of doing this and with different params?

lib/msprime.h
```diff
@@ -123,13 +123,31 @@ typedef struct {
...
         double *time;
         size_t num_steps;
     } trajectory;
+
+    uint32_t locus;
+    union {
+        /* Other selection trajectory simulators would go here */
+        genic_selection_trajectory_t genic_selection_trajectory;
+    } trajectory_params;
+    int (*generate_trajectory)(struct _single_sweep_t *self, struct _msp_t *simulator,
```
jeromekelleher: This is the declaration of the function pointer that for the single_sweep_t simulation model. I'm imagining this is set in the ``set_single_sweep_model`` code path somehow.

jeromekelleher:
A quick outline here @andrewkern. Is this level of generality needed, or will we just need a single type of trajectory with a fixed list of parameters for the foreseeable future? (In which case, we flatten all this down and add the trajectory parameters directly to the model).

lib/msprime.c
andrewkern: great. using the function pointer makes all the sense in the world.

lib/msprime.h
andrewkern: yes definitely we should anticipate different kinds. as you aptly named this one, we are doing genic selection, so we can ignore h, the dominance parameter that we might put in to a different kind of trajectory.

andrewkern:
i think this level of generality is really good. there are a few trajectories we could imagine generating (e.g. soft sweeps vs hard) that could be associated with different parameters sets. i'll game to work in this framework, although i'm a bit unsure about how that function pointer will be set as written.

jeromekelleher:
OK, I've made a bunch of changes here to implement this idea. Unfortunately I don't have time to finish it out properly (checking out for a few days now), but hopefully you can take a look and see how it's going so far and see if you spot any problems @andrewkern.

jeromekelleher:
I've made good progress here @andrewkern, and I think we're very close to done as far as the major infrastructure is concerned. There's still stuff to be done, but I think we should merge this much and hopefully you can make some progress on fixing the low-level problems. 

The outstanding issue is that I can't get the sweep to finish without fully coalescing  the sample, so I can't develop the infrastructure needed to finish out the simulation in Hudson mode. I'm not sure if this is because I'm choosing stupid parameters, or this is an error, or it's a fundamental aspect of the model. I've left a few questions for you lying around the place too.

Can you take a quick look through this, and if you're happy there's nothing too stupid in there we merge? I'm busy for the next few days so won't be able to get back to this for a bit. It'd be great if you could pick it up in a follow-on PR.

lib/msprime.c
```diff
@@ -4698,6 +4706,104 @@ dtwf_model_rate_to_generation_rate(simulation_model_t *MSP_UNUSED(model), double
...
+        time[num_steps] = t;
+        allele_frequency[num_steps] = x;
+        x = 1.0 - genic_selection_stochastic_forwards(
+                trajectory.dt, 1.0 - x, trajectory.alpha * current_size,
+                gsl_rng_uniform(rng));
+        t += trajectory.dt;
+        num_steps++;
+    }
+    assert(num_steps < max_steps); /* num_steps + 1 above guarantees this */
+    /* JK: question to Andy: Does this bookending make sense?
```
andrewkern: yeah i think this makes sense. need to have a harder think about it, but my knee jerk impression is this should work fine

lib/msprime.c
```diff
@@ -4950,73 +5055,57 @@ msp_set_simulation_model_beta(msp_t *self, double population_size, double alpha,
...
-        }
+    if (start_frequency >= end_frequency) {
+        ret = MSP_ERR_BAD_TRAJECTORY_START_END;
+        goto out;
+    }
+    if (dt <= 0) {
+        ret = MSP_ERR_BAD_TIME_DELTA;
+        goto out;
     }
+    /* TODO any limitations on alpha? */
```
andrewkern: alpha just needs to be positive

lib/msprime.c
```diff
@@ -4950,73 +5055,57 @@ msp_set_simulation_model_beta(msp_t *self, double population_size, double alpha,
...
-            allele_frequency, num_steps * sizeof(double));
-    memcpy(model->params.single_sweep.trajectory.time,
-            time, num_steps * sizeof(double));
-    /* TODO we need to rescale the times into model units here */
+    model->params.sweep.locus = locus;
+    model->params.sweep.generate_trajectory = genic_selection_generate_trajectory;
+    model->params.sweep.print_state = genic_selection_print_state;
+    trajectory->start_frequency = start_frequency;
+    trajectory->end_frequency = end_frequency;
+    /* JK: question for Andy: what are the dimensions of alpha? Does it
```
andrewkern: \alpha = 2Ns. no reason to rescale it.

jeromekelleher: Ah, OK. For consistency I think we should change the top-level parameter to s. There are no other scaled parameters in the Python API, and s is a lot more user-friendly than alpha. I'll leave it for the moment though. and merge this now so you can get working.

jeromekelleher:
Great, thanks @andrewkern! I'm going to leave the questions in the code - can you delete them if you're happy we've resolved the issues on your next pass please?

PR #759, opened 2019-05-15
Allow multiple historical samples per generation in DTWF

DomNelson:
Similar to what was happening with demographic events, only one historical sample would be added per generation. Now make sure all with `time <= self->time` are added.


jeromekelleher:
Nice catch, thanks @DomNelson. Merging.

PR #763, opened 2019-05-24
added a few more tests to verify.py

andrewkern:
minor additions to the verification.py tests using `discoal`

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -274,10 +274,25 @@ class SimulationVerifier(object):
             self._run_mutation_stats(key, command_line)
         self._instances[key] = f
 
-    def _run_mutation_discoal_stats(self, key, args):
+    def _discoal_str_to_ms(self, args):
         # convert discoal string to msprime string
         tokens = args.split(" ")
-        msp_str = " ".join(tokens[0:2]+tokens[3:])
+        # cut out sites param
+        del(tokens[2])
+        # adjust popIDs
+        for i in range(len(tokens)):
+            # pop size change case
+            if tokens[i] == "-en":
+                tokens[i+2] = str(int(tokens[i + 2]) + 1)
+            # migration rate case
+            if tokens[i] == "-m":
+                tokens[i+1] = str(int(tokens[i + 1]) + 1)
+                tokens[i+2] = str(int(tokens[i + 2]) + 1)
+        msp_str = " ".join(tokens)
+        return(msp_str)
+
+    def _run_mutation_discoal_stats(self, key, args):
+        msp_str = self._discoal_str_to_ms(args)
         df_msp = self._run_msprime_mutation_stats(msp_str)
         df_d = self._run_discoal_mutation_stats(args)
         self._plot_stats(key, "mutation", df_d, df_msp)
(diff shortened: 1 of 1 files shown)


verification.py
```diff
@@ -272,10 +272,25 @@ def f():
             self._run_mutation_stats(key, command_line)
         self._instances[key] = f
 
-    def _run_mutation_discoal_stats(self, key, args):
+    def _discoal_str_to_ms(self, args):
         # convert discoal string to msprime string
         tokens = args.split(" ")
-        msp_str = " ".join(tokens[0:2]+tokens[3:])
+        # cut out sites param
```
jeromekelleher: Not sure what this is doing @andrewkern - is the idea that we're creating an ms-like command line for msprime to run? If so we need to take a different approach I think, as I'm not planning on adding support for sweeps to the mspms interface (it's already super complicated, and I don't want to touch it unless absolutely necessary).

verification.py
```diff
@@ -272,10 +272,25 @@ def f():
...
+        del(tokens[2])
```
jeromekelleher: FYI, ``del`` is a statement not a function so the brackets are uncessary here.

verification.py
```diff
@@ -272,10 +272,25 @@ def f():
...
+        for i in range(len(tokens)):
+            # pop size change case
+            if tokens[i] == "-en":
+                tokens[i+2] = str(int(tokens[i + 2]) + 1)
+            # migration rate case
+            if tokens[i] == "-m":
+                tokens[i+1] = str(int(tokens[i + 1]) + 1)
+                tokens[i+2] = str(int(tokens[i + 2]) + 1)
+        msp_str = " ".join(tokens)
+        return(msp_str)
```
jeromekelleher: Similary for return - should be ``return msp_str``

verification.py
andrewkern: ugh sorry.

verification.py
andrewkern: yes that's the idea-- i'm translating a `discoal` string to `ms`. how should i proceed instead? btw these checks that i added all looked good to me, so at least `msprime` and `discoal` are giving similar results for these neutral examples

andrewkern:
happy to change to that. should I take a stab at it and then you can review my revisions?

jeromekelleher:
> happy to change to that. should I take a stab at it and then you can review my revisions?

Sounds great, thanks. You might be able to add some new options to the parser for ms command lines [here](https://github.com/tskit-dev/msprime/blob/master/msprime/cli.py#L862) which might make life a bit easier. .e.g.
```
parser = msprime.cli.get_mspms_parser()
# Add new options to parser
args = parser.parse_args(cmd.split())
# Do something with args
```

If there's only a few extra options and the changes needed are small, I could be convinced to add support for ``discoal`` compatability to ``mspms``. That would make this verification stuff easy to do then. Is this something you'd be interested in @andrewkern?

jeromekelleher:
I'm going to merge this @andrewkern to clear up. If there's still some outstanding issues can you open a follow-up PR please?

PR #770, opened 2019-06-04
Likelihood

JereKoskela:
I've managed to simplify both likelihood functions a fair bit, and also included a much more substantial test.

### DIFF
--- a/msprime/likelihood.py
+++ b/msprime/likelihood.py
@@ -19,9 +19,9 @@
 """
 Module responsible for computing likelihoods.
 """
+import collections
 import math
 import msprime
-import numpy as np
 
 
 def unnormalised_log_mutation_likelihood(arg, theta):
@@ -29,110 +29,62 @@ def unnormalised_log_mutation_likelihood(arg, theta):
     # that depends on the pattern of observed mutations, but not on the ARG
     # or the mutation rate
     tables = arg.tables
-    number_of_edges = len(tables.edges)
-    edges_above = {}
-    edge = 0
+    edges_above = collections.defaultdict(list)
+    edges_below = collections.defaultdict(list)
     total_material = 0
-    while edge < number_of_edges:
-        child = tables.edges[edge].child
-        right = tables.edges[edge].right
-        left = tables.edges[edge].left
-        parent = tables.edges[edge].parent
-        total_material += ((right - left) *
-                           (tables.nodes[parent].time - tables.nodes[child].time))
-        edges_above[child] = [edge]
-        edge += 1
-        if edge == number_of_edges:
-            break
-        while tables.edges[edge].child == child:
-            right = tables.edges[edge].right
-            left = tables.edges[edge].left
-            parent = tables.edges[edge].parent
-            total_material += ((right - left) *
-                               (tables.nodes[parent].time - tables.nodes[child].time))
-            edges_above[child].append(edge)
-            edge += 1
-            if edge == number_of_edges:
-                break
-    edges_below = {}
-    edge = 0
-    while edge < number_of_edges:
-        parent = tables.edges[edge].parent
-        edges_below[parent] = [edge]
-        edge += 1
-        if edge == number_of_edges:
-            break
-        while tables.edges[edge].parent == parent:
-            edges_below[parent].append(edge)
-            edge += 1
-            if edge == number_of_edges:
-                break
+    time = tables.nodes.time
+    for edge in tables.edges:
+        total_material += ((edge.right - edge.left) *
+                           (time[edge.parent] - time[edge.child]))
+        edges_above[edge.child].append(edge)
+        edges_below[edge.parent].append(edge)
     number_of_mutations = len(tables.mutations)
-    if number_of_mutations == 0 and theta == 0:
-        ret = -float("inf")
+    if theta == 0:
+        if number_of_mutations == 0:
+            ret = 0
+        else:
+            ret = -float("inf")
     else:
         ret = (number_of_mutations * math.log(total_material * theta) -
                total_material * theta)
     for mut in arg.mutations():
         mutant_location = tables.sites[mut.site].position
-        ind = 0
-        edge = edges_above[mut.node][0]
-        while tables.edges[edge].right < mutant_location:
-            ind += 1
-            edge = edges_above[mut.node][ind]
-        parent = tables.edges[edge].parent
-        child = tables.edges[edge].child
-        potential_branch_length = tables.nodes[parent].time - tables.nodes[child].time
-        continue_downwards = True
-        if tables.nodes[child].time == 0:
-            continue_downwards = False
+        index = 0
+        to_check = edges_above[mut.node]
+        while to_check[index].right < mutant_location:
+            index += 1
+        edge = edges_above[mut.node][index]
+        potential_branch_length = time[edge.parent] - time[edge.child]
+        continue_leafwards = time[edge.child] > 0
         tmp_edge = edge
-        while continue_downwards:
+        while continue_leafwards:
             count = 0
-            to_check = edges_below[child]
-            for e in to_check:
-                if (tables.edges[e].left <= mutant_location and
-                        mutant_location < tables.edges[e].right):
+            for e in edges_below[tmp_edge.child]:
+                if e.left <= mutant_location and mutant_location < e.right:
                     count += 1
                     tmp_edge = e
             if count == 1:
-                parent = tables.edges[tmp_edge].parent
-                child = tables.edges[tmp_edge].child
-                potential_branch_length += (tables.nodes[parent].time -
-                                            tables.nodes[child].time)
-                if tables.nodes[child].time == 0:
-                    continue_downwards = False
+                potential_branch_length += time[tmp_edge.parent] - time[tmp_edge.child]
+                continue_leafwards = time[tmp_edge.child] > 0
             else:
-                continue_downwards = False
-        continue_upwards = True
+                continue_leafwards = False
         tmp_edge = edge
-        parent = tables.edges[edge].parent
-        child = tables.edges[edge].child
-        if parent not in edges_above:
-            continue_upwards = False
-        while continue_upwards:
-            count_up = 0
-            to_check = edges_above[parent]
-            for e in to_check:
-                if (tables.edges[e].left <= mutant_location and
-                        mutant_location < tables.edges[e].right):
-                    count_up += 1
-                    tmp_edge = e
-            count_down = 0
-            to_check = edges_below[parent]
-            for e in to_check:
-                if (tables.edges[e].left <= mutant_location and
-                        mutant_location < tables.edges[e].right):
-                    count_down += 1
-            if count_up == 1 and count_down == 1:
-                parent = tables.edges[tmp_edge].parent
-                child = tables.edges[tmp_edge].child
-                potential_branch_length += (tables.nodes[parent].time -
-                                            tables.nodes[child].time)
-                if parent not in edges_above:
-                    continue_upwards = False
+        continue_rootwards = tmp_edge.parent in edges_above
+        while continue_rootwards:
+            count = 0
+            for e in edges_below[tmp_edge.parent]:
+                if e.left <= mutant_location and mutant_location < e.right:
+                    count += 1
+            if count == 1:
+                to_check = edges_above[tmp_edge.parent]
+                index = 0
+                while to_check[index].right < mutant_location:
+                    index += 1
+                tmp_edge = to_check[index]
+                potential_branch_length += time[tmp_edge.parent] - time[tmp_edge.child]
+                continue_rootwards = tmp_edge.parent in edges_above
             else:
-                continue_upwards = False
+                continue_rootwards = False
         ret += math.log(potential_branch_length / total_material)
     return ret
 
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Looks good @JereKoskela. Here's some changes we'd want to get it to fit in to the overall framework:

1. Put the main functions we're using in a new file, ``msprime/likelihood.py`` (which includes the standard headers). You'll need to update ``msprime/__init__.py`` to make the functions in this module visible in the msprime module namespace.
2. Take the testing code and put it into a file ``tests/test_likelihood.py``. See, e.g., https://github.com/tskit-dev/msprime/blob/master/tests/test_demography.py for an example of how these tests are laid out usually. I'm assuming this is an example that you've worked out by hand, so we could have a single test to start with that builds this tree sequence and compares the output (using np.allclose, or similar) to the pre-calculated value. We wouldn't want any plotting in here (or the main code module).

I wouldn't worry so much about performance for now, just getting the APIs and as many tests as we can think of in place so that we can easily test the output when we start thinking about performance.

Does this sound OK?

JereKoskela:
> Looks good @JereKoskela. Here's some changes we'd want to get it to fit in to the overall framework:
> 
> 1. Put the main functions we're using in a new file, `msprime/likelihood.py` (which includes the standard headers). You'll need to update `msprime/__init__.py` to make the functions in this module visible in the msprime module namespace.
> 2. Take the testing code and put it into a file `tests/test_likelihood.py`. See, e.g., https://github.com/tskit-dev/msprime/blob/master/tests/test_demography.py for an example of how these tests are laid out usually. I'm assuming this is an example that you've worked out by hand, so we could have a single test to start with that builds this tree sequence and compares the output (using np.allclose, or similar) to the pre-calculated value. We wouldn't want any plotting in here (or the main code module).
> 
> I wouldn't worry so much about performance for now, just getting the APIs and as many tests as we can think of in place so that we can easily test the output when we start thinking about performance.
> 
> Does this sound OK?

Thanks, that sounds good to me. I should be able to get a new version up relatively quickly, possibly even tomorrow.

JereKoskela:
Ok, so an update took a fair while longer than I had hoped. The main reason was a conceptual flaw in the mutation likelihood: when we store common ancestor events and recombinations as edges, it is no longer true that each infinite sites mutation takes place on a unique edge. Rather, it can happen on a set of contiguous edges. The new method finds all of these edges for each mutation in the continue_upwards and continue_downwards loops, and thus integrates over all possible locations of each mutation.

I'm envisaging that the dictionaries I build in the mutation likelihood (lines 11-46 of likelihood.py) will eventually be part of the data structure, presumably stored as binary trees so that we can search and modify them efficiently, and thus won't need to be built in the function.

I'll fix the CircleCI niggles in a further update.

msprime/likelihood.py
```diff
@@ -0,0 +1,195 @@
...
+
+def unnormalised_log_mutation_likelihood(arg, theta):
+    # log_likelihood of mutations on a given ARG up to a normalising constant
+    # that depends on the pattern of observed mutations, but not on the ARG
+    # or the mutation rate
+    tables = arg.tables
+    number_of_edges = len(tables.edges)
+    edges_above = {}
+    edge = 0
+    total_material = 0
```
jeromekelleher: It would save a bit of typing and be more efficient to do it like this:
```python
edges_above = collections.defaultdict(list)
edges_below = collections.defaultdict(list)
time = tables.nodes.time
for j, edge in enumerate(tables.edges):   
    total_material += (edge.right - edge.left) * (time[edge.parent] - time[edge.child]])
    edges_above[edge.child].append(j)
    edges_below[edge.parent].append(j)    
```

msprime/likelihood.py
```diff
@@ -0,0 +1,195 @@
...
+        if number_of_mutations == 0:
+            ret = 0
+        else:
+            ret = -float("inf")
+    else:
+        ret = (number_of_mutations * math.log(total_material * theta) -
+               total_material * theta)
+    for mut in arg.mutations():
+        mutant_location = tables.sites[mut.site].position
+        ind = 0
```
jeromekelleher: What's ``ind`` for here?

msprime/likelihood.py
jeromekelleher: Also, might be easier to add the actual edge objects to these lists rather than the indexes. Would save looking up the tables again afterwards.

msprime/likelihood.py
```diff
@@ -0,0 +1,195 @@
...
+            ind += 1
+            edge = edges_above[mut.node][ind]
+        parent = tables.edges[edge].parent
+        child = tables.edges[edge].child
+        potential_branch_length = tables.nodes[parent].time - tables.nodes[child].time
+        continue_downwards = True
+        if tables.nodes[child].time == 0:
+            continue_downwards = False
+        tmp_edge = edge
+        while continue_downwards:
```
jeromekelleher: Can you explain in words what this is doing? I don't quite get it, and I think it can be made a bit more concise.

msprime/likelihood.py
```diff
@@ -0,0 +1,195 @@
...
+            if gap == 0:
+                # we evaluate the density rather than the probability if there is no gap
+                gap = 1
+            if rho == 0:
+                ret -= float("inf")
+            else:
+                ret += math.log(rho * gap)
+        else:
+            segment_length_in_children = -tables.edges.left[edge]
+            while tables.edges[edge].child == child:
```
jeromekelleher: more concise and idiomatic to say:
```python
while edge < num_edges and edges[edge].child == child:
    edge += 1
```
rather than having a break at the tail of the loop.

msprime/likelihood.py
```diff
@@ -0,0 +1,195 @@
...
+                edge += 1
+                if edge == number_of_edges:
+                    break
+            segment_length_in_children += (tables.edges.right[edge - 1] -
+                                           tables.edges.left[edge])
+            child = tables.edges[edge].child
+            while tables.edges[edge].child == child:
```
jeromekelleher: Ditto here.

msprime/likelihood.py
JereKoskela: I agree about storing the edge objects rather than indices. As for building both dictionaries in one loop, I don't think that would be straightforward. In general, every entry of edges_above and edges_below will be a list of multiple edges, and the behaviour of the body of the loop needs to be different based on whether the current edge will be a new entry in the dictionary, or appended to an existing one. Currently, the while-loops on lines 47 and 65 scroll through edges that don't need new entries, and append them to the correct existing entry. The loops will trigger different numbers of times for edges_below and edges_above, respectively, so combining them into one would would involve extra bookkeeping at a minimum, and probably require looking through both dictionaries at each iteration to check whether or not a given edge already has an entry.

msprime/likelihood.py
JereKoskela: A dummy variable for finding the edge in `edges_above[mut.node]` that is ancestral at `tables.sites[mut.site].position` using the loop at line 83. There's probably a more elegant way to do that though.

msprime/likelihood.py
JereKoskela: For the sake of paranoia: the second condition won't get checked if the first one fails? I never remember which language does what in those cases.

msprime/likelihood.py
JereKoskela: Each infinite sites mutation can appear on one of a set of contiguous edges, and I'm assuming it has been recorded on an arbitrary valid edge. The job of the `while continue_downwards` and `while continue_upwards` loops starting on respective lines 93 and 116 is to find the extremal valid edges; more specifically, the total amount of branch length in between. Continue_downwards goes towards the leaves, continue_upwards towards the root. I suppose that already suggests better variable names for the next update.

Going leafwards, the valid window of edges ends is if the current child node is the parent of a coalescence event, and both of its children carry ancestral material at the mutant site. That means the mutation couldn't have happened any further leafward, while still appearing on all necessary leaves. The variable `count` counts how many leafward edges carry ancestral material, and we stop looking when it fails to equal one. The amount of available branch length for the mutation gets stored in `potential_branch_length`.

Going rootwards, the window of valid edges can end in two ways. The current parent node could have two offspring (one of which is the current child node), both of which carry ancestral material at the mutant site (analogously to how the leafward case terminated). Alternatively, no edge above the current parent node carries ancestral material at the mutant site, in which case it has reached its MRCA. Now that I write that out, I realise that the former condition is implied by the latter, so this can indeed be simplified. As it is, the variable `count_down` checks the former condition, and `count_up` the latter.

msprime/likelihood.py
jeromekelleher: This is handled automatically by the defaultdict though, isn't it? C code will look quite different anyway, the aim here would be to get the cleanest, simplest Python expression of the algorithm IMO.

msprime/likelihood.py
jeromekelleher: This is safe, ``and`` guarantees the order in which things are executed and short-circuits. C has the same behaviour.

msprime/likelihood.py
jeromekelleher: OK, that's helpful, thanks. I think another iteration where we store the edge objects themselves in the lookup dicts will help to clear things up a little. Would you mind making these changes please?

JereKoskela: Not at all. In fact, many of them are already made. I'll read up on defaultdict and see if it's cleverer than I'm giving it credit for, and then upload a new version.

JereKoskela:
Ok @jeromekelleher, I've made all the tweaks we talked about above. defaultdict did make things considerably cleaner, and I've also added an `edges_above` dictionary to the `log_arg_likelihood` method to handle the check on whether a given parent node is an MRCA, or has further parents of its own (line 133 of likelihood.py). That makes the two log likelihood functions more similar, and it will be easy to cleanly remove the loop that builds `edges_above` if/when the dictionary is more permanently in place.

msprime/likelihood.py
```diff
@@ -0,0 +1,141 @@
...
+        ret = (number_of_mutations * math.log(total_material * theta) -
+               total_material * theta)
+    for mut in arg.mutations():
+        mutant_location = tables.sites[mut.site].position
+        index = 0
+        to_check = edges_above[mut.node]
+        while to_check[index].right < mutant_location:
+            index += 1
+        edge = edges_above[mut.node][index]
+        potential_branch_length = time[edge.parent] - time[edge.child]
```
jeromekelleher: Wouldn't it be simpler to compute this using the trees? We're basically iterating up and down the tree at the position for each mutation, aren't we? For each edge we examine, we're only looking at the branch length right, not the 'area' of the edge?

So,
```python
for tree in arg.trees():
    for site in tree.sites():
        assert len(site.mutations) == 1
        mutation = site.mutations[0]
        # Compute ``potential_branch_length`` by traversing up and down the tree?
        u = tree.parent(mutation.node)
        potential_branch_length = 0
        # Traverse up
        while u != tskit.NULL:
             potential_branch_length += tree.branch_length(u) # branch length of a root is 0
         # Traverse down (but we could do this without traversals if this is what we want)
         for u in tree.nodes(mutation.node):
             potential_branch_length += tree.branch_length(u)
```

JereKoskela: It is be possible to iterate over trees as well, and you're right that we only want the branch length, not the edge area. It's not quite as simple as your code though, because we still need to detect the start and end points of the window of valid of edges that are ancestral to exactly the right set of leaves at the mutant site. That window will typically not span the whole tree unless I'm dramatically misunderstanding the data structure. I can have a think about trying to write the function this way, and see if it makes a difference.

I had missed that `branch_length` is an implemented method. That will certainly make the code a little more readable as well.

jeromekelleher: Ah, I see. This [issue](https://github.com/tskit-dev/tskit/issues/186) has some discussion on how to get edge information for tree nodes, which may be useful.

I think it's probably useful to recast this in terms of trees all right though since we are inherently doing tree traversals.

JereKoskela:
You were right @jeromekelleher: iterating over trees is significantly simpler. On a fixed tree, the end points of the valid window of edges can be detected based solely on numbers of children. As soon as a node has two children, then the number of leaves its lineage subtends changes, and we know the window is over.

jeromekelleher:
> You were right @jeromekelleher: iterating over trees is significantly simpler. On a fixed tree, the end points of the valid window of edges can be detected based solely on numbers of children. As soon as a node has two children, then the number of leaves its lineage subtends changes, and we know the window is over.

Well done, it's really beautiful now!

I'm happy to merge this now if you like, and we can then work on finalising the user interface and C API as a follow up. Or, if you prefer to keep working on this PR that's fine too.

We will want to rebase and clean up the history a little bit at some point. See [here](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow) for a write up on how to rebase and squash commits. We don't need to squash down to one commit, but but it would be good to weed out commits that are just to keep PEP8 happy, etc.

JereKoskela:
I'm happy to get this merged. Likelihood functions seem like useful features to have available in general.

Do we need to rebase before merging, or can it be done afterwards as well? It will take me a little while to get to grips with what I'm supposed to do to make it happen.

Regarding next steps, would a sensible start be to build similar functions in C into lib/msprime.c, or something else?

jeromekelleher:
We need to rebase and clean up the history before merging. Once something is merged into upstream/master the history is never modified afterwards. Follow the steps in the guide I wrote above on rebasing, it should be fairly thorough. Ping me again once it's rebased and force-pushed and I'll merge.

> Regarding next steps, would a sensible start be to build similar functions in C into lib/msprime.c, or something else?

I think we should probably have a ``likelihood.c`` file - msprime.c is a bit of a beast, don't want to make it any bigger and this is a logically separate thing. We should be able to use the [tskit C API](https://tskit.readthedocs.io/en/stable/c-api.html) to follow these Python functions pretty closely.

JereKoskela:
Ping @jeromekelleher. I think I managed to follow your instructions (you were right: not as cryptic as I feared). The number of commits has changed and nothing seems to have broken, so I think my time warp worked.

jeromekelleher:
Thanks @JereKoskela; merging!

PR #774, opened 2019-06-12
Increase SLiM generations to 10x first coalescence time

DomNelson:
Previously ran until coalescence then continuing for the same number of generations, but this still biased coalescence times in some scenarios. Now run for 10x the time to first coalescence.


jeromekelleher:
Great, thanks @DomNelson.

PR #779, opened 2019-06-21
Coalescence time calculation

petrelharp:
For stdpopsim, I have implemented a method that uses the 'ground truth' coalescence rate calculations to compute mean time until coalescence, for comparison with methods that infer a single Ne at a particular time. It just uses the relationship that
 E[T] = int_0^infty P(T > t) dt
and we're already computing P(T > t).

That said, in implementing this I (a) found and fixed a number of subtle bugs in our calculations of P(T > t), and (b) got terribly mixed up with endpoints and the like. But, now it's working.  Also note that it does automatic breakpoint computation.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -28,6 +28,7 @@ import sys
 import os
 import warnings
 import copy
+import logging
 
 import tskit
 import numpy as np
@@ -50,6 +51,9 @@ from _msprime import MutationGenerator
 # _msprime.restore_gsl_error_handler(), which will set the error handler to
 # the value that it had before this function was called.
 _msprime.unset_gsl_error_handler()
+
+
+logger = logging.getLogger(__name__)
 
 
 Sample = collections.namedtuple(
@@ -1366,7 +1370,7 @@ def _matrix_exponential(A):
     Yinv = np.linalg.pinv(Y)
     D = np.diag(np.exp(d))
     B = np.matmul(Y, np.matmul(D, Yinv))
-    return B
+    return np.real_if_close(B, tol=1000)
 
 
 class DemographyDebugger(object):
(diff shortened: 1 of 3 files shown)


msprime/simulations.py
```diff
@@ -1528,6 +1528,88 @@ def population_size_trajectory(self, steps):
...
+                    + np.sum(dt[nz] * dP[nz] / dlogP[nz]))
+
+        if steps is None:
+            epochs = [e.start_time for e in self.epochs]
+            steps = sorted(list(set(np.linspace(0, 1.2 * max(epochs) + 1,
+                                                101)).union(set(epochs))))
+        p_diff = m_diff = np.inf
+        last_P = np.inf
+        n = 0
+        while n < max_iter and (last_P > rtol or p_diff > rtol/20 or m_diff > rtol/20):
```
jeromekelleher: What's 20 here? Is it the same 20 as below in np.linspace? It's be nice to de-magic number this a little bit.

msprime/simulations.py
```diff
@@ -1528,6 +1528,88 @@ def population_size_trajectory(self, steps):
...
+        """
+
+        def mean_time(steps, P):
+            # Mean is int_0^infty P(T > t) dt, which we estimate by discrete integration
+            # assuming that f(t) = P(T > t) is piecewise exponential:
+            # if f(u) = a exp(bu) then b = log(f(t)/f(s)) / (t-s) for each s < t, so
+            # \int_s^t f(u) du = (a/b) \int_s^t exp(bu) b du = (a/b)(exp(bt) - exp(bs))
+            #    = (t - s) * (f(t) - f(s)) / log(f(t) / f(s))
+            # unless b = 0, of course.
+            assert(steps[0] == 0)
```
jeromekelleher: Nit picking here, but ``assert`` isn't a function. Should write ``assert steps[0] == 0``.

msprime/simulations.py
petrelharp: It is not the same as the 20 below, and I unfortunately don't have a good explanation for it: clearly we need to be approximating P to somewhat better than rtol to get an answer that is within rtol, and there's no good reason for the factor of 20 other than that it seemed to work. I will add a comment.

msprime/simulations.py
petrelharp: I always forget which.

petrelharp:
I've added a verification (which caught an additional sometimes-issue where matrix exponential was returning floating-point imaginary values):
mean_coaltimes

petrelharp:
I think we're good now; this is pretty well tested. I will wait for tests to run then rebase.

petrelharp:
Rebased; I think we should merge this now?

andrewkern:
when this gets merged, we will want to make sure to track the version onto the stdpopsim/analysis requirements

petrelharp:
Whoops; one last change that improves the initial conditions and the error output. Now this works:
```
import msprime
from stdpopsim import homo_sapiens

model = homo_sapiens.GutenkunstThreePopOutOfAfrica()
ddb = msprime.DemographyDebugger(**model.asdict())
m22 = ddb.mean_coalescence_time([0,0,2], rtol=.05, verbose=True)
```

msprime/simulations.py
```diff
@@ -1528,6 +1528,107 @@ def population_size_trajectory(self, steps):
...
+                            steps=steps, num_samples=num_samples,
+                            min_pop_size=min_pop_size,
+                            double_step_validation=False)
+            m2 = mean_time(steps, P2)
+            keep_steps = np.in1d(steps, last_steps)
+            p_diff = max(np.abs(P1 - P2[keep_steps]))
+            m_diff = np.abs(m1 - m2)/m2
+            last_P = P2[-1]
+            n += 1
+            if verbose:
```
jeromekelleher: Not so sure about this ``verbose`` option - this information should be logged really (i.e., logging.debug(...)). Outputting via prints makes it a pain to test. The output is pretty opaque anyway unless you're actually reading the code, so not sure how useful this is in general. I would vote to just comment out these prints to keep them handy for debugging in the future and remove the ``verbose`` option.

petrelharp: Hm. I agree, it feels hacky, but we need something to give users a clue what's going on when the automatic discretisation-finder doesn't converge. Pretty much every optimiser or numerical integrator has an option like this, to print out semi-opaque information about what it's doing under the hood. It's messy, but general-purpose numerical integration is unavoidably messy sometimes. Other options:

1. I document the refinement method and what's being printed better. Downside: this will be lengthy.
2. The diagnostic information only gets printed if a "failure to converge" error is thrown.
3. The `verbose` option is changed to `_verbose` and undocumented.

I don't like (2) because it's helpful to look at that output to make sure that nothing wierd is happening even if it thinks it has converged.

About testing: I don't think we need to actually test that it is outputting the correct numbers, since as you say it's useful for debugging and probably needs looking at the code. I have put in a test that *uses* the `verbose` option to make sure it doesn't fail. So, my vote is I guess for a bit more documentation and then leave it alone?

jeromekelleher: I've added a commit using the logger to capture this information; there's an example there as well to show how to use it. There's no need to test the content of the debug output.

If you prefer the existing way, just rebase and drop my commit. Please make sure that the ``verbose`` code path tests don't write out strings to stdout when running ``python3 -m nose -vs``, though.

petrelharp:
Thanks for the logging: good call. The instructions in the docstring didn't work for me, though; I've changed it to something that does work for me. If tests pass, I'll rebase/squash and then I think we're ready to merge?

jeromekelleher:
> Thanks for the logging: good call. The instructions in the docstring didn't work for me, though; I've changed it to something that does work for me. If tests pass, I'll rebase/squash and then I think we're ready to merge?

Hmm, the current version doesn't work for me. I'm surprised the original version didn't work for you - this should be the simplest way of turning on logging to stderr. I don't see how it could work without calling ``basicSetup()``, and I don't think the user should be interacting with the module's logger (I was amazed this didn't crash when I looked at it, but then realised everything in ``msprime/simulations.py`` is exported into the global namespace).

jeromekelleher:
Huh, now I remember why I started using [daiquiri](https://daiquiri.readthedocs.io/en/latest/); Python logging setup is complicated!

We could change the guidance to 
```python
import daiquiri
daiquiri.setup(level="DEBUG")
# do stuff
```` 

This *always* works in my experience, and you get nice coloured output which makes everyone happy.

petrelharp:
wtf. What I found was happening was that changing the logging level with `basicSetup()` was affecting `logging.root`'s level (ie the logger in the global namespace) but *not* `msprime.logging.root` (ie the one that msprime was using).  I'll give daiquiri a try.

petrelharp:
`daiquiri` worked for me. I've updated the guidance. Are we good to merge if the tests pass?

jeromekelleher:
Great, thanks @petrelharp!

PR #781, opened 2019-07-05
adds gene conversion (python version)

fbaumdicker:
This adds gene conversion to the python algorithm. The rates are computed just as in Hudson's ms. I.e. there are two rates for gene conversion events.
One rate corresponds to gene conversions where the left starting point falls within one of the segments or between two consecutive segments and the length is geometrically distributed.
The second rate corresponds to gene conversions where the left starting point is left of the first segment and the length is geometrically distributed conditioned such that the right endpoint of the conversion tract falls within or between segments.
Future versions might split the first rate into two further rates to speed up the simulation.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -896,22 +896,22 @@ class Simulator(object):
         """
         h = random.randint(1, self.L[label].get_total())
         # generate tracklength
-        l = np.random.geometric(1/self.tracklength)
+        tl = np.random.geometric(1/self.tracklength)
         # Get the segment containing the h'th link
         y = self.segments[self.L[label].find(h)]
         k = y.right - self.L[label].get_cumulative_frequency(y.index) + h - 1
         # check if the gene conversion falls between segments --> no effect
-        if y.left >= k+l:
+        if y.left >= k+tl:
             # print("noneffective GCI EVENT")
             return None
         self.num_gc_events += 1
         x = y.prev
         # both breaks are within the same segment
-        if k+l < y.right:
+        if k+tl < y.right:
             if k <= y.left:
                 y.prev = None
                 z2 = self.alloc_segment(
-                    k+l, y.right, y.node, y.population, x, y.next)
+                    k+tl, y.right, y.node, y.population, x, y.next)
                 if x is not None:
                     x.next = z2
                     lhs = x.right
@@ -923,14 +923,14 @@ class Simulator(object):
                 if y.next is not None:
                     y.next.prev = z2
                 y.next = None
-                y.right = k+l
-                self.L[label].increment(y.index, k + l - z2.right)
+                y.right = k+tl
+                self.L[label].increment(y.index, k + tl - z2.right)
                 z = y
             elif k > y.left:
                 z = self.alloc_segment(
-                    k, k+l, y.node, y.population, None, None)
+                    k, k+tl, y.node, y.population, None, None)
                 z2 = self.alloc_segment(
-                    k+l, y.right, y.node, y.population, y, y.next)
+                    k+tl, y.right, y.node, y.population, y, y.next)
                 if y.next is not None:
                     y.next.prev = z2
                 y.next = z2
@@ -942,7 +942,7 @@ class Simulator(object):
         else:
             # Get the segment y2 containing the end of the conversion tract
             y2 = y
-            while y2 is not None and k + l >= y2.right:
+            while y2 is not None and k+tl >= y2.right:
                 y2 = y2.next
             # process left break
             if k <= y.left:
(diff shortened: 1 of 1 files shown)


algorithms.py
```diff
@@ -825,6 +890,160 @@ def hudson_recombination_event(self, label, return_heads=False):
...
+            if k <= y.left:
+                y.prev = None
+                z2 = self.alloc_segment(
+                    k+tl, y.right, y.node, y.population, x, y.next)
+                if x is not None:
+                    x.next = z2
+                    lhs = x.right
+                    lhs_tail = x
+                else:
+                    print("this should never be possible")
```
jeromekelleher: An assert would be better here. Also, why the line of code immediately after?

algorithms.py
```diff
@@ -307,7 +333,8 @@ def __init__(
             population_growth_rate_changes, population_size_changes,
             migration_matrix_element_changes, bottlenecks, model='hudson',
             from_ts=None, max_segments=100, num_labels=1, sweep_trajectory=None,
-            full_arg=False, time_slice=None):
+            full_arg=False, time_slice=None, geneconversion_rate=0.0,
```
jeromekelleher: Let's change this to ``gene_conversion_rate``  and ``gene_conversion_length`` for consistency please.

algorithms.py
```diff
@@ -825,6 +890,160 @@ def hudson_recombination_event(self, label, return_heads=False):
...
+            # process right break
+            if y2 is not None:
+                if y2.left < k + tl:
+                    z2 = self.alloc_segment(
+                        k + tl, y2.right, y2.node, y2.population, lhs_tail, y2.next)
+                    if lhs_tail is not None:
+                        lhs_tail.next = z2
+                        lhs = lhs_tail.right
+                    else:
+                        print("this should never be possible(2)")
```
jeromekelleher: Ditto comment above about assert.

algorithms.py
```diff
@@ -825,6 +890,160 @@ def hudson_recombination_event(self, label, return_heads=False):
...
+                self.L[label].increment(y.index, k - z2.right)
+                lhs_tail = y
+        # breaks are in separate segments
+        else:
+            # Get the segment y2 containing the end of the conversion tract
+            y2 = y
+            while y2 is not None and k+tl >= y2.right:
+                y2 = y2.next
+            # process left break
+            if k <= y.left:
```
jeromekelleher: This if/else block looks very similar to the one above - could this logic be pulled out into a helper function?

algorithms.py
```diff
@@ -825,6 +890,160 @@ def hudson_recombination_event(self, label, return_heads=False):
...
+        """
+        Implements a gene conversion event that started left of a first segment.
+        """
+        self.num_gc_events += 1
+        h = random.uniform(0, self.get_cleft_total(self.tracklength))
+        # Get segment where gene conversion starts from left and length of the individual
+        index, distance = self.find_cleft_individual(h, self.tracklength)
+        y = self.segments[index]
+        # generate tracklength
+        k = y.left + math.floor(1.0 +
```
jeromekelleher: What's the distribution of the value we're adding to y.left? I thought it would be a geometric of some sort.

algorithms.py
fbaumdicker: done

algorithms.py
fbaumdicker: I changed this to an assert.
This case is not impossible but handled in a different branch of the algorithm. The line of code immediately after would be necessary if the case would get to this point.

algorithms.py
fbaumdicker: see above

algorithms.py
fbaumdicker: Good idea. I wrote a helper function cut_right_break to do so.

algorithms.py
fbaumdicker: It is a geometrically distributed RV, but conditioned to be smaller than distance. This way the gene conversion will never cover the whole individual from left to right, which would have no effect anyway.

fbaumdicker:
I would go for 2. (implementing this as it is in C first) as well.
Then we can use the verification.py script to compare it efficiently with ms output.
Or is there a similar verification script for the python version @jeromekelleher?
With the verification.py script we could compare it to ms output. In addition, we should check for some of the theoretical results for the covariance between branch lengths from Wiufs TPB paper doi:10.1006/tpbi.2000.1462 .

jeromekelleher:
OK, let's got with option 2 then @fbaumdicker. I'll take a another look through this and get back to you with a proposal for how to move forward.

Are you going to SMBE by any chance?

fbaumdicker:
Great! I am looking forward to your response. Unfortunately, I am not going to SMBE.

jeromekelleher:
OK, I'm going to merge this much so @fbaumdicker. I'll see what needs to be done next then. Thanks - this is great progress!

Issue #783, opened 2019-07-09
Very rarely, waiting time until coalesence = 0. What to do?


jeromekelleher:
In very, very, very rare cases msprime can generate a zero waiting time until the next coalescent event, resulting in an error being thrown. This script for example, does it:

```python
import msprime as msp

def ancient_sample_test(
        num_modern=1000, anc_pop = 0, anc_num = 1, anc_time = 200, split_time_anc = 400,
        Ne0 = 10000, Ne1 = 10000, length = 1000):
    samples = [msp.Sample(population = 0, time = 0)]*num_modern
    samples.extend([msp.Sample(population = anc_pop, time = anc_time)]*(2*anc_num))
    pop_config = [msp.PopulationConfiguration(initial_size = Ne0), msp.PopulationConfiguration(initial_size = Ne1)]
    divergence = [msp.MassMigration(time = split_time_anc, source = 1, destination = 0, proportion = 1.0)]
    seed = 94320219
    sims = msp.simulate(
        samples=samples,Ne=Ne0,population_configurations=pop_config,
        demographic_events=divergence, length=length,
        random_seed=seed)

if __name__ == "__main__":
    num_ind = 10
    ancient_sample_test(
        num_modern=100,anc_pop=0,anc_num=num_ind,Ne0=3000,Ne1=3000,anc_time=5419,split_time_anc=5919,length=500)
```

we get 
```
  File "ancient_genotypes_simulation.py", line 19, in <module>
    num_modern=100,anc_pop=0,anc_num=num_ind,Ne0=3000,Ne1=3000,anc_time=5419,split_time_anc=5919,length=500)
  File "ancient_genotypes_simulation.py", line 14, in ancient_sample_test
    random_seed=seed)
  File "/home/jk/work/github/msprime/msprime/simulations.py", line 485, in simulate
    sim, mutation_generator, 1, provenance_dict, end_time))
  File "/home/jk/work/github/msprime/msprime/simulations.py", line 163, in _replicate_generator
    sim.run(end_time)
  File "/home/jk/work/github/msprime/msprime/simulations.py", line 704, in run
    self.ll_sim.run(end_time)
_msprime.LibraryError: The simulation model supplied resulted in a parent node having a time value <= to its child. This can occur either as a result of multiple bottlenecks happening at the same time or because of numerical imprecision with very small population sizes.
```

[...]

PR #785, opened 2019-07-12
Zero coalescence time

molpopgen:
Candidate fix for #783.  The attempt is to use `nextafter` to solve the problem.

AFAIK, I've only made changes to the Hudson algorithm.  I've not walked through the other models.


molpopgen:
I see no evidence of a difference in performance with the fix.

Timings based on msprime/master:

```

In [2]: %%timeit
   ...: msprime.simulate(500, recombination_rate=250., random_seed=42)
   ...: 
10 loops, best of 5: 55.1 ms per loop

In [3]: %%timeit
   ...: msprime.simulate(500, recombination_rate=2500., random_seed=42)
   ...: 
1 loop, best of 5: 2.38 s per loop

In [4]: %%timeit
   ...: msprime.simulate(2500, recombination_rate=2500., random_seed=42)
   ...: 
1 loop, best of 5: 2.62 s per loop

In [5]: %%timeit
   ...: msprime.simulate(2500, recombination_rate=5000., random_seed=42)
   ...: 
1 loop, best of 5: 9.46 s per loop
```

Timings based on this PR:

```

In [2]: %%timeit
   ...: msprime.simulate(500, recombination_rate=250., random_seed=42)
   ...: 
10 loops, best of 5: 55.4 ms per loop

In [3]: %%timeit
   ...: msprime.simulate(500, recombination_rate=2500., random_seed=42)
   ...: 
1 loop, best of 5: 2.42 s per loop

In [4]: %%timeit
   ...: msprime.simulate(2500, recombination_rate=2500., random_seed=42)
   ...: 
1 loop, best of 5: 2.62 s per loop

In [5]: %%timeit
   ...: msprime.simulate(2500, recombination_rate=5000., random_seed=42)
   ...: 
1 loop, best of 5: 9.39 s per loop
```

jeromekelleher:
Thanks for this @molpopgen, the ``nextafter`` solution works really well. I changed your code a little bit to catch the problem at the source (when gsl_ran_exponential returns 0 computing coalescence times), and I think this avoids all the tricky issues with detecting zero increments with absurdly small population sizes.

I removed the test case also as it's basically untestable - #782 will change the random trajectory anyway, so we won't be able to reproduce this afterwards.

PR #787, opened 2019-07-14
Some infrastructure for gene conversion in C.

jeromekelleher:
Work in progress. 

@fbaumdicker, I've put in some work to get the basic infrastructure in place for GC support in the C code. I'll ping you via email on how we can move on from here.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3458,6 +3458,19 @@ out:
 }
 
 static PyObject *
+Simulator_get_num_gene_conversion_events(Simulator  *self)
+{
+    PyObject *ret = NULL;
+    if (Simulator_check_sim(self) != 0) {
+        goto out;
+    }
+    ret = Py_BuildValue("n",
+        (Py_ssize_t) msp_get_num_gene_conversion_events(self->sim));
+out:
+    return ret;
+}
+
+static PyObject *
 Simulator_get_num_migration_events(Simulator  *self)
 {
     PyObject *ret = NULL;
@@ -4112,6 +4125,9 @@ static PyMethodDef Simulator_methods[] = {
     {"get_num_recombination_events",
             (PyCFunction) Simulator_get_num_recombination_events, METH_NOARGS,
             "Returns the number of recombination_events" },
+    {"get_num_gene_conversion_events",
+            (PyCFunction) Simulator_get_num_gene_conversion_events, METH_NOARGS,
+            "Returns the number of gene_conversion_events" },
     {"get_num_migration_events",
             (PyCFunction) Simulator_get_num_migration_events, METH_NOARGS,
             "Returns the number of migration events" },
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3003,7 +3003,8 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
         "tables", "population_configuration", "pedigree", "migration_matrix", "demographic_events",
         "model", "avl_node_block_size", "segment_block_size",
         "node_mapping_block_size", "store_migrations", "start_time",
-        "store_full_arg", "num_labels", NULL};
+        "store_full_arg", "num_labels", "gene_conversion_rate",
+        "gene_conversion_track_length", NULL};
     PyObject *py_samples = NULL;
     PyObject *migration_matrix = NULL;
     PyObject *population_configuration = NULL;
@@ -3024,11 +3025,13 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
     int store_migrations = 0;
     int store_full_arg = 0;
     double start_time = -1;
+    double gene_conversion_rate = 0;
+    double gene_conversion_track_length = 1.0;
 
     self->sim = NULL;
     self->random_generator = NULL;
     self->recombination_map = NULL;
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!O!O!O!|O!OO!O!O!nnnidin", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!O!O!O!|O!O!O!O!nnnidindd", kwlist,
             &PyList_Type, &py_samples,
             &RecombinationMapType, &recombination_map,
             &RandomGeneratorType, &random_generator,
@@ -3040,7 +3043,8 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
             &PyDict_Type, &py_model,
             &avl_node_block_size, &segment_block_size,
             &node_mapping_block_size, &store_migrations, &start_time,
-            &store_full_arg, &num_labels)) {
+            &store_full_arg, &num_labels,
+            &gene_conversion_rate, &gene_conversion_track_length)) {
         goto out;
     }
     self->random_generator = random_generator;
@@ -3105,6 +3109,12 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
             (size_t) node_mapping_block_size);
     if (sim_ret != 0) {
         handle_input_error("node_mapping_block_size", sim_ret);
+        goto out;
+    }
+    sim_ret = msp_set_gene_conversion_rate(self->sim, gene_conversion_rate,
+            gene_conversion_track_length);
+    if (sim_ret != 0) {
+        handle_input_error(sim_ret);
         goto out;
     }
 
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3000,8 +3000,8 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
     int ret = -1;
     int sim_ret;
     static char *kwlist[] = {"samples", "recombination_map", "random_generator",
-        "tables", "population_configuration", "pedigree", "migration_matrix", "demographic_events",
-        "model", "avl_node_block_size", "segment_block_size",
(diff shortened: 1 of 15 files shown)


jeromekelleher:
I've updated the code here to add some minor changes @fbaumdicker. Most of it was just style stuff:

- There were some very long lines, I tried to break these up/simplify the logic
- The spacing around brackets and braces was inconsistent, I tidied this up
- There was a weird usage of ``assert`` --- this would have meant the code wouldn't work when compiled in production mode!
- I added checks after segments were allocated for memory failures.

Otherwise, it looks great!

jeromekelleher:
I've added a test case in C for the GC code @fbaumdicker, and it's currently failing. Run it like:
```
$ valgrind ./build/tests test_gene_conversion_simulation
```

which gives
```
  Test: test_gene_conversion_simulation ...==20424== Invalid write of size 8
==20424==    at 0x12E46A: msp_gene_conversion_left_event (msprime.c:1728)
==20424==    by 0x1318CE: msp_run_coalescent (msprime.c:2911)
==20424==    by 0x133469: msp_run (msprime.c:3553)
==20424==    by 0x1173B7: test_gene_conversion_simulation (tests.c:1745)
==20424==    by 0x4E3FD36: ??? (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==20424==    by 0x4E405A7: CU_run_test (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==20424==    by 0x128EAD: main (tests.c:4165)
==20424==  Address 0x28 is not stack'd, malloc'd or (recently) free'd
==20424==
==20424==
==20424== Process terminating with default action of signal 11 (SIGSEGV)
==20424==  Access not within mapped region at address 0x28
==20424==    at 0x12E46A: msp_gene_conversion_left_event (msprime.c:1728)
==20424==    by 0x1318CE: msp_run_coalescent (msprime.c:2911)
==20424==    by 0x133469: msp_run (msprime.c:3553)
==20424==    by 0x1173B7: test_gene_conversion_simulation (tests.c:1745)
==20424==    by 0x4E3FD36: ??? (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==20424==    by 0x4E405A7: CU_run_test (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==20424==    by 0x128EAD: main (tests.c:4165)
==20424==  If you believe this happened as a result of a stack
==20424==  overflow in your program's main thread (unlikely but
==20424==  possible), you can try to increase the size of the
==20424==  main thread stack using the --main-stacksize= flag.
==20424==  The main thread stack size used in this run was 8388608.
```

So, there's a logic error in here somewhere. We'll want to make sure that all code paths in the C code are covered by the tests. Can you fix this bug and make sure the test gets into all the code paths please?

fbaumdicker:
I fixed the bug in PR mcveanlab/msprime#2 . Not sure about full coverage of the code paths in the C code. As gc is a random process and there are some rare cases it might not suffice to generate a long sequence with high conversion rates. I will have to think about this.

jeromekelleher:
Let's see how this looks after the CI has finished. I'll finish up the basic infrastructure then if it's all looking good and aim to merge soon.

jeromekelleher:
I've made some updates here @fbaumdicker:

- Made some more cleanups, removing  ``} else if (negation of if condition)``. Can you have a look at these and see if you agree please?
- Got rid of the values storing p and logp. Can you check these also please?
- Added a basic Python interface 
- Rescaled the input parameters. Rates are all expressed in generations as input and in 'model time' within the simulation. The changing back and forth is done here. The track_length is tricky. We have to assume that this is expressed in physical units because that's what all the other inputs are, so we convert it into a genetic distance. 

The main problem that seems to be outstanding is that the code doesn't currently work with the default recombination map, which gives a very large number of loci, and with a recombination rate of 0, will map any non-zero genetic map value to UINT32_MAX. This is probably causing overflows in the GC code.

jeromekelleher:
Update: note, I've rebased this branch, so make sure you ``git fetch mcveanlab`` and check out a fresh copy of the ``gc-initial`` branch.

fbaumdicker:
>     * Made some more cleanups, removing  `} else if (negation of if condition)`. Can you have a look at these and see if you agree please?

Looks good to me. I was keeping the (negation of if condition) for better readability. What do you think about adding a comment line "//negation of if condition"?
 
>     * Got rid of the values storing p and logp. Can you check these also please?

This should be correct.

>     * Added a basic Python interface
> 
>     * Rescaled the input parameters. Rates are all expressed in generations as input and in 'model time' within the simulation. The changing back and forth is done here. The track_length is tricky. We have to assume that this is expressed in physical units because that's what all the other inputs are, so we convert it into a genetic distance.
>

I would argue, that the track_length is given as the expected number of bases a conversion track spans and should thus not be scaled if the Length of the sequence equals the number of bases.
Is that what the function recomb_map_phys_to_genetic at https://github.com/mcveanlab/msprime/blob/99a6feb43567910ca4846d5807695f0f66c33ddf/lib/msprime.c#L244 does?
But in this case, it should be independent of the total recombination rate.
 
> 
> The main problem that seems to be outstanding is that the code doesn't currently work with the default recombination map, which gives a very large number of loci, and with a recombination rate of 0, will map any non-zero genetic map value to UINT32_MAX. This is probably causing overflows in the GC code.

In my opinion, the most reasonable way would be to add a flag to either handle gc independently of the recombination map, i.e. with a constant rate, or use the same recombination map as recombination but maybe linearly scaled. In either case, the track_length should not differ along the genome and be independent of the recombination map. Otherwise, gene conversions will cause longer conversion tracks in regions with low recombination rates.
Does this make sense?

jeromekelleher:
> Looks good to me. I was keeping the (negation of if condition) for better readability. What do you think about adding a comment line "//negation of if condition"?

Sure, putting in the comment won't hurt. Keeping it as an else-if would be confusing though, as a reader would expect this to be something other than the negation of the if condition (anyway, it's inefficient, since you're evaluating an expression that will always be true).

> I would argue, that the track_length is given as the expected number of bases a conversion track spans and should thus not be scaled if the Length of the sequence equals the number of bases.
Is that what the function recomb_map_phys_to_genetic at https://github.com/mcveanlab/msprime/blob/99a6feb43567910ca4846d5807695f0f66c33ddf/lib/msprime.c#L244 does?
But in this case, it should be independent of the total recombination rate.

This is an annoying corner case, and I think we'll just have to work around it in some way. I'm not sure what the best way is, tbh.

> In my opinion, the most reasonable way would be to add a flag to either handle gc independently of the recombination map, i.e. with a constant rate, or use the same recombination map as recombination but maybe linearly scaled. In either case, the track_length should not differ along the genome and be independent of the recombination map. Otherwise, gene conversions will cause longer conversion tracks in regions with low recombination rates.
@jeromekelleher Does this make sense?

I agree. We can either throw an error when there's a variable recombination rate (in practise, when the map has more than two elements), or we could raise a warning saying the GC is approximate.

I guess what we should decide here is, what the minumum you need to get testing this? If I implemented the ms-like interface, would this be enough for you to start writing some statistical tests?

fbaumdicker:
I totally agree.

The minimum would be to have a constant gc rate and expected track length just as in ms. So the ms-like interface would certainly be enough to start statistical tests. Before we start thinking about including maps for gene conversion, we should run these statistical tests.

jeromekelleher:
OK, I'll see if I can get the ms interface working and at least throw errors for the corner cases we're not handling properly. Hopefully we can merge then and get on with statistical tests.

fbaumdicker:
Awesome, I am looking forward to it.

jeromekelleher:
Here's a rough first pass at the ms interface for GC @fbaumdicker --- expect bugs! We still can't run the zero recombination rate case, but the non-zero recombination rate should be working now.

There's one sticky issue though. There's a subtle difference in the way that ms and msprime does things which causes some headaches. In ms, if a recombination occurs, two trees will always be printed **even if** those trees are identical. In msprime, these invisible recombinations don't appear in the tree sequence (since they're redundant). So, for strict ms compatability, we keep a list of the recombination breakpoints that occur and output the positions of trees based on this, writing out identical trees if we need to. 

The problem is we're not currently tracking the breakpoints in the GC code at the moment. It should be a straightforward enough thing to do: we just need to put in calls like this [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L1381) to msp_insert_breakpoint at the appropriate times in the GC code. Would you mind looking into this @fbaumdicker? 

I've made it a bit easier now by adding the function ``msp_has_breakpoint`` and added a check for the breakpoints in the low-level C tests.

jeromekelleher:
This needs to be rebased and get a final review, but should be ready to merge into the main codebase then. We can see what needs doing in follow-ups.

jeromekelleher:
I think this is more-or-less ready to merge @fbaumdicker, what do you think? I'll clear up the remaining issues here after rebasing, and then we just need to log an issue to track the zero recombination rate thing. How does that sound?

fbaumdicker:
That sounds great. Let's do it.

jeromekelleher:
> That sounds great. Let's do it.

OK, let's get #771 landed first and I'll merge this PR then.

jeromekelleher:
I think this is ready to merge @fbaumdicker --- can you have one last look over it please to make sure I haven't introduced some issues during the merge (which was quite hairy!)

fbaumdicker:
I had a look over it. Looks good to me.

PR #791, opened 2019-07-23
Update MassMigration descriptions in simulations.py

marianne-aspbury:
Adding a brief clarification to description for MassMigration class which appears in the docs. Also adding information to debugger output, to clarify direction of migrations, but this part may be a bit messy. 
In the docs it'd probably be helpful to add a small schematic of a simple migration.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -1063,8 +1063,8 @@ class MassMigration(DemographicEvent):
     A mass migration event in which some fraction of the population in
     one deme simultaneously move to another deme, viewed backwards in
     time. Each lineage currently present in the source population
-    moves to the destination population with probability equal to
-    ``proportion``.
+    moves to the destination population (backwards in time) with
+    probability equal to ``proportion``.
 
     This event class generalises the population split (``-ej``) and
     admixture (``-es``) events from ``ms``. Note that MassMigrations
@@ -1098,9 +1098,13 @@ class MassMigration(DemographicEvent):
 
     def __str__(self):
         return (
-            "Mass migration: lineages move from {} to {} with "
-            "probability {}".format(
-                self.source, self.dest, self.proportion))
+            "Mass migration: "
+            "Lineages moved with probability {} backwards in time with "
+            "source {} & dest {}"
+            "\n                     "
+            "(equivalent to migration from {} to {} forwards in time)".format(
+                self.proportion, self.source, self.dest,
+                self.dest, self.source))
 
 
 class SimulationModelChange(DemographicEvent):


jeromekelleher:
Doc update is a good clarification, thanks @marianne-aspbury. I'm not sure about providing both the forwards and backwards time interpretation across two lines in the debugger though - people might interpret this as two events. Any thoughts here @hyanwong?

marianne-aspbury:
Yes I thought so too but didn't know how to be clear. Thinking about it more, an alternative could be just a re-iteration like in the docs? e.g.
```
    def __str__(self):
        return (
            "Mass migration: "
            "Lineages move from {} to {} (backwards in time) "
            "with probability {} \n ".format(
                self.source, self.dest, self.proportion))
```
I can update this as a commit if you agree

hyanwong:
I agree with @jeromekelleher about putting it on a single line, but I think @marianne-aspbury 's reasoning for putting both the forward and backward time view is sensible. This seems simply a phrasing problem - it might be possible to put both the forwards and backwards interpretations on the same line, as long as it's not too long. Something like:

```
"Mass migration: lineages moved with probability {p} backwards in time with source population {s} & destination population {d} (equivalent to migration from {d} to {s} forwards in time)"
```

or something more  condensed but more confusing like

```
"Mass migration: lineages in source population {} assigned from destination population {} with probability {} "
```

jeromekelleher:
If the line doesn't get too long, then clarifying the forwards time interpretation is good. We don't want it more than say 90 chars long though. Something too terse trying to cram everything in would be worse than just a clear backwards-time statement like @marianne-aspbury has above.

marianne-aspbury:
I think it's difficult to clarify both on a single line less than 90 characters, another possibility is indenting the next line to make it clear that it's the same event? 
Adapting @hyanwong's text it looks something like this (the first and longer line is 89 characters including spaces):

image


code:
```
    def __str__(self):
        return (
            "Mass migration: "
            "Lineages moved with probability {} backwards in time with "
            "source {} & dest {}"
            "\n                     "
            "(equivalent to migration from {} to {} forwards in time)".format(
                self.proportion, self.source, self.dest,
                self.dest, self.source))
```

jeromekelleher:
Good idea @marianne-aspbury, I like it.

marianne-aspbury:
Alright, glad to hear it @jeromekelleher. I've updated the branch by committing the above code

jeromekelleher:
Thanks @marianne-aspbury, looks good. Can you rebase this branch and squash the commits please? This [guide](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow) should help. It's a good idea to get used to doing it here on a nice simple change like this one.

@hyanwong will be able to help if you get into any trouble.

marianne-aspbury:
Alright! I think I succeeded with the help of the guide, thanks @jeromekelleher

jeromekelleher:
Great work, thanks @marianne-aspbury!

Issue #590, opened 2018-08-24
Replace unicode trees in tutorial with SVGs


jeromekelleher:
Some browsers don't have proper  monospace fonts so the trees aren't rendering correctly.

Fix #589 first though.

jeromekelleher:
Be nice to do this for 1.0. Any volunteers??

jeromekelleher:
This will be closed once we have ported all the docs to the jupyter style started in #1258

PR #799, opened 2019-08-08
Add census event(s) to msprime simulation

gtsambos:
This is just a baby pull request I made so I can log some thoughts and issues as I go.

- `add_census_event()` will have to do something a little bit different in the case that `census_time` is the same as one of the randomly generated coalescence/migration/gene conversion etc. times.
- So far have been thinking about the Hudson simulator - might need to be modified if DTWF model is used instead.
 - Should make it so user can input a list of census times, rather than just a single time.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -266,6 +266,14 @@ class Population(object):
         """
         return iter(self._ancestors[label])
 
+    def iter_ancestors(self):
+        """
+        Iterates over all ancestors in a population.
+        """
+        for ancestors in self._ancestors:
+            for ancestor in ancestors:
+                yield ancestor
+
     def find_indv(self, indv):
         """
         find the index of an ancestor in population
@@ -331,10 +339,10 @@ class Simulator(object):
             self, sample_size, num_loci, recombination_rate, migration_matrix,
             sample_configuration, population_growth_rates, population_sizes,
             population_growth_rate_changes, population_size_changes,
-            migration_matrix_element_changes, bottlenecks, model='hudson',
-            from_ts=None, max_segments=100, num_labels=1, sweep_trajectory=None,
-            full_arg=False, time_slice=None, gene_conversion_rate=0.0,
-            gene_conversion_length=1):
+            migration_matrix_element_changes, bottlenecks, census_times,
+            model='hudson', from_ts=None, max_segments=100, num_labels=1,
+            sweep_trajectory=None, full_arg=False, time_slice=None,
+            gene_conversion_rate=0.0, gene_conversion_length=1):
         # Must be a square matrix.
         N = len(migration_matrix)
         assert len(sample_configuration) == N
@@ -423,6 +431,8 @@ class Simulator(object):
         for time, pop_id, intensity in bottlenecks:
             self.modifier_events.append(
                 (time, self.bottleneck_event, (int(pop_id), 0, intensity)))
+        for time in census_times:
+            self.modifier_events.append((time[0], self.census_event, time))
         self.modifier_events.sort()
 
     def initialise_from_ts(self, ts):
@@ -1148,6 +1158,18 @@ class Simulator(object):
                 u = u.next
             print(s)
 
+    def census_event(self, time):
+        for pop in self.P:
+            for ancestor in pop.iter_ancestors():
+                seg = ancestor
+                self.flush_edges()
+                u = self.tables.nodes.add_row(time=time, flags=18, population=pop._id)
+                while seg is not None:
+                    # Add an edge joining the segment to the new node.
+                    self.store_edge(seg.left, seg.right, u, seg.node)
+                    seg.node = u
+                    seg = seg.next
+
     def bottleneck_event(self, pop_id, label, intensity):
         # self.print_state()
         # Merge some of the ancestors.
@@ -1525,7 +1547,7 @@ def run_simulate(args):
         population_sizes, args.population_growth_rate_change,
         args.population_size_change,
         args.migration_matrix_element_change,
-        args.bottleneck, args.model, from_ts=args.from_ts,
+        args.bottleneck, args.census_time, args.model, from_ts=args.from_ts,
         max_segments=10000, num_labels=num_labels, full_arg=args.full_arg,
         sweep_trajectory=sweep_trajectory, time_slice=args.time_slice,
         gene_conversion_rate=gamma, gene_conversion_length=mean_tracklength)
@@ -1571,6 +1593,8 @@ def add_simulator_arguments(parser):
         action="append", default=[])
     parser.add_argument(
         "--bottleneck", type=float, nargs=3, action="append", default=[])
+    parser.add_argument(
+        "--census-time", type=float, nargs=1, action="append", default=[])
     parser.add_argument(
         "--trajectory", type=float, nargs=3, default=None,
         help="Parameters for the allele frequency trajectory simulation")
--- a/algorithms.py
+++ b/algorithms.py
@@ -1163,7 +1163,8 @@ class Simulator(object):
             for ancestor in pop.iter_ancestors():
                 seg = ancestor
                 self.flush_edges()
-                u = self.tables.nodes.add_row(time=time, flags=18, population=pop._id)
+                u = self.tables.nodes.add_row(
+                        time=time, flags=msprime.NODE_IS_CEN_EVENT, population=pop._id)
                 while seg is not None:
                     # Add an edge joining the segment to the new node.
                     self.store_edge(seg.left, seg.right, u, seg.node)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -111,6 +111,8 @@ more details on the migration rates.
 
 .. autoclass:: msprime.PopulationConfiguration
 
+.. _sec_api_demographic_events:
+
 ******************
 Demographic Events
 ******************
@@ -127,6 +129,9 @@ and all rates are per-generation.
 .. autoclass:: msprime.MigrationRateChange
 .. autoclass:: msprime.MassMigration
 .. autoclass:: msprime.SimulationModelChange
+
+.. _sec_api_demographic_events_census:
+
 .. autoclass:: msprime.CensusEvent
 
 ++++++++++++++++++++++++++++
@@ -355,7 +360,7 @@ flags values are defined:
 
 .. data:: msprime.NODE_IS_CEN_EVENT
 
-    The node was created by a `msprime.CensusEvent`.
+    The node was created by a :class:`msprime.CensusEvent`.
 
 
 ********************
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -127,6 +127,7 @@ and all rates are per-generation.
 .. autoclass:: msprime.MigrationRateChange
 .. autoclass:: msprime.MassMigration
 .. autoclass:: msprime.SimulationModelChange
+.. autoclass:: msprime.CensusEvent
 
 ++++++++++++++++++++++++++++
 Debugging demographic models
@@ -351,6 +352,10 @@ flags values are defined:
     The node is an ARG migration event identifying the individual that migrated.
     Can be used in combination with the ``record_migrations`` argument to
     :func:`.simulate`.
+
+.. data:: msprime.NODE_IS_CEN_EVENT
+
+    The node was created by a `msprime.CensusEvent`.
 
 
 ********************
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -618,6 +618,102 @@ Once you are satisfied that the demographic history that you have built
 is correct, it can then be simulated by calling the :func:`.simulate`
 function.
 
+-------------
+Census events
+-------------
+
+Census events allow you to add a node to each branch of the tree sequence at a given time during the simulation. This can be useful when you wish to study haplotypes that are ancestral to your simulated sample.
+
+For instance, the following code specifies a simulation with two samples drawn from each of two populations. There are two demographic events: a migration rate change and a census event. At generation 100 and earlier, the two populations exchange migrants at a rate of 0.05. At generation 5000, a census is performed.
+
+.. code-block:: python
+
+    >>> pop_config = msprime.PopulationConfiguration(sample_size=2, initial_size=1000)
+    >>> mig_rate_change = msprime.MigrationRateChange(time=100, rate=0.05)
+    >>> ts = msprime.simulate(
+                population_configurations=[pop_config, pop_config],
+                length=1000,
+                demographic_events=[mig_rate_change, msprime.CensusEvent(time=5000)],
+                recombination_rate=1e-7,
+                random_seed=141)
+
+The resulting tree sequence has nodes on each tree at the specified census time. These are the nodes with IDs 9, 10, 11, 12 and 13:
+
+.. code-block:: python
+
+    >>> for tree in ts.trees():
+        print(tree.draw(format="unicode"))
+
+      14   
+     ┏━┻━┓ 
+     9  12 
+     ┃   ┃ 
+     5   ┃ 
+    ┏┻┓  ┃ 
+    ┃ ┃  4 
+    ┃ ┃ ┏┻┓
+    1 3 0 2
+
+     14     
+    ┏━┻━┓   
+    10 13   
+    ┃   ┃   
+    ┃   6   
+    ┃  ┏┻━┓ 
+    ┃  ┃  4 
+    ┃  ┃ ┏┻┓
+    3  1 0 2
+
+      14   
+     ┏━┻━┓ 
+    13  11 
+     ┃   ┃ 
+     ┃   7 
+     ┃  ┏┻┓
+     6  ┃ ┃
+    ┏┻┓ ┃ ┃
+    1 0 2 3
+
+      15   
+     ┏━┻━┓ 
+     8  11 
+     ┃   ┃ 
+     ┃   7 
+     ┃  ┏┻┓
+     6  ┃ ┃
+    ┏┻┓ ┃ ┃
+    1 0 2 3
+
+This tells us that the genetic material ancestral to the present day sample was held within 5 haplotypes at time 5000. The node table shows us that three of these haplotypes (nodes 9, 10 and 11) were in population 0 at this time, and two of these haplotypes (nodes 12 and 13) were in population 1 at this time.
+
+.. code-block:: python
+
+    >>> print(ts.tables.nodes)
+    id  flags   population  individual  time    metadata
+    0   1       0   -1  0.00000000000000    
+    1   1       0   -1  0.00000000000000    
+    2   1       1   -1  0.00000000000000    
+    3   1       1   -1  0.00000000000000    
+    4   0       1   -1  2350.08685279051815 
+    5   0       1   -1  3759.20387382847684 
+    6   0       0   -1  4234.97992185234671 
+    7   0       1   -1  4598.83898042243527 
+    8   1048576 0   -1  5000.00000000000000 
+    9   1048576 0   -1  5000.00000000000000 
+    10  1048576 0   -1  5000.00000000000000 
+    11  1048576 0   -1  5000.00000000000000 
+    12  1048576 1   -1  5000.00000000000000 
+    13  1048576 1   -1  5000.00000000000000 
+    14  0       1   -1  5246.90282987397495 
+    15  0       0   -1  8206.73121309170347
+
+If we wish to study these ancestral haplotypes further, we can simplify the tree sequence with respect to the census nodes and perform subsequent analyses on this simplified tree sequence:
+
+.. code-block:: python
+
+    >>> nodes = [i.id for i in list(ts.nodes()) if i.flags==msprime.NODE_IS_CEN_EVENT]
+    >>> ts_anc = ts.simplify(samples=nodes)
+
 ******************
 Recombination maps
 ******************
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -618,13 +618,21 @@ Once you are satisfied that the demographic history that you have built
 is correct, it can then be simulated by calling the :func:`.simulate`
 function.
 
+.. _sec_tutorial_demography_census:
+
 -------------
 Census events
 -------------
 
-Census events allow you to add a node to each branch of the tree sequence at a given time during the simulation. This can be useful when you wish to study haplotypes that are ancestral to your simulated sample.
+Census events allow you to add a node to each branch of the tree sequence at a given time
+during the simulation. This can be useful when you wish to study haplotypes that are
+ancestral to your simulated sample, or when you wish to know which lineages were present in
+which populations at specified times.
 
-For instance, the following code specifies a simulation with two samples drawn from each of two populations. There are two demographic events: a migration rate change and a census event. At generation 100 and earlier, the two populations exchange migrants at a rate of 0.05. At generation 5000, a census is performed.
+For instance, the following code specifies a simulation with two samples drawn from each of
+two populations. There are two demographic events: a migration rate change and a census
+event. At generation 100 and earlier, the two populations exchange migrants at a rate of
+0.05. At generation 5000, a census is performed:
 
 .. code-block:: python
 
@@ -637,54 +645,18 @@ For instance, the following code specifies a simulation with two samples drawn f
                 recombination_rate=1e-7,
                 random_seed=141)
 
-The resulting tree sequence has nodes on each tree at the specified census time. These are the nodes with IDs 9, 10, 11, 12 and 13:
+The resulting tree sequence has nodes on each tree at the specified census time.
+These are the nodes with IDs 8, 9, 10, 11, 12 and 13:
 
 .. code-block:: python
 
-    >>> for tree in ts.trees():
-        print(tree.draw(format="unicode"))
+    >>> display(SVG(ts.draw_svg()))
 
-      14   
-     ┏━┻━┓ 
-     9  12 
-     ┃   ┃ 
-     5   ┃ 
-    ┏┻┓  ┃ 
-    ┃ ┃  4 
-    ┃ ┃ ┏┻┓
-    1 3 0 2
+.. image:: _static/ts_with_census_nodes.svg
+   :width: 800px
+   :alt: A tree sequence with census nodes.
 
-     14     
-    ┏━┻━┓   
-    10 13   
-    ┃   ┃   
-    ┃   6   
-    ┃  ┏┻━┓ 
-    ┃  ┃  4 
-    ┃  ┃ ┏┻┓
-    3  1 0 2
-
-      14   
-     ┏━┻━┓ 
-    13  11 
-     ┃   ┃ 
-     ┃   7 
-     ┃  ┏┻┓
-     6  ┃ ┃
-    ┏┻┓ ┃ ┃
-    1 0 2 3
-
-      15   
-     ┏━┻━┓ 
-     8  11 
-     ┃   ┃ 
-     ┃   7 
-     ┃  ┏┻┓
-     6  ┃ ┃
-    ┏┻┓ ┃ ┃
-    1 0 2 3
-
-This tells us that the genetic material ancestral to the present day sample was held within 5 haplotypes at time 5000. The node table shows us that three of these haplotypes (nodes 9, 10 and 11) were in population 0 at this time, and two of these haplotypes (nodes 12 and 13) were in population 1 at this time.
+This tells us that the genetic material ancestral to the present day sample was held within 5 haplotypes at time 5000. The node table shows us that four of these haplotypes (nodes 8, 9, 10 and 11) were in population 0 at this time, and two of these haplotypes (nodes 12 and 13) were in population 1 at this time.
 
 .. code-block:: python
 
@@ -707,12 +679,18 @@ This tells us that the genetic material ancestral to the present day sample was 
     14  0       1   -1  5246.90282987397495 
     15  0       0   -1  8206.73121309170347
 
-If we wish to study these ancestral haplotypes further, we can simplify the tree sequence with respect to the census nodes and perform subsequent analyses on this simplified tree sequence:
+If we wish to study these ancestral haplotypes further, we can simplify the tree sequence
+with respect to the census nodes and perform subsequent analyses on this simplified tree
+sequence.
+In this example, ``ts_anc`` is a tree sequence obtained from the original tree sequence
+``ts`` by labelling the census nodes as samples and removing all nodes and edges that are 
+not ancestral to these census nodes.
 
 .. code-block:: python
 
-    >>> nodes = [i.id for i in list(ts.nodes()) if i.flags==msprime.NODE_IS_CEN_EVENT]
+    >>> nodes = [i.id for i in ts.nodes() if i.flags==msprime.NODE_IS_CEN_EVENT]
     >>> ts_anc = ts.simplify(samples=nodes)
+
 
 ******************
 Recombination maps
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4249,7 +4249,7 @@ msp_census_event(msp_t *self, demographic_event_t *event)
             node = ancestors->head;
 
             while (node != NULL){
-                seg = node->item;
+                seg = (segment_t *) node->item;
 
                 while (seg != NULL) {
                     // Add an edge to the edge table.
@@ -4263,9 +4263,9 @@ msp_census_event(msp_t *self, demographic_event_t *event)
                     if (ret < 0) {
                         goto out;
                     }
-                    u = ret;
+                    u = (tsk_id_t) ret;
                     // Add an edge joining the segment to the new node.
-                    ret = msp_store_edge(self, seg->left, seg->right, ret, seg->value);
+                    ret = msp_store_edge(self, seg->left, seg->right, u, seg->value);
                     if (ret != 0) {
                         goto out;
                     }
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4260,19 +4260,24 @@ msp_census_event(msp_t *self, demographic_event_t *event)
                     ret = tsk_node_table_add_row(&self->tables->nodes,
                             MSP_NODE_IS_CEN_EVENT, time, (population_id_t) i, TSK_NULL,
                             NULL, 0);
+                    if (ret < 0) {
+                        goto out;
+                    }
                     u = ret;
                     // Add an edge joining the segment to the new node.
                     ret = msp_store_edge(self, seg->left, seg->right, ret, seg->value);
+                    if (ret != 0) {
+                        goto out;
+                    }
                     // Modify segment node id.
                     seg->value = u;
-                    ret = 0;
                     seg = seg->next;
                 }
                 node = node->next;
             }
         }
     }
-
+    ret = 0;
 out:
     return ret;
 }
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2533,7 +2533,6 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
         if (self->next_demographic_event != NULL) {
             demographic_event_time = self->next_demographic_event->time;
         }
-
         /* The simulation state is can only changed from this point on. If
          * any of the events would cause the time to be >= max_time, we exit
          */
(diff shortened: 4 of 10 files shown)


algorithms.py
```diff
@@ -642,6 +657,9 @@ def hudson_simulate(self):
                 elif min_time == t_ca:
                     # print("CA EVENT")
                     self.common_ancestor_event(ca_population, 0)
+                elif min_time == t_cen:
```
jeromekelleher: We don't want to do it this way --- we want to register the census event as one of the ``modifier_events`` and use the general infrastructure for dispatching these events. We shouldn't be touching the main loop of the simulation here at all --- all we want is

1) a function that runs the census event (modeled after a population parameters change, say)
2) We an event referencing this function to the ``modifier_events`` list so that this function gets executed at the correct time.

gtsambos:
Hi @jeromekelleher, when you've got a sec could you please take a look at the new placeholder code and let me know if I've put anything in the wrong place? I'll now think a bit more about what `add_census_event` should actually do...

gtsambos:
btw, I think the trickiest thing here will be figuring out what to do in the case that one of the coalescence/recombination events happens at the exact same time as a census event. Let's talk about this in person sometime.

jeromekelleher:
> btw, I think the trickiest thing here will be figuring out what to do in the case that one of the coalescence/recombination events happens at the exact same time as a census event. Let's talk about this in person sometime.

We'll get zero branch lengths and the tree sequence will be invalid. Let's not worry about it for now.

jeromekelleher:
> Hi @jeromekelleher, when you've got a sec could you please take a look at the new placeholder code and let me know if I've put anything in the wrong place? I'll now think a bit more about what add_census_event should actually do...

Yes, this is exactly it.

algorithms.py
```diff
@@ -1148,6 +1150,22 @@ def print_heaps(self, L):
                 u = u.next
             print(s)
 
+    def census_event(self, time):
+        for pop in self.P:
+            # pop.print_state()
+            for a in pop._ancestors:
```
jeromekelleher: Would help to add an ``all_ancestors`` method to Population:
```python
def all_ancestors(self):
    """ 
    Iterate over all ancestors in this population
    """
    for each ancestor: # psuedocode
         yield ancestor
```
then this can be 
```python
for pop in self.P:
    for ancestor_head in pop.all_ancestors():
         seg = ancestor_head
        while seg is not None:
             # etc.
```


Might help to use more descriptive names:
```python

File: algorithms.py [hunk 3]
```diff
@@ -1148,6 +1150,22 @@ def print_heaps(self, L):
...
+                for s in a:
+                    while s is not None:
+                        self.flush_edges()
+                        u = self.tables.nodes.add_row(time=time, flags=18,
```
jeromekelleher: This will be something like ``flags=msprime.NODE_IS_CENSUS_EVENT``. But, it's the bit position 18 we're setting anyway, so for now it's ``1<<18``.

gtsambos:
FYI, here are the things I could find that looked like existing flags in msprime/tskit.

From argtable3.c in tskit [(link)](https://github.com/tskit-dev/msprime/blob/master/lib/dev-tools/argtable3.c#L324-L326):
```
#define FLAG_PERMUTE	0x01	/* permute non-options to the end of argv */
#define FLAG_ALLARGS	0x02	/* treat non-options as args to option "-1" */
#define FLAG_LONGONLY	0x04	/* operate as getopt_long_only */
```
From tables.h in tskit: [(link)](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.h#L74)
```
#define TSK_DEBUG                       (1u << 31)


/** @brief Do not initialise the parameter object. */
#define TSK_NO_INIT                     (1u << 30)


/**@} */


/* Flags for simplify() */
#define TSK_FILTER_SITES                 (1 << 0)
#define TSK_FILTER_POPULATIONS           (1 << 1)
#define TSK_FILTER_INDIVIDUALS           (1 << 2)
#define TSK_REDUCE_TO_SITE_TOPOLOGY      (1 << 3)
#define TSK_KEEP_UNARY                   (1 << 4)


/* Flags for check_integrity */
#define TSK_CHECK_OFFSETS                (1 << 0)
#define TSK_CHECK_EDGE_ORDERING          (1 << 1)
#define TSK_CHECK_SITE_ORDERING          (1 << 2)
#define TSK_CHECK_SITE_DUPLICATES        (1 << 3)
#define TSK_CHECK_MUTATION_ORDERING      (1 << 4)
#define TSK_CHECK_INDEXES                (1 << 5)
#define TSK_CHECK_ALL                    \
    (TSK_CHECK_OFFSETS | TSK_CHECK_EDGE_ORDERING | TSK_CHECK_SITE_ORDERING | \
     TSK_CHECK_SITE_DUPLICATES | TSK_CHECK_MUTATION_ORDERING | TSK_CHECK_INDEXES)


/* Flags for dump tables */
#define TSK_NO_BUILD_INDEXES            (1 << 0)


/* Flags for load tables */
#define TSK_BUILD_INDEXES               (1 << 0)
```

From trees.h in tskit: [(link)](https://github.com/tskit-dev/tskit/blob/master/c/tskit/trees.h#L39-L48)
```
#define TSK_SAMPLE_COUNTS  (1 << 0)
#define TSK_SAMPLE_LISTS   (1 << 1)


#define TSK_STAT_SITE               (1 << 0)
#define TSK_STAT_BRANCH             (1 << 1)
#define TSK_STAT_NODE               (1 << 2)


/* Leave room for other stat types */
#define TSK_STAT_POLARISED          (1 << 10)
#define TSK_STAT_SPAN_NORMALISE     (1 << 11)
```

jeromekelleher:
These are all flags defining different options for functions, not node flags. [This](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.h#L49) is what you're looking for.

gtsambos:
thanks for this help Jerome! 😅 Now that the python bit looks okay, I'm going to trawl through the actual C code to see how where this fits in. I took some notes of our last conversation about this but will ping you if I'm stuck on anything.

gtsambos:
Hi @jeromekelleher, if you're happy with the most recent commits, we should be ready to link up with Python + make some Python tests tomorrow.

gtsambos:
Oh, btw, I still haven't bothered fixing up population ID and flags in the table output - will do this tomorrow as well.

algorithms.py
```diff
@@ -266,6 +266,15 @@ def iter_label(self, label):
         """
         return iter(self._ancestors[label])
 
+    def iter_ancestors(self):
+        """
+        Iterates over all ancestors in a population.
+        """
+        # return iter(self._ancestors)
```
jeromekelleher: stray line here

algorithms.py
```diff
@@ -1148,6 +1159,20 @@ def print_heaps(self, L):
                 u = u.next
             print(s)
 
+    def census_event(self, time):
+        for pop in self.P:
+            # pop.print_state()
+            for ancestor in pop.iter_ancestors():
+                seg = ancestor
+                while seg is not None:
+                    self.flush_edges()
```
jeromekelleher: I think you're flushing edges too much --- just needs to be done once per ancestor.

algorithms.py
```diff
@@ -1148,6 +1159,20 @@ def print_heaps(self, L):
...
+                    u = self.tables.nodes.add_row(
```
jeromekelleher: Ah, also, adding too many nodes. This is one node per segment, definitely too much!

lib/msprime.c
```diff
@@ -4223,6 +4222,82 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+                    goto out;
+                }
+                ret = tsk_node_table_add_row(&self->tables->nodes, 0, time, 0,
+                        TSK_NULL, NULL, 0);
+
+                // Add an edge joining the segment to the new node.
+                ret = msp_store_edge(self, seg->left, seg->right, ret, seg->label);
+
+                // Modify segment node id.
+                seg->label = ret;
```
jeromekelleher: Why are you setting the label? It's the node we update.

lib/msprime.c
```diff
@@ -4223,6 +4222,82 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+
+                // Add an edge to the edge table.
+                ret = msp_flush_edges(self);
+                if (ret != 0) {
+                    goto out;
+                }
+                ret = tsk_node_table_add_row(&self->tables->nodes, 0, time, 0,
+                        TSK_NULL, NULL, 0);
+
+                // Add an edge joining the segment to the new node.
```
jeromekelleher: Missing ``while (seg != NULL) `` loop here. You're just tagging the first segment in each ancestor here (will work fine in single locus sim).

lib/msprime.c
```diff
@@ -4223,6 +4222,82 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+            event->time);
+}
+
+int MSP_WARN_UNUSED
+msp_add_census_event(msp_t *self, double time)
+{
+    int ret = 0;
+    demographic_event_t *de;
+
+    if (time <= 0) {
```
jeromekelleher: Is this caugnt in ``add_demographic_event``? If so, we don't need it.

lib/msprime.h
```diff
@@ -273,6 +273,10 @@ typedef struct {
     double strength;
 } instantaneous_bottleneck_t;
 
+typedef struct {
```
jeromekelleher: Looks like we don't actually need this because there's no parameters. We can delete this struct and its place in the union. (Time is stored already in demographic_event_t)

lib/tests/tests.c
```diff
@@ -956,6 +956,60 @@ test_demographic_events(void)
...
+    CU_ASSERT_TRUE(no_census_nodes > 1);
+
+    /* Free things. */
+    ret = msp_free(msp);
+    CU_ASSERT_EQUAL(ret, 0);
+    gsl_rng_free(rng);
+    free(msp);
+    free(samples);
+    recomb_map_free(&recomb_map);
+    tsk_table_collection_free(&tables);
```
jeromekelleher: Spare blank lines here.

algorithms.py
gtsambos: > This is one node per segment, definitely too much!

Perhaps I should go back and read the msprime paper in a bit more detail because I still don't quite understand this. Here's some output of this current version of `algorithms.py` with one census event, and census nodes 9, 10, 11 and 12. It looks okay to me.

are you saying that the census nodes on the last tree (10, 12) should be the same as on the first few trees (9, 11)?

Screen Shot 2019-08-23 at 10 56 50 am

lib/msprime.h
gtsambos: 👍

lib/msprime.c
gtsambos: The only check of time I can see in `msp_add_demographic_event` is [this one](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L3635-L3640), which is about sortedness rather than times being < 0:

```
    if (self->demographic_events_tail != NULL) {
        if (time < self->demographic_events_tail->time) {
            ret = MSP_ERR_UNSORTED_DEMOGRAPHIC_EVENTS;
            goto out;
        }
    }
```
so maybe it would be better to move this check of time > 0 into `add_demographic_event` instead?

jeromekelleher: Yes, maybe. Try it out and see if it breaks stuff. If it does then forget about it.

lib/msprime.c
gtsambos: ah yep, oops 😅

lib/msprime.c
gtsambos: I've put a check of time < 0 into `add_demographic_event` (time == 0 breaks it).

Perhaps I could also put in a check that there aren't any existing nodes at the census time for `add_census_event`? This is a more general test that would also catch the time == 0 case. I think you said that some there is already some error that might be thrown if the census time is identical to a coalescence time, but perhaps it would be useful to have a more specific/interpretable error message for this situation?

gtsambos:
I'm going to work on the Python/C boilerplate now.

lib/msprime.c
gtsambos: NB. esp for the DTWF situation where users might be inclined to set a census at an integer time.

algorithms.py
```diff
@@ -1148,6 +1158,18 @@ def print_heaps(self, L):
                 u = u.next
             print(s)
 
+    def census_event(self, time):
+        for pop in self.P:
+            for ancestor in pop.iter_ancestors():
+                seg = ancestor
+                self.flush_edges()
+                u = self.tables.nodes.add_row(time=time, flags=18, population=pop._id)
```
jeromekelleher: Need to use msprime.NODE_IS_CENSUS_EVENT here

lib/msprime.c
```diff
@@ -4223,6 +4227,80 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+    node_id_t u;
+
+    for (i = 0; i < (int) self->num_populations; i++) {
+        for (j = 0; j < (int) self->num_labels; j++) {
+
+            // Get segment from an ancestor in a population.
+            ancestors = &self->populations[i].ancestors[j];
+            node = ancestors->head;
+
+            while (node != NULL){
```
jeromekelleher: Slightly more concise way of doing the for loop over avl nodes:
```c

for (node = ancestors->head; node != NULL; node = node->next) {
    seg = (segment_t *) node->item;
    // as before
}

File: lib/msprime.c [hunk 14]
```diff
@@ -4223,6 +4227,80 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+                    ret = msp_flush_edges(self);
+                    if (ret != 0) {
+                        goto out;
+                    }
+                    ret = tsk_node_table_add_row(&self->tables->nodes,
+                            MSP_NODE_IS_CEN_EVENT, time, (population_id_t) i, TSK_NULL,
+                            NULL, 0);
+                    u = ret;
+                    // Add an edge joining the segment to the new node.
+                    ret = msp_store_edge(self, seg->left, seg->right, ret, seg->value);
```
jeromekelleher: Need to check the value of ``ret`` here and after node_table_add_row --- an error could have happen during either of these calls and we'd miss them currently.

lib/msprime.c
```diff
@@ -4223,6 +4227,80 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+                    // Modify segment node id.
+                    seg->value = u;
+                    ret = 0;
```
jeromekelleher: Why change the value of ret here?

msprime/simulations.py
```diff
@@ -1201,6 +1201,21 @@ def __str__(self):
                 self.population, self.strength))
 
 
+class CensusEvent(DemographicEvent):
```
jeromekelleher: Need some documentation here. What is this for and how do you use it?

tests/test_demography.py
```diff
@@ -2498,3 +2498,49 @@ def test_multi_population(self):
...
+
+class TestCensusEvent(unittest.TestCase):
+    """
+    Tests of the census demographic event.
+    """
+    def verify(self, ts, census_time):
+        """
+        Verifies that a census event has been added correctly.
+        """
+        # It would be better to get the census nodes using node flags.
```
jeromekelleher: Agreed - we should do this, and check that the time is the right time.

tests/test_demography.py
```diff
@@ -2498,3 +2498,49 @@ def test_multi_population(self):
...
+        census_ids = np.where(ts.tables.nodes.time == census_time)[0]
+        self.assertGreater(len(census_ids), 1)
+        # Check that all samples have a census ancestor on each tree.
+        for tree in ts.trees():
+            leaves = []
+            census_nodes = [u for u in census_ids if u in list(tree.nodes())]
+            for node in census_nodes:
+                le = list(tree.leaves(node))
```
jeromekelleher: Also, check that len(ts.children(node)) == 1

gtsambos:
Hi again @jeromekelleher 👋 I'm going to continue chugging along and hopefully finish this up in the next few days.

Btw, I'm having a few problems compiling this with ninja. I'm getting the following error:
```
In file included from ../tests/tests.c:24:
../msprime.h:27:10: fatal error: 'gsl/gsl_rng.h' file not found
#include <gsl/gsl_rng.h>
```
(btw, I've checked that I have `gsl` and it's up to date). I've noticed that one of the CI tests is failing for a gsl-related reason too. 
Weirdly though, the build isn't failing everywhere - it works fine on my laptop, for example. wondering if it might be to do with versions of clang?

gtsambos:
Full error is here:
```
6200D-132482-M:builddir gtsambos$ ninja
[1/4] Compiling C object 'dev-cli@exe/dev-tools_dev-cli.c.o'.
FAILED: dev-cli@exe/dev-tools_dev-cli.c.o 
clang -Idev-cli@exe -I. -I.. -Isubprojects/kastore -I../subprojects/kastore -Isubprojects/tskit -
I../subprojects/tskit -I../subprojects/tskit/tskit -I/usr/local/Cellar/libconfig/1.7.2/include -Xclang -fcolor-
diagnostics -pipe -Wall -Winvalid-pch -g -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -
fPIC -fPIE -fstack-protector-strong -O2 -D_FORTIFY_SOURCE=2 -mmacosx-version-min=10.9 -Dlint -
MD -MQ 'dev-cli@exe/dev-tools_dev-cli.c.o' -MF 'dev-cli@exe/dev-tools_dev-cli.c.o.d' -o 'dev-
cli@exe/dev-tools_dev-cli.c.o' -c ../dev-tools/dev-cli.c
../dev-tools/dev-cli.c:28:10: fatal error: 'gsl/gsl_math.h' file not found
#include <gsl/gsl_math.h>
         ^~~~~~~~~~~~~~~~
```

jeromekelleher:
Welcome back @gtsambos! What sort of machine are you running into problems compiling this on? Can you delete the ``build`` directory and copy in the output of ``meson build`` please?

The CI issue seems to be an upstream problem with Windows GSL,  which we're dealing with in #803.

gtsambos:
>  What sort of machine are you running into problems compiling this on? 

Mac OS X 10.14.5

> Can you delete the build directory and copy in the output of meson build please?

On it's own, `meson build` doesn't work - I get the following message:
```
6200D-132482-M:lib gtsambos$ meson builddir
The Meson build system
Version: 0.51.1
Source dir: /Users/gtsambos/Projects/msprime/lib
Build dir: /Users/gtsambos/Projects/msprime/lib/builddir
Build type: native build
Project name: msprime
Project version: undefined
Appending CFLAGS from environment: '-march=core2 -mtune=haswell -mssse3 -ftree-vectorize -
fPIC -fPIE -fstack-protector-strong -O2 -pipe'
Appending LDFLAGS from environment: '-Wl,-pie -Wl,-headerpad_max_install_names -Wl,-
dead_strip_dylibs'
Appending CPPFLAGS from environment: '-D_FORTIFY_SOURCE=2 -mmacosx-version-min=10.9'

meson.build:1:0: ERROR: Compiler x86_64-apple-darwin13.4.0-clang can not compile programs.
```
But I did some googling and saw that this seemed to be similar a problem that [others were having](https://github.com/ContinuumIO/anaconda-issues/issues/9096) (see how many others referenced this issue!!!), so I (thought I) got around it by explicitly specifying the compiler I wanted meson to use like this, which I found out how to do from [this](https://github.com/mesonbuild/meson/issues/1752).
```
CC=clang CXX=clang++ meson build-clang
```
Executing this gave me the following output:

```
6200D-132482-M:lib gtsambos$ CC=clang CXX=clang++ meson builddir
The Meson build system
Version: 0.51.1
Source dir: /Users/gtsambos/Projects/msprime/lib
Build dir: /Users/gtsambos/Projects/msprime/lib/builddir
Build type: native build
Project name: msprime
Project version: undefined
Appending CFLAGS from environment: '-march=core2 -mtune=haswell -mssse3 -ftree-vectorize -
fPIC -fPIE -fstack-protector-strong -O2 -pipe'
Appending LDFLAGS from environment: '-Wl,-pie -Wl,-headerpad_max_install_names -Wl,-
dead_strip_dylibs'
Appending CPPFLAGS from environment: '-D_FORTIFY_SOURCE=2 -mmacosx-version-min=10.9'
C compiler for the build machine: clang (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
Appending CFLAGS from environment: '-march=core2 -mtune=haswell -mssse3 -ftree-vectorize -
fPIC -fPIE -fstack-protector-strong -O2 -pipe'
Appending LDFLAGS from environment: '-Wl,-pie -Wl,-headerpad_max_install_names -Wl,-
dead_strip_dylibs'
Appending CPPFLAGS from environment: '-D_FORTIFY_SOURCE=2 -mmacosx-version-min=10.9'
C compiler for the host machine: clang (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
Build machine cpu family: x86_64
Build machine cpu: x86_64

|Executing subproject kastore method meson 
|
|Project name: kastore
|Project version: undefined
|C compiler for the build machine: clang (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
|Appending CXXFLAGS from environment: '-march=core2 -mtune=haswell -mssse3 -ftree-vectorize -
fPIC -fPIE -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -std=c++14 -
fmessage-length=0'
|Appending LDFLAGS from environment: '-Wl,-pie -Wl,-headerpad_max_install_names -Wl,-
dead_strip_dylibs'
|Appending CPPFLAGS from environment: '-D_FORTIFY_SOURCE=2 -mmacosx-version-min=10.9'
|C++ compiler for the build machine: clang++ (clang 4.0.1 "clang version 4.0.1 
(tags/RELEASE_401/final)")
|C compiler for the host machine: clang (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
|Appending CXXFLAGS from environment: '-march=core2 -mtune=haswell -mssse3 -ftree-vectorize -
fPIC -fPIE -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -std=c++14 -
fmessage-length=0'
|Appending LDFLAGS from environment: '-Wl,-pie -Wl,-headerpad_max_install_names -Wl,-
dead_strip_dylibs'
|Appending CPPFLAGS from environment: '-D_FORTIFY_SOURCE=2 -mmacosx-version-min=10.9'
|C++ compiler for the host machine: clang++ (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
|Build targets in project: 1
|Subproject kastore finished.


|Executing subproject tskit method meson 
|
|Project name: tskit
|Project version: undefined
|C compiler for the build machine: clang (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
|C++ compiler for the build machine: clang++ (clang 4.0.1 "clang version 4.0.1 
(tags/RELEASE_401/final)")
|C compiler for the host machine: clang (clang 4.0.1 "clang version 4.0.1 (tags/RELEASE_401/final)")
|C++ compiler for the host machine: clang++ (clang 4.0.1 "clang version 4.0.1 
(tags/RELEASE_401/final)")
|Library m found: YES
|Build targets in project: 2
|Subproject tskit finished.

Library m found: YES
Found pkg-config: /usr/local/bin/pkg-config (0.29.2)
Run-time dependency gsl found: YES 2.5
Run-time dependency cunit found: YES 2.1-3
Run-time dependency libconfig found: YES 1.7.2
Build targets in project: 6
Found ninja-1.9.0.git.kitware.dyndep-1.jobserver-1 at /Users/gtsambos/miniconda3/bin/ninja
6200D-132482-M:lib gtsambos$ 
```

Here's some info about my gcc/clang versions btw

```
6200D-132482-M:lib gtsambos$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-
dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacO
SX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
6200D-132482-M:lib gtsambos$ clang --version
clang version 4.0.1 (tags/RELEASE_401/final)
Target: x86_64-apple-darwin18.6.0
Thread model: posix
InstalledDir: /Users/gtsambos/miniconda3/bin
```

jeromekelleher:
Good detective work @gtsambos --- so, it looks like it's working now? If meson succeeds in finding the GSL dependency, it should all work when you build. If not, then I'm sure how much help I can be (other than the, unhelpful, "use linux"!).

gtsambos:
The meson build works (with the above hack), but ninja doesn't. If all else fails, I think I'm going to try and change all my versions of things to the ones that work on my laptop - and if that fails, Linux it is!!

jeromekelleher:
> The meson build works (with the above hack), but ninja doesn't. If all else fails, I think I'm going to try and change all my versions of things to the ones that work on my laptop - and if that fails, Linux it is!!

Sounds like things are a bit messy on that machine all right, trying to get to the same state as your laptop is a good idea. [ struggles with urge to rant about crappiness of macs as a C development environment ]

gtsambos:
Yeah, I was coerced into lots of system updates when I turned on my office computer again after months of it being off. The things I do for my pretty desktop...

gtsambos:
After a long day of making, changing and editing VMs, I went back and tried making this meson build with gcc instead of clang:
```
CC=gcc CXX=g++ meson builddir
```
and this worked, and ninja seems happy too!
I may edit the developers docs to suggest this for Mac OS X users so they don't mess around as much as I did (or perhaps just as a reminder to future me)

gtsambos:
(now to get on with the actual code...)

molpopgen:
The presence of this line is informative about what is happening on OSX, I think:

```
x86_64-apple-darwin13.4.0-clang
```

That implies that you have one or both of the following installed:

```
clang_osx-64
clangxx_os-64
```

These are the [new](https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html) compiler tools used to build conda packages for OS X.  Once installed, they over-ride many shell variables in order to attempt to "just work".  For example, they alias CC to to longer string I pasted above.

However, unlike their Linux analogs, it seems incredibly difficult to get these working correctly on a "regular" person's OS X machine.  I went down this rabbit hole last week.  There seem to be steps missing in the [docs](https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html) on how to get these properly working.  Or, it is possible that build systems need to be modified to be aware of this bizarro world and do special things to accommodate it.  If I sort this out on my end, I'll post something about the full solution on my website

The reason why `CC=gcc CXX=g++` works is because Xcode provides symlinks from clang and clang++ to the GNU compiler names by default.  In essence, you are telling the system to avoid the conda compilers and use the Xcode ones instead.  That will be safe until it isn't.

A simpler solution for now is:

```
conda remove clang_osx-64  clangxx_osx-64
```

gtsambos:
Thanks @molpopgen. Yeah, I realised last week that there were lots of other things screwing up with this new clang compiler (spent Monday having lots of grief with R packages :sadface) 

I made a pull request that adds some extra notes and 'gotchas' to the docs for developers - will put this tip in there too when I'm next at my computer. (I think the gcc option is also worth mentioning, in case users don't want to remove the new clang completely). If there are any other tips you think should be in these docs, feel free to modify it.

jeromekelleher:
Glad to see you figured it out @gtsambos --- can you ping me when this is ready for final review please?  Looks like we need to push the test coverage up a little bit.

gtsambos:
@jeromekelleher ready for review 🚀

gtsambos:
@jeromekelleher ping in case you've forgotten (but no worries if not and you're just busy!) 🐘

jeromekelleher:
Sorry for the delay @gtsambos! I'm still struggling with a GitHub inbox backlog...

docs/api.rst
```diff
@@ -352,6 +353,10 @@ flags values are defined:
     Can be used in combination with the ``record_migrations`` argument to
     :func:`.simulate`.
 
+.. data:: msprime.NODE_IS_CEN_EVENT
+
+    The node was created by a `msprime.CensusEvent`.
```
jeromekelleher: Maybe link to the documentation with :class:`.CensusEvent`?

docs/tutorial.rst
```diff
@@ -618,6 +618,102 @@ Once you are satisfied that the demographic history that you have built
 is correct, it can then be simulated by calling the :func:`.simulate`
 function.
 
+-------------
+Census events
+-------------
+
+Census events allow you to add a node to each branch of the tree sequence at a given time during the simulation. This can be useful when you wish to study haplotypes that are ancestral to your simulated sample.
```
jeromekelleher: These lines are very long --- can you break them up a bit? This makes it easier to do diffs and stuff later.

jeromekelleher: Good to note that this allows to figure out *where* all the lineages are at this time as well, as this is the main application I think.

docs/tutorial.rst
```diff
@@ -618,6 +618,102 @@ Once you are satisfied that the demographic history that you have built
...
+    >>> ts = msprime.simulate(
+                population_configurations=[pop_config, pop_config],
+                length=1000,
+                demographic_events=[mig_rate_change, msprime.CensusEvent(time=5000)],
+                recombination_rate=1e-7,
+                random_seed=141)
+
+The resulting tree sequence has nodes on each tree at the specified census time. These are the nodes with IDs 9, 10, 11, 12 and 13:
+
+.. code-block:: python
```
jeromekelleher: We're trying to phase out using text trees in the documentation (#590) because they dont render properly in many browsers. Maybe we could do this with ``ts.draw_svg()``, and save the svg file (should be other examples of this, perhaps see the tskit tutorial for an example?)

docs/tutorial.rst
```diff
@@ -618,6 +618,102 @@ Once you are satisfied that the demographic history that you have built
...
+    12  1048576 1   -1  5000.00000000000000 
+    13  1048576 1   -1  5000.00000000000000 
+    14  0       1   -1  5246.90282987397495 
+    15  0       0   -1  8206.73121309170347
+
+If we wish to study these ancestral haplotypes further, we can simplify the tree sequence with respect to the census nodes and perform subsequent analyses on this simplified tree sequence:
+
+.. code-block:: python
+
+    >>> nodes = [i.id for i in list(ts.nodes()) if i.flags==msprime.NODE_IS_CEN_EVENT]
```
jeromekelleher: You don't need the ``list()`` here I think.

docs/tutorial.rst
```diff
@@ -618,6 +618,102 @@ Once you are satisfied that the demographic history that you have built
...
+    >>> ts_anc = ts.simplify(samples=nodes)
+
```
jeromekelleher: Maybe print out this simplified ts, and discuss how what it is? I think the last chunk will be a bit mysterious to many users unless you explain it a little bit.

lib/msprime.c
```diff
@@ -4225,6 +4229,85 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+
+    for (i = 0; i < (int) self->num_populations; i++) {
+        for (j = 0; j < (int) self->num_labels; j++) {
+
+            // Get segment from an ancestor in a population.
+            ancestors = &self->populations[i].ancestors[j];
+            node = ancestors->head;
+
+            while (node != NULL){
+                seg = node->item;
```
jeromekelleher: I think we need a cast here, ``seg = (segment_t *) node->item``.

lib/msprime.c
```diff
@@ -4225,6 +4229,85 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+                    if (ret != 0) {
+                        goto out;
+                    }
+                    ret = tsk_node_table_add_row(&self->tables->nodes,
+                            MSP_NODE_IS_CEN_EVENT, time, (population_id_t) i, TSK_NULL,
+                            NULL, 0);
+                    if (ret < 0) {
+                        goto out;
+                    }
+                    u = ret;
```
jeromekelleher: cast here too. ``u = (tsk_id_t) ret;``

lib/msprime.c
```diff
@@ -4225,6 +4229,85 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+                    // Add an edge joining the segment to the new node.
+                    ret = msp_store_edge(self, seg->left, seg->right, ret, seg->value);
```
jeromekelleher: Use u rather than ret.

lib/tests/tests.c
```diff
@@ -956,6 +962,60 @@ test_demographic_events(void)
...
+test_census_event(void)
+{
+    int ret;
+    uint32_t n = 10;
+    sample_t *samples = malloc(n * sizeof(sample_t));
+    msp_t *msp = malloc(sizeof(msp_t));
+    gsl_rng *rng = gsl_rng_alloc(gsl_rng_default);
+    recomb_map_t recomb_map;
+    tsk_table_collection_t tables;
+    int no_census_nodes = 0;
```
jeromekelleher: I read this as "no census nodes``, ie, as a boolean which would be true if there were no census nodes. ``num_census_nodes`` would follow the local conventions better.

msprime/simulations.py
```diff
@@ -1201,6 +1202,30 @@ def __str__(self):
...
+    """
+    An event that adds a node to each branch of every tree at a given time
+    during the simulation. This may be used to record all ancestral haplotypes
+    present at that time, and to extract other information related to these
+    haplotypes: for instance to trace the local ancestry of a sample back to a
+    set of contemporaneous ancestors, or to assess whether a subset of samples
+    has coalesced more recently than the census time.
+
```
jeromekelleher: Add in a link to the tutorial here somewhere.

docs/tutorial.rst
gtsambos: I wasn't able to draw the simplified ts because of [this problem](https://github.com/tskit-dev/tskit/issues/236) but have added a bit of extra explanation.

gtsambos:
Hi @jeromekelleher, I assume I did something to mess up the CI while I was fixing these other things -- it's hard to know what though, as none of these tests fail on my computer! And a bit strange as I was mostly only touching the docs. I don't have time to look closer now but will do again tomorrow.

docs/tutorial.rst
```diff
@@ -618,6 +618,79 @@ Once you are satisfied that the demographic history that you have built
...
+    >>> pop_config = msprime.PopulationConfiguration(sample_size=2, initial_size=1000)
+    >>> mig_rate_change = msprime.MigrationRateChange(time=100, rate=0.05)
+    >>> ts = msprime.simulate(
+                population_configurations=[pop_config, pop_config],
+                length=1000,
+                demographic_events=[mig_rate_change, msprime.CensusEvent(time=5000)],
+                recombination_rate=1e-7,
+                random_seed=141)
+
+The resulting tree sequence has nodes on each tree at the specified census time. These are the nodes with IDs 8, 9, 10, 11, 12 and 13:
```
jeromekelleher: Wide lines here and below

lib/msprime.c
```diff
@@ -4225,6 +4229,85 @@ msp_add_instantaneous_bottleneck(msp_t *self, double time, int population_id,
...
+                    }
+                    ret = tsk_node_table_add_row(&self->tables->nodes,
+                            MSP_NODE_IS_CEN_EVENT, time, (population_id_t) i, TSK_NULL,
+                            NULL, 0);
+                    if (ret < 0) {
+                        goto out;
+                    }
+                    u = (tsk_id_t) ret;
+                    // Add an edge joining the segment to the new node.
+                    u = msp_store_edge(self, seg->left, seg->right, ret, seg->value);
```
jeromekelleher: This is wrong. I meant use u as the parameter to store_edge:
```
ret = msp_store_edge(self, seg->left, seg->right, u, seg->value);
if (ret != 0) {
}
```

gtsambos: ah 🤦‍♂

jeromekelleher:
Excellent! Should we wait on a fix for the SVG tree sequence output, or is the one used here unaffected?

gtsambos:
The one in here at the moment is affected by the odd scaling of node heights issue ([issue 319](https://github.com/tskit-dev/tskit/issues/319) in tskit), but it's otherwise okay, and I think the text and printed table makes it somewhat clear what this output *should* look like.

I think it would make more sense to keep the SVG issues for a separate pull request, as this one is already quite long and unwieldy. In this other pull request we could keep a log of existing code that is affected by `draw_svg` (including this one). I'm not sure exactly when I'll get round to the SVG stuff as it's a public holiday here tomorrow -- and only 2 more workdays before my 30ish hrs of travel next week 🙃

jeromekelleher:
OK, let's merge then. I'll open an issue to regenerate the SVG here.

Great work, thanks @gtsambos!

PR #800, opened 2019-08-10
Likelihood

JereKoskela:
This is a fairly direct reimplementation of the Python prototype. I've tried to follow the readthedocs style guide, but let me know if some gremlins have sneaked in.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,14 @@
 ********************
+[0.7.3] - 2019-08-03
+********************
+
+**Bug fixes**:
+
+- Support for SMC models coupled with the record_full_arg feature was
+  erroneously removed in a previous version (:issue:`795`). The feature
+  has been resinstated (:pr:`796`).
+
+********************
 [0.7.2] - 2019-07-30
 ********************
 


jeromekelleher:
This looks good, thanks @JereKoskela. I'll need to spend a little time working through it though, and I'm away at the moment. Is it OK if I go through this next week?

JereKoskela:
> This looks good, thanks @JereKoskela. I'll need to spend a little time working through it though, and I'm away at the moment. Is it OK if I go through this next week?

Yeah, that's fine. There's no particular hurry.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+    ret = tsk_treeseq_get_node(ts, edge->parent, parent);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_tree_init(tree, ts, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    do {
+        total_material += (edge->right - edge->left)
```
jeromekelleher: I'd do this slightly differently:
```c
static double 
get_total_material(tsk_treeseq_t *ts)
{
    double ret = 0.0;
    tsk_edge_table_t *edges = ts->tables.edges;
    tsk_size_t j;

    for (j = 0; j < edges->num_rows; j++) {
         ret += (edges->right[j] - edges->left[j]) * (// etc)
    }
    return ret;
}
```

Saves quite a bit of faffing and is more efficient.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+        total_material += (edge->right - edge->left) 
+            * (parent->time - child->time);
+    } while (msp_increment_edge(ts, edge, parent, child) == 0);
+    if (theta > 0) {
+        *lik = num_mutations * log(total_material * theta) -
+            total_material * theta;
+        for (it = tsk_tree_first(tree); it == 1; it = tsk_tree_next(tree)) {
+            for (j = 0; j < tree->sites_length; j++) {
+                mut = tree->sites[j].mutations[0];
+                ret = tsk_treeseq_get_node(ts, mut.node, child);
```
jeromekelleher: We can save a bit of hassle here too with
```
// in the header
const double *node_time = ts->tables.nodes->time;

// later; parent and child are the integer node ids.
 branch_length = node_time[parent] - node_time[child];

File: lib/likelihood.c [hunk 3]
```diff
@@ -0,0 +1,304 @@
...
+                if (ret != 0) {
+                    goto out;
+                }
+                ret = tsk_tree_get_parent(tree, child->id, &node_id);
```
jeromekelleher: Might as well use ``tree->parent[child]`` here, we're sure the IDs are good so we don't nee the error checking.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+                if (ret != 0) {
+                    goto out;
+                }
+                ret = tsk_treeseq_get_node(ts, node_id, parent);
+                if (ret != 0) {
+                    goto out;
+                }
+                branch_length = parent->time - child->time;
+                ret = tsk_tree_get_parent(tree, parent->id, &node_id);
+                while (tsk_tree_get_parent(tree, parent->id, &node_id) == 0
```
jeromekelleher: Ditto ``tree->parent`` here rather than the function.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+    unsigned int j;
+    tsk_mutation_t mut;
+    tsk_size_t num_mutations = tsk_treeseq_get_num_mutations(ts);
+    tsk_id_t node_id;
+    double total_material = 0;
+    double branch_length;
+    tsk_tree_t *tree = NULL;
+    tsk_edge_t *edge = NULL;
+    tsk_node_t *parent = NULL;
+    tsk_node_t *child = NULL;
```
jeromekelleher: Space here between variable declarations and code.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+    int i;
+    int lineages = (int)tsk_treeseq_get_num_samples(ts);
+    tsk_id_t parent_id;
+    double sim_time = 0;
+    double material = lineages * tsk_treeseq_get_sequence_length(ts);
+    double material_in_children;
+    double material_in_parent;
+    double rate;
+    double gap;
+    int first_parent_edge[ts->tables->nodes.num_rows];
```
jeromekelleher: We need to malloc these rather than allocating on the stack. You can only really allocate small bounded things on the stack safely.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+    parent = malloc(sizeof *parent);
+    if (parent == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    child = malloc(sizeof *child);
+    if (child == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
```
jeromekelleher: An easy and fast way to set these is 
```
memset(first_parent_edge, 0xff, sizeof(first_parent_edge) * num_nodes);
```

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+    for (i = 0; i < (int)ts->tables->nodes.num_rows; i++) {
+        first_parent_edge[i] = -1;
+        last_parent_edge[i] = -1;
+    }
+    for (i = 0; i < (int)ts->tables->edges.num_rows; i++) {
+        ret = tsk_treeseq_get_edge(ts, i, edge);
+        if (ret != 0) {
+            goto out;
+        }
+        if (first_parent_edge[edge->child] == -1) {
```
jeromekelleher: Might be good to use TSK_NULL here rather than -1. Might be better to use ``tsk_id_t`` types rather than plain ints also.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+            if (ret != 0) {
+                goto out;
+            }
+            material -= gap;
+            lineages++;
+            if (!(gap > 0)) {
+                // we evaluate the density rather than probability
+                gap = 1;
+            }
+            if (!(rho > 0)) {
```
jeromekelleher: Why not ``rho <= 0``?

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+                *lik -= DBL_MAX;
```
jeromekelleher: Why -=  DBL_MAX? Wouldn't you be setting it to DBL_MAX either way?

lib/tests/tests.c
```diff
@@ -346,7 +347,7 @@ test_single_locus_historical_sample(void)
     ret = tsk_table_collection_init(&tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    for (j = 0; j < 2; j++) {
+    for (j = 1; j < 2; j++) {
```
jeromekelleher: I think these changes are brought in because you merged upstream rather than rebased. Git is lying to you when it tells you you should merge from an upstream remote. Always rebase...

lib/tests/tests.c
```diff
@@ -1649,6 +1591,473 @@ test_dtwf_multi_locus_simulation(void)
...
+    for (i = 0; i < 0; i++) {
+        ll_exact = (5 * log(tree_length * theta[i]) - tree_length * theta[i]);
+        ll_exact -= 2 * log(4 * tree_length);
+        ll_exact -= 2 * log(tree_length);
+        ll_exact += log(3 / (4 * tree_length));
+        ret = msp_unnormalised_log_likelihood_mut(&ts, theta[i], &lik);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        CU_ASSERT_DOUBLE_EQUAL(ll_exact, lik, tol);
+    }
+    // Second test case: two MRCAs
```
jeromekelleher: Maybe split these into three separate functions? Would be easier to follow then I think.

lib/likelihood.c
JereKoskela: You're right. That should be *lik = -DBL_MAX;

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+                    node_id = tree->left_child[child->id];
+                    ret = tsk_treeseq_get_node(ts, node_id, child);
+                    if (ret != 0) {
+                        goto out;
+                    }
+                    branch_length += parent->time - child->time;
+                }
+                *lik += log(branch_length / total_material);
+            }
+        }
```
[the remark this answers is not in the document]
JereKoskela: Done. I put this right at the top of the body of the for loop to catch it as early as possible.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+                                    double *lik) {
+    int ret = 1;
+    int it;
+    unsigned int j;
+    tsk_mutation_t mut;
+    tsk_size_t num_mutations = tsk_treeseq_get_num_mutations(ts);
+    tsk_id_t node_id;
+    double total_material = 0;
+    double branch_length;
+    tsk_tree_t *tree = NULL;
```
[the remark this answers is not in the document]
JereKoskela: I've tried to follow this example throughout. Let me know if I've misinterpreted anything.

lib/likelihood.c
```diff
@@ -0,0 +1,304 @@
...
+    ret = tsk_tree_init(tree, ts, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    do {
+        total_material += (edge->right - edge->left) 
+            * (parent->time - child->time);
+    } while (msp_increment_edge(ts, edge, parent, child) == 0);
+    if (theta > 0) {
+        *lik = num_mutations * log(total_material * theta) -
```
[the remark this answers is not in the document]
JereKoskela: Sounds sensible. Done.

lib/tests/tests.c
JereKoskela: I've changed the for loop back to for (j = 0; j < 2; j++) {

Is that enough, or do I need to do something fancier to set the record straight?

lib/tests/tests.c
JereKoskela: Do you mean a separate test function for each test case, in which case there would be five? Or some other split?

JereKoskela:
Thanks @jeromekelleher. I've followed your suggestions, and the resulting code is a lot shorter and tidier.

I'm not sure what (if anything) I should do about the changes brought in by my lack of rebasing. I'd appreciate your guidance. I also haven't yet split up anything in the likelihood test, because I wasn't quite sure what you meant. Everything else you commented on should now be fixed.

lib/tests/tests.c
jeromekelleher: No, you still need to rebase to get the history sorted out. See the documentation [here](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow) for guidance. We shouldn't see any merge commits in the history for this PR. The [network view](https://github.com/tskit-dev/msprime/network) is helpful for visualising. Your branch should fork off once, directly from the tip of upstream/master.

lib/tests/tests.c
jeromekelleher: If the tests are fairly independent, then it'd be good to split them up into separate functions. Five seems good --- good to know we're covering different situations.

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+msp_unnormalised_log_likelihood_mut(tsk_treeseq_t *ts, double theta,
+                                    double *r_lik) {
+    int ret = 1;
+    tsk_id_t it;
+    tsk_size_t j;
+    tsk_mutation_t mut;
+    tsk_size_t num_mutations = tsk_treeseq_get_num_mutations(ts);
+    const double total_material = get_total_material(ts);
+    const double *node_time = ts->tables->nodes.time;
+    double branch_length;
```
jeromekelleher: can declare branch_length and lik on the same line here, they're simple local variables that don't need initialialisation

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+    double lik;
+    tsk_tree_t tree;
+    tsk_id_t parent;
```
jeromekelleher: Ditto parent, child and ``it`` here.

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+    
+    ret = tsk_tree_init(&tree, ts, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    if (theta > 0) {
+        lik = num_mutations * log(total_material * theta) -
+            total_material * theta;
+        for (it = tsk_tree_first(&tree); it == 1; it = tsk_tree_next(&tree)) {
+            if (it < 0) {
```
jeromekelleher: This doesn't work I'm afraid. We only enter the loop body if ``it == 1``, so this condition can never be true. The check needs to happen immediately **after** the loop.

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+                ret = it;
+                goto out;
+            }
+            for (j = 0; j < tree.sites_length; j++) {
```
jeromekelleher: We're assuming here that there's exactly one mutation per site. We should put in a check here to make sure it's true (we could segfault on zero mutations, anyway). We can return MSP_ERR_BAD_PARAM_VALUE for now, I guess, unless you see a more appropriate error code there.

lib/likelihood.c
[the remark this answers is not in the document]
jeromekelleher: See comment above.

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+int
+msp_log_likelihood_arg(tsk_treeseq_t *ts, double rho, double *r_lik) {
+    int ret = 1;
+    tsk_id_t i;
+    tsk_size_t lineages = tsk_treeseq_get_num_samples(ts);
+    double sim_time = 0;
+    double material = lineages * tsk_treeseq_get_sequence_length(ts);
+    double material_in_children;
```
jeromekelleher: Can combine some of these declarations that don't need initialisiation together to save a bit of vertical space.

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+    double material_in_parent;
+    double rate;
+    double gap;
+    double lik = 0;
+    tsk_edge_table_t *edges = &ts->tables->edges;
+    tsk_node_table_t *nodes = &ts->tables->nodes;
+    tsk_id_t *first_parent_edge;
```
jeromekelleher: These must be initialised to NULL:

```
tsk_id_t *first_parent_edge = NULL;
```

This is because we conditionally free the memory based on whether it's null or not. If an error occured allocating first parent edge here, we'd currently free ``last_parent_edge`` (which would be junk left on the stack), resulting in a segfault or similar.

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+    
+    first_parent_edge = malloc(nodes->num_rows * sizeof(tsk_id_t));
+    if (first_parent_edge == NULL) {
+        goto out;
+    }
+    last_parent_edge = malloc(nodes->num_rows * sizeof(tsk_id_t));
+    if (last_parent_edge == NULL) {
+        goto out;
+    }
+    memset(first_parent_edge, TSK_NULL, sizeof(tsk_id_t) * nodes->num_rows);
```
jeromekelleher: I didn't expect this to work, but it does now that I think about it.  Probably more readable than 0xff; I'll use this as the more readable alternative from now on!

lib/likelihood.c
```diff
@@ -0,0 +1,199 @@
...
+            first_parent_edge[edges->child[i]] = i;
+        }
+        last_parent_edge[edges->child[i]] = i;
+    }
+    while (lineages > 0) {
+        rate = lineages * (lineages - 1) / 2 + material * rho;
+        parent = edges->parent[edge];
+        lik -= rate * (nodes->time[parent] - sim_time);
+        sim_time = nodes->time[parent];
+        if (nodes->flags[parent] == MSP_NODE_IS_RE_EVENT) {
```
jeromekelleher: Flag are bitwise, so we should probably check just this flag to be more robust:
```c
if (nodes->flags[parent] & MSP_NODE_IS_RE_EVENT) {
}
``

File: lib/likelihood.c [hunk 24]
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+        if (lineages > 0) {
+            edge++;
+        }
+    }
+    *r_lik = lik;
+    ret = 0;
+out:
+    if (first_parent_edge != NULL) {
```
jeromekelleher: shorthand for this is 
```c
out:
    msp_safe_free(first_parent_edge);
```

Issue #804, opened 2019-09-04
meson build instructions on Mac OS X


gtsambos:
see emotional rollercoaster [here](https://github.com/tskit-dev/msprime/pull/799). Edit documentation for developers to suggest this for Mac OS C users.

gtsambos:
@jeromekelleher, would it be going too far to explicitly recommend that Mac users compile with `gcc` instead of `clang`? I don't have much understanding of the differences between different compilers and I wasn't haven't problems with `clang` on my laptop, but perhaps that was just lucky. And perhaps it might make it easier for you/others to help troubleshoot if we are all compiling things with the same compiler?

jeromekelleher:
I don't really have an opinion on this stuff I'm afraid @gtsambos --- I'm happy to go with what OSX users think in terms of recommendations for development. Ordinary users should be using conda I think, so choice of compiler isn't going to make any difference.

molpopgen:
I made a comment on this in the PR (#805).  It is still actually possible to `conda install gcc` on OS X?  I just tried on my mac laptop (Sierra, latest miniconda).  The command fails at the "solving environment" step.

When you are referring to `gcc` on your mac, what is the path of that compiler?  `which gcc` will tell you that?  I think the answer will be `/usr/bin/gcc`?  If so, that is the xcode clang.  You can see that in your comments in #799.  The output of `gcc --version` is referring to clang, and not to any actual `gcc` version.

gtsambos:
Hi @molpopgen, sorry, somehow I missed this amongst a flurry of other stuff last week. You're right, it's in `/usr/bin/gcc`. And yeah, I'm pretty sure I got my version of (what I thought was...) clang via Xcode, so that would make sense. I've just updated the PR - are there any changes to that text that you would suggest?

gtsambos:
ps. I'm on a slightly later version of OS X - Mojave. 

>  It is still actually possible to conda install gcc on OS X? I just tried on my mac laptop (Sierra, latest miniconda) 

Will try this later (need to be careful to ensure I don't break things that are currently working for me!)

molpopgen:
> Will try this later (need to be careful to ensure I don't break things that are currently working for me!)

I wouldn't try too hard!  Attempting to `conda install gcc` on OS X with the "latest" miniconda causes the system to go into a long process where it tries to back-calculate how far it needs to downgrade in order to make your request work.  I killed it after about 5 minutes.

[...]

Issue #801, opened 2019-08-23
Mac instructions for making the Makefile


gtsambos:
On machines that compile with clang (eg. Macs by default), just typing `make` produces errors, but `make ext3` works. This note should be added to the developer documentation.
<img width="506" alt="Screen Shot 2019-08-23 at 1 34 09 pm" src="https://user-images.githubusercontent.com/17755074/63592993-d3cc3b00-c5aa-11e9-8545-cefc50553744.png">

jeromekelleher:
Closed in #805.

PR #803, opened 2019-09-02
Update appveyor configutation.

jeromekelleher:

PR #805, opened 2019-09-05
Updated developer docs with info for Mac users.

gtsambos:
See #804 and #801.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -225,14 +225,14 @@ directory, run
 
       ERROR: Compiler x86_64-apple-darwin13.4.0-clang can not compile programs.
 
-   This is due to incompatibilities between meson and some versions of ``clang`` (specifically: ``clang_osx-64`` and ``clangxx_osx-64``), which is the default compiler on Mac OS X systems. If you have this problem, you have a few options.
-   One option is to install the ``gcc`` compiler from Xcode. You can then explicitly specify your meson build to use gcc instead of clang:
+   This is due to setup issues with some versions of ``clang`` (specifically: ``clang_osx-64`` and ``clangxx_osx-64``), which is the default compiler on Mac OS X systems. If you have this problem, you have a few options.
+   One option is to install Xcode. You can then explicitly specify your meson build to use the  ``clang`` compiler from Xcode via its alias, ``gcc``.
 
    .. code-block:: bash
 
       $ CC=gcc CXX=g++ meson build
 
-   If you wish, you can remove these problematic versions of clang from the conda environment completely. Take care, though - this will also remove any packages that depend on these versions of clang!
+   If you wish, you can remove the problematic default versions of clang from the conda environment completely. Take care, though - this will also remove any packages that depend on them!
 
    .. code-block:: bash
 
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -218,6 +218,26 @@ directory, run
     $ cd lib 
     $ meson build
 
+.. note::
+   Some Mac OS X users may run into errors like this one when running ``meson build``:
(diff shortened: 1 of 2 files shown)


docs/development.rst
```diff
@@ -26,6 +26,10 @@ Quickstart
 - Install the :ref:`basic requirements <sec_installation_system_requirements>`.
 - Install the Python development requirements using ``pip install -r requirements/development.txt``.
 - Build the low level module by running ``make`` in the project root.
+
+.. note::
+   Mac OS X users should run ``make ext3`` instead.
```
jeromekelleher: I think we should change the makefile here so that the default to build without lots of checks, and add a target ``allchecks`` which does the current default. This will make this advice unnecessary.

We can either do this here (if you're happy to do it) or create an issue to follow up on. Either way, we should remove this note and the other one at the end.

gtsambos: yup, happy to look into how to do this - I've never made a makefile before but not afraid of having a poke around.

jeromekelleher:
Thanks @gtsambos. Don't worry about the Appveyor builds failing btw, this is an unrelated issue (#803).

docs/development.rst
gtsambos: I've just changed the order of the existing rules around -- so by default, `make` will run the `ext3` rule, and the `allchecks` rule (previously the default) is only run when explicitly specified by the user. Is this what you had in mind?

gtsambos:
Hi @jeromekelleher, I think this is ready for review 🎸

gtsambos:
Maybe you too @molpopgen 🎹

jeromekelleher:
LGTM, thanks @gtsambos! @molpopgen, do you agree with this advice for OSX users?

molpopgen:
i made a minor comment re: clarifying what gcc really is on os x.  @gtsambos -- this stuff is definitely confusing! :)

docs/development.rst
jeromekelleher: Perfect, thanks.

jeromekelleher:
OK great; ping me when this is ready to merge.

gtsambos:
@molpopgen - I can't see your comment, would you mind reposting it?

molpopgen:
@gtsambos D'oh--another github review fail!  I just submitted the review.  You should see it now.

docs/development.rst
```diff
@@ -218,6 +218,26 @@ directory, run
...
 
+.. note::
+   Some Mac OS X users may run into errors like this one when running ``meson build``:
+
+   .. code-block:: bash
+
+      ERROR: Compiler x86_64-apple-darwin13.4.0-clang can not compile programs.
+
+   This is due to incompatibilities between meson and some versions of ``clang`` (specifically: ``clang_osx-64`` and ``clangxx_osx-64``), which is the default compiler on Mac OS X systems. If you have this problem, you have a few options.
+   One option is to install the ``gcc`` compiler from Xcode. You can then explicitly specify your meson build to use gcc instead of clang:
```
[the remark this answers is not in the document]
gtsambos: great -- thanks for this! 🎶

gtsambos:
ping @jeromekelleher @molpopgen

molpopgen:
Looks good to me.  @jeromekelleher will have to merge.

I think removing the new clang packages should be safe.  They should not be getting installed as a run-time dependency.  If so, that should be considered a recipe error.  At least on Linux, the run time for the compiler and the compiler itself are different conda packages.  We'd need to test on OSX, though.

jeromekelleher:
Merged - thanks @gtsambos and @molpopgen!

PR #810, opened 2019-09-15
Basic C implementation of likelihood evaluation

JereKoskela:
The history looks ok at my end this time, and I think I've caught everything we discussed in the previous PR. Apologies that the edits in the likelihood files are somewhat imprecise in comparison to what they would have been if I hadn't deleted the earlier PR. I hope I haven't created too much extra work.

### DIFF
--- a/lib/likelihood.c
+++ b/lib/likelihood.c
@@ -1,0 +1,193 @@
+/*
+** Copyright (C) 2019 University of Oxford
+**
+** This file is part of msprime.
+**
+** msprime is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** msprime is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with msprime.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/*
+ * Implementation of log-likelihood functions for full-ARG.
+ */
+
+#include <float.h>
+#include <math.h>
+#include <stdlib.h>
+#include "likelihood.h"
+#include "msprime.h"
+
+static double
+get_total_material(tsk_treeseq_t *ts) {
+    double ret = 0;
+    tsk_size_t j;
+    tsk_edge_table_t *edges = &ts->tables->edges;
+    tsk_node_table_t *nodes = &ts->tables->nodes;
+    tsk_id_t parent;
+    tsk_id_t child;
+    
+    for (j = 0; j < edges->num_rows; j++) {
+        parent = edges->parent[j];
+        child = edges->child[j];
+        ret += (edges->right[j] - edges->left[j])
+            * (nodes->time[parent] - nodes->time[child]);
+    }
+    return ret;
+}
+
+int
+msp_unnormalised_log_likelihood_mut(tsk_treeseq_t *ts, double theta,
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Apologies for the delay on this @JereKoskela, I'm at a conference right now. I'll look at this at the end of the week.

JereKoskela:
No rush. I've been slow as well, and in fact I'm also at a workshop this week so unlikely to respond to comments until next week.

lib/likelihood.c
```diff
@@ -0,0 +1,193 @@
...
+                child = parent;
+                parent = tree.parent[parent];
+                while (parent != TSK_NULL && tree.left_child[child]
+                        == tree.right_child[child]) {
+                    branch_length += node_time[parent] - node_time[child];
+                    child = parent;
+                    parent = tree.parent[parent];
+                }
+                child = mut.node;
+                while (tree.left_child[child] >= 0 && tree.left_child[child]
```
jeromekelleher: I'm not following this loop; it looks like you're iterating over the siblings of ``child``, but I think that's what it's doing. Can you explain what you're doing here?

lib/likelihood.c
```diff
@@ -0,0 +1,193 @@
...
+    tsk_id_t *last_parent_edge = NULL;
+    tsk_id_t edge = 0;
+    tsk_id_t parent;
+    
+    first_parent_edge = malloc(nodes->num_rows * sizeof(tsk_id_t));
+    if (first_parent_edge == NULL) {
+        goto out;
+    }
+    last_parent_edge = malloc(nodes->num_rows * sizeof(tsk_id_t));
+    if (last_parent_edge == NULL) {
```
jeromekelleher: Can combine these two:
```
first_parent_edge = malloc(nodes->num_rows * sizeof(tsk_id_t));
last_parent_edge = malloc(nodes->num_rows * sizeof(tsk_id_t));
if (first_parent_edge == NULL || last_parent_edge) {...}
```

lib/likelihood.c
JereKoskela: This iterates leafwards from the mutation, and checks how much branch length there is in which the mutation could have taken place without changing the pattern of observed SNPs. There are two ways in which that window can end: either we reach a leaf, or there is a bifurcation, in which case the mutation cannot have happened any further leafwards. 

`tree.left_child[child] >= 0` catches when we reach a leaf, and `tree.left_child[child] == tree.right_child[child]` catches when a bifurcation happens, i.e. when the valid branch length for the given mutation ends. Unary nodes have `tree.left_child[node] == tree.right_child[node]`.

jeromekelleher: I see, this is subtle and clever. Can you put this in as a comment please?

It would be a bit more idiomatic to catch the leaf case with ``tree.left_child[child]  != TSK_NULL``.

jeromekelleher:
I think we should merge this much after these last couple of comments are dealt with @JereKoskela. Then we can think about wiring it all up with a user interface (which will be straightforward; I can do it if you like, or I can point you in the right direction if you'd prefer).

One thing we need to decide though: should we use the scaled rho and theta in the UI, or use Ne, recombination_rate, mutation_rate? I'd be heavily in favour of the latter in the interest of consistency, but don't want to be too prescriptive about it.

JereKoskela:
I've dealt with the comments, and am happy for you to merge if you are @jeromekelleher. I'm also happy for you to set up the user interface, particularly as it sounds like you already know what you're aiming for.

There's no downside to using Ne, recombination_rate, and mutation_rate in the UI as far as I can see, so let's go for that. There are examples of both conventions in earlier Monte Carlo algorithms (never mind the fact that some use 2Ne, some use 4Ne, etc.) so we'll be consistent with some and inconsistent with others no matter what we do.

jeromekelleher:
Great, thanks @JereKoskela, merging.

PR #812, opened 2019-09-25
Try executing gsl-config in shell

hyanwong:
Fixes #803 (hopefully)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,7 @@ install:
   - cmd: python --version
 
 build_script:
+  - cmd: SET MSP_CONDA_PREFIX=%PYTHON%
   - cmd: python setup.py build_ext --inplace
   - cmd: python -m pip install newick
   - cmd: python -m nose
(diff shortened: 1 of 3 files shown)


hyanwong:
Doesn't work yet, disappointingly.

jeromekelleher:
Does running gsl-config on windows produce anything useful? (Can you check please @marianne-aspbury?)

marianne-aspbury:
Not really, we can't get it to run. Have tried various ways of trying to interpret it as a bash script but ultimately windows complains: 
`
'gsl-config' is not recognized as an internal or external command,
operable program or batch file.
`
It definitely finds it, just can't open it. e.g. specifying `shell=True` in `subprocess.call()` just opens the file in a text editor. Otherwise it just complains the above.

jeromekelleher:
OK, thanks. I wouldn't waste time with it, I don't think it's written to run on windows. It's in config.py as a "on the rate ocasions this works it's useful, but we can't rely on it".

hyanwong:
It would obviously be neater to be able to run `gsl-config` on windows (it works fine on OS X). There must be a way. I'm just investigating.

PR #813, opened 2019-09-25
Plumbing for the log likelihood C code.

jeromekelleher:
Here's the plumbing for part of the log likelihood code @JereKoskela, can you take a look through please?

I didn't hook up the mutation code as it's not clear to me why we're separating the two at this point. It seems more user-friendly to have a single function that'll return a single likelihood value rather than breaking it up into two. When we're computing lots of these things on slightly modified tree sequences then yes, we'll need it to be modular, but for now I think simpler is better. (In that case, we'll want an object to maintain state anyway, so it'll need a different interface; we want the simple functional approach as a building block anyway though).

I'd suggest we merge this much anyway, and hopefully you can update it with what's needed for the mutation stuff as a follow-up.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -30,6 +30,7 @@
 #include <gsl/gsl_errno.h>
 
 #include "msprime.h"
+#include "likelihood.h"
 
 /* We keep a reference to the gsl_error_handler so it can be restored if needed */
 static gsl_error_handler_t *old_gsl_error_handler;
@@ -77,6 +78,12 @@ static void
 handle_library_error(int err)
 {
     PyErr_SetString(MsprimeLibraryError, msp_strerror(err));
+}
+
+static void
+handle_tskit_library_error(int err)
+{
+    PyErr_SetString(MsprimeLibraryError, tsk_strerror(err));
 }
 
 static void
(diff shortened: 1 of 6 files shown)


jeromekelleher:
I've made a small change here to make input recombination rates <= 0 an error. Doesn't seem worth complicating the C code by bothering  with?

JereKoskela:
Thanks @jeromekelleher. Nothing looks out of place to me overall. Three comments:

1) As you say, the reason I wrote the two log likelihood methods separately was that we could try multiple mutation rates on the same trees without recomputing the arg likelihood. I think it also makes the individual methods more readable. A simple way to combine them into one would be to make a `log_likelihood(arg, recombination_rate, mutation_rate, Ne=0.25)` that returns the sum `_msprime.log_likelihood_arg(...) + _msprime.log_mutation_likelihood(...)`. Would that make sense? 

2) I think we will want to allow rho = 0, which should result in zero likelihood (or -inf log likelihood) when there are recombinations in the ancestry. That will help e.g. build methods for testing a rho = 0 hypothesis without needing to change the plumbing in the likelihoods. rho < 0 is unambiguously an error though.

3) On the need to scale branch lengths according to Ne: does tskit currently store branch lengths in units of generations by default? That didn't come up in any of the tests, print statements etc that I ran while building the likelihoods.

jeromekelleher:
> As you say, the reason I wrote the two log likelihood methods separately was that we could try multiple mutation rates on the same trees without recomputing the arg likelihood. I think it also makes the individual methods more readable. A simple way to combine them into one would be to make a log_likelihood(arg, recombination_rate, mutation_rate, Ne=0.25) that returns the sum _msprime.log_likelihood_arg(...) + _msprime.log_mutation_likelihood(...). Would that make sense?

Yes, this sounds like the right approach. We can keep it as two functions in the C code (as you say, this is better to read), but let's expose a single high-level Python function first. (I think we'll need more state when we want to do this multiple times efficiently anyway, so we'll need an object.)

> think we will want to allow rho = 0, which should result in zero likelihood (or -inf log likelihood)

Sounds perfectly reasonable. We can catch this as a special case early on I think (if it doesn't just fall out automatically) to avoid complicating the main code.

> On the need to scale branch lengths according to Ne: does tskit currently store branch lengths in units of generations by default? That didn't come up in any of the tests, print statements etc that I ran while building the likelihoods

Tskit is agnostic to how branch lengths are stored --- it's up to you to know what the units are. Msprime outputs branch lengths in generations, so I think we should assume that the input tree sequence has branch lengths in generations.

If this much looks OK, would you mind if we merge and you can follow up this issues when suits? I'm under a bit of time pressure at the moment!

JereKoskela:
> If this much looks OK, would you mind if we merge and you can follow up this issues when suits? I'm under a bit of time pressure at the moment!

Yes, merging is fine by me.

jeromekelleher:
Great, done.

PR #830, opened 2019-11-12
clarify migration matrix; closes #828

### LINKED ISSUE #828: docs for migration_matrix do not define source/dest pops
In the Python API docs, neither the `migration_matrix` parameter for `msprime.simulate()`, nor the `matrix_index` parameter of `msprime.MigrationRateChange`, suggest which dimension of the matrix corresponds to the migration source and which to the destination. Unfortunately the examples provided use symmetric migration matrices, so do not clarify this.

petrelharp:
As discussed in #828, the docs for `simulate` don't specify which way the matrix goes. This rectifies that.


jeromekelleher:
LGTM, thanks @petrelharp.

PR #832, opened 2019-11-22
Update Doxyfile

hyanwong:
I have updated the Doxyfile to the newest format, and set `USE_MATHJAX = YES` rather than `NO` as was (strangely to me) the case before. I hope this might fix https://github.com/tskit-dev/tskit/issues/150


jeromekelleher:
I don't think the doxyfile does anything --- maybe this is just something left over from an older version? If so, we should just delete it.

jeromekelleher:
Doxygen is used in tskit to generate C API docs. But we're not doing that here.

hyanwong:
Good point. I forgot that msprime doesn't use Doxygen. I'll delete the file in this PR and update.

PR #833, opened 2019-11-26
Fix problem in very-low recombination DTWF.

### LINKED ISSUE #831: Runtime error with `model='dtwf'` and recombination if `length=` is too small.
This causes an immediate crash on version 0.7.3:

```
msp.simulate(10, Ne=1e3,
             model="dtwf", 
             mutation_rate=1e-8, 
             recombination_rate=1e-9
            )
```

The error message is:

> Assertion failed: (k > 0 && "Negative distance: Overflow in gsl_ran_exponential!"), function msp_dtwf_recombine, file lib/msprime.c, line 1279.

Switching to Hudson's algorithm, and/or setting the `length=` parameter to something large, behaves as expected.

jeromekelleher:
Closes #831.

@DomNelson, can you take a look through this please?

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1263,21 +1263,32 @@ msp_defrag_segment_chain(msp_t *self, segment_t *z)
     return 0;
 }
 
+static int64_t
+msp_dtwf_generate_breakpoint(msp_t *self, int64_t start)
+{
+    double mu = 1.0 / self->recombination_rate;
+    double x = gsl_ran_exponential(self->rng, mu);
+
+    /* Prevent overflowing into a negative number when we convert to int
+     * and clamp to UINT32_MAX since we can't have anything bigger than
+     * this anyway. */
+    x = GSL_MIN(UINT32_MAX, x);
+    assert(x >= 0);
+    /* TODO: is floor the correct operation here or should we round? */
+    return 1 + start + (int64_t) x;
+}
+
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Ping @DomNelson --- any chance of a quick review here please?

lib/msprime.c
```diff
@@ -1263,21 +1263,32 @@ msp_defrag_segment_chain(msp_t *self, segment_t *z)
...
+msp_dtwf_generate_breakpoint(msp_t *self, int64_t start)
+{
+    double mu = 1.0 / self->recombination_rate;
+    double x = gsl_ran_exponential(self->rng, mu);
+
+    /* Prevent overflowing into a negative number when we convert to int
+     * and clamp to UINT32_MAX since we can't have anything bigger than
+     * this anyway. */
+    x = GSL_MIN(UINT32_MAX, x);
+    assert(x >= 0);
```
DomNelson: Any benefit to just setting `x = UINT32_MAX` if `x < 0`? We'd avoid having an assertion then, but maybe that's masking too much.

jeromekelleher: I don't think x can be negative at this point, as gsl_ran_exponential should always return something positive. We only got negative values because we cast to an int, and the value was too large for an int64, causing it to flip negative. But, just to be sure, I stuck in an assert!

PR #834, opened 2019-11-26
Add mathjax path

hyanwong:
May fix https://github.com/tskit-dev/tskit/issues/150#issuecomment-536281106

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -371,3 +371,5 @@ intersphinx_mapping = {
 # Github repo
 issues_github_path = "tskit-dev/msprime"
 
+# Mathjax path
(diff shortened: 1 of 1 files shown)


hyanwong:
Thanks @jeromekelleher - I don't know if this will fix it, but it's probably worth a try. I note that the API docs are still not corrected on https://msprime.readthedocs.io/en/stable/api.html but I don't know when that gets updated from the master repo.

jeromekelleher:
I don't think this is a problem any more, see the comment here https://github.com/tskit-dev/tskit/issues/150

hyanwong:
Actually, it was not rendering correctly 5 mins ago, but is now, so I think my fix worked, and it just hadn't quite synced up when I checked. Closing

Issue #836, opened 2019-12-04
Incorrect length in mspms with invisible recombinations


jeromekelleher:
When invisible recombinations occur (i.e., a recombination that doesn't result in a distinct tree), mspms outputs incorrect lengths for the trees. e.g.:

```
mspms_dev.py 5 -T -r 2 100 1
887837739 1348004305 543203144

//
[10]((2:0.090,(1:0.040,3:0.040):0.050):0.208,(4:0.156,5:0.156):0.141);
[28]((3:0.090,(1:0.029,2:0.029):0.061):0.208,(4:0.156,5:0.156):0.141);
[6]((3:0.090,(1:0.029,2:0.029):0.061):0.971,(4:0.156,5:0.156):0.905);
[15]((3:0.090,(1:0.029,2:0.029):0.061):0.971,(4:0.156,5:0.156):0.905);
[13]((3:0.090,(1:0.029,2:0.029):0.061):0.661,(4:0.156,5:0.156):0.594);
[37]((3:0.090,(1:0.029,2:0.029):0.061):0.661,(4:0.156,5:0.156):0.594);
[10]((3:0.090,(1:0.029,2:0.029):0.061):0.661,(4:0.185,5:0.185):0.566);
jk@tempest$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 10 + 28 + 6 + 15 + 13 + 37 + 10
119
```
The lengths here sum to 119 not 100, as they should.

jeromekelleher:
Thanks to @fbaumdicker for spotting this issue!

PR #837, opened 2019-12-04
Test case for mspms incorrect length output.

jeromekelleher:
See #836


jeromekelleher:
This fixes the problem I think @fbaumdicker?

fbaumdicker:
This should indeed fix the problem, as long as for every
`right`
from
`left, right = tree.interval`
there is a breakpoint at right. Which should always be the case.
Otherwise, some trees could be omitted.

jeromekelleher:
Great, thanks @fbaumdicker. The breakpoints must be a superset of the left,right intervals, so we're safe there. Merging.

PR #838, opened 2019-12-04
Fixed memory allocation error in mutgen.

### LINKED ISSUE #806: mutate fails with a large number of mutations per site
Hello there, I have a treeseq I've produced in slim with a single mutation which is under selection. I've been trying to overlay neutral mutations upon that tree but without any success in keeping the aforementioned selected mutation.

Briefly, if I do the following after loading in my tree with pyslim:
`nonWF_mut = msprime.mutate(nomu_nonWF, rate=2.36e-8, random_seed=1, keep=True)`
 I obtain the following error:
`Assertion failed: (size < self->chunk_size), function tsk_blkalloc_get, file lib/subprojects/tskit/tskit/core.c, line 413.`
`Abort trap: 6`

If I adjust my overlaying of mutations to `keep=False` then it works fine. Unfortunately for what I'm trying to model, keeping the mutation is crucial. Any ideas as to what may be causing me to get the above error message?

jeromekelleher:
Closes #806

### DIFF
--- a/lib/dev-tools/dev-cli.c
+++ b/lib/dev-tools/dev-cli.c
@@ -652,7 +652,7 @@ run_simulate(const char *conf_file, const char *output_file, int verbose, int nu
     }
     get_configuration(rng, &msp, &tables, &mutation_params, &recomb_map, conf_file);
     ret = mutgen_alloc(&mutgen, mutation_params.mutation_rate, rng,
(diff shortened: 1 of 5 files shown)


jeromekelleher:
Missing coverage in the C diff can't be helped --- it's a small diff, and there are untestable error conditions in there.

PR #839, opened 2019-12-04
Debugging bizarre leak in ll module.

jeromekelleher:
Opening a PR so I don't forget about this. This seems to be a minor and unpredictable memory leak, where we have to run the stress_lowlevel script several times for it to show up. Sometimes it runs and there's no memory leak; other times, the RSS keeps incrementing steadily. Weird.

It's not a serious leak, but we should track it down at some point.


jeromekelleher:
Fix for the problem over in #840.

PR #840, opened 2019-12-05
Fix memory leak in low-level module.

jeromekelleher:
The census-event string was not decrefed and we therefore leaked memory.

This was not released, and the leak was pretty minor, so no big deal.


jeromekelleher:
Going to merge this, despite the Circle CI problems. These are infrastructure issues and the tests are fine.

PR #845, opened 2019-12-17
Don't assume that the start of a genetic map is at 0.

### LINKED ISSUE #844: mean_recombination_rate() assumes the start of the map is position 0
Given a genetic map like the example provided in the documentation (reproduced below), `RecombinationMap.mean_recombination_rate()` should give zero weight to the region 0-55550. Ie. it should be calculated to match `124.482178 / 100 / (183630013 - 55550)`, but current behaviour is to (effectively) calculate `124.482178 / 100 / 183630013`.

```
Chromosome  Position(bp)    Rate(cM/Mb)     Map(cM)
chr1        55550           2.981822        0.000000
chr1        82571           2.082414        0.080572
chr1        88169           2.081358        0.092229
chr1        254996          3.354927        0.439456
chr1        564598          2.887498        1.478148
...
chr1        182973428       2.512769        122.832331
chr1        183630013       0.000000        124.482178
```

grahamgower:
Closes #844.

### DIFF
--- a/tests/test_highlevel.py
+++ b/tests/test_highlevel.py
@@ -637,6 +637,38 @@ class TestSimulatorFactory(unittest.TestCase):
         mean_rr = recomb_map.mean_recombination_rate
         self.assertEqual(mean_rr, 0.0)
 
+        # Test mean_recombination_rate is correct after reading from
+        # a hapmap file. RecombinationMap.read_hapmap() ignores the cM
+        # field, so here we test against using the cM field directly.
+        def hapmap_rr(hapmap_file):
+            first_pos = 0
+            with open(hapmap_file) as f:
+                next(f)  # skip header
+                for line in f:
+                    pos, rate, cM = map(float, line.split()[1:4])
+                    if cM == 0:
+                        first_pos = pos
+            return cM / 100 / (pos - first_pos)
+
+        hapmap = """chr pos        rate                    cM
+                    1   4283592    3.79115663174456        0
+                    1   4361401    0.0664276817058413      0.294986106359414
+                    1   7979763   10.9082897515584         0.535345505591925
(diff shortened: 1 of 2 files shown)


grahamgower:
So a `tempfile.NamedTemporaryFile()` can't be opened read/write on windows?

grahamgower:
Maybe if I move the call to RecombinationMap.read_hapmap() off line 666 it will work?

jeromekelleher:
Windows is a lot stricter about filehandle hygiene than *nix systems seem to be. IIRC the simplest thing to do is to make a temporary directory, and write some files in there. Have a look around the tests for other places tempfiles are used, there should be a pattern.

tests/test_highlevel.py
```diff
@@ -637,6 +637,41 @@ def test_mean_recombination_rate(self):
...
+                    1   4283592    3.79115663174456        0
+                    1   4361401    0.0664276817058413      0.294986106359414
+                    1   7979763   10.9082897515584         0.535345505591925
+                    1   8007051    0.0976780648822495      0.833010916332456
+                    1   8762788    0.0899929572085616      0.906829844052373
+                    1   9477943    0.0864382908650907      0.971188757364862
+                    1   9696341    4.76495005895746        0.990066707213216
+                    1   9752154    0.0864316558730679      1.25601286485381
+                    1   9881751    0.0                     1.26721414815999"""
+        temp_dir = tempfile.mkdtemp(prefix="msp_test_mean_rr")
```
jeromekelleher: A nicer pattern here rather than manually deleting the tempdir is to use:
```python
with tempfile.TemporaryDirectory() as temp_dir:
    hapfile = os.path.join(temp_dir, "hapmap.txt")
    with open(...)
```

grahamgower: Done.

PR #847, opened 2020-01-08
Recomb map

### LINKED ISSUE #798: Simulations with recombination map slow
It looks like simulations that use a recombination map are much slower than those that don't. It appears that ``recomb_map_genetic_to_phys`` is getting called a **lot**. This is a pretty serious performance issue, so we should resolve it. There should be no reason that recombination map simulations are any slower, as in principle all the remapping should be done at the end.

ivan-krukov:
This should resolve #798.
Now the simulation with a recombination map have comparable runtimes to those with a fixed recomb rate.

Added a binary search for the recombination map positions in `recomb_map_genetic_to_phys`. Also needed to add a `cumulative` member to `recomb_map_t` that keeps track of the total recombination distance up to a position - this is what `s` was referring to [here](https://github.com/tskit-dev/msprime/blob/873a2124f0b6a5dd4295d5b8cf3ecf7597531abe/lib/recomb_map.c#L236)

### DIFF
--- a/lib/recomb_map.c
+++ b/lib/recomb_map.c
@@ -56,7 +56,7 @@ recomb_map_alloc(recomb_map_t *self, uint32_t num_loci, double sequence_length,
         double *positions, double *rates, size_t size)
 {
     int ret = MSP_ERR_BAD_RECOMBINATION_MAP;
-    double length;
+    double length, s;
     size_t j;
 
     memset(self, 0, sizeof(recomb_map_t));
@@ -69,7 +69,9 @@ recomb_map_alloc(recomb_map_t *self, uint32_t num_loci, double sequence_length,
     }
     self->positions = malloc(size * sizeof(double));
     self->rates = malloc(size * sizeof(double));
-    if (self->positions == NULL || self->rates == NULL) {
+    self->cumulative = malloc(size * sizeof(double));
+    
+    if (self->positions == NULL || self->rates == NULL || self->cumulative == NULL) {
         ret = MSP_ERR_NO_MEMORY;
         goto out;
     }
@@ -92,6 +94,13 @@ recomb_map_alloc(recomb_map_t *self, uint32_t num_loci, double sequence_length,
         self->rates[j] = rates[j];
         self->positions[j] = positions[j];
     }
+    
+    s = 0;
+    self->cumulative[0] = 0;
+    for (j = 1; j < size; j ++) {
+        s += (self->positions[j] - self->positions[j - 1]) * self->rates[j - 1];
+        self->cumulative[j] = s;
+    }
     ret = 0;
 out:
     return ret;
@@ -105,6 +114,9 @@ recomb_map_free(recomb_map_t *self)
     }
     if (self->rates != NULL) {
         free(self->rates);
+    }
+    if (self->cumulative != NULL) {
+        free(self->cumulative);
     }
     return 0;
 }
@@ -231,12 +243,10 @@ recomb_map_genetic_to_phys(recomb_map_t *self, double genetic_x)
          * map back into physical coordinates. */
         x = (genetic_x / num_loci) * self->total_recombination_rate;
         if (x > 0) {
-            s = 0;
-            k = 0;
-            while (s < x && k < self->size - 1) {
-                s += (p[k + 1] - p[k]) * r[k];
-                k++;
-            }
+            
+            k = msp_binary_interval_search(x, self->cumulative, self->size);
+            s = self->cumulative[k];
+            assert((s >= x) || (k >= self->size - 1));
             assert(k > 0);
             excess = 0;
             if (r[k - 1] > 0) {
--- a/lib/tests/tests.c
+++ b/lib/tests/tests.c
@@ -3175,6 +3175,60 @@ test_recomb_map_errors(void)
     recomb_map_free(&recomb_map);
 }
 
+/* Linear version of the genetic_to_phys coordinate translation, to
+   verify against the binary search version in
+   recomb_map_genetic_to_phys */
+static double
+recomb_map_genetic_to_phys_linear(recomb_map_t *self, double genetic_x)
+{
+    size_t k;
+    double ret = 0.0;
+    double x, s, excess;
+    double *p = self->positions;
+    double *r = self->rates;
+    double num_loci = self->num_loci;
+
+    assert(num_loci >= 1);
+    assert(genetic_x >= 0);
+    assert(genetic_x <= num_loci);
+    if (self->total_recombination_rate == 0) {
+        /* When we have a 0 total rate, anything other than m maps to 0. */
+        ret = genetic_x >= self->num_loci? self->sequence_length: 0;
+    } else if (self->size == 2) {
+        /* Avoid roundoff when num_loci == self->sequence_length */
+        ret = genetic_x;
+        if (self->sequence_length != num_loci) {
+            ret = (genetic_x / num_loci) * self->sequence_length;
+        }
+    } else if (genetic_x == num_loci) {
+        /* Map the right end of the chromosome to the right end of the
+         * chromosome, even if there are zero-recomb regions on the end. */
+        ret = self->sequence_length;
+    } else {
+        /* genetic_x is in the range [0,num_loci], and so we rescale
+         * this into [0,total_recombination_rate] so that we can
+         * map back into physical coordinates. */
+        x = (genetic_x / num_loci) * self->total_recombination_rate;
+        if (x > 0) {
+
+            s = 0;
+            k = 0;
+            while (s < x && k < self->size - 1) {
+                s += (p[k + 1] - p[k]) * r[k];
+                k++;
+            }
+            assert((s >= x) || (k >= self->size - 1));
+            assert(k > 0);
+            excess = 0;
+            if (r[k - 1] > 0) {
+                excess = (s - x) / r[k - 1];
+            }
+            ret = p[k] - excess;
+        }
+    }
+    return ret;
+}
+
 static void
 verify_recomb_map(uint32_t num_loci, double length, double *positions,
         double *rates, size_t size)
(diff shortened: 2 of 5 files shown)


ivan-krukov:
I also wanted to give a shoutout to @DomNelson who gave the idea for the fix and helped with 44f6055

ivan-krukov:
The binary search over the recombination map increases the performance of the sims with a map to be comparable with a uniform recomb rate. This translates to 2.38 seconds from 38.55 seconds on my machine® (GRCh37 chr1 map, Ne=1000, sample_size=1000).

@jeromekelleher, maintainers, a few questions:

1) The original #798 suggestion for calling `msp_genetic_to_phys` "in the end" didn't quite work - still needed to do an expensive linear search for each entry in the edge table. Maybe I missed the point? My attempt is on this commit [here](https://github.com/ivan-krukov/msprime/commit/053f5a268bae97eb005ff78d14a6cdf4ccf2990c).

2) As you can see the code coverage still fails, mostly due to the new function I added. With the unity testing framework, I am guessing the function needs to be "public" (it's currently `static`, and [not predeclared ](https://github.com/ivan-krukov/msprime/blob/recomb_map/lib/recomb_map.c#L215). Where would `binary_interval_search` fit? `util.c`? `msprime.c`? Should it be in the `recomb_map_` namespace?

Thanks

lib/recomb_map.c
```diff
@@ -196,6 +208,50 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
...
+            index = m;
+            goto fin;
+        } else if ((values[m] < query) && (m != l)) {
+            l = m;
+        } else if ((values[m] > query) && (m != r)) {
+            r = m;
+        } else {
+            l ++;
+            r = l;
+        }
```
daniel-goldstein: Could you avoid the linear search by adjusting this logic to something like `values[m] < query => l = m + 1` and otherwise `r = m`? You also wouldn't need the boundary checks since the `while` loop should catch that.

I'm new to the group so am poking around some issues/PRs (so please correct me if I'm missing anything)!

ivan-krukov: I am also new here!

The suggestion seems to work as far as not needing to take the branch above (237-240). However, I still need the ugly iteration [248-251](https://github.com/ivan-krukov/msprime/blob/recomb_map/lib/recomb_map.c#L248), since I want the leftmost value on the recombination map (I think, I might be missing the point). The tests affected are `test_zero_rate_two_intervals (tests.test_recombination_map.TestCoordinateConversion)` [here](https://github.com/ivan-krukov/msprime/blob/recomb_map/tests/test_recombination_map.py#L180).

daniel-goldstein: Ah, part of my suggestion included removing the if statement that checks equality between the value at `m` and `query`, leaving just the less than and else cases. This allows `l` and `r` to converge onto the leftmost equivalent element as opposed to stopping when any one is found which would require further iteration (this is the same as finding the "insertion point" of `query`).

A small additional note: you can remove the need for the if statements before and after the loop by initializing `r = n_values - 1`.

ivan-krukov: I see! This is great, it really simplifies things. Not sure why I settled on the verbose version in retrospect. 

Thank you.

lib/recomb_map.c
```diff
@@ -196,6 +208,31 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
     return ret;
 }
 
+/* Find the `index` of the interval within `values` the `query` fits, such that
+ * values[index-1] < query <= values[index]
+ * If necessary, will find the leftmost such index
+ */
+size_t
+recomb_map_binary_interval_search(double query, double *values, size_t n_values)
```
jeromekelleher: Since this isn't a "method" of recomb_map (as the first parameter isn't a ``recomb_map_t``) we should rename it.  It's just a binary search, isn't it, where we find the leftmost value <= query? So, ``msp_binary_search``, perhaps?

lib/tests/tests.c
```diff
@@ -4098,6 +4137,7 @@ main(int argc, char **argv)
         {"test_simple_recombination_map", test_simple_recomb_map},
         {"test_recombination_map_errors", test_recomb_map_errors},
         {"test_recombination_map_examples", test_recomb_map_examples},
+        {"test_recombination_map_examples", test_binary_interval_search},
```
jeromekelleher: Need to rename to ``test_binary_search`` -- not being tested at the moment.

lib/tests/tests.c
```diff
@@ -3028,6 +3028,45 @@ test_recomb_map_examples(void)
     verify_recomb_map(UINT32_MAX, 16.0, p2, r2, 7);
 }
 
+static void
+test_binary_interval_search(void)
+{
+        double values[] = {0, 10, 20, 30};
```
jeromekelleher: Really nice, good tests. I think we just want a few tests checking what happens in boundary cases also:

- What happens when we have lists of length 0, 1 and 2?

- What happens if we have a list of all zeros?

- What if we include negative numbers also?

It might make sense to split these tests up a little bit too.

jeromekelleher:
> As you can see the code coverage still fails, mostly due to the new function I added. With the unity testing framework, I am guessing the function needs to be "public" (it's currently static, and not predeclared . Where would binary_interval_search fit? util.c? msprime.c? Should it be in the recomb_map_ namespace?

I think ``util.c`` with name ``msp_binary_search`` is the best option.

lib/recomb_map.c
jeromekelleher: I'm not sure what we should do at the top of the range here. Probably it's best if we return an error if the query is > values[n_value - 1], otherwise we could have silent errors. Therefore, we'll need to change the return type to ``int64_t`` and return MSP_ERR_BAD_BINARY_SEARCH_QUERY, I think.

lib/recomb_map.c
```diff
@@ -196,6 +208,31 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
...
+{
+
+    size_t l = 0;
+    size_t r = n_values - 1;
+    size_t m, index;
+
+    while (l < r) {
+        m = (size_t)(floor((double)(l + r) / 2L));
```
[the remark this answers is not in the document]
ivan-krukov: I thought so as well, but `gcc` complains about losing precision here (circleci from 668c669  [link](https://circleci.com/gh/tskit-dev/msprime/1708?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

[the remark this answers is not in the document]
jeromekelleher:
```
  m = (size_t)((long long)(floor((l + r) / 2)));
```
There's no need for the floor (which converts to floating point) though: integer division is defined as the floor. You should be able to do (assuming these are all int64_t) ``m = (l + r) / 2ULL``.

lib/recomb_map.c
ivan-krukov: Should the function be decorated with `MSP_WARN_UNUSED`? In this case, should `recomb_map_genetic_to_phys` return an error code as well?

ivan-krukov: I think that the current implementation also allows the return of the last index at top of the range [source](https://github.com/tskit-dev/msprime/blob/master/lib/recomb_map.c#L236):

```
s = 0;
k = 0;
while (s < x && k < self->size - 1) {
    s += (p[k + 1] - p[k]) * r[k];
    k++;
}
assert(k > 0);
```
if `x` is larger than the last value of `s`, `k` can point to the last element of `p`. This is the behaviour that the `binary_interval_search` has as well (tested [here](https://github.com/tskit-dev/msprime/blob/c596ba848f20473f46043b4a8efed9caf2899630/lib/tests/tests.c#L3119)). This seems consistent to me, as the function looks for an appropriate interval where the query fits. If we look for a very large value, we find the last (open) interval. Maybe I am missing the point though :/

lib/recomb_map.c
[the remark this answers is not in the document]
ivan-krukov: 96125a5

lib/tests/tests.c
ivan-krukov: 9da4c4c

lib/recomb_map.c
jeromekelleher: No, you're right, let's keep the interface consistent with the existing code.

lib/util.h
```diff
@@ -74,6 +75,7 @@
 #define MSP_ERR_DTWF_ZERO_POPULATION_SIZE                           -38
 #define MSP_ERR_DTWF_UNSUPPORTED_BOTTLENECK                         -39
 #define MSP_ERR_BAD_PROPORTION                                      -40
+#define MSP_ERR_BAD_BINARY_SEARCH_QUERY                             -41
```
jeromekelleher: Excellent - just need to define the string that goes with this in ``util.c`` and we're done.

ivan-krukov: removed this error code since currently the binary search does not return errors (it is guaranteed to return a valid index, as discussed above).

jeromekelleher:
This is really excellent stuff, thanks @ivan-krukov! Merging.

ivan-krukov:
Glad I can contribute :D

petrelharp:
Excelllent!!!!!

hyl317:
It seems like this change about recomb map hasn't been integrated into the latest release. When will that be done? Wanna use that feature to do some simulation. Thx!

molpopgen:
@hyl317 install the master branch.  Definitely stable enough.

jeromekelleher:
> It seems like this change about recomb map hasn't been integrated into the latest release. When will that be done? Wanna use that feature to do some simulation. Thx!

It might be a while before this gets released unfortunately @hyl317, as we've got a lot of new features coming through the pipelines and are gearing up for a 1.0 release. Installing from master *should* be OK for now, but I definitely wouldn't recommend it for the next couple of weeks (or in general really; there could be subtle bugs, we don't do statistical tests on every commit...)

PR #862, opened 2020-01-27
Simulate in physical coordinates

daniel-goldstein:
Simulations now are computed in physical coordinates. The Fenwick tree used to track and sample segments measures cumulative recombination mass. Supports both discrete and continuous physical space (using the `discrete` flag on `RecombinationMap`).


lib/recomb_map.c
```diff
@@ -214,48 +194,39 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
 double
 recomb_map_genetic_to_phys(recomb_map_t *self, double genetic_x)
 {
```
jeromekelleher: Ditto here, we want this function as the higher level APIs that we're supporting still provide it.

lib/recomb_map.c
```diff
@@ -214,48 +194,39 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
...
-                excess = (s - x) / r[k - 1];
-            }
-            ret = p[k] - excess;
-        }
-    }
-    return ret;
+double
+recomb_map_position_to_recomb_mass(recomb_map_t *self, double pos)
+{
+    if (pos == self->positions[0]) return 0;
```
jeromekelleher: Always use braces:
```c
if (pos == self->positions[0]) {
    return 0;
}
```

lib/recomb_map.c
```diff
@@ -214,48 +194,39 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
...
-            excess = 0;
-            if (r[k - 1] > 0) {
-                excess = (s - x) / r[k - 1];
-            }
-            ret = p[k] - excess;
-        }
-    }
-    return ret;
+double
+recomb_map_position_to_recomb_mass(recomb_map_t *self, double pos)
```
jeromekelleher: Rename to ``recomb_map_position_to_mass()``

A quick comment describing what it does would help too.

lib/recomb_map.c
```diff
@@ -214,48 +194,39 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
...
+{
+    if (pos == self->positions[0]) return 0;
+    if (pos >= self->positions[self->size - 1]) return self->cumulative[self->size - 1];
+
+    size_t right = msp_binary_interval_search(pos, self->positions, self->size);
```
jeromekelleher: ``left`` and ``right`` arent great names. What about 
```c
double offset;
size_t index = msp_binary_interval_search(pos, self->positions, self->size);
assert(index != 0)
index--;
offset = ..
```

lib/recomb_map.c
```diff
@@ -214,48 +194,39 @@ recomb_map_phys_to_discrete_genetic(recomb_map_t *self, double x, uint32_t *locu
...
+    size_t left = right - 1;
+    double offset = pos - self->positions[left];
+    return self->cumulative[left] + offset * self->rates[left];
+}
+
+static double
+recomb_map_recomb_mass_to_position(recomb_map_t *self, double recomb_mass)
```
jeromekelleher: rename ``recomb_map_mass_to_position``

lib/fenwick.c
```diff
@@ -49,12 +49,12 @@ fenwick_alloc_buffers(fenwick_t *self)
 
     self->tree = NULL;
     self->values = NULL;
-    self->tree = calloc((1 + self->size), sizeof(int64_t));
+    self->tree = calloc((1 + self->size), sizeof(self->tree));
```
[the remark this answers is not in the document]
daniel-goldstein: yup

lib/recomb_map.c
```diff
@@ -166,30 +168,8 @@ recomb_map_get_num_loci(recomb_map_t *self)
 double
 recomb_map_phys_to_genetic(recomb_map_t *self, double x)
```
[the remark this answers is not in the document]
daniel-goldstein: I'll keep these around and remove unnecessary calls to them instead. I thought we were going to remove them entirely but either way this is cleaner for now.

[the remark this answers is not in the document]
jeromekelleher: Sounds good. Making them no-ops was a good quick hack to see if this all worked, but we want to start doing things properly now.

daniel-goldstein:
Addressed comments. I'm a bit wary of the effect on GC and other events although I _think_ I removed only the necessary translations of genetic -> phys. I believe GC should still operate fine in genetic space. If things look ok I'll squash.

msprime/simulations.py
```diff
@@ -967,17 +988,19 @@ def get_per_locus_recombination_rate(self):
     def get_size(self):
         return self._ll_recombination_map.get_size()
 
+    @deprecated(reason="Use sequence_length instead")
```
daniel-goldstein: remove

tests/test_lowlevel.py
```diff
@@ -36,15 +36,12 @@
 NULL_NODE = -1
 
 
-def uniform_recombination_map(num_loci=1, rate=0, L=None):
+def uniform_recombination_map(L=1, rate=0):
```
daniel-goldstein: skip failing num_loci tests

msprime/simulations.py
```diff
@@ -175,6 +175,7 @@ def simulator_factory(
         length=None,
         recombination_rate=None,
         recombination_map=None,
+        finite_recombination_sites=False,
```
jeromekelleher: Let's remove this parameter for now and have a think about what we want the API to be.

msprime/simulations.py
```diff
@@ -819,57 +825,77 @@ class RecombinationMap(object):
...
-    """
-    def __init__(self, positions, rates, num_loci=None):
-        if num_loci is None:
-            num_loci = self.DEFAULT_NUM_LOCI
+    def __init__(self, positions, rates, num_loci=None, discrete=False):
+        if num_loci is not None:
+            if num_loci == positions[-1]:
+                warnings.warn(
+                        "num_loci is deprecated and should not be used. "
+                        "The default behavior is that the number of loci "
```
jeromekelleher: Remove the second sentence.

msprime/simulations.py
```diff
@@ -819,57 +825,77 @@ class RecombinationMap(object):
...
-    def uniform_map(cls, length, rate, num_loci=None):
+    def uniform_map(cls, length, rate, num_loci=None, discrete=False):
         """
         Returns a :class:`.RecombinationMap` instance in which the recombination
         rate is constant over a chromosome of the specified length. The optional
-        ``num_loci`` controls the number of discrete loci in the underlying
-        simulation, and is by default large enough to be effectively be
-        a continuous model.
+        ``discrete`` controls whether recombination sites can occur only on integer
+        positions or can take continuous values. The optional ``num_loci`` is
```
jeromekelleher: It's not so much deprecated as removed. "The legacy ``num_loci`` option is no longer supported and should not be used".

msprime/simulations.py
```diff
@@ -819,57 +825,77 @@ class RecombinationMap(object):
...
+        deprecated and should not be used.
 
         The following map can be used to simulate a true finite locus model
         with a fixed number of loci ``m``::
 
-            >>> recomb_map = RecombinationMap.uniform_map(m, rate, num_loci=m)
+            >>> recomb_map = RecombinationMap.uniform_map(m, rate, discrete=True)
+
+        .. warning::
```
jeromekelleher: I don't think we need a warning here, can delete.

msprime/simulations.py
```diff
@@ -819,57 +825,77 @@ class RecombinationMap(object):
...
+            The ``num_loci`` parameter is deprecated. To set a discrete number of
+            possible recombination sites along the sequence, scale ``positions``
+            to the desired number of sites and set ``discrete=True`` to ensure
+            recombination occurs only at integer values.
 
         :param float length: The length of the chromosome.
         :param float rate: The rate of recombination per unit of sequence length
             along this chromosome.
-        :param int num_loci: The number of discrete loci in the underlying
+        :param int num_loci: **This parameter is deprecated**.
```
jeromekelleher: this parameter is no longer supported

jeromekelleher: Delete the old description -- we just want to tell the user that this positional parameter is there, and that it shouldn't be used.

jeromekelleher:
This is getting pretty close and could use some more eyes on it. Most things work now, except for recombination in the DTWF which we'll have to rethink a bit. (I think we should probably do this as a follow up PR). The genic selection code seems to have been roughed up a bit too.

@petrelharp, @andrewkern, @DomNelson, @fbaumdicker, any chance you could take a look through please?

daniel-goldstein:
Updated DTWF so that breakpoints are sampled using a rate equivalent to the average rate of recombination from [x->left, sequence_length). Tests for DTWF are now passing and no longer skipped

jeromekelleher:
Awesome, thanks @daniel-goldstein. Let's clear up any remaining skips if we can, and get ready to merge.

Has the algorithms.py file been updated to use the new logic?

daniel-goldstein:
> Has the algorithms.py file been updated to use the new logic?

it should be mostly there because it doesn't have much of the typing issues.  Will make sure it's all synced.

daniel-goldstein:
No new skips now. `algorithms.py` now reflects the backend logic

algorithms.py
```diff
@@ -1705,27 +1802,28 @@ def verify(self):
...
                             assert u.prev.label == u.label
                         else:
-                            s = u.right - u.left - 1
-                        if self.model != 'dtwf' and self.model != 'wf_ped':
-                            assert s == self.L[u.label].get_frequency(u.index)
+                            s = self.recomb_map.mass_between_left_exclusive(
+                                            u.left, u.right)
                         right = u.right
+                        if self.model != 'dtwf' and self.model != 'wf_ped':
+                            assert abs(s - self.L[l].get_frequency(u.index)) < 1e-6
```
jeromekelleher: Use math.isclose() here.

jeromekelleher: And elsewhere

daniel-goldstein:
Yup, everything seems to check out fine for `ms` and `mshot` comparisons. Here's the breakpoint distribution on the latest commit for `mshot-high-recomb-variance`
recomb map coalescent_breakpoints

jeromekelleher:
Can you run some of the dtwf checks as well please? Once these pass we're good to merge I think.

daniel-goldstein:
> Can you run some of the dtwf checks as well please? Once these pass we're good to merge I think.

They all seem pretty consistent. Here's the `num_trees` and `tmrca_mean` for

`dtwf_vs_coalescent_long_region`:
num_trees
tmrca_mean


`dtwf_vs_coalescent_3_pops_asymm_mig`:
num_trees
tmrca_mean

jeromekelleher:
OK, looking good. I'm going to leave this for a day or so in case someone wants to review, but I think we merge then.

jeromekelleher:
OK, we're good to go @daniel-goldstein. I made some tiny cosmetic tweaks and rebased. Will merge once CI passes. This is a huge jump forward, thanks!

gtsambos:
Yay what an exciting feature! Thanks for doing it @daniel-goldstein!

I'm just wondering, does this feature mean there's now a more elegant solution to the issue of simulating multiple chromosomes (documented [here](https://msprime.readthedocs.io/en/stable/tutorial.html#multiple-chromosomes) )?

Ie. does this mean that if I specify my recombination map with
```
positions = [chr1, chr2-1, chr2, chr3-1]
rates = [rho, 1, rho, 0]
```
that the resulting tree sequence will be of two independent chromosomes, with a tree breakpoint at coordinate `chr2`?

Asking because

 - I'm currently working on some new docs for msprime and can add this to the list of things to change
 - I'm currently designing some simulations that would work much better if I could do this...

jeromekelleher:
I don't think this will make any fundamental difference @gtsambos, as it's essentially the same model we're simulating as we've documented there. But, we will need to update the docs. I'll open an issue so we don't forget about it.

DomNelson:
I think it will be slightly cleaner though @gtsambos @jeromekelleher, since previously the hotspot added between chromosomes would end up with many distinct breakpoints/trees all jammed into it, since during the simulations it would be treated the same as any other region with the same recombination length. That meant you needed to be a bit careful iterating through all trees, since the hotspots would have a huge number of trees < 1 bp apart, and it also wasted a decent amount of simulation time generating those extra trees.

In physical coords all that mess is avoided, no? As I understand it you'd just pass directly from one chromosome to another, which does seem like an improvement. Also the rescaling to avoid bad edges (as described in the docs) would be unnecessary.

jeromekelleher:
Thinking about it again, the "phantom trees" thing shouldn't have existed anyway if we were doing the num_loci == sequence_length thing properly in the old code, so I don't know why that happened. I guess it should be fixed now anyway --- anyone want to try it out?

DomNelson:
Ran a quick check and things seem to be working for the Hudson but not DTWF - put things in a comment on #876

gtsambos:
Sorry for the late reply all, been a bit busy moving house.

> I think it will be slightly cleaner though @gtsambos @jeromekelleher, since previously the hotspot added between chromosomes would end up with many distinct breakpoints/trees all jammed into it,

Yes this is what I was thinking (and what I hoped would be fixed with this new feature!) I'm trying to simulate a chromosome with some intervals masked out -- it doesn't make sense to simulate the chunks independently since there's long-range LD, so I'm currently trying to approach shown in the current docs. You can set the space between the segments to be small enough that there are no (or negligibly many) mutations on them, but you still end up simulating a gazillion squished up trees in the space. As is , you still effectively have to simulate the tree topologies for the whole chromosome, which is v wasteful if you only care about certain small segments of the chromosome. But it looks like this feature does away with this (if I've understood correctly?)

jeromekelleher:
This should work cleanly now with ``discrete=True`` @gtsambos (although I still don't understand why we were getting lots of trees in the non-recombining bit if num_loci == sequence_length in the old code...)

PR #863, opened 2020-01-27
Add verification tests for recombination maps with msHOT

daniel-goldstein:
Pulled in msHOT and modified it to produce summary stats just like with the ms installation. Added an option to `mspms` that allows the specification of recombination "hotspots" identical to the msHOT cli. Added the following tests to the verification suite:
- simple sanity check test that msHOT and ms produce the same result on a simulation with uniform recombination rate
- simulation from the [msHOT documentation](https://uchicago.app.box.com/s/l3e5uf13tikfjm7e1il1eujitlsjdx13/file/68787014053) with recombination hotspots. This test checks out both on master and my other current [PR](https://github.com/tskit-dev/msprime/pull/862) that computes in physical space
- a hotspot with a recombination rate of 0
- multiple recombination hotspots across the sequence, with varying rates of recombination, some zero

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -1,4 +1,4 @@
-all: scrm ms slim
+all: scrm ms msHOT slim
 
 slim:
 	rm -fR SLiM*
@@ -21,6 +21,11 @@ ms:
 	cp msdir/ms_summary_stats ./
 	cp msdir/sample_stats ./
 
+msHOT:
+	make -C msHOTdir
+	cp msHOTdir/msHOT ./
+	cp msHOTdir/msHOT_summary_stats ./
+
 discoal:
 	wget https://github.com/kern-lab/discoal/archive/v0.1.4.tar.gz
 	tar -xf v0.1.4.tar.gz
@@ -28,4 +33,4 @@ discoal:
 	cp discoal-0.1.4/discoal ./
 
 clean:
-	rm -fR scrm* ms ms_summary_stats sample_stats discoal* v0.1.4.tar.gz*
+	rm -fR scrm* ms ms_summary_stats msHOT msHOT_summary_stats sample_stats discoal* v0.1.4.tar.gz*
--- a/msprime/cli.py
+++ b/msprime/cli.py
@@ -53,6 +53,15 @@ mscompat_recombination_help = (
     "of sites between which recombination can occur")
 mscompat_gene_conversion_help = (
     "TODO")
+mshotcompat_hotspot_help = (
+    "Recombination hotspots defined according to the msHOT format. This is "
+    "defined as a sequence: n (start stop scale)+ where n is the number of "
+    "hotspots and each hotspot spans [start, stop) where the recombination "
+    "rate is the background recombination rate times scale. Adjacent hotspots "
+    "may stop and start at the same position but must otherwise be non-overlapping "
+    "and specified in ascending order."
+        )
+
 msprime_citation_text = """
 If you use msprime in your work, please cite the following paper:
 Jerome Kelleher, Alison M Etheridge and Gilean McVean (2016), "Efficient
@@ -111,6 +120,46 @@ def get_single_seed(seeds):
     return int(m.hexdigest(), 16) % (2**32)
 
 
+def hotspots_to_recomb_map(hotspots, background_rate, seq_length):
+    """
+    Translates the background recombination rate and recombination
+    hotspots according to the msHOT cli spec to a msprime
+    recombination map.
+
+    hotspots is of the form: [n] + [start, stop, factor]+
+    where n is the number of hotspots and each hotspot spans [start, stop)
+    with a recombination rate of background_rate * factor. Intervals between
+    hotspots have a recombination rate of background_rate.
+    """
+    assert len(hotspots) > 0
+    n_hotspots = hotspots[0]
+    assert len(hotspots[1:]) == 3 * n_hotspots
+    positions = []
+    rates = []
+    if hotspots[1] != 0:
+        positions.append(0)
+        rates.append(background_rate)
+    for i in range(1, len(hotspots) - 2, 3):
+        [start, stop, factor] = hotspots[i:i+3]
+        # Beginning hotspot
+        positions.append(start)
+        rates.append(factor * background_rate)
+
+        if i == len(hotspots) - 3 or stop != hotspots[i+3]:
+            # Ending hotspot, back to normal recombination rate
+            positions.append(stop)
+            if stop != seq_length:
+                rates.append(background_rate)
+            else:
+                rates.append(0)
+
+    if positions[-1] != seq_length:
+        positions.append(seq_length)
+        rates.append(0)
+
+    return msprime.RecombinationMap(positions, rates, int(positions[-1]))
+
+
 class SimulationRunner(object):
     """
     Class to run msprime simulation and output the results.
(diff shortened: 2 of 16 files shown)


data/Makefile
```diff
@@ -21,11 +21,16 @@ ms:
 	cp msdir/ms_summary_stats ./
 	cp msdir/sample_stats ./
 
+mshot:
+	make -C msHOTdir
+	cp msHOTdir/msHOT ./
+	cp msHOTdir/msHOT_summary_stats ./
```
jeromekelleher: msHot's version of summary_stats is probably identical to ms's, so there's no point in including it. We should probably just remove the code also, no point in keeping the files in our repo.

data/msHOTdir/clms
```diff
@@ -0,0 +1,2 @@
+gcc -O2 -o ms  ms.c  rand1.c streec.c -lm
```
jeromekelleher: Don't think we need this file

data/msHOTdir/col1
```diff
@@ -0,0 +1 @@
+awk ' { print $'$1' }'
```
jeromekelleher: Don't think we need this file

data/msHOTdir/dist3.c
```diff
@@ -0,0 +1,69 @@
+/* *********************** Probability of s segregating sites (no recombination) *********
```
jeromekelleher: Don't think we need this file

data/msHOTdir/microsat.c
```diff
@@ -0,0 +1,114 @@
+/*  microsat.c:   This function takes ms output and converts it is microsatellite
```
jeromekelleher: Don't think we need this file

data/msHOTdir/migmat
```diff
@@ -0,0 +1,5 @@
+-ma
```
jeromekelleher: Don't think we need this file

msprime/cli.py
```diff
@@ -121,15 +121,38 @@ def __init__(
...
         self._sample_size = sample_size
         self._num_loci = num_loci
         self._num_replicates = num_replicates
         self._recombination_rate = scaled_recombination_rate
         self._mutation_rate = scaled_mutation_rate
+        self._hotspots = hotspots
         # For strict ms-compability we want to have m non-recombining loci
-        recomb_map = msprime.RecombinationMap.uniform_map(
-            num_loci, self._recombination_rate, num_loci)
+        if (hotspots is None):
```
jeromekelleher: Brackets redundant here

msprime/cli.py
```diff
@@ -121,15 +121,38 @@ def __init__(
...
+            recomb_map = msprime.RecombinationMap.uniform_map(
+                num_loci, self._recombination_rate, num_loci)
+        else:
+            assert(len(hotspots) > 0)
```
jeromekelleher: Brackets redundant here (assert isn't a function)

data/Makefile
daniel-goldstein: do you mean sample stats? `msHOT_summary_stats` is compiled with the necessary flags to produce stats for verification.

jeromekelleher: yes I do, my bad.

data/Makefile
```diff
@@ -21,11 +21,16 @@ ms:
 	cp msdir/ms_summary_stats ./
 	cp msdir/sample_stats ./
 
+mshot:
```
jeromekelleher: Looks like the target should be msHot to match the filename

data/msHOTdir/params
```diff
@@ -0,0 +1 @@
+  -t 3.0 -s 12
```
daniel-goldstein: still a couple files left to clean up including this one

data/msHOTdir/seedms
```diff
@@ -0,0 +1 @@
+40859 29729 63930
```
jeromekelleher: Don't need this file

data/msHOTdir/stats.c
```diff
@@ -0,0 +1,74 @@
+/*  stats.c : Calculates mean, standard deviation for a list of numbers.
```
jeromekelleher: Or this file

data/msHOTdir/tajd.c
```diff
@@ -0,0 +1,120 @@
+/************************* tajima.c *************************************************************
+ This program calculates Tajima's D when number of sequences, number of segregating sites,
```
jeromekelleher: Or this one

jeromekelleher:
Excellent! One more update and I think we're good to merge. Can we 

- Add an extra  plot for the tests where we plot the density of breakpoints along the chromosome for msprime and the ms-like instance? We can do this for all ms instances. Basically, we want to see if the breakpoints are falling the right places.
- Add a few more hotspot tests where we try out variations with different rates at a number of places along the genome. We want to see peaks of breakpoints of different heights along the genome; this will be a really good check on whether we're getting the rate calculations right.

daniel-goldstein:
Will do. Those will be really nice to have!

daniel-goldstein:
Histograms are looking pretty good. Here's the output for the `msdoc-recomb-ex` (uniform recomb rate) and `mshot-high-recomb-variance` tests respectively:
coalescent_breakpoints

variable recombination rate coalescent_breakpoints

msprime/cli.py
```diff
@@ -161,6 +167,35 @@ def __init__(
         self._mutation_generator = msprime.MutationGenerator(
             self._random_generator, self._mutation_rate)
 
+    def hotspots_to_recomb_map(self, hotspots, recomb_rate, num_loci):
```
jeromekelleher: We're not using ``self`` here at all, so this is more naturally a function I think. The function should have a docstring defining the input and outout. Then we can add a few unit tests in the test_cli module
```python
class TestHotspotsToRecombMap(..)
     def test_one_hotspot(...)

```
Just a few different tests to make sure we're getting all the corner cases in for this function.

jeromekelleher:
Excellent, great work @daniel-goldstein. It would be nice to see a bit more clearly the two distributions. Perhaps seaborn would help here? Maybe a kernel density smoothing would give us two lines we could more easily compare? (But these introduce noise too, so I'm not too sure). See also above comment on adding unit tests for the hotspot_to_recomb_map function.

daniel-goldstein:
Here's what the same tests would look like using Seaborn's kernel density estimate. We can restrict how loosely it estimates the breakpoints equivalently to adjusting the number of bins in the histogram. Not smoothing at all is pretty inscrutable. These plots are using the default heuristics for fitting the curves:

coalescent_breakpoints

recomb map coalescent_breakpoints

msprime/cli.py
```diff
@@ -121,15 +161,21 @@ def __init__(
...
         self._mutation_rate = scaled_mutation_rate
         # For strict ms-compability we want to have m non-recombining loci
-        recomb_map = msprime.RecombinationMap.uniform_map(
-            num_loci, self._recombination_rate, num_loci)
+        if hotspots is None:
+            self._recomb_map = msprime.RecombinationMap.uniform_map(
+                num_loci, self._recombination_rate, num_loci)
+        else:
+            self._recomb_map = hotspots_to_recomb_map(
+                    hotspots, scaled_recombination_rate, num_loci)
```
jeromekelleher: probably better to use ``self._recombination_rate`` so the reader isn't wondering why the two cases are different.

msprime/cli.py
```diff
@@ -628,6 +681,7 @@ def get_mspms_parser(error_handler=None):
         "--gene-conversion", "-c", type=float, nargs=2, default=(0, 1),
         metavar=("gc_recomb_ratio", "track_length"),
         help=mscompat_gene_conversion_help)
+    group.add_argument("--hotspots", "-v", type=float, nargs="+", default=None)
```
jeromekelleher: Needs some help text. Can say something like "Recombination hotspots defined according to msHOT format. This is blah blah``

tests/test_cli.py
```diff
@@ -296,6 +313,49 @@ class CustomExceptionForTesting(Exception):
     """
 
 
+class TestHotspotsToRecombMap(TestCli):
+
+    def verify_map(self, recomb_map, expected_positions, expected_rates):
```
jeromekelleher: Perfect.

jeromekelleher:
Awesome, merging. Thanks @daniel-goldstein, this is great stuff!

PR #871, opened 2020-02-03
Update verification.py

TPPSellinger:
I added simple a simple test to check ARGs distribution when simulating data under xi_beta or xi_dirac coalescent. Code will create qq plots. If ARGs make sense, a "good" fit to the expected distribution should be obtained.

PS: Sequence of length 10^5 bp are simulated for calculation time purposes. When simulating sequences under strong multiple merger, may be the sequence length should be increased to 10^7 in order to have enough data.

Thibaut

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -30,6 +30,8 @@ import argparse
 import msprime.cli as cli
 import msprime
 
+import pylab 
+import scipy.stats as stats
 
 def flatten(l):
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Thanks for this @TPPSellinger. I've refactored this a bit so it follows the coding style of the project and to remove some redundancy. The tests look good,  I think.

PR #872, opened 2020-02-05
adding statistical tests of sweeps (WIP)

andrewkern:
I added an example statistical test between the msprime implementation and my old discoal implementation. I'm not 100% confident that all the scaling conversions are good to go between discoal and msprime however. @petrelharp looked over my should a bit today and we got it pretty close but I'm not sure how to scale the selection coefficient between discoal and msprime. 

from playing with it way too much it looks like the \alpha_discoal * 2 = \alpha_msprime, but I'm scratching my head as to why.

### DIFF
--- a/lib/dev-tools/example.cfg
+++ b/lib/dev-tools/example.cfg
@@ -32,8 +32,8 @@ model = {
     #truncation_point = 1.0;
 
     # The discrete time Wright Fisher
-    /* name = "dtwf"; */
-    /* population_size = 10.0; */
+    # name = "dtwf"; 
+    # population_size = 10.0; 
 
     # The genic selection sweep model
     /* name = "sweep_genic_selection"; */ 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4415,15 +4415,6 @@ msp_run_sweep(msp_t *self)
             total_rate = sweep_pop_tot_rate;
             event_prob *= 1.0 - total_rate;
             curr_step++;
-            /* sanity check print statement */
-            //printf("p: %g p_coal_b: %g p_coal_B: %g p_rec_b: %g p_rec_B: %g total_rate: %g event_prob: %g event_rand:%g\n", allele_frequency[curr_step],p_coal_b, 
-            //        p_coal_B, p_rec_b, p_rec_B, total_rate, event_prob, event_rand);
-           // printf("curr_step: %ld num_steps: %ld\n", curr_step, num_steps);
-            //printf("sweep_pop_sizes: %g %g pop_size: %g\n", sweep_pop_sizes[0], sweep_pop_sizes[1], pop_size);
-            /* JK->Andy: is this what it means when the total_rate goes to 0? */
-            /* AK-> JK: it means that all of the alleles have coalesced onto a
-             * background and we can stop the structured phase of the
-             * simulation */
             sweep_over = total_rate == 0;
         }
         if (sweep_over) {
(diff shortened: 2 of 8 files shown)


verification.py
```diff
@@ -363,6 +363,58 @@ def f():
...
+        seqlen = 1e4
+        mu = 2.5e-4
+        rho = 2.5e-4
+        mod.position = np.floor(seqlen/2)
+        data = collections.defaultdict(list)
+        replicates = msprime.simulate(10,  model=mod, length=seqlen,
+                                      num_labels=2, recombination_rate=rho,
+                                      num_replicates=nreps)
+        for x in replicates:
+            ts = msprime.simulate(from_ts=x, recombination_rate=rho)
```
andrewkern: this is a bit ugly, but I'm forced to do this as our current implementation of the `SweepGenicSelection()` model doesn't finish all the coalescing for us. so what i've done here is start a second round of simulation off from the tree sequence that the sweep simulation ends with. 

@jeromekelleher what do you think here?

verification.py
```diff
@@ -363,6 +363,58 @@ def f():
...
+            data["D"].append(tsm.Tajimas_D())
+            data["ss"].append(tsm.segregating_sites(span_normalise=False))
+        data["pi"] = np.array(data["pi"]).flatten()
+        data["D"] = np.array(data["D"]).flatten()
+        data["ss"] = np.array(data["ss"]).flatten()
+        df = pd.DataFrame.from_dict(data)
+        df = df.fillna(0)
+        df_d = self._run_discoal_mutation_stats(args)
+        df_df = df_d[["pi", "D", "ss"]]
+        print("msp pi mean: ", df['pi'].mean())
```
andrewkern: this is just sanity checking stuff that is helping me debug scaling. I imagine I'll take it out before we merge

lib/dev-tools/example.cfg
```diff
@@ -32,17 +32,17 @@ model = {
     #truncation_point = 10.0;
 
     # The discrete time Wright Fisher
-    /* name = "dtwf"; */
```
andrewkern: not sure what happened to these comments on master... shouldn't they be #'s?

lib/msprime.c
```diff
@@ -6284,8 +6292,8 @@ msp_set_simulation_model_sweep_genic_selection(msp_t *self, double reference_siz
         ret = MSP_ERR_BAD_SWEEP_POSITION;
         goto out;
     }
-    if (start_frequency < 0.0 || start_frequency > 1.0
-            || end_frequency < 0.0 || end_frequency > 1.0) {
+    if (start_frequency <= 0.0 || start_frequency >= 1.0
```
andrewkern: need to change these to avoid weird behavior in trajectory simulation

verification.py
jeromekelleher: How about adding  a ``demographic_events=SimulationModelChange(time=None, model="hudson")`` ? That was the idea anyway, I'm not sure if I got around to implementing this. We should change the default value for ``time`` to None if this works, so we can just say

```
ts = msprime.simulate(
      10,  
      model=selection_model, 
      demographic_events=[SimulationModelChange(model="hudson")])
```
and it'll do what we want.

andrewkern: Yeah it doesn't look like a `time=None` option has been implemented for this situation where you want it to change models at some arbitrary time.

andrewkern: seems to me like setting `time=np.Inf` is doing the correct thing but i'm not 100% certain

jeromekelleher:
@andrewkern, can we update this and get it merged soon please? We can register issues to follow up on, but it'd be nice to get this merged (we want to do some reformatting of the code using Black and clang-format, so any open PRs will be disrupted)

jeromekelleher:
@andrewkern, looks like the actual changes here are small. Mind if I clean this up a bit and merge it? We can pick up the actual testing stuff again afterwards.

andrewkern:
@jeromekelleher sorry been missing these notifications! yes merge away!!

pinging @daniel-goldstein here as the verification params I am using should be appropriate

jeromekelleher:
I've tidied this up @andrewkern and opened up a few issues (#941, #942, #943) to follow up on. I changed few things back to the way they were, most notably the default params for the sweep model. I think it's a Bad Idea to put in arbitrary values here and users are much better off being forced to decide what they should be.

Issue #852, opened 2020-01-14
Investigating if sims can be done in physical coordinates


jeromekelleher:
@daniel-goldstein and I are investigating if msprime can be updated to run simulations directly in physical space rather than simulating in genetic space as we currently do. The advantage of working directly in physical coordinates is that it would get rid of a lot of annoying corner cases, especially when it comes to supporting gene conversion. The disadvantage is that it might potentially substantially slow down simulations with variable recombination rates, as we'll need to iterate over subsets of the genetic map each time we insert a segment into the Fenwick tree.

The plan is to make the minimal changes we can to try out simulating directly in physical space for the basic Hudson model, and then do a performance analysis. If it turns out that we lose a lot of performance (more than say, 50%), we go back to the drawing board/abandon the idea.

If not, then we hope to push the changes needed through the whole codebase. There may be some small semantic changes incurred, but if there are we can tag the next release 1.0 and clearly document the semantic changes incurred in the major version bump.

Cleaning up the current mess that we have due to simulating in genetic space would be a great thing to do, and an excellent start for development on the 1.0 series, I think.

cc @fbaumdicker, @andrewkern, @petrelharp, @DomNelson, @ivan-krukov

daniel-goldstein:
I've started experimenting with this [here](https://github.com/daniel-goldstein/msprime/tree/recomb-phys-coord) by changing the semantics of segments to represent discrete physical space. The underlying Fenwick Tree accumulates continuous recombination mass from which we select recombination breakpoints **assuming a uniform recombination rate**. These breakpoints are then floored since segments have integer ends, but can easily be changed if segments move to continuous physical space.

This doesn't really add any more work so we wouldn't expect the performance to change but as a quick sanity check here's a rough benchmark of recombination and the overall runtime.

Simulation was ran with a few trials of the following:
```msprime.simulate(10, Ne=10000, random_seed=1, recombination_map=msprime.RecombinationMap.uniform_map(length=10 ** 8,         num_loci=10 ** 8, rate=1e-8)```

|                                         | msp_recombination_event | msp_run |
-------------------------|---------------------------|-----------
|  master                            |          ~7.7s (20.2%)             |     ~38s   |
|  recomb-phys-coord     |          ~8.3s (21.4%)             |     ~38s   |

Note that with this change, the **number of loci must be equal to the sequence length**. However, under the continuous model loci lose their utility so just working around this for now by setting the number of loci manually.

[...]

PR #885, opened 2020-02-06
Store left and right endpoint masses in segments

### LINKED ISSUE #873: Performance tuning for physical coordinate sims
#852 changed the internals of msprime to simulate in physical space. An analysis of the initial implementation in #852 showed there was a bit of a performance penalty over the existing approach in the case of a large recombination map.

Investigate if this can be improved, without too much code change.

daniel-goldstein:
- Add the masses of left and right endpoints in the segment struct. This avoids doing lookups for segment masses when merging segment chains in methods like `merge_two_ancestors`. This change takes the new methods introduced in #862 like `recomb_map_mass_between` off the critical path. The special case of updating the mass of a segment at the start of a segment chain in discrete coordinates still requires a lookup, but the impact is negligible.

- Some small refactoring of update methods for masses on the Fenwick tree.

Resolves #873.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -117,6 +117,8 @@ class Segment(object):
     def __init__(self, index):
         self.left = None
         self.right = None
+        self.left_mass = None
+        self.right_mass = None
         self.node = None
         self.prev = None
         self.next = None
@@ -635,7 +637,10 @@ class Simulator(object):
                 sample_size = sample_configuration[pop_index]
                 for k in range(sample_size):
                     j = len(self.tables.nodes)
-                    x = self.alloc_segment(0, self.m, j, pop_index)
+                    x = self.alloc_segment(
+                        0, self.m, 0,
+                        self.recomb_map.position_to_mass(self.m),
+                        j, pop_index)
                     self.set_single_segment_mass(x)
                     self.P[pop_index].add(x)
                     self.tables.nodes.add_row(
@@ -707,7 +712,11 @@ class Simulator(object):
                 for root in tree.roots:
                     population = ts.node(root).population
                     if root_segments_head[root] is None:
-                        seg = self.alloc_segment(left, right, root, population)
+                        seg = self.alloc_segment(
+                                left, right,
+                                self.recomb_map.position_to_mass(left),
+                                self.recomb_map.position_to_mass(right),
+                                root, population)
                         root_segments_head[root] = seg
                         root_segments_tail[root] = seg
                     else:
@@ -715,7 +724,11 @@ class Simulator(object):
                         if tail.right == left:
                             tail.right = right
                         else:
-                            seg = self.alloc_segment(left, right, root, population, tail)
+                            seg = self.alloc_segment(
+                                    left, right,
+                                    self.recomb_map.position_to_mass(left),
+                                    self.recomb_map.position_to_mass(right),
+                                    root, population, tail)
                             tail.next = seg
                             root_segments_tail[root] = seg
         self.S[self.m] = -1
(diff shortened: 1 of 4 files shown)


daniel-goldstein:
Here's an update of the plot in #852. It shows the performance of msprime prior to the change to physical coordinates, the current release, current master, and the branch on this PR.

Screen Shot 2020-02-05 at 7 03 59 PM

jeromekelleher:
Thanks @daniel-goldstein, this looks good. I'll need to work through this so I understand what's happening. Can I get back to you on it next week?

It would be useful to know if there's a significant memory penalty for adding the extra 16 bytes to the segment struct (there are zillions of them). Can you compare the peak RAM usage for 0.7.4, master and this for the same sims as you've done above please?

algorithms.py
```diff
@@ -1553,10 +1610,14 @@ def merge_ancestors(self, H, pop_id, label):
             if len(X) == 1:
                 x = X[0]
                 if len(H) > 0 and H[0][0] < x.right:
+                    next_l_mass = self.recomb_map.position_to_mass(H[0][0])
```
jeromekelleher: OK, so this is the only lookup we need to do. But, aren't all the values in H a result of being either left or right coordinates? Whenever we push a coordinate into H there's also a segment along with it, so we can get the mass from there, avoiding this lookup.

algorithms.py
```diff
@@ -1186,12 +1204,23 @@ def migration_event(self, j, k):
...
-    def increment_segment_mass(self, seg, left, right):
-        mass = self.recomb_map.mass_between(left, right)
+    def set_segment_left_endpoint(self, seg, pos):
+        seg.left = pos
+        seg.left_mass = self.recomb_map.position_to_mass(pos)
+
+    def increment_segment_mass_between(self, seg, l_mass, r_mass):
+        self.L[seg.label].increment(seg.index, r_mass - l_mass)
+
+    def increment_segment_mass(self, seg, seg2):
```
jeromekelleher: Maybe ``add_segment_mass(seg1, seg2)`` with a comment saying "adds the mass subtended between the endpoints of seg2 to seg1" or something.

algorithms.py
```diff
@@ -1186,12 +1204,23 @@ def migration_event(self, j, k):
...
+        mass = seg2.right_mass - seg2.left_mass
+        self.L[seg.label].increment(seg.index, mass)
+
+    def decrement_segment_mass(self, seg, seg2):
```
jeromekelleher: Similarly "subtract_segment_mass"

jeromekelleher:
LGTM, thanks @daniel-goldstein. Can we get a plot of the final performance/memory figures here please?

Have you run the verification checks?

daniel-goldstein:
> LGTM, thanks @daniel-goldstein. Can we get a plot of the final performance/memory figures here please?

Here are the final figures. Performance even improved after the last change.
Screen Shot 2020-02-11 at 10 52 02 AM

RAM usage saw no significant difference between the two branches. There's a large spike in memory usage right at the end of simulations (dealing with the output?). I've included both, the first figure showing the max RAM consumption during the majority of the simulation and the second showing the final spike.

Screen Shot 2020-02-11 at 10 51 42 AM

Screen Shot 2020-02-11 at 11 00 55 AM

> Have you run the verification checks?

Verification tests still look good.

jeromekelleher:
Superb, thanks @daniel-goldstein! Great news there's no memory penalty, and the perf is excellent.

@fbaumdicker, can you take a look at the GC code changes here please?

fbaumdicker:
I had a careful look at the GC part. Looks good to me.
I really like the new function names. Much more intuitive.
But this needs to be combined with #896 for GC to give the expected results. 
I suggest merging this now and #896 later. After that, I can verify whether the GC code is acting as expected.

jeromekelleher:
OK, thanks @fbaumdicker and @daniel-goldstein, merging.

Issue #884, opened 2020-02-06
[sweeps] Document model change event to Hudson and test


jeromekelleher:
We need a simple way to transition to the Hudson model after a sweep has completed. This should be done with a ModelChangeEvent, but we need a way to leave the time unspecified. Using np.inf should work here.

Test this approach to make sure it works as expected.

andrewkern:
was trying to think of a reasonable test for this. it should be that under the SweepGenicSelection() model many trees are not fully coalesced whereas if we let them then transition models to StandardCoalescent() everything. maybe we could use `Tree.num_roots` here?

jeromekelleher:
Yep, that sounds exactly right. We need to choose the parameters so that the sweep *some* coalescing, but not too much. We should have 1 < tree.num_roots < ts.num_samples. Then rerun the same sim with the ModelChange event and tree.num_roots should be == 1.

andrewkern:
i can work on this today. should i put such a test in verification.py or somewhere else?

jeromekelleher:
In [this class](https://github.com/tskit-dev/msprime/blob/master/tests/test_models.py#L552), ideally. Should be quite small to keep things quick.

andrewkern:
can you assign me to this issue so that it shows up on my queue?

jeromekelleher:
Done

PR #886, opened 2020-02-06
test of changing models from sweep to hudson using time set to INF

andrewkern:
added a simple test of model change events. this PR will close https://github.com/tskit-dev/msprime/issues/884

also added a bit of documentation about using `np.Inf` as a time for model change events.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -141,14 +141,14 @@ that are currently supported. Note that all times are measured in
 generations, all sizes are absolute (i.e., *not* relative to :math:`N_e`),
 and all rates are per-generation.
 
+Model change events are also considered to be demographic events;
+see the :ref:`sec_api_simulation_models_change_events` section for details.
+
 .. autoclass:: msprime.PopulationParametersChange
 .. autoclass:: msprime.MigrationRateChange
 .. autoclass:: msprime.MassMigration
-.. autoclass:: msprime.SimulationModelChange
-
-.. _sec_api_demographic_events_census:
-
 .. autoclass:: msprime.CensusEvent
+
 
 ++++++++++++++++++++++++++++
 Debugging demographic models
@@ -198,6 +198,11 @@ and
     msprime.simulate(10, model=msprime.DiscreteTimeWrightFisher(1000))
 
 define the same simulation.
+
+.. note:: When ``Ne`` and an instance of :class:`.SimulationModel` with
+    ``reference_size`` set are both provided as arguments to :func:`.simulate`,
+    the ``Ne`` parameter is ignored.
+
 
 .. todo:: Add a discussion of population sizes here, describing
     what Ne/model.reference_size really means, and how it interacts
@@ -272,6 +277,27 @@ To use this option, set the flag ``model="dtwf"`` as in the following example::
 All other parameters can be set as usual.
 
 .. autoclass:: msprime.DiscreteTimeWrightFisher
+
+
+.. _sec_api_simulation_models_selective_sweeps:
+
+++++++++++++++++
+Selective sweeps
+++++++++++++++++
+
+.. todo:: Document the selective sweep models.
+
+
+.. _sec_api_simulation_models_change_events:
+
+++++++++++++++++++++++++
+Simulation model changes
+++++++++++++++++++++++++
+
+.. todo:: Describe the subtleties of how simulation model changes work
+    along with how times are computed.
+
+.. autoclass:: msprime.SimulationModelChange
 
 
 .. _sec_api_simulate_from:
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4378,7 +4378,12 @@ msp_run(msp_t *self, double max_time, unsigned long max_events)
         ret = MSP_ERR_UNSUPPORTED_OPERATION;
         goto out;
     }
-    if (self->model.type == MSP_MODEL_DTWF) {
+
+    if (msp_is_completed(self)) {
+        /* If the simulation is completed, run() is a no-op for
+         * all models. */
+        ret = 0;
+    } else if (self->model.type == MSP_MODEL_DTWF) {
         ret = msp_run_dtwf(self, scaled_time, max_events);
     } else if (self->model.type == MSP_MODEL_WF_PED) {
         if (self->pedigree == NULL ||
@@ -4399,11 +4404,12 @@ msp_run(msp_t *self, double max_time, unsigned long max_events)
             goto out;
         }
     } else if (self->model.type == MSP_MODEL_SWEEP) {
-        /* FIXME making sweep atomic for now as it's non-renentrant */
+        /* FIXME making sweep atomic for now as it's non-rentrant */
         ret = msp_run_sweep(self);
     } else {
         ret = msp_run_coalescent(self, scaled_time, max_events);
     }
+
     if (ret < 0) {
         goto out;
     }
@@ -6461,6 +6467,8 @@ msp_set_simulation_model_sweep_genic_selection(msp_t *self, double reference_siz
     /* dt value is expressed in generations for consistency; translate to
      * model time. */
     trajectory->dt = self->model.generations_to_model_time(&self->model, dt);
+
+    ret = msp_rescale_model_times(self);
 out:
     return ret;
 }
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -243,9 +243,7 @@ def simulator_factory(
     if start_time is not None and start_time < 0:
         raise ValueError("start_time cannot be negative")
 
-    if num_labels is None:
-        num_labels = 1
-    if num_labels < 1:
+    if num_labels is not None and num_labels < 1:
         raise ValueError("Must have at least one structured coalescent label")
 
     if from_ts is not None:
@@ -560,7 +558,7 @@ class Simulator(object):
         self.model_change_events = []
         self.store_migrations = False
         self.store_full_arg = False
-        self.num_labels = 1
+        self.num_labels = None
         # We always need at least n segments, so no point in making
         # allocation any smaller than this.
         num_samples = (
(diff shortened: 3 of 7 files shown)


tests/test_models.py
```diff
@@ -609,7 +609,8 @@ def test_sweep_start_time_complete(self):
             num_labels=2, random_seed=2)
         self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees()))
 
-    @unittest.skip("Can't get sweep to be incomplete")
```
jeromekelleher: Just delete the skip if it's working.

msprime/simulations.py
```diff
@@ -1517,6 +1517,9 @@ class SimulationModelChange(DemographicEvent):
     :param float time: The time at which the simulation model changes
         to the new model, in generations. After this time, all internal
         tree nodes, edges and migrations are the result of the new model.
+        If time is set to infinity using ``np.Inf`` this has the effect
```
jeromekelleher: It's not particularly intuitive though. It would be better if we made ``time=None`` the default, and then said "If time is unspecified or None, the model change occurs immediately after the previous model completes". We need to make this work for chaining multiple sweeps together too and I don't think the np.inf thing is going to work there, we'll have to detect None explicitly in the model change logic.

andrewkern:
if you want to go for it-- do it. I'd rather spend my time right now dialing in the sweep implementation and making sure it lines up with discoal. I can currently use the `np.Inf` trick to do that.

jeromekelleher:
OK, leave this one with me then @andrewkern.

jeromekelleher:
I've made some updates here @andrewkern, and I think the interface is looking pretty solid. I've gotten rid of the need to type in ``num_labels=2``, which is a win.

What do you think, looking good?

tests/test_highlevel.py
```diff
@@ -901,3 +857,61 @@ def test_discrete(self):
         for truthy in [True, False, {}, None, "ser"]:
             rm = msprime.RecombinationMap.uniform_map(1, 0, discrete=truthy)
             self.assertEqual(rm.discrete, bool(truthy))
+
+    def test_zero_recombination_map(self):
```
andrewkern: did you intend to include these zero recombination tests on this PR? totally cool if so

tests/test_models.py
```diff
@@ -583,23 +597,21 @@ def test_incorrect_num_labels(self):
...
         self.assertEqual(ts.num_trees, 1)
         for tree in ts.trees():
             self.assertEqual(tree.num_roots, 1)
 
-    @unittest.skip("Not clear what's happening now")
     def test_sweep_coalescence_recomb(self):
         model = msprime.SweepGenicSelection(
-            position=0.5, start_frequency=0.6, end_frequency=0.7, alpha=0.01,
-            dt=0.1)
+            position=0.5, start_frequency=0.01, end_frequency=0.999, alpha=1.0,
```
andrewkern: set alpha to something like 1000 for faster sweeps

tests/test_models.py
```diff
@@ -608,39 +620,77 @@ def test_sweep_coalescence_same_seed(self):
...
+    def test_many_sweeps(self):
+        sweep_models = [
+            msprime.SweepGenicSelection(
+                reference_size=0.25, position=j, start_frequency=0.69,
+                end_frequency=0.7, alpha=1e-5, dt=0.1)
+            for j in range(10)]
+        ts = msprime.simulate(
+            10, Ne=0.25, length=10,
+            recombination_rate=0.2,
+            demographic_events=[
```
andrewkern: nice. a cool thing to do next here would be to set a rate of sweeps, or any arbitrary model change, going back in time. i.e. we could imagine that sweeps occur in time according to a poisson point process

tests/test_highlevel.py
jeromekelleher: Yeah, spotted they were in the wrong place and moved them to the right place. Documented in the commit message.

tests/test_models.py
jeromekelleher: Cool, I'll have a go at that. Lots of very fast sweeps, occurring at random times...

andrewkern: it will be just like the original recurrent hitchhiking model!

lib/msprime.c
```diff
@@ -6461,6 +6467,8 @@ msp_set_simulation_model_sweep_genic_selection(msp_t *self, double reference_siz
     /* dt value is expressed in generations for consistency; translate to
      * model time. */
     trajectory->dt = self->model.generations_to_model_time(&self->model, dt);
+
+    ret = msp_rescale_model_times(self);
```
jeromekelleher: Note: this was a bug here. Any times we were getting before will be wrong, unless we happened to always be using Ne=1/4 or something.

tests/test_models.py
```diff
@@ -608,39 +620,124 @@ def test_sweep_coalescence_same_seed(self):
...
+        self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees()))
+
+    def test_many_sweeps_regular_times_model_change(self):
+        demographic_events = []
+        for j in range(10):
+            sweep_model = msprime.SweepGenicSelection(
+                position=j, start_frequency=0.69, end_frequency=0.7,
+                alpha=1000, dt=0.0125)
+            # Start the sweep after 0.01 generations of Hudson
+            demographic_events.append(msprime.SimulationModelChange(
```
jeromekelleher: So, this is the proposal for doing lots of sweeps. The ``event.time`` field is a callable, which returns the time at which the sweep should start, given the current simulation time at the start of the previous model as a parameter. (This is slightly confusing, but general, I think). This way we don't have to make decisions before the simulation is run and can have randomness in there too.

tests/test_models.py
```diff
@@ -608,39 +620,124 @@ def test_sweep_coalescence_same_seed(self):
...
+                time=lambda t: t + 0.01, model=sweep_model))
+            # Revert back to Hudson until the next sweep
+            demographic_events.append(msprime.SimulationModelChange())
```
jeromekelleher: Reverting to Hudson with Ne is easy: ``msprime.SimulationModelChange()`` with no params.

lib/msprime.c
andrewkern: wait - i've been trying to track down time scaling issues in this implementation for about two weeks now... what will the msp_rescale_model_times to in this case? go from 4N gens to gens? also would this have affected mutation numbers (due to time scaling)? 

been meaning to bother you for help in tracking down these scaling issues but grant reviewing responsibilities have me by the throat at the moment

jeromekelleher:
> looks really good. it looks like by default if there are multiple time=None events they are dealt with in the exact order in which they are passed in the array, is that right?

Yes, model changes will always occur in the order they were presented in the array. Anything out of order raises a ValueError

tests/test_models.py
andrewkern: > given the current simulation time at the start of the previous model as a parameter. (This is slightly confusing, but general, I think). 

this is a bit confusing. why not just hand it the current time of the simulation at exit from the previous model?

andrewkern:
yes i agree lets merge this. with this bugfix I need to go back to the verification side of things.

tests/test_models.py
jeromekelleher: > this is a bit confusing. why not just hand it the current time of the simulation at exit from the previous model?

Because we're generating the time that the previous model finishes. The way to think of it is this: we're about to run a Hudson phase at time t, so what time should this Hudson phase finish so we can start a sweep?

lib/msprime.c
jeromekelleher: This call scales all times and rates that are in the simulator from generations into model time (4N generations, here, I guess).

jeromekelleher:
OK, will wait for some feedback from  @petrelharp on the interface and merge if he doesn't scream in horror.

grahamgower:
This is rather offtopic, but I thought I'd bring it up while you're thinking about interfaces for selection...

Undoubtedly it will be nicer to have selection api in stdpopsim that can do both msprime and slim. I suspect these will end up being engine specific, but its worth considering the possbility of a uniform interface. Here's some (incomplete) code I threw together for the stdpopsim slim engine to do adaptive introgression in test_selection.py. Note that the time and population for mutation introduction is decoupled from the time and population in which selection occurs. https://github.com/grahamgower/stdpopsim/commit/b12b9aacbd19719e8608cf7735cedd64812f386b

docs/api.rst
```diff
@@ -199,6 +199,11 @@ and
 
 define the same simulation.
 
+.. note:: When ``Ne`` and an instance of :class:`.SimulationModel` with
+    ``reference_size`` set are both provided as arguments to :func:`.simulate`,
+    the ``Ne`` parameter is ignored.
```
petrelharp: shouldn't this throw a warning at least?

msprime/simulations.py
```diff
@@ -1530,14 +1563,21 @@ class SimulationModelChange(DemographicEvent):
     :param float time: The time at which the simulation model changes
         to the new model, in generations. After this time, all internal
         tree nodes, edges and migrations are the result of the new model.
+        If time is set to None (the default), the model change will occur
+        immediately after the previous model has completed. If time is a
+        callable, the time at which the simulation model changes is the result
+        of calling this function with the current simulation time
+        as a parameter.
```
petrelharp: what is the "current" time? the time the previous model starts?

msprime/simulations.py
```diff
@@ -1530,14 +1563,21 @@ class SimulationModelChange(DemographicEvent):
...
         a simulation model class (e.g, ``msprime.DiscreteTimeWrightFisher(100)``.
         Please see the :ref:`sec_api_simulation_models` section for more details
         on specifying simulations models. If the argument is a string, the
         reference population size is set from the top level ``Ne`` parameter
-        to :func:`.simulate`. If this is None (the default) we revert to
-        the standard coalescent with the reference population size set by ``Ne``.
+        to :func:`.simulate`. If this is None (the default) the model is
+        changed to the standard coalescent with
+        provided to :func:`msprime.simulate` or the standard coalescent
+        with a reference_size of Ne (if model was not specified).
```
petrelharp: this got garbled - delete between the two `with`s

tests/test_models.py
petrelharp: I like it, but it needs to be a bit better explained in the docs

petrelharp:
I like it - just the docs need a bit of improvement.

docs/api.rst
jeromekelleher: Can't I'm afraid, we have no way of detecting it because the default for Ne is 1 in the top-level function. Changing this to None could have all sorts of subtle side effects, which I don't want to get into, tbh.

tests/test_models.py
jeromekelleher: OK, will open an issue to track this doc task.

jeromekelleher:
I merged this much so @andrewkern can get going with other fixes and statistical checks, and made some issues to track the lack of docs.

jeromekelleher:
> Undoubtedly it will be nicer to have selection api in stdpopsim that can do both msprime and slim. I suspect these will end up being engine specific, but its worth considering the possbility of a uniform interface. Here's some (incomplete) code I threw together for the stdpopsim slim engine to do adaptive introgression in test_selection.py. Note that the time and population for mutation introduction is decoupled from the time and population in which selection occurs. grahamgower/stdpopsim@b12b9aa

Totally agree we would like some sort of high-level interface in stdpopsim that would allow us to describe sweeps in msprime and in SLiM @grahamgower, always good to keep in mind. Please do jump in if you see issues with what we're doing here.

PR #892, opened 2020-02-12
Fix recombination site sampling in DTWF

### LINKED ISSUE #876: [stats tests] Add checks for variable recombination rate in DTWF
Add some plots comparing the Hudson and DTWF models in terms of the breakpoint distribution, under a reasonably complex hotspot model.

daniel-goldstein:
## Bug Fix
Recombination sites for DTWF were generated by sampling from a Poisson distribution in physical coordinates where lambda was the average recombination rate moving forward on the sequence. Doing this in physical space is very problematic when using non-uniform recombination rates. All spatial knowledge is lost and it allows recombination breakpoints to fall in areas of no recombination.

Instead, we now sample a point in recombination space at which the following breakpoint will occur. In this space, the lambda for the Poisson is just 1. We can then translate this to the corresponding physical coordinate after our starting point to find the next breakpoint.

## Testing
- Added plots of breakpoint distributions for all DTWF vs coalescent comparisons.
- Added tests for DTWF using non-uniform recombination maps, in discrete and continuous coordinates
- Added test from #876 to confirm that a single forced breakpoint in discrete space creates two and only two trees, ensuring those trees are consistent between Hudson and DTWF.

Resolves #876

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -529,9 +529,6 @@ class RecombinationMap(object):
         left_bound = left + 1 if self.discrete else left
         return self.mass_between(left_bound, right)
 
-    def rate_between(self, left, right):
-        return self.mass_between(left, right) / (right - left)
-
     def position_to_mass(self, pos):
         if pos == self.positions[0]:
             return 0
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
Examples of recombination distributions for uniform low recombination (an existing test), variable recombination rates (continuous and discrete) and the single forced recombination showing the tree intervals.

breakpoints

breakpoints

breakpoints

breakpoints

jeromekelleher:
Awesome, thanks @daniel-goldstein. Can we get a quick unit test also of the "multiple chromosome" simulation example that Dom did in ``test_models.py``. This just runs a small simulation with the zero recombination bit in the middle of the map, and checks that we don't get any breakpoints within it.

@DomNelson, can we get your eyes on this please?

PR #893, opened 2020-02-13
Species-tree parsing

mmatschiner:
Parsing of species trees in Newick or Nexus format. If in the latter and written by StarBEAST, population-size estimates per branch are used in the definition of population configurations.

### DIFF
--- a/lib/tables.c
+++ b/lib/tables.c
@@ -4724,8 +4724,8 @@ table_collection_compute_mutation_parents(table_collection_t *self, int flags)
     tk = 0;
     site = 0;
     mutation = 0;
-    while (tj < M) {
-        left = edges.left[I[tj]];
+    left = 0;
+    while (tj < M || left < self->sequence_length) {
         while (tk < M && edges.right[O[tk]] == left) {
             parent[edges.child[O[tk]]] = MSP_NULL_NODE;
             tk++;
@@ -4735,9 +4735,13 @@ table_collection_compute_mutation_parents(table_collection_t *self, int flags)
             tj++;
         }
         right = self->sequence_length;
-        if (tk < M) {
-            right = edges.right[O[tk]];
+        if (tj < M) {
+            right = MSP_MIN(right, edges.left[I[tj]]);
         }
+        if (tk < M) {
+            right = MSP_MIN(right, edges.right[O[tk]]);
+        }
+
         /* Tree is now ready. We look at each site on this tree in turn */
         while (site < (site_id_t) sites.num_rows && sites.position[site] < right) {
             /* Create a mapping from mutations to nodes. If we see more than one
@@ -4782,6 +4786,8 @@ table_collection_compute_mutation_parents(table_collection_t *self, int flags)
             }
             site++;
         }
+        /* Move on to the next tree */
+        left = right;
     }
 
 out:
--- a/lib/tests.c
+++ b/lib/tests.c
@@ -2671,6 +2671,54 @@ test_simplest_holey_tree_sequence(void)
 }
 
 static void
+test_simplest_holey_tree_sequence_mutation_parents(void)
+{
+    const char *nodes_txt =
+        "1  0   0\n"
+        "1  0   0\n"
+        "0  1   0";
+    const char *edges_txt =
+        "0  1   2   0\n"
+        "2  3   2   0\n"
+        "0  1   2   1\n"
+        "2  3   2   1\n";
+    const char *sites_txt =
+        "0.5  0\n"
+        "1.5  0\n"
+        "2.5  0\n";
+    const char *mutations_txt =
+        "0    0     1\n"
+        "0    0     1\n"
+        "1    1     1\n"
+        "1    1     1\n"
+        "2    2     1\n"
+        "2    2     1\n";
+    tree_sequence_t ts;
+    table_collection_t tables;
+    int ret;
+
+    tree_sequence_from_text(&ts, 0, nodes_txt, edges_txt, NULL, sites_txt,
+            mutations_txt, NULL);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_sites(&ts), 3);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_mutations(&ts), 6);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_trees(&ts), 3);
+    ret = table_collection_alloc(&tables, MSP_ALLOC_TABLES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tree_sequence_dump_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = table_collection_compute_mutation_parents(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tables.mutations.parent[0], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[1], 0);
+    CU_ASSERT_EQUAL(tables.mutations.parent[2], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[3], 2);
+    CU_ASSERT_EQUAL(tables.mutations.parent[4], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[5], 4);
+    table_collection_free(&tables);
+    tree_sequence_free(&ts);
+}
+
+static void
 test_simplest_initial_gap_tree_sequence(void)
 {
     const char *nodes =
@@ -2719,6 +2767,51 @@ test_simplest_initial_gap_tree_sequence(void)
         CU_ASSERT_STRING_EQUAL(haplotype, haplotypes[j]);
     }
     hapgen_free(&hapgen);
+    tree_sequence_free(&ts);
+}
+
+static void
+test_simplest_initial_gap_tree_sequence_mutation_parents(void)
+{
+    const char *nodes_txt =
+        "1  0   0\n"
+        "1  0   0\n"
+        "0  1   0";
+    const char *edges_txt =
+        "2  3   2   0,1\n";
+    const char *sites_txt =
+        "0.5  0\n"
+        "1.5  0\n"
+        "2.5  0\n";
+    const char *mutations_txt =
+        "0    0     1\n"
+        "0    0     1\n"
+        "1    1     1\n"
+        "1    1     1\n"
+        "2    2     1\n"
+        "2    2     1\n";
+    tree_sequence_t ts;
+    table_collection_t tables;
+    int ret;
+
+    tree_sequence_from_text(&ts, 0, nodes_txt, edges_txt, NULL, sites_txt,
+            mutations_txt, NULL);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_sites(&ts), 3);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_mutations(&ts), 6);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_trees(&ts), 2);
+    ret = table_collection_alloc(&tables, MSP_ALLOC_TABLES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tree_sequence_dump_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = table_collection_compute_mutation_parents(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tables.mutations.parent[0], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[1], 0);
+    CU_ASSERT_EQUAL(tables.mutations.parent[2], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[3], 2);
+    CU_ASSERT_EQUAL(tables.mutations.parent[4], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[5], 4);
+    table_collection_free(&tables);
     tree_sequence_free(&ts);
 }
 
@@ -2774,6 +2867,50 @@ test_simplest_final_gap_tree_sequence(void)
     tree_sequence_free(&ts);
 }
 
+static void
+test_simplest_final_gap_tree_sequence_mutation_parents(void)
+{
+    const char *nodes_txt =
+        "1  0   0\n"
+        "1  0   0\n"
+        "0  1   0";
+    const char *edges_txt =
+        "0  2   2   0,1\n";
+    const char *sites_txt =
+        "0.5  0\n"
+        "1.5  0\n"
+        "2.5  0\n";
+    const char *mutations_txt =
+        "0    0     1\n"
+        "0    0     1\n"
+        "1    1     1\n"
+        "1    1     1\n"
+        "2    0     1\n"
+        "2    0     1\n";
+    tree_sequence_t ts;
+    table_collection_t tables;
+    int ret;
+
+    tree_sequence_from_text(&ts, 3, nodes_txt, edges_txt, NULL, sites_txt,
+            mutations_txt, NULL);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_sites(&ts), 3);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_mutations(&ts), 6);
+    CU_ASSERT_EQUAL(tree_sequence_get_num_trees(&ts), 2);
+    ret = table_collection_alloc(&tables, MSP_ALLOC_TABLES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tree_sequence_dump_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = table_collection_compute_mutation_parents(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tables.mutations.parent[0], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[1], 0);
+    CU_ASSERT_EQUAL(tables.mutations.parent[2], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[3], 2);
+    CU_ASSERT_EQUAL(tables.mutations.parent[4], -1);
+    CU_ASSERT_EQUAL(tables.mutations.parent[5], 4);
+    table_collection_free(&tables);
+    tree_sequence_free(&ts);
+}
 
 static void
 test_simplest_bad_records(void)
@@ -7070,8 +7207,14 @@ main(int argc, char **argv)
         {"test_simplest_back_mutations", test_simplest_back_mutations},
         {"test_simplest_general_samples", test_simplest_general_samples},
         {"test_simplest_holey_tree_sequence", test_simplest_holey_tree_sequence},
+        {"test_simplest_holey_tree_sequence_mutation_parents",
+            test_simplest_holey_tree_sequence_mutation_parents},
         {"test_simplest_initial_gap_tree_sequence", test_simplest_initial_gap_tree_sequence},
+        {"test_simplest_initial_gap_tree_sequence_mutation_parents",
+            test_simplest_initial_gap_tree_sequence_mutation_parents},
         {"test_simplest_final_gap_tree_sequence", test_simplest_final_gap_tree_sequence},
+        {"test_simplest_final_gap_tree_sequence_mutation_parents",
+            test_simplest_final_gap_tree_sequence_mutation_parents},
         {"test_simplest_bad_records", test_simplest_bad_records},
         {"test_simplest_overlapping_parents", test_simplest_overlapping_parents},
         {"test_simplest_contradictory_children", test_simplest_contradictory_children},
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -410,6 +410,12 @@ class TestHoleyTreeSequences(TopologyTestCase):
         for t in ts.trees():
             observed.append((t.interval, t.parent_dict))
         self.assertEqual(expected, observed)
+        # Test simple algorithm also.
+        observed = []
+        for interval, parent in tsutil.algorithm_T(ts):
+            parent_dict = {j: parent[j] for j in range(ts.num_nodes) if parent[j] >= 0}
+            observed.append((interval, parent_dict))
+        self.assertEqual(expected, observed)
 
     def test_simple_hole(self):
         nodes = six.StringIO("""\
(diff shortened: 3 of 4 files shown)


msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+                    newick_string_tail_start_pos = pos
+                    newick_string_tail = working_newick_string[pos+1:]
+                    break
+            if newick_string_tail.count(",") == 0:
+                nstsp = newick_string_tail_start_pos
+                working_newick_string = working_newick_string[1:nstsp+1]
+
+        # Parse the bifurcating part of the tree.
+        err = 'It appears that the tree string does not include branch lengths!'
+        assert ":" in working_newick_string, err
```
jeromekelleher: We should raise a ValueError here rather than an assert, I think. I think asserts are for internal sanity checks ("this assumption I'm making should definitely be true"), not input validation. So, this would be changed to:
```python
if ":" not in working_newick_string:
    raise ValueError("Species tree strings must include branch lenghts")
```

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+                    level -= 1
+                if level == 1 and pos > 1:
+                    newick_string_tail_start_pos = pos
+                    newick_string_tail = working_newick_string[pos+1:]
+                    break
+            if newick_string_tail.count(",") == 0:
+                nstsp = newick_string_tail_start_pos
+                working_newick_string = working_newick_string[1:nstsp+1]
+
+        # Parse the bifurcating part of the tree.
```
jeromekelleher: does it have to be strictly binary? What happens if someone passes in a tree with more than 2 children at a node?

msprime/species_tree.py
jeromekelleher: There's a bunch of other places with the same pattern, I won't comment on them.

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+                            total_edge_lengths.append(total_edge_length)
+                        else:
+                            this_edge = parent_edge
+        # Make sure that the tree is at least roughly ultrametric.
+        # A maximum difference of 1% between the shortest and longest
+        # root-to-tip distances is allowed.
+        max_total_edge_length = max(total_edge_lengths)
+        if (max_total_edge_length - min(total_edge_lengths) >
+                0.01 * max_total_edge_length):
+            raise NotImplementedError(
```
jeromekelleher: This is a ValueError --- NotImplementedError is used when a subclass hasn't implemented an abstract method from a superclass

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+    def get_node_ids(self):
+        return self.node_ids
+
+    def get_is_extant(self):
+        if self.node_ids[1][0:12] == 'internalNode':
+            return False
+        else:
+            return True
+
+    def set_length(self, length):
```
jeromekelleher: I don't think we need all the Java-style getters and setters here, we can just access and set the ``length`` attribute directly (particularly since this is an internal API, and won't be exposed to the user)

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+    def get_id(self):
+        return self.id
+
+    def set_node_ids(self, node_ids):
+        self.node_ids = node_ids
+
+    def get_node_ids(self):
+        return self.node_ids
+
+    def get_is_extant(self):
```
jeromekelleher: I think we can implement this as a property:
```python
@property()
def is_extant(self):
    return self.node_ids[1][0:12] != 'internalNode'
```

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+    else:
+        err = 'With branch lengths in units of '
+        err += '"{}", generation_time must be specified.'.format(branch_length_units)
+        assert generation_time, err
+
+    # Get the number of generations per branch length unit.
+    if branch_length_units == "gen":
+        generations_per_branch_length_unit = 1
+    elif branch_length_units == "myr":
+        generations_per_branch_length_unit = 1000000/generation_time
```
jeromekelleher: 10**6 is more readable here

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+    else:
+        generations_per_branch_length_unit = 1/generation_time
+
+    # Read the input file.
+    tree_string = None
+    starbeast_format = False
```
jeromekelleher: I don't think we should try to detect the format --- it's much simpler and more robust to make user do it (they'll know the format, presumably). 

I think we should split this into two functions: ``parse_species_tree`` and ``parse_starbeast``. That would simplify the parameterisation (we don't have to explain that one parameter is expected if it's in one format but not if it's in the other, etc), and make the code more readable.

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+    starbeast_root_pop_size = None
+    infile = open(filename)
+    inlines = infile.readlines()
+    assert len(inlines) > 0, 'File {} is empty.'.format(filename)
+    if inlines[0][0:6].lower() == '#nexus':
+        # Assume the input is in nexus format. Maximally one tree string is read.
```
jeromekelleher: Let's split this out into a function, ``parse_nexus``

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+        info_string += 'Edge extant progeny ids:'.ljust(28)
+        for extant_progeny_id in self.extant_progeny_ids:
+            info_string += '{}, '.format(extant_progeny_id)
+        info_string = info_string[:-2]
+        info_string += '\n'
+        return info_string
+
+
+def parse_species_tree(
+        filename=None,
```
jeromekelleher: I think it would be simpler for now to pass a string as the argument rather than insisting on a file path. It makes things pretty tedious for testing if we have to write a bunch of tempfiles. It's easy enough for the user to write
```python
with open("treefile", "r") as f:
     parse_species_tree(f.read())
```

We can add an additional ``filename`` argument also later if you like, but for the initial version let's get it working with a string parameter.

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+                lists_are_sorted = False
+                break
+
+    # Determine at which time which populations should merge.
+    indices1 = []
+    indices2 = []
+    internal_pop_sizes = []
+    divergence_times = []
+    for edge in species_tree.get_edges():
+        if edge.get_is_extant() is False:
```
jeromekelleher: It's not a good idea to compare booleans with object identity. 
```python
if edge.is_extant:
```
is more idiomatic.

msprime/species_tree.py
```diff
@@ -0,0 +1,834 @@
...
+                    proportion=1.0))
+        demographic_events.append(
+            msprime.MassMigration(
+                time=divergence_times[-1]*generations_per_branch_length_unit,
+                source=indices2[-1],
+                destination=indices1[-1],
+                proportion=1.0))
+
+    # Return a tuple of population_configurations and demographic_events.
+    return (population_configurations, demographic_events)
```
jeromekelleher: The brackets aren't needed here.

msprime/__init__.py
```diff
@@ -51,3 +51,4 @@
 from msprime.exceptions import *
 from msprime.mutations import *
 from msprime.likelihood import *
+from msprime.species_tree import *
```
jeromekelleher: Not sure we want to import everything into the msprime namespace here, as there's going to be quite a bit of private stuff in here we don't want the user to see. I think we should import ``parse_species_tree`` and parse_startbeast`` here.

tests/data/species_trees/primates.tre
```diff
@@ -0,0 +1 @@
+(((human:5.6,chimpanzee:5.6):3.0,gorilla:8.6):9.4,orangutan:18.0)
```
jeromekelleher: Not really worth storing these as files; we could just keep them as module level variables in ``test_species_tree.py``, like
```python
primates_tree = "(((human:5.6,chimpanzee:5.6):3.0,gorilla:8.6):9.4,orangutan:18.0)"
```

msprime/__init__.py
mmatschiner: Makes sense!

msprime/species_tree.py
mmatschiner: Fixed.

msprime/species_tree.py
mmatschiner: Fixed.

msprime/species_tree.py
mmatschiner: Fixed.

msprime/species_tree.py
mmatschiner: It does. The parser throws an error when a node is not binary. Unless you think it should be changed,  I'ld prefer to keep this restriction to binary trees, because in practice, species trees inferred with all or at least most methods are binary, and because I'm a bit worried that changes to the parser might make it less robust.

petrelharp: I vote for not restricting to binary trees. What sort of non-robustness are you worried about?

jeromekelleher: Let's keep it to binary trees for the initial version then, and see about generalising later. The restriction is fine I think, once we raise a good error when we encounter a non-binary tree.

jeromekelleher:
I've had a bit of a play with this @mmatschiner, and I think our lives would be a lot simpler if we used the [newick](https://pypi.org/project/newick/) package to do the parsing. I inserted this into the ``parse_extended_newick_string``:
```python
        print("self.newick_string = ", self.newick_string)
        root = newick.loads(self.newick_string)[0]
        for node in root.walk():
            print(node.length, node.name)
```
and we get 
```
self.newick_string =  ((((((spc12[&dmv={0.11865530745434695}]:2.1888253779862743,spc11[&dmv={0.3619158652399699}]:2.1888253779862743)[&dmv={0.8111402760207012}]:0.5650137097072183,((spc08[&dmv={0.15615846438202152}]:1.1573795518978671,spc06[&dmv={0.449247825255664}]:1.1573795518978671)[&dmv={0.22731541596814978}]:0.7339367424408518,(spc07[&dmv={0.7068472628949976}]:0.06103616637684422,spc05[&dmv={0.6657390159580989}]:0.06103616637684422)[&dmv={0.08988914060435668}]:1.8302801279618748)[&dmv={0.25563233175600886}]:0.8625227933547737)[&dmv={0.3262773143035577}]:4.492254220433834,(spc04[&dmv={0.2580836901510064}]:5.7741246038893665,spc03[&dmv={0.21825140538284654}]:5.7741246038893665)[&dmv={0.23175238045171034}]:1.47196870423796)[&dmv={0.16637277604025583}]:1.3138908669207447,spc02[&dmv={0.3822654051014509}]:8.55998417504807)[&dmv={0.32647418064987677}]:9.417574569831729,(spc09[&dmv={0.1799564092258708}]:16.063369096903575,spc01[&dmv={0.02494323374786103}]:16.063369096903575)[&dmv={0.6254159122303401}]:1.914189647976226)[&dmv={1.1800837491137275}]:3.2008553127180406,spc10[&dmv={0.019049791450860193}]:21.17841405759784)
0.0 None
3.2008553127180406 [&dmv={1.1800837491137275}]
9.417574569831729 [&dmv={0.32647418064987677}]
1.3138908669207447 [&dmv={0.16637277604025583}]
4.492254220433834 [&dmv={0.3262773143035577}]
0.5650137097072183 [&dmv={0.8111402760207012}]
2.1888253779862743 spc12[&dmv={0.11865530745434695}]
2.1888253779862743 spc11[&dmv={0.3619158652399699}]
0.8625227933547737 [&dmv={0.25563233175600886}]
0.7339367424408518 [&dmv={0.22731541596814978}]
1.1573795518978671 spc08[&dmv={0.15615846438202152}]
1.1573795518978671 spc06[&dmv={0.449247825255664}]
1.8302801279618748 [&dmv={0.08988914060435668}]
0.06103616637684422 spc07[&dmv={0.7068472628949976}]
0.06103616637684422 spc05[&dmv={0.6657390159580989}]
1.47196870423796 [&dmv={0.23175238045171034}]
5.7741246038893665 spc04[&dmv={0.2580836901510064}]
5.7741246038893665 spc03[&dmv={0.21825140538284654}]
8.55998417504807 spc02[&dmv={0.3822654051014509}]
1.914189647976226 [&dmv={0.6254159122303401}]
16.063369096903575 spc09[&dmv={0.1799564092258708}]
16.063369096903575 spc01[&dmv={0.02494323374786103}]
21.17841405759784 spc10[&dmv={0.019049791450860193}]
```

So it seems like we can just parse the extended newick annotations out of the ``node.name`` attribute? This seems a **lot** easier to me than parsing newick...

mmatschiner:
That's right. I'll rewrite species_tree.py to use the newick package.

msprime/species_tree.py
mmatschiner: Thanks, I didn't know that. Should now be fixed in my fork.

msprime/species_tree.py
mmatschiner: By using the newick package for parsing now, binary trees are allowed. They are randomly resolved by injecting zero-length branches.

msprime/species_tree.py
mmatschiner: Ok, thanks. All these are no longer included.

msprime/species_tree.py
mmatschiner: This also is no longer included.

msprime/species_tree.py
mmatschiner: I now split the code into these two functions.

msprime/species_tree.py
mmatschiner: The new functions `parse_species_tree` and `parse_starbeast` now allow input that is either a one-line Newick string or a Nexus file written by StarBEAST. Nexus files that are not in StarBEAST format are not allowed. If we stick to this, the change that @jeromekelleher suggests here is no longer relevant.

msprime/species_tree.py
mmatschiner: I agree, this is now changed.

msprime/species_tree.py
mmatschiner: Thanks, this is now changed.

msprime/__init__.py
mmatschiner: Now changed.

msprime/species_tree.py
```diff
@@ -0,0 +1,489 @@
...
+    species_tree_string = get_species_tree_string(species_tree_lines[0])
+
+    # Parse the newick tree string.
+    root = newick.loads(species_tree_string)[0]
+
+    # Remove nodes that have only a single child if there should be any.
+    root.remove_redundant_nodes()
+
+    # Resolve polytomies by injecting zero-length branches.
+    root.resolve_polytomies()
```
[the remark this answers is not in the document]
mmatschiner: It seems `resolve_polytomies()` isn't exactly doing what I thought it would. Not sure if the ancestor property is not getting updated as it should:
```
import newick

def print_node_heights(tree):
    root = tree
    # Set node depths (distances from root)
    max_depth = 0
    for node in root.walk():
        depth = 0
        moving_node = node
        while moving_node.ancestor:
            depth += moving_node.length
            moving_node = moving_node.ancestor
        node.depth = depth
        if depth > max_depth:
            max_depth = depth
    # Set node heights (distances from present).
    for node in root.walk():
        node.height = max_depth - node.depth
        print(node.height)

root = newick.loads("((human:1,chimpanzee:1,gorilla:1):2,orangutan:3)")[0]
print_node_heights(root)
print()
root.resolve_polytomies()
print_node_heights(root)
```
Anyway, the best way to avoid this issue may be not to use resolve_polytomies, so I'll follow your advice and remove the dependence on bifurcating trees (at least for the newick parsing, not for the starbeast parsing though).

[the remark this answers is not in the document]
mmatschiner: The output of the above code is:
```
3.0
1.0
0.0
0.0
0.0
0.0

```(before calling resolve_polytomies)
and 
```
3.0
1.0
0.0
3.0
2.0
2.0
0.0

```
(after calling it). So somehow nodes end up having a height of  2 (which is not in the input tree) and fewer have a height of 0 (even though four nodes are extant).

File: tests/data/species_trees/primates.tre (hunk 13 above)
Comment (mmatschiner) [reply tests/data/species_trees/primates.tre]: Done, except for two large input trees (one in starbeast format, one with over 100 species).

File: msprime/species_tree.py [hunk 15]
```diff
@@ -0,0 +1,489 @@
...
+    if branch_length_units == "gen":
+        generations_per_branch_length_unit = 1
+    elif branch_length_units == "myr":
+        generations_per_branch_length_unit = 10**6/generation_time
+    else:
+        generations_per_branch_length_unit = 1/generation_time
+    return generations_per_branch_length_unit
+
+
+def get_species_tree_string(species_tree_line=None):
```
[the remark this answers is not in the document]
mmatschiner: I now removed the checks for equal numbers of opening and closing parentheses and for having at least one parenthesis open throughout the string. The error messages that the newick package throws could be better, but I agree, let's leave this responsability to that package.

msprime/species_tree.py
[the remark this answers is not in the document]
mmatschiner: Ok, resolve_polytomies() is no longer used, and neither is remove_redundant_nodes(). Instead the code now works with trees that have nodes with any number of descendants.

mmatschiner:
Tests are going to come next.

jeromekelleher:
Looking really nice now @mmatschiner, thanks. I'm looking forward to the tests (as I say, I can help out a bit here, just let me know).

jeromekelleher:
Tests looks good @mmatschiner, probably needs a bit more on parsing bad data though.

I can take it from here if you like?

mmatschiner:
Feel free to! I'm new to testing, but I did what seemed to make sense for me. I'll leave this to you now but please let me know if I should do anything else.

jeromekelleher:
OK, cool, I'll update things here soon and ping you for a review.

Just so you know, I'll need to write to your fork's master branch when I'm [rebasing](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing), so it will get out of sync with your local version. This is one of the reasons that using [topic branches](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow) is a good idea. You basically then never use the master branch of your fork, so don't need to worry about keeping it synced.

Anyway, doesn't make any difference here, just wanted to give you a heads up in case you encountered weirdness later.

jeromekelleher:
I've spent a good bit of time with this @mmatschiner, and updated to simplify the algorithm a bit and to update the tests. Hopefully this is an improvement. I think that the tests for parse_species_tree are pretty good now (although we do need to make sure we're round tripping correctly with generation time units). There's still a lot of work to be done in parse_startbeast though. There's a **lot** of code here, and a lot of places for things to go wrong, and we need to test and verify all those places. Hopefully what I've done here will serve as a useful platform for updating the code with some more tests.

Here's what we need to do I think:

1. Test round tripping with different branch_length_units and generation_time values
2. Lots of tests for malformed nexus intput for starbeast. The newick side of things is pretty well covered, as we're using the same code now for both, but we need to verify what happens when we have malformed translation blocks and all sorts of stuff. It's messy.
3. Check for malformed annotations in starbeast, plus unparseable dmv values. Also those where dmv is present for only some nodes (so the initial regexp doesn't catch the error)
4. Test that we're doing the back translating species names correctly when doing translations (although, we're not actually using the names now...)
5. Probably add ``metadata={"species_name": node.name}`` into the PopulationConfiguration object so that we can track that we're capturing the structure correctly (and make the information available to the user)
6. Document the resulting simulation more clearly. Basically we need to paraphrase the translation algorithm so that users will know what each population means at each point in time. I.e, we always merge into the left-most population, which then becomes the ancestral pop.

That's about it I think. Apologies is this seems like a lot more work, but parsing text well is hard, and it needs a lot of testing. You should be able to add more commits to this branch if you run ``git pull`` locally, and then push new commits as before.

mmatschiner:
Hi Jerome, thanks a lot for all the efforts. I haven't had a chance yet to go through everything, sorry for that. Regarding StarBEAST tree parsing; I see two things to consider:

1. We could assume that StarBEAST trees are really just written by the program StarBEAST and not manually by the user or by other tools. If we only expect trees written by StarBEAST, we know how they will be written (https://github.com/genomescale/starbeast2/blob/master/src/starbeast2/SpeciesTreeLogger.java) and we would not need to worry too much about malformed input.

2. If you think the StarBEAST part is too difficult to test and maintain properly I wouldn't insist on keeping it in msprime. Would it make more sense to distribute the function with StarBEAST instead so the responsability is not with the msprime developers?

I'll work on addressing your comments over the next few days.

jeromekelleher:
Your point about starbeast input being machine generated is a good one @mmatschiner, so I guess we don't need to be *too* worried about raising error messages and just using lots of ``asserts`` to sanity check is a good approach. 

One thing we could do then is to be quite strict about **only** accepting stuff that has come from starbeast (for example, does it always use a translation block?), which could simplify the code a bit too.

What do you think?

mmatschiner:
Good, I agree. I'll contact the StarBEAST developers about our plans to check which aspects of the format they consider stable.

jeromekelleher:
Hey @mmatschiner, we'd like to get this merged pretty soon. Perhaps we could get the small things finished up, and then open some issues for things like improving starbeast test coverage?

mmatschiner:
Sorry for the delay. A number of other projects unexpectedly need my attention. But I'll prioritize this now. Your suggestion sounds fine!

mmatschiner:
Took me a bit to figure out how to modify tree object, but https://tskit.readthedocs.io/en/latest/tutorial.html#working-with-tables helped.

mmatschiner:
Addressing point 6 next...

jeromekelleher:
Great, please ping me when you'd like input @mmatschiner.

mmatschiner:
Following https://github.com/tskit-dev/msprime/pull/893#issuecomment-597548838 I added asserts for parse_starbeast. Please @jeromekelleher let me know if you think this is not going in the right direction. Otherwise I'll continue with
1. extending TestStarbeastRoundTrip
2. adding more asserts to the translation block parsing.

Additionally, I wonder if we should add something like this:
```
        for y in range(len(root.get_leaves())):
            if y != source:
                demographic_events.append(
                    msprime.MigrationRateChange(
                        time=divergence_times[x]*generations_per_branch_length_unit,
                        rate=0,
                        matrix_index=(source, y)))
                demographic_events.append(
                    msprime.MigrationRateChange(
                        time=divergence_times[x]*generations_per_branch_length_unit,
                        rate=0,
                        matrix_index=(y, source)))

```
with every MassMigration. It would facilitate the use of migration rate matrices by stopping migration into lineages after (before in forward time) the MassMigration. On the other hand, it would implicitly assign the migration rates of the left-most childs to ancestral species which may also not be wanted.

jeromekelleher:
The changes look good, thanks @mmatschiner, I think you're going in the right direction. Let's not worry about migration rates for now, we can do this in a follow up issue I think.

mmatschiner:
I think points 2-4 and 6 of https://github.com/tskit-dev/msprime/pull/893#issuecomment-595210458 are now also addressed. Let me know if anything more is missing, @jeromekelleher.

jeromekelleher:
Great, thanks @mmatschiner. I'll finish things up from here and open issues for any stuff that I think needs following up on.

PR #894, opened 2020-02-13
Implement verify_overlaps for continuous and discrete coordinates

### LINKED ISSUE #875: [C code] Fix verify_overlaps to work on continuous coordinates
We don't have any C level test coverage for the continuous coordinates and we should make sure that the basic mechanisms are working.

I suggest we update the verify_overlaps algorithm to use a simple linked list based approach, where we use a list of segments to count how many of the extant segments overlap with each interval. We start off with a single interval [0, L] which has 0 segments overlapping it. We then iterate through all the individuals and their segments and update this list, performing a linear search each time to find the appropriate interval, and breaking it as needed. This shouldn't be too complicated to do, and can reuse the ``msp_alloc_segment/free_segment`` infrastructure to handle the memory management (using the ``value`` field on msp_segment_t as the counter).

It'd be sensible to do this in Python in algorithms.py first, where we also have some verification logic.

How does this sounds @daniel-goldstein?

daniel-goldstein:
For `verify_overlaps`, instead of checking overlaps per locus (which only works in discrete coordinates), we now create a segment chain that counts how many segments cover particular intervals and check that against the node mappings. The algorithm follows roughly that described in #875.

Resolves #875

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -572,6 +572,66 @@ class RecombinationMap(object):
         return left
 
 
+class OverlapCounter(object):
+    def __init__(self, seq_length):
+        self.seq_length = seq_length
+        self.overlaps = self._make_segment(0, seq_length, 0)
+
+    def overlaps_at(self, pos):
+        assert 0 <= pos < self.seq_length
+        curr_interval = self.overlaps
+        while curr_interval is not None:
+            if curr_interval.left <= pos < curr_interval.right:
+                return curr_interval.node
+            curr_interval = curr_interval.next
+        raise ValueError("Bad overlap count chain")
+
+    def increment_interval(self, left, right):
+        """
+        Increment the count that spans the interval
+        [left, right), creating additional intervals in overlaps
+        if necessary.
+        """
+        curr_interval = self.overlaps
+        while left < right:
+            if curr_interval.left == left:
+                if curr_interval.right <= right:
+                    curr_interval.node += 1
+                    left = curr_interval.right
+                    curr_interval = curr_interval.next
+                else:
+                    self._split(curr_interval, right)
+                    curr_interval.node += 1
+                    break
+            else:
+                if curr_interval.right < left:
+                    curr_interval = curr_interval.next
+                else:
+                    self._split(curr_interval, left)
+                    curr_interval = curr_interval.next
+
+    def _split(self, seg, breakpoint):
+        """
+        Split the segment at breakpoint and add in another segment
+        from breakpoint to seg.right. Set the original segment's
+        right endpoint to breakpoint
+        """
+        right = self._make_segment(breakpoint, seg.right, seg.node)
+        if seg.next is not None:
+            seg.next.prev = right
+            right.next = seg.next
+        right.prev = seg
+        seg.next = right
+        seg.right = breakpoint
+
+    def _make_segment(self, left, right, count):
+        seg = Segment(0)
+        seg.left = left
+        seg.right = right
+        seg.node = count
+        return seg
+
+
 class Simulator(object):
     """
     A reference implementation of the multi locus simulation algorithm.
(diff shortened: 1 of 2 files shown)


algorithms.py
```diff
@@ -572,6 +572,61 @@ def _search(self, values, query):
         return left
 
 
+def make_segment(left, right, node):
```
jeromekelleher: These functions together feel like a class. Maybe a higher level interface like OverlapClounter or something would be good? It could have a method ``increment(left, right)`` which would do that, and hide away the details of how it's done.

lib/msprime.c
```diff
@@ -807,53 +807,123 @@ msp_verify_segments(msp_t *self, bool verify_breakpoints)
     }
 }
 
-/* FIXME logic doesn't make sense in continuous space */
+/* Find the segment that spans the given position */
```
jeromekelleher: Same logic can apply in the c code here. We can make a 'local' segment_finder_t struct (or "class") which would do the same thing as the Python one. It'll simplify things a bit for someone reading the code if this functionality is grouped together and not part of the ``msp_`` namespace (which is poorly named btw, should be something like ``msp_sim_t``, but that's another story)

daniel-goldstein: Yes this is much cleaner.

algorithms.py
daniel-goldstein: Yes this is much cleaner. Thanks!

lib/msprime.c
```diff
@@ -807,53 +807,147 @@ msp_verify_segments(msp_t *self, bool verify_breakpoints)
...
+        right_seg->next = seg->next;
+        seg->next->prev = right_seg;
+    }
+    right_seg->prev = seg;
+    seg->next = right_seg;
+    seg->right = breakpoint;
+}
+
+/* Increment the number of segments that span
+ * [left, right], creating additional intervals if necessary.
```
jeromekelleher: Same comment on notation here.

lib/msprime.c
```diff
@@ -807,53 +807,147 @@ msp_verify_segments(msp_t *self, bool verify_breakpoints)
...
-/* FIXME logic doesn't make sense in continuous space */
+typedef struct {
+    double seq_length;
+    segment_t *overlaps;
+} overlap_counter_t;
+
+static overlap_counter_t*
+overlap_counter_alloc(double seq_length, int initial_count)
+{
+    overlap_counter_t *counter = malloc(sizeof(overlap_counter_t));
```
jeromekelleher: Just for the sake of keeping things consistent, it's probably better to make the first parameter of this ``self``, and the return type an int. Future readers might wonder why this object is following a different convention to other objects, when it doesn't really matter.

lib/msprime.c
```diff
@@ -807,53 +807,147 @@ msp_verify_segments(msp_t *self, bool verify_breakpoints)
...
-    double left, right;
-    size_t remaining_samples = self->num_sampling_events - self->next_sampling_event;
-    /* We check for every locus, so obviously this rules out large numbers
-     * of loci. This code should never be called except during testing,
-     * so we don't need to recover from malloc failure.
-     */
-    uint32_t *overlaps = calloc((uint32_t) self->sequence_length, sizeof(uint32_t));
+    uint32_t j, label;
+    int remaining_samples = (int) (self->num_sampling_events - self->next_sampling_event);
+    overlap_counter_t *counter = overlap_counter_alloc(
```
jeromekelleher: If we make this 
```
overlap_counter_t counter;

overlap_counter_alloc(&counter, self->sequence_length, remaining_samples);
```
then we keep the conventional API structure and avoid a malloc.

daniel-goldstein:
Changes addressed. Realized though that with the interface you requested of `increment(left, right)` on the `OverlapCounter`, we have to search from the beginning of overlaps again, instead of starting where the previous segment in the chain was inserted. Unit tests run closer to 6 seconds instead of 4.5. If this is fine it's a bit simpler. Or I can add a whole segment chain at a time to not look from the beginning for each segment, only each chain.

jeromekelleher:
I think the tests are quick enough, let's keep the interface a bit simpler. Will merge, thanks.

Issue #890, opened 2020-02-07
Fix error in beta coalescent time scaling


jeromekelleher:
From @JereKoskela:

> The correct one is beta_model_rate_to_generation_rate(), which is just calculating c_N from the equation in [Lemma 13]( https://reader.elsevier.com/reader/sd/pii/S0304414903000280?token=A3ABF6C4FDCBBC3D293361DECEF28CE281102B8C678CDB37A63ECF837C65DE083BE365C73955BD8AA7AA42D8EF303ED0) (set phi = 1 the code to see that easily, which should be its default value anyway).

jeromekelleher:
See also #889.

Issue #186, opened 2017-04-12
Time rescaling needed for Dirac coalescent


jeromekelleher:
We need to rescale the time in the Dirac coalescent into generations. Currently we're scaling by 4Ne, which is wrong.

Similarly, our input recombination rate is in units of generations, but is not being rescaled into the appropriate time units for the Dirac coalescent.

eldonb:
we rescale by N^2 (N squared); then we have the mixture process of  multiple mergers  +  the kingman part.   However, recombination is scaled in units of N; this follows from  the generation of 1  recombinant offspring at most each generation.  

bjarki

jeromekelleher:
See the code [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L5453).

At moment we're scaling by N in one direction and N^2 in the other, which can't be right...

JereKoskela:
> we rescale by N^2 (N squared); then we have the mixture process of multiple mergers + the kingman part. However, recombination is scaled in units of N; this follows from the generation of 1 recombinant offspring at most each generation.

Are you sure that rescaling by N^2 is right @eldonb? Granted, the pre-limiting Moran model scales to the ARG when time is scaled in units of N^2 steps. But I thought the usual convention for "a generation" in a Moran model was N steps, so that everyone gets replaced once on average. Overall, that would leave a scaling factor proportional to N for both time and recombination rates when converting coalescent time to generations.

eldonb:
N^2 is the correct scaling of the time in a Moran-type model -  it  clearly follows from the scaling of c_N, the probability of  two individuals (gene  copies) sampled at the same time having the same parent; however,  we need to scale  recombination by N, this follows from the arrangement of having recombination occurring in conjunction with a "small" reproduction event, when  only one offspring is produced, an event which happens with probability 1 - \epsilon_N, and  \epsilon_N \to 0 as N \to \infty.   So the "effective size" Ne  is of order N^2.  The timescaling is discussed in the Genetics ARG-paper with Birkner and Blath; I will dig this out and write it up in the section part

JereKoskela:
I completely agree, and if we were converting between coalescent time and discrete Moran time steps, then everything would be crystal clear. But that's not the conversion we are making - we are converting between coalescent time and something we call "generations".

[...]

PR #898, opened 2020-02-17
Unit test for consistency of model time changes.

### LINKED ISSUE #889: [C code] Unit test consistency of model time changes
Add a unit test in C to check that all model time change functions are consistent, e.g.
```
    beta_generations_to_model_time(beta_model_time_to_generations(t)) == t,
    beta_model_time_to_generations(t) == beta_generation_rate_to_model_rate(t),
```

grahamgower:
Closes #889.

The test passes, so I'm not sure what #890 is suggesting. Probably I'm missing something here---the mulitple-merger coalescent maths is all way over my head.

Also, I didn't see any info on how the cunit tests should be grouped/ordered, so possibly I put this in a less than ideal location.

### DIFF
--- a/lib/tests/tests.c
+++ b/lib/tests/tests.c
@@ -852,8 +852,8 @@ test_demographic_events(void)
             MSP_ERR_POPULATION_OUT_OF_BOUNDS);
 
         CU_ASSERT_EQUAL(
-        	msp_add_census_event(&msp, -0.5),
-        	MSP_ERR_BAD_PARAM_VALUE);
+            msp_add_census_event(&msp, -0.5),
+            MSP_ERR_BAD_PARAM_VALUE);
 
         ret = msp_add_census_event(&msp, 0.05);
         CU_ASSERT_EQUAL(ret, 0);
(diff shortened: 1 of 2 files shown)


lib/tests/tests.c
```diff
@@ -3117,6 +3117,84 @@ compute_beta_coalescence_rate_fails(unsigned int num_ancestors, double alpha)
...
+    double g, g2, t2;
+    g = msp->model.model_time_to_generations(&msp->model, t);
+    t2 = msp->model.generations_to_model_time(&msp->model, g);
+    CU_ASSERT_DOUBLE_EQUAL(t, t2, tol);
+    g2 = msp->model.generation_rate_to_model_rate(&msp->model, t);
+    CU_ASSERT_DOUBLE_EQUAL(g, g2, tol);
+}
+
+static void
+test_model_time_change_consistency1(double population_size, double t)
```
jeromekelleher: The convention is to use the ``test_`` prefix for top-level tests (like in unittest), so let's change this to ``check_model_time_consistency``

lib/tests/tests.c
```diff
@@ -3117,6 +3117,84 @@ compute_beta_coalescence_rate_fails(unsigned int num_ancestors, double alpha)
...
+    tsk_table_collection_free(&tables);
+    free(samples);
+    gsl_rng_free(rng);
+}
+
+static void
+test_model_time_change_consistency(void)
+{
+    double N;
+    for (N=1.0; N<1e9; N+=N) {
```
jeromekelleher: This is a **lot** of Ns! Something like ``double Ns = {1e-6, 1e-1, 1, 2, 3, 10, 1e4, 1e9}`` would probably cover all the interesting cases I'd say.

lib/tests/tests.c
```diff
@@ -3117,6 +3117,84 @@ compute_beta_coalescence_rate_fails(unsigned int num_ancestors, double alpha)
...
+    CU_ASSERT_DOUBLE_EQUAL(g, g2, tol);
+}
+
+static void
+test_model_time_change_consistency1(double population_size, double t)
+{
+    int ret;
+    msp_t msp;
+    const unsigned int n = 10;
+    const double alpha = 1.5;
```
jeromekelleher: We should probably check this for the cross product of a few plausible values of alpha and truncation point, ie.

```c
double alphas[] = {1e-6, 0.1, 1, 1.5, 2};
double truncation_points[] = {1e-6, 0.1, 1, 5};

for (j = 0; j < sizeof(alphas) / sizeof(*alphas); j++) {
    for (k = 0; k < sizeof(...)) {
        ret = msp_set_simulation_model_beta(&msp, population_size, alphas[j], truncation_points[k]);
￼        CU_ASSERT_EQUAL_FATAL(ret, 0);
￼        check_time_change(&msp, t);
    }
}

**Comment by grahamgower** [comment]
I added a check that `model_rate_to_generation_rate` and `generation_rate_to_model_rate` are inverses. This fails for each of the dirac and beta test cases.

I also changed `check_time_change()` to be a macro instead of a function, as this more faithfully preserves line numbers for the failed asserts.

File: lib/tests/tests.c [hunk 4]
```diff
@@ -3117,6 +3117,94 @@ compute_beta_coalescence_rate_fails(unsigned int num_ancestors, double alpha)
...
+    ret = msp_set_simulation_model_smc_prime(&msp, population_size);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    check_time_change(&msp, t);
+    ret = msp_set_simulation_model_dtwf(&msp, population_size);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    check_time_change(&msp, t);
+    ret = msp_set_simulation_model_wf_ped(&msp, population_size);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    check_time_change(&msp, t);
+    ret = msp_set_simulation_model_dirac(&msp, population_size, psi, c);
```
jeromekelleher: Be nice to do the same thing here with plausible values for psi and c

jeromekelleher:
> I added a check that model_rate_to_generation_rate and generation_rate_to_model_rate are inverses. This fails for each of the dirac and beta test cases.

Good, thanks. That's reassuring.

> I also changed check_time_change() to be a macro instead of a function, as this more faithfully preserves line numbers for the failed asserts.

Clever, I like it.

lib/tests/tests.c
grahamgower: Happy to do that if you can suggest some numbers.

jeromekelleher: Let's say psi = {1e-6, 0.01, 0.5, 0.9, 1 -1e-6} and c = {0.01, 1, 5}

jeromekelleher:
Hey @grahamgower, just wondering what the status of this one is? Be nice if we could rebase and get it merged soon.

grahamgower:
I'll update the parameter values to be in the accepted ranges and rebase.

grahamgower:
For moderate-large `reference_size` values, the following expression is zero (which is problematic because we divide by this scalar when doing time/rate conversions).
```
scalar = exp(log(alpha) - alpha * log(m) - (alpha - 1) * reference_size)
             * acceptance_rate;
```

I checked the equation, and it faithfully reproduces the previous code before my changes in #904. Is it reasonable to have large values for reference_size, or should that ordinarily be closer to 1?

jeromekelleher:
I would have thought reference_size of 10^6 would be perfectly reasonable. @JereKoskela, any thoughts here?

JereKoskela:
Sorry, I should have noticed this sooner. The expression should be

`scalar = exp(log(alpha) - alpha * log(m) - (alpha - 1) * log(reference_size))
             * acceptance_rate;`

That is, the reference size enters polynomially like `N^(alpha - 1)`, not exponentially. That should take care of any underflows to zero.

The question of default values is not trivial, because the drawback of the Schweinsberg model is that it requires astronomical effective population sizes for predicting reasonable amounts of diversity. Perhaps we should just put in a default of 10^6 and highlight this in the documentation, but practitioners should expect to need numbers closer to 10^20 in many settings.

grahamgower:
I made the change to `log(reference_size)`, and this triggered some gsl integration errors in other tests. I'll keep chipping away at it.

grahamgower:
The beta coalescent bits are happy now. But there are still tests/tests.c failures because the dirac coalescent time/rate conversion functions are not yet inverses.

jeromekelleher:
Great, thanks @grahamgower. How about we skip those dirac tests (comment them out I guess) and register the problem as an issue? I'd like to get these open PRs merged as we want to start using Black/clang-format for formatting the code.

grahamgower:
Ok. Squashed and ready for merging once circle CI is happy.

jeromekelleher:
Done, thanks @grahamgower. I'll make a note on #186 to say we should enable these tests once it's fixed.

PR #899, opened 2020-02-18
[stats tests] Add breakpoint verification tests for continuous coordinates

### LINKED ISSUE #877: [stats tests] Add checks for hotspots when simulating in continuous coords
All existing hotspot tests are in the case of discrete coords as we're comparing the msHOT. Run a test comparing breakpoint positions in msprime Hudson when we have discrete and continuous coords (should probably check in DTWF also, see #876)

daniel-goldstein:
Added tests compare breakpoint distributions and num_trees between simulations run in continuous and discrete coordinates, both on the same large scale as well as different sequence lengths/recombination rates that should have the same distribution and are scaled post-simulation. Tests are implemented for Hudson and DTWF.

Resolves #877

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -36,6 +36,12 @@ def flatten(l):
     return [x for sublist in l for x in sublist]
 
 
+def scale_breakpoints(df, factor):
+    def scale(points):
+        return [factor * x for x in points]
+    df['breakpoints'] = df['breakpoints'].map(scale)
+
+
 def harmonic_number(n):
     return np.sum(1 / np.arange(1, n + 1))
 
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
### Output Examples
Of the two simulations run, one was in discrete coordinates over 2Mb and the other over a sequence of length 1 with recombination rate scaled appropriately. First plot shows breakpoint distributions over a 2Mb range (the continuous simulation was scaled up afterward). Second plot shows num_trees.

#### Hudson
compare continuous and discrete coordinates_breakpoints
compare continuous and discrete coordinates_num_trees


#### DTWF

compare continuous and discrete coordinates_breakpoints
compare continuous and discrete coordinates_num_trees

jeromekelleher:
Excellent, thanks @daniel-goldstein. Merging.

PR #903, opened 2020-02-20
Add RecombinationMap.slice() to get a recomb map subset.

### LINKED ISSUE #901: Add method to return a slice of a recombination map
It is useful to return a subset of a recombination map. We could have a method like

```
def slice(self, start=None, end=None, remove_flanks=False):
    """
    Returns a subset of this recombination map between the specified end points. If 
    start is not specified or None, it defaults to 0. ... If remove_flanks is True,
    remove the flanking zero recombination rate regions such that the sequence length
    of the new recombination map is end - start.
    """
```

I'm not sure about the name ``slice``, since that clashes with the concept of a slice in Python, but it's quite descriptive here.

grahamgower:
Closes #901.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -29,6 +29,7 @@ import os
 import warnings
 import copy
 import logging
+import bisect
 
 import tskit
 import numpy as np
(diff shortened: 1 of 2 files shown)


msprime/simulations.py
```diff
@@ -963,6 +964,69 @@ def mean_recombination_rate(self):
...
+                else:
+                    new_positions.insert(0, 0)
+                    new_rates.insert(0, 0.0)
+            if new_positions[-1] != positions[-1]:
+                new_positions.append(positions[-1])
+                new_rates.append(0)
+        return self.__class__(
+                new_positions, new_rates, discrete=self.discrete)
+
+    def __getitem__(self, key):
```
jeromekelleher: Very nice idea, I like it! What happens when we give non-integer arguments like ``m[0.5:0.9]``? Very cool if it works!

msprime/simulations.py
```diff
@@ -963,6 +964,69 @@ def mean_recombination_rate(self):
...
+        rates = self.get_rates()
+
+        if start is None:
+            i = 0
+            start = 0
+        if end is None:
+            end = positions[-1]
+            j = len(positions)
+
+        # TODO: Is there interest in supporting negative indexes? Eg. make
```
jeromekelleher: I guess we could right? In the case where the recombination map is discrete it makes perfect sense, and I guess the same logic should hold for floating point values. Here's the standard approach:
```
    def __getitem__(self, index):
        if index < 0:
            index += len(self)
        if index < 0 or index >= len(self):
            raise IndexError("Index out of bounds")
        return self.getter(index)
```

msprime/simulations.py
grahamgower: I didn't try it before, but it seems to work just fine.
```
>>> m = msprime.RecombinationMap.read_hapmap("chr1_average_canFam3.1.txt")
>>> m[10e6:11e6].get_positions()
[0.0, 8027.0, 683500.0, 780593.0, 807939.0, 934077.0, 961544.0, 1000000.0]
>>> m[10e6+0.001:11e6+0.7].get_positions()
[0.0, 8026.998999999836, 683499.9989999998, 780592.9989999998, 807938.9989999998, 934076.9989999998, 961543.9989999998, 1000000.6989999991]
```

jeromekelleher: Beautiful! OK, let's write a few test cases for this behaviour and merge. This is awesome.

grahamgower:
Hmmm... I think I bodged one of the tests here. I'll take a look with fresh eyes tomorrow.

msprime/simulations.py
```diff
@@ -963,6 +964,72 @@ def mean_recombination_rate(self):
 
         return np.average(rates, weights=weights)
 
+    def slice(self, start=None, end=None, remove_flanks=False):
```
jeromekelleher: Actually, can we rename this parameter to ``trim`` please? This would make it consistent with some tskit functions that do similar things, e.g. https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.ltrim

grahamgower: Yup. slice is now trim, so 'slice' just means python slice syntax.

grahamgower:
Just out of curiousity, why would one want to trim with remove_flanks=False?

msprime/simulations.py
jeromekelleher: Sorry Graham, I should have been more explicit here, I meant to change the ``remove_flanks`` parameter to ``trim``, so the signature would be ``slice(start=None, end=None, trim=False)``

grahamgower: n'th time's a charm... :)

jeromekelleher: :+1:

jeromekelleher:
> Just out of curiousity, why would one want to trim with remove_flanks=False?

You might want to keep the coordinate systems the same, and just not simulate stuff to the left and right. Say if you wanted to split a chromosome up into 10 chunks and simulate each of those independently, and then show some agregated results by genome coordinate. This'll make it a bit easier (probably).

jeromekelleher:
Merged, thanks @grahamgower!

PR #904, opened 2020-02-21
Make beta coalescent time and rate conversion functions inverses.

### LINKED ISSUE #890: Fix error in beta coalescent time scaling

grahamgower:
Closes #890.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -5961,9 +5961,9 @@ beta_integrand(double x, void *p)
     for (l = 0; l <= m; l++)  {
         ret += exp(r[l] - r_max);
     }
-    exponent = (-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x;
+    exponent = (-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x
+                + log(4) - gsl_sf_lnbeta(2 - alpha, alpha);
     ret = exp(exponent) - exp(exponent + r_max + log(ret));
-    ret *= 4 / gsl_sf_beta(2 - alpha, alpha);
     return ret;
 }
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -42,6 +42,51 @@
 #define MSP_STATE_SIMULATING 2
 #define MSP_STATE_DEBUGGING 3
 
+/* Draw a random variable from a truncated Beta(a, b) distribution,
+ * by rejecting draws above the truncation point x.
+ */
+static double
+ran_inc_beta_rej(gsl_rng *r, double a, double b, double x)
+{
+    double ret;
+    do {
+        ret = gsl_ran_beta(r, a, b);
+    } while(ret > x);
+    return ret;
+}
+
+/* Draw a random variable from a truncated Beta(a, b) distribution,
+ * using the inverse transform sampling method. A uniform variable
+ * is drawn and then transformed into a Beta(a, b) distribution using
+ * the inverse CDF. Truncation is obtained by defining the upper bound
+ * of the uniform variable using the incomplete beta function. I.e.
+ *      upper_bound = gsl_sf_beta_inc(a, b, x),
+ * where x is the truncation point.
+ */
+static double
+ran_inc_beta_its(gsl_rng *r, double a, double b, double upper_bound)
+{
+    double u = gsl_ran_flat(r, 0, upper_bound);
+    return gsl_cdf_beta_Pinv(u, a, b);
+}
+
+/* Draw a random variable from a truncated Beta(a, b) distribution,
+ * using either the rejection method or the inverse transform sampling method.
+ * The rejection method is quicker for a moderate number of rejections, but if
+ * the acceptance probability is low, we instead choose inverse transform
+ * sampling. The acceptance probability for ran_inc_beta_rej() is equivalent to
+ * the upper bound used with ran_inc_beta_its().
+ */
+static double
+ran_inc_beta(gsl_rng *r, double a, double b, double x, double upper_bound)
+{
+    if (upper_bound < 0.1) {
+        return ran_inc_beta_its(r, a, b, upper_bound);
+    } else {
+        return ran_inc_beta_rej(r, a, b, x);
+    }
+}
+
 static bool
 doubles_almost_equal(double a, double b, double eps)
 {
@@ -5942,7 +5987,7 @@ msp_compute_beta_integral(msp_t *self, unsigned int num_ancestors, double alpha,
     F.function = &beta_integrand;
     F.params = &params;
     ret = gsl_integration_qags(&F, 0, truncation_point, epsabs, epsrel,
-		    workspace_size, w, result, &err);
+            workspace_size, w, result, &err);
     if (ret != 0) {
         /* It's ugly, but it should only happen while tuning models so we output
          * the GSL error to stderr */
@@ -6022,35 +6067,6 @@ msp_multi_merger_common_ancestor_event(msp_t *self, double x,
 
 out:
     return ret;
-}
-
-/* Rejection method. */
-static double
-ran_inc_beta_rej(gsl_rng *r, double a, double b, double x)
-{
-	double ret;
-	do {
-		ret = gsl_ran_beta(r, a, b);
-	} while(ret > x);
-	return ret;
-}
-
-/* Inverse transform sampling. */
-static double
-ran_inc_beta_its(gsl_rng *r, double a, double b, double upper_bound)
-{
-	double u = gsl_ran_flat(r, 0, upper_bound);
-	return gsl_cdf_beta_Pinv(u, a, b);
-}
-
-/* Choose between rejection or ITS based on the rejection probability. */
-static double
-ran_inc_beta(gsl_rng *r, double a, double b, double x, double upper_bound)
-{
-	if (upper_bound < 0.1)
-		return ran_inc_beta_its(r, a, b, upper_bound);
-	else
-		return ran_inc_beta_rej(r, a, b, x);
 }
 
 static int MSP_WARN_UNUSED
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -25,6 +25,7 @@
 #include <gsl/gsl_rng.h>
 #include <gsl/gsl_math.h>
 #include <gsl/gsl_randist.h>
+#include <gsl/gsl_cdf.h>
 #include <gsl/gsl_statistics_int.h>
 #include <gsl/gsl_sf.h>
 #include <gsl/gsl_integration.h>
@@ -5894,18 +5895,19 @@ beta_integrand(double x, void *p)
     double b = num_ancestors;
     double log_x = gsl_sf_log(x);
     double log_1_minus_x = gsl_sf_log(1 - x);
+    double exponent;
 
     m = GSL_MIN(num_ancestors, 4);
     /* An underflow error occurs because of the large exponent (b-l). We use the
      * LSE trick to approximate this calculation. For details, see at
      * https://en.wikipedia.org/wiki/LogSumExp
      */
-    r[0] = b * gsl_sf_log(1 - x);
+    r[0] = b * log_1_minus_x;
     r_max = r[0];
     for (l = 1; l <= m; l++){
         r[l] = gsl_sf_lnchoose(num_ancestors, l)
                 + compute_falling_factorial_log(l)
-                - l * 1.386294      /* log(4) = 1.386294 */
+                - l * log(4)
                 + l * log_x
                 + (b - l) * log_1_minus_x;
         r_max = GSL_MAX(r_max, r[l]);
@@ -5914,8 +5916,8 @@ beta_integrand(double x, void *p)
     for (l = 0; l <= m; l++)  {
         ret += exp(r[l] - r_max);
     }
-    ret = 1 - exp(r_max + log(ret));
-    ret *= exp((-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x);
+    exponent = (-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x;
+    ret = exp(exponent) - exp(exponent + r_max + log(ret));
     ret *= 4 / gsl_sf_beta(2 - alpha, alpha);
     return ret;
 }
@@ -5930,6 +5932,7 @@ msp_compute_beta_integral(msp_t *self, unsigned int num_ancestors, double alpha,
     size_t workspace_size = self->model.params.beta_coalescent.integration_workspace_size;
     double epsrel = self->model.params.beta_coalescent.integration_epsrel;
     double epsabs = self->model.params.beta_coalescent.integration_epsabs;
+    double truncation_point = self->model.params.beta_coalescent.truncation_point;
     struct beta_integral_params params = {num_ancestors, alpha};
 
     if (w == NULL) {
@@ -5938,7 +5941,8 @@ msp_compute_beta_integral(msp_t *self, unsigned int num_ancestors, double alpha,
     }
     F.function = &beta_integrand;
     F.params = &params;
-    ret = gsl_integration_qags(&F, 0, 1, epsabs, epsrel, workspace_size, w, result, &err);
+    ret = gsl_integration_qags(&F, 0, truncation_point, epsabs, epsrel,
+		    workspace_size, w, result, &err);
     if (ret != 0) {
         /* It's ugly, but it should only happen while tuning models so we output
          * the GSL error to stderr */
@@ -5977,7 +5981,6 @@ msp_beta_get_common_ancestor_waiting_time(msp_t *self, population_id_t pop_id,
         /* An error occured, and we signal this back using a negative waiting time */
         result = ret;
     } else {
-        lambda *= 2; /* JK: Why do we multiply lambda by 2 here? */
         result = msp_get_common_ancestor_waiting_time_from_rate(self, pop, lambda);
     }
     return result;
@@ -6021,6 +6024,34 @@ out:
     return ret;
 }
 
+/* Rejection method. */
+static double
+ran_inc_beta_rej(gsl_rng *r, double a, double b, double x)
+{
+	double ret;
+	do {
+		ret = gsl_ran_beta(r, a, b);
+	} while(ret > x);
+	return ret;
+}
+
+/* Inverse transform sampling. */
+static double
+ran_inc_beta_its(gsl_rng *r, double a, double b, double upper_bound)
+{
+	double u = gsl_ran_flat(r, 0, upper_bound);
+	return gsl_cdf_beta_Pinv(u, a, b);
+}
+
+/* Choose between rejection or ITS based on the rejection probability. */
+static double
+ran_inc_beta(gsl_rng *r, double a, double b, double x, double upper_bound)
+{
+	if (upper_bound < 0.1)
+		return ran_inc_beta_its(r, a, b, upper_bound);
+	else
+		return ran_inc_beta_rej(r, a, b, x);
+}
 
 static int MSP_WARN_UNUSED
 msp_beta_common_ancestor_event(msp_t *self, population_id_t pop_id, label_id_t label)
@@ -6067,9 +6098,12 @@ msp_beta_common_ancestor_event(msp_t *self, population_id_t pop_id, label_id_t l
     }
     avl_init_node(q_node, y);
     q_node = avl_insert_node(&Q[4], q_node);
-    beta_x = gsl_ran_beta(self->rng,
+
+    beta_x = ran_inc_beta(self->rng,
                  2.0 - self->model.params.beta_coalescent.alpha,
-                 self->model.params.beta_coalescent.alpha);
+                 self->model.params.beta_coalescent.alpha,
+                 self->model.params.beta_coalescent.truncation_point,
+                 self->model.params.beta_coalescent.acceptance_rate);
 
     ret = msp_multi_merger_common_ancestor_event(self, beta_x, ancestors, Q);
     if (ret < 0) {
@@ -6505,6 +6539,7 @@ msp_set_simulation_model_beta(msp_t *self, double reference_size, double alpha,
     int ret = 0;
     double m;
     double scalar;
+    double acceptance_rate;
 
     /* Numerical instability occurs for values above 2 - 0.0005. */
     if (alpha <= 1.0 || alpha >= 2.0) {
@@ -6522,19 +6557,22 @@ msp_set_simulation_model_beta(msp_t *self, double reference_size, double alpha,
         goto out;
     }
 
-    m = 2.0 + exp(alpha * 0.6931472 + (1 - alpha) * 1.098612 - log(alpha - 1));
-    scalar = exp(log(alpha) - alpha * log(m)
-             + gsl_sf_beta_inc(2 - alpha, alpha, truncation_point)
-             - (alpha - 1) * reference_size);
+    m = 2.0 + exp(alpha * log(2) + (1 - alpha) * log(3) - log(alpha - 1));
+    acceptance_rate = gsl_sf_beta_inc(2 - alpha, alpha, truncation_point);
+    scalar = exp(log(alpha) - alpha * log(m) - (alpha - 1) * reference_size)
+             * acceptance_rate;
 
     assert(!isnan(m));
+    assert(!isnan(acceptance_rate));
     assert(!isnan(scalar));
     assert(!isinf(m));
+    assert(!isinf(acceptance_rate));
     assert(!isinf(scalar));
 
     self->model.params.beta_coalescent.alpha = alpha;
     self->model.params.beta_coalescent.truncation_point = truncation_point;
     self->model.params.beta_coalescent.scalar = scalar;
+    self->model.params.beta_coalescent.acceptance_rate = acceptance_rate;
 
     /* TODO we probably want to make these input parameters, as there will be situations
      * where integration fails and being able to tune them will be useful */
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -6508,12 +6508,12 @@ msp_set_simulation_model_beta(msp_t *self, double reference_size, double alpha,
 
     /* Numerical instability occurs for values above 2 - 0.0005. */
     if (alpha <= 1.0 || alpha >= 2.0) {
-        ret = MSP_ERR_BAD_PARAM_VALUE;
+        ret = MSP_ERR_BAD_ALPHA;
         goto out;
     }
 
     if (truncation_point <= 0.0 || truncation_point > 1.0) {
-        ret = MSP_ERR_BAD_PARAM_VALUE;
+        ret = MSP_ERR_BAD_TRUNCATION_POINT;
         goto out;
     }
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -5846,31 +5846,25 @@ out:
 static double
 beta_model_time_to_generations(simulation_model_t *model, double t)
 {
-    return 4 * model->reference_size * t;
+    return 4 * model->params.beta_coalescent.scalar * t;
 }
 
 static double
 beta_generations_to_model_time(simulation_model_t *model, double g)
 {
-    return g / (4 * model->reference_size);
+    return g / (4 * model->params.beta_coalescent.scalar);
 }
 
 static double
 beta_generation_rate_to_model_rate(simulation_model_t *model, double rate)
 {
-    return rate * 4 * model->reference_size;
+    return rate * 4 * model->params.beta_coalescent.scalar;
 }
 
 static double
 beta_model_rate_to_generation_rate(simulation_model_t *model, double rate)
 {
-    double alpha = model->params.beta_coalescent.alpha;
-    double m = model->params.beta_coalescent.m;
-    double phi = model->params.beta_coalescent.phi;
-    double scalar = exp(log(alpha) - alpha * log(m)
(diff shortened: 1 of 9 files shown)


grahamgower:
Currently fails in `test_gsl_error_handling_beta_coalescent` because it sets alpha=1e20. Without the !isnan(scalar) assertion, this fails another assertion in a more obscure location.

grahamgower:
It seems the CI is showing the more obscure location. Maybe there are other invalid parameter values are being passed in test_lowlevel.py?
```
(gdb) run -m nose -xvs --with-coverage --cover-package msprime --cover-branches tests/test_lowlevel.py
Starting program: /usr/bin/python -m nose -xvs --with-coverage --cover-package msprime --cover-branches tests/test_lowlevel.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Detaching after fork from child process 1164544]
test_state_machine_errors (tests.test_lowlevel.TestDemographyDebugger) ... ok
test_zero_events (tests.test_lowlevel.TestDemographyDebugger) ... ok
test_bad_tables (tests.test_lowlevel.TestLikelihood) ... ok
test_interface (tests.test_lowlevel.TestLikelihood) ... ok
test_simple_example (tests.test_lowlevel.TestLikelihood) ... ok
test_gsl_error_handler (tests.test_lowlevel.TestModule) ... ok
test_library_versions (tests.test_lowlevel.TestModule) ... ok
test_alphabet (tests.test_lowlevel.TestMutationGenerator) ... ok
test_basic_constructor (tests.test_lowlevel.TestMutationGenerator) ... ok
test_generate_interface (tests.test_lowlevel.TestMutationGenerator) ... ok
test_keep_mutations_block_size_ancestral_state (tests.test_lowlevel.TestMutationGenerator) ... ok
test_keep_mutations_block_size_derived_state (tests.test_lowlevel.TestMutationGenerator) ... ok
test_keep_mutations_block_size_metadata (tests.test_lowlevel.TestMutationGenerator) ... ok
test_keep_mutations_block_size_mutations (tests.test_lowlevel.TestMutationGenerator) ... ok
test_mutation_rate (tests.test_lowlevel.TestMutationGenerator) ... ok
test_rng (tests.test_lowlevel.TestMutationGenerator) ... ok
test_time_interval (tests.test_lowlevel.TestMutationGenerator) ... ok
test_constructor (tests.test_lowlevel.TestRandomGenerator) ... ok
test_seed (tests.test_lowlevel.TestRandomGenerator) ... ok
test_constructor (tests.test_lowlevel.TestRecombinationMap) ... ok
test_get_discrete (tests.test_lowlevel.TestRecombinationMap) ... ok
test_total_rate (tests.test_lowlevel.TestRecombinationMap) ... ok
test_demographic_events (tests.test_lowlevel.TestSimulationState) ... ok
test_event_by_event (tests.test_lowlevel.TestSimulationState) ... ok
test_random_sims (tests.test_lowlevel.TestSimulationState) ... ok
test_small_sims (tests.test_lowlevel.TestSimulationState) ... ok
test_bad_demographic_event_messages (tests.test_lowlevel.TestSimulator) ... ok
test_bad_demographic_event_types (tests.test_lowlevel.TestSimulator) ... ok
test_bad_demographic_event_values (tests.test_lowlevel.TestSimulator) ... ok
test_bad_migration_matrix (tests.test_lowlevel.TestSimulator) ... ok
test_bad_parameters (tests.test_lowlevel.TestSimulator) ... ok
test_bad_population_configurations (tests.test_lowlevel.TestSimulator) ... ok
test_bad_run_args (tests.test_lowlevel.TestSimulator) ... ok
test_bad_sample_configurations (tests.test_lowlevel.TestSimulator) ... ok
test_bad_samples (tests.test_lowlevel.TestSimulator) ... ok
test_beta_simulation_model (tests.test_lowlevel.TestSimulator) ... python: lib/msprime.c:3302: msp_initialise_from_samples: Assertion `k == self->num_sampling_events' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff7e1fce5 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff7e1fce5 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff7e09857 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff7e09727 in __assert_fail_base.cold () from /usr/lib/libc.so.6
#3  0x00007ffff7e18426 in __assert_fail () from /usr/lib/libc.so.6
#4  0x00007fffe2b26bbe in msp_initialise_from_samples (self=0x555555ed1050) at lib/msprime.c:3398
#5  msp_initialise (self=0x555555ed1050) at lib/msprime.c:3375
#6  0x00007fffe2b19d59 in Simulator_init (self=0x7fffe278c9f0, args=<optimized out>,
    kwds=<optimized out>) at _msprimemodule.c:3074
#7  0x00007ffff7b5a847 in ?? () from /usr/lib/libpython3.8.so.1.0
#8  0x00007ffff7b533a0 in PyObject_Call () from /usr/lib/libpython3.8.so.1.0
#9  0x00007ffff7c15d03 in _PyEval_EvalFrameDefault () from /usr/lib/libpython3.8.so.1.0
#10 0x00007ffff7c00e3b in _PyEval_EvalCodeWithName () from /usr/lib/libpython3.8.so.1.0
#11 0x00007ffff7c0224b in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#12 0x00007ffff7b5bde4 in ?? () from /usr/lib/libpython3.8.so.1.0
#13 0x00007ffff7b31a9f in ?? () from /usr/lib/libpython3.8.so.1.0
#14 0x00007ffff7c027a6 in ?? () from /usr/lib/libpython3.8.so.1.0
#15 0x00007ffff7b5bde4 in ?? () from /usr/lib/libpython3.8.so.1.0
#16 0x00007ffff7b318d0 in ?? () from /usr/lib/libpython3.8.so.1.0
#17 0x00007ffff7c0206d in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#18 0x00007ffff7b5bde4 in ?? () from /usr/lib/libpython3.8.so.1.0
#19 0x00007ffff7b31185 in ?? () from /usr/lib/libpython3.8.so.1.0
#20 0x00007ffff7c00e3b in _PyEval_EvalCodeWithName () from /usr/lib/libpython3.8.so.1.0
#21 0x00007ffff7c0224b in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#22 0x00007ffff7c02a7b in ?? () from /usr/lib/libpython3.8.so.1.0
#23 0x00007ffff7b5330d in PyObject_Call () from /usr/lib/libpython3.8.so.1.0
#24 0x00007ffff7c15d03 in _PyEval_EvalFrameDefault () from /usr/lib/libpython3.8.so.1.0
#25 0x00007ffff7c00e3b in _PyEval_EvalCodeWithName () from /usr/lib/libpython3.8.so.1.0
#26 0x00007ffff7c0224b in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#27 0x00007ffff7ba0f14 in _PyObject_FastCallDict () from /usr/lib/libpython3.8.so.1.0
#28 0x00007ffff7ba1143 in _PyObject_Call_Prepend () from /usr/lib/libpython3.8.so.1.0
#29 0x00007ffff7c60d99 in ?? () from /usr/lib/libpython3.8.so.1.0
#30 0x00007ffff7b5bad2 in _PyObject_MakeTpCall () from /usr/lib/libpython3.8.so.1.0
#31 0x00007ffff7b318d0 in ?? () from /usr/lib/libpython3.8.so.1.0
#32 0x00007ffff7c0206d in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#33 0x00007ffff7b5bde4 in ?? () from /usr/lib/libpython3.8.so.1.0
#34 0x00007ffff7b31185 in ?? () from /usr/lib/libpython3.8.so.1.0
#35 0x00007ffff7c0206d in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#36 0x00007ffff7c02a7b in ?? () from /usr/lib/libpython3.8.so.1.0
#37 0x00007ffff7b5330d in PyObject_Call () from /usr/lib/libpython3.8.so.1.0
#38 0x00007ffff7c15d03 in _PyEval_EvalFrameDefault () from /usr/lib/libpython3.8.so.1.0
#39 0x00007ffff7c00e3b in _PyEval_EvalCodeWithName () from /usr/lib/libpython3.8.so.1.0
#40 0x00007ffff7c0224b in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#41 0x00007ffff7ba0f14 in _PyObject_FastCallDict () from /usr/lib/libpython3.8.so.1.0
#42 0x00007ffff7ba1143 in _PyObject_Call_Prepend () from /usr/lib/libpython3.8.so.1.0
#43 0x00007ffff7c60d99 in ?? () from /usr/lib/libpython3.8.so.1.0
#44 0x00007ffff7b5bad2 in _PyObject_MakeTpCall () from /usr/lib/libpython3.8.so.1.0
#45 0x00007ffff7b318d0 in ?? () from /usr/lib/libpython3.8.so.1.0
#46 0x00007ffff7c0206d in _PyFunction_Vectorcall () from /usr/lib/libpython3.8.so.1.0
#47 0x00007ffff7c02a7b in ?? () from /usr/lib/libpython3.8.so.1.0
#48 0x00007ffff7b5330d in PyObject_Call () from /usr/lib/libpython3.8.so.1.0
--Type <RET> for more, q to quit, c to continue without paging--q
Quit
(gdb) frame 4
#4  0x00007fffe2b26bbe in msp_initialise_from_samples (self=0x555555ed1050) at lib/msprime.c:3398
3398        return ret;
(gdb) p ret
$1 = 0
(gdb) p self->model
$2 = {type = 3, reference_size = 2, params = {beta_coalescent = {alpha = 1,
      truncation_point = 0.10000000000000001, m = inf, scalar = 0,
      integration_epsabs = 0, integration_epsrel = 0.001,
      integration_workspace_size = 1000, integration_workspace = 0x555555efd780},
    dirac_coalescent = {psi = 1, c = 0.10000000000000001}, sweep = {locus = 1,
      trajectory_params = {genic_selection_trajectory = {
--Type <RET> for more, q to quit, c to continue without paging--
          000000001, end_frequency = inf, alpha = 0, dt = 0}},
      generate_trajectory = 0x3f50624dd2f1a9fc, print_state = 0x3e8}},
  free = 0x7fffe2b1e0f0 <beta_model_free>,
  model_time_to_generations = 0x7fffe2b1d470 <beta_model_time_to_generations>,
  generations_to_model_time = 0x7fffe2b1d490 <beta_generations_to_model_time>,
  generation_rate_to_model_rate = 0x7fffe2b1d4b0 <beta_generation_rate_to_model_rate>,
  model_rate_to_generation_rate = 0x7fffe2b1e9f0 <beta_model_rate_to_generation_rate>}
```

jeromekelleher:
Thanks @grahamgower. @JereKoskela and @TPPSellinger, can you help Graham with this please?

TPPSellinger:
I will look at it and let you know if I find anything.

grahamgower:
Perhaps I shouldn't have naiively jumped in here without understanding the basic idea. So, uh, what is this beta coalescent model? I've been looking through Schweinsberg (2003), and wondering if this code is just intended for `1<a<2`. (I'm assuming that `a` in the manuscript is `alpha` in the msprime code, but I could be wrong here too)

jeromekelleher:
I'm hazy on the details here @grahamgower, hopefully Thibaud and Jere can help. Thanks for jumping in though, good to get this moving!

Re the aborts, I'm guessing we're in some weird part of parameter space that we should be raising an error earlier on for.

TPPSellinger:
Yes, a (or alpha) should be >1 and <2 . If alpha equals to 1 it's a different model (Bolthausen-Sznitman) and if  alpha = 2 it's the "normal" Kingman model. Yes we should check alpha values earlier. An error should be set when alpha <=1  and when alpha=2 we should simply jump back to the kingman (if you force Beta coalescent with alpha = 2, you will most likeli face definition errors).

grahamgower:
Excellent! Thanks @TPPSellinger! So does this mean that the dirac coalescent model is for 0<a<1?

TPPSellinger:
Sorry for late answer. I don't think there should be an a (or alpha ) for the dirac coalescent.  Otherwise it's a strange notation choice or error in the script. Parameters for the Xi_dirac should be psi and c. They should both be positive. Psi correspond to a probability (so I guess it should be between 0 and 1) of individual participating in a multiple merger event (if I understood the underlying model correctly). c is link to the probability of multiple merger events (but is not a probability, at least in the paper). If I understood correctly and it's implemented like described in the paper, c should be between 0 and N² (Where N is the population size). So you can see psi as the proportion of individual participating in a multiple merger event, and c gives you the probability of multiple merger event (which should be c/N²). However, you should ask Bjarki or Jere if you want a better answer. You can find (way) more details in Birkner et at (2013).

jeromekelleher:
Possibly it's better if someone closer to the mutiple merger world takes the lead on this @grahamgower, as the model we're simulating hasn't been published yet (hence the lack of details). This is a great start though, and should provide a good basis for sorting these issues out.

grahamgower:
Yeah, no problem @jeromekelleher. I'll leave it here. FYI, I was able to provoke gsl integration errors for values between 2.0-0.0005 and 2.0.

grahamgower:
I coulnd't help myself, I think I got the last one fixed. Definitely stopping now.

JereKoskela:
Thanks @grahamgower, this looks really helpful! Like @TPPSellinger said above, alpha needs to be in (1, 2), and corresponds to the `a` in Schweinsberg (2003). The `a <= 1` and `a = 2` cases play no role in the model implemented here. 

The merger rate of the Beta-coalescent model diverges as alpha approaches two. That's what you're seeing as gsl_integration errors, and it is great to get some idea of how close we can get before that starts happening.

The Dirac coalescent is formally unconnected to the Beta one. They are parametrised differently, and scale in different ways obtained from different pre-limiting sequences of Cannings models.

jeromekelleher:
Excellent, thanks @grahamgower! Looks great to me. @JereKoskela, can you review please?

@TPPSellinger, also good to get your thoughts on the fix please.

TPPSellinger:
Hi. I quickly checked. I think the fix is good and useful. Should we do something similar for the xi_dirac coalescent ?

jeromekelleher:
> Hi. I quickly checked. I think the fix is good and useful. Should we do something similar for the xi_dirac coalescent ?

Yes, this would be very useful, I think. Are you volunteering? :wink:

TPPSellinger:
Sure I can try to do it, except if @grahamgower is already working on it?

grahamgower:
I've not yet done the equivalent for the dirac model.

Before this is merged though, I have two more questions.

1. The truncation_point (phi) is only used for the incomplete Beta function to calculate the scaling factor for the time and rate conversion functions. Should the truncation_point also be used as an upper limit of integration in `msp_compute_beta_integral()`? What about the random variable drawn from a Beta distribution in `msp_beta_common_ancestor_event()` (see call to `gsl_ran_beta()`)?

2. In #186, there is a suggestion that the recombination rate ought to be scaled differently to the rate of common ancestor events. I assume this comment also applies to the Xi_beta coalescent? Does this mean the main event loop in `msp_run_coalescent()` should choose the waiting time until recombination by scaling the recombination rate differently than the time/rate conversion functions I've changed in this PR?

JereKoskela:
> 1. The truncation_point (phi) is only used for the incomplete Beta function to calculate the scaling factor for the time and rate conversion functions. Should the truncation_point also be used as an upper limit of integration in `msp_compute_beta_integral()`? What about the random variable drawn from a Beta distribution in `msp_beta_common_ancestor_event()` (see call to `gsl_ran_beta()`)?

I think yes to both. The random variable generation is probably easiest to do by drawing from the corresponding Beta density, and rejecting realisations above the threshold. I can't seem to tag @eldonb, but it would be good to get his input to make sure. I'll email him to point out the thread.

> 2. In #186, there is a suggestion that the recombination rate ought to be scaled differently to the rate of common ancestor events. I assume this comment also applies to the Xi_beta coalescent? Does this mean the main event loop in `msp_run_coalescent()` should choose the waiting time until recombination by scaling the recombination rate differently than the time/rate conversion functions I've changed in this PR?

No, this definitely doesn't apply here. In fact, I'm not sure it's right in #186 either, but I'll take that conversation there so as to not mix up unrelated issues.

lib/msprime.c
```diff
@@ -6375,19 +6367,38 @@ msp_set_simulation_model_beta(msp_t *self, double reference_size, double alpha,
...
-    /* TODO bounds check alpha and truncation_point: what are legal values? */
     ret = msp_set_simulation_model(self, MSP_MODEL_BETA, reference_size);
     if (ret != 0) {
         goto out;
     }
+
+    m = 2.0 + exp(alpha * 0.6931472 + (1 - alpha) * 1.098612 - log(alpha - 1));
+    scalar = exp(log(alpha) - alpha * log(m)
+             + gsl_sf_beta_inc(2 - alpha, alpha, truncation_point)
+             - (alpha - 1) * reference_size);
```
JereKoskela: I think this is a mistake inherited from earlier versions, but this doesn't look quite right. The RHS should be `exp(log(alpha) - alpha * log(m) - (alpha - 1) * log(reference_size)) * gsl_sf_beta_inc(2 - alpha, alpha)`.

JereKoskela: I think It would also make sense to change all the hard-coded decimal expressions to their respective exact formulas, e.g. `0.6931472` is really `log(2)`, `1.098612` is really `log(3)`, and so on. I think `log(4)` appears somewhere similarly too. I'm not sure why they've been written this way as it is, but if we want this to be stable for the greatest possible range of alphas, then they should be as accurate as (reasonably) possible. I think the `log()` versions are also more readable.

jeromekelleher: Totally agree on putting in log(2) etc @JereKoskela, putting in constants is a premature optimisation (and an unreadable one).

jeromekelleher:
Thanks for the input @JereKoskela, and the coding effort here @grahamgower. If you don't have time/don't want to implement the latest batch of comments we can get this much merged and log them as issues. Just let me know your preference.

grahamgower:
Thanks very much @JereKoskela, I'll work through your requested changes.

@jeromekelleher If you want to merge this pr, I can address the other suggestions later in the week. I'll leave the analogous Xi_dirac changes to @TPPSellinger.

lib/msprime.c
grahamgower: Agreed regarding the log()s. The log(4) is hiding inside `beta_integrand()`. If optimisation is really important (I doubt it in this case), we could add our own #defines for log(3) and log(4) (math.h has M_LOG2E for log(2)).

grahamgower: FYI, log is usually implemented as a lookup table, and I'm guessing the compiler can inline the function call. So, I'd be surprised if log(constant) required any explicit optimisation.

jeromekelleher: Totally agree @grahamgower, any compiler worth its salt is going to pre-compute these.

jeromekelleher:
OK, well there's no point in merging this yet if we have known problems with it. Ping me when it's ready to go @grahamgower (no particular rush).

@TPPSellinger, are you happy going through the changes for dirac in another PR?

TPPSellinger:
Yes, I will try to do analogous changes for the xi_dirac. But because of schedule I might not be able to do it before Friday.

grahamgower:
Regarding the generation of random beta variables, I've done a quick experiment to compare the rejection method (ran_inc_beta1 below) against using inverse transform sampling (ran_inv_beta2 below). Suprisingly to me, the inverse transform sampling is rather slow. Probably `gsl_ran_beta()` is well optimised, and it seems the inverse cdf function `gsl_cdf_beta_Pinv()` is iterating on something until convergence. Actually, with alpha>=1.9, truncation_point=not tiny, `gsl_cdf_beta_Pinv` fails to converge at all.

But I worry about regions of the (alpha, truncation_point) parameter space where rejection rates will be high. The acceptance probability is just `gsl_sf_beta_inc(2 - alpha, alpha, truncation_point)`, the same as the upper bound for the uniform variate in ran_inc_beta2. Luckily, 
`gsl_cdf_beta_Pinv()` is stable when the rejection rate is high.

Maybe this region of parameter space isn't so interesting? In any event, I think we should choose between either function at an appropriate tipping point (I've chosen 10% acceptance rate below, which seems about right based on eyeballing the timings).

```c
/* Rejection method. */
double
ran_inc_beta1(gsl_rng *r, double a, double b, double x)
{
	double ret;
	do {
		ret = gsl_ran_beta(r, a, b);
	} while(ret > x);
	return ret;
}

/* Uniform random variable with inverse CDF. */
double
ran_inc_beta2(gsl_rng *r, double a, double b, double x, double upper_bound)
{
	double u = gsl_ran_flat(r, 0, upper_bound);
	return gsl_cdf_beta_Pinv(u, a, b);
}

/* Choose 1 or 2 above, depending on the acceptance rate. */
double
ran_inc_beta3(gsl_rng *r, double a, double b, double x, double upper_bound)
{
	if (upper_bound < 0.1)
		return ran_inc_beta2(r, a, b, x, upper_bound);
	else
		return ran_inc_beta1(r, a, b, x);
}

...

// acceptance rate, or upper bound on the uniform random variable
upper_bound = gsl_sf_beta_inc(2 - alpha, alpha, truncation_point);
```

Full test code attached. Some example output below for 10000 draws from each function, where t is clock cycle count (lower is better).
```
$ ./ran_inc_beta 1.9 .1
acceptance rate, F(truncation_point; 2 - alpha, alpha): 0.861018
1: mean=0.00858891, var=0.00036654, min=1.00513e-39, max=0.0999581, t=151
2: mean=nan, var=nan, min=nan, max=nan, t=15548
3: mean=0.00842383, var=0.000359212, min=2.23275e-40, max=0.0999427, t=156
Correlation of quantiles for 1 and 2: pearson=nan
$ ./ran_inc_beta 1.5 .1
acceptance rate, F(truncation_point; 2 - alpha, alpha): 0.395819
1: mean=0.0332165, var=0.000883012, min=1.22194e-08, max=0.099988, t=380
2: mean=0.0331753, var=0.000877041, min=4.06719e-09, max=0.0999756, t=2992
3: mean=0.0330672, var=0.000885674, min=1.79538e-10, max=0.099973, t=357
Correlation of quantiles for 1 and 2: pearson=0.999972
$ ./ran_inc_beta 1.5 1
acceptance rate, F(truncation_point; 2 - alpha, alpha): 1
1: mean=0.249387, var=0.062933, min=1.39891e-10, max=0.998615, t=161
2: mean=0.248843, var=0.0621535, min=1.79028e-08, max=0.998531, t=3808
3: mean=0.248068, var=0.0636976, min=1.52632e-08, max=0.997572, t=135
Correlation of quantiles for 1 and 2: pearson=0.999946
$ ./ran_inc_beta 1.01 0.1
acceptance rate, F(truncation_point; 2 - alpha, alpha): 0.103293
1: mean=0.0497672, var=0.000838226, min=7.90814e-07, max=0.0999973, t=1403
2: mean=0.0502882, var=0.000828112, min=1.50764e-05, max=0.0999819, t=1700
3: mean=0.0495294, var=0.000827957, min=1.16442e-05, max=0.0999997, t=1215
Correlation of quantiles for 1 and 2: pearson=0.999931
$ ./ran_inc_beta 1.01 0.01
acceptance rate, F(truncation_point; 2 - alpha, alpha): 0.0105748
1: mean=0.00497946, var=8.22856e-06, min=3.22193e-06, max=0.00999925, t=11758
2: mean=0.0049212, var=8.38085e-06, min=2.23145e-06, max=0.00999853, t=1466
3: mean=0.00494175, var=8.2637e-06, min=2.77973e-07, max=0.00999817, t=1478
Correlation of quantiles for 1 and 2: pearson=0.999813
```


[ran_inc_beta.c.txt](https://github.com/tskit-dev/msprime/files/4260728/ran_inc_beta.c.txt)

lib/msprime.c
```diff
@@ -5954,6 +5949,34 @@ msp_multi_merger_common_ancestor_event(msp_t *self, double x,
     return ret;
 }
 
+/* Rejection method. */
```
jeromekelleher: Looks good @grahamgower. The sake of legibility, I think it'd be good to put these functions towards the top of the file with some other general utilities, and to add a quick docstring-like comment about what each function does.

lib/msprime.c
```diff
@@ -5954,6 +5949,34 @@ msp_multi_merger_common_ancestor_event(msp_t *self, double x,
...
+static double
+ran_inc_beta_rej(gsl_rng *r, double a, double b, double x)
+{
+	double ret;
```
jeromekelleher: The indentation is off here: should be 4 spaces.

lib/msprime.c
```diff
@@ -5954,6 +5949,34 @@ msp_multi_merger_common_ancestor_event(msp_t *self, double x,
...
+{
+	double u = gsl_ran_flat(r, 0, upper_bound);
+	return gsl_cdf_beta_Pinv(u, a, b);
+}
+
+/* Choose between rejection or ITS based on the rejection probability. */
+static double
+ran_inc_beta(gsl_rng *r, double a, double b, double x, double upper_bound)
+{
+	if (upper_bound < 0.1)
```
jeromekelleher: Style issue: we always use braces here.

lib/msprime.c
```diff
@@ -5839,8 +5841,8 @@ beta_integrand(double x, void *p)
     for (l = 0; l <= m; l++)  {
         ret += exp(r[l] - r_max);
     }
-    ret = 1 - exp(r_max + log(ret));
-    ret *= exp((-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x);
+    exponent = (-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x;
```
JereKoskela: Can we not add `log(4) - gsl_sf_lnbeta(2 - alpha, alpha)` to `exponent` and delete line 5846? That is probably more stable for `alpha` close to 2 since it avoids computing a near-singular beta function.

lib/msprime.c
```diff
@@ -41,6 +42,51 @@
...
+ * the inverse CDF. Truncation is obtained by defining the upper bound
+ * of the uniform variable using the incomplete beta function. I.e.
+ *      upper_bound = gsl_sf_beta_inc(a, b, x),
+ * where x is the truncation point.
+ */
+static double
+ran_inc_beta_its(gsl_rng *r, double a, double b, double upper_bound)
+{
+    double u = gsl_ran_flat(r, 0, upper_bound);
+    return gsl_cdf_beta_Pinv(u, a, b);
```
JereKoskela: Is it obvious that this is the right thing to do? A naive version of the inverse CDF method would use a `u ~ U(0,1)` uniform variate, and the CDF

`F(x) = B(x; a, b) / B(T; a, b)`,

where `B(x; a, b)` is the incomplete Beta function with parameters `a` and `b`, and `T` is the truncation point. Setting `F(x) = u` and doing a little algebra gives

`B(x; a, b) / B(1; a, b) = u B(T; a, b) / B(1; a, b)`,

which suggests that the right way to sample is 

`double u = gsl_rng_uniform(gen) * gsl_sf_beta_inc(alpha, beta, T) / gsl_sf_beta(alpha, beta);
`return gsl_cdf_beta_Pinv(u , alpha, beta);`

Is your way equivalent? If it is, it is definitely simpler.

JereKoskela:
Looks good @grahamgower, thanks! I've made a few minor comments above. And sorry for how long it has taken to look over your code.

lib/msprime.c
grahamgower: I followed the approach by Wikipedia et al. https://en.wikipedia.org/wiki/Inverse_transform_sampling#Truncated_distribution

I generated 10000 samples for each of `ran_inc_beta_rej` and `ran_inc_beta_its`, sorted the samples, and found their pearson correlation was >0.999 for a few different values of alpha and beta. See the ran_inc_beta.c.txt attached in an earlier comment.

lib/msprime.c
grahamgower: Yep, done.

lib/msprime.c
grahamgower: I think gsl_sf_beta_inc is already normalised. In which case your suggestion and my code are equivalent.
https://www.gnu.org/software/gsl/doc/html/specfunc.html#c.gsl_sf_beta_inc

JereKoskela: Ah, you're right. That's what I was missing, and makes your way obviously better. Thanks!

grahamgower:
Thanks for the detailed reviews @JereKoskela. I think this is good to merge now @jeromekelleher. I've just rebased. Do you want me to squash any/all of the commits?

lib/msprime.c
jeromekelleher: I feel like framing this discussion as an example of "How Scientific Software Development Should Be Done"!

jeromekelleher:
Superb stuff, thanks @grahamgower and @JereKoskela! Really makes me happy to see people get into the serious detail here!

PR #905, opened 2020-02-21
Fix memory and Fenwick tree updates in dtwf_recombine

daniel-goldstein:
The current implementation of a DTWF generation is not entirely consistent between Python and C but overall they both perform the following:
- remove all the individuals from the population
- perform recombination events (`dtwf_recombine`)
- merge lineages (`merge_ancestors`)
     - `merge_ancestors` adds remaining segments back into the population

`dtwf_recombine` was creating segments whose left ends pointed to stale sentinel values on the stack and was not properly updating the Fenwick tree to track the weights of those segments. This went unnoticed because these segments were never inserted into the population. `merge_ancestors` would reconstruct all segments (sometimes needlessly) and erase any evidence of the bad intermediate state.

To mend the intermediate state and remove the implicit dependency of `dtwf_recombine` on `merge_ancestors` (and allow us to easily avoid merging in unnecessary cases), we:
- Don't remove individuals from the population until they are about to be merged
- Update the Fenwick tree values immediately during `dtwf_recombine`
- Leave recombine in a verifiable state*

*This is not _exactly_ true because we still need to add the potentially new segment that results from `dtwf_recombine` to the population, and the priority queue used to track children causes the checks about avl node allocations to fail. It might be nice to add/remove individuals in recombine/merge where their segments are allocated/freed so that there isn't this temporary misalignment. For now I kept the population management in `dtwf_generation` because otherwise it could interfere with other methods that use `dtwf_recombine`/`merge_ancestors`.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -680,6 +680,12 @@ msp_free_segment(msp_t *self, segment_t *seg)
     fenwick_set_value(&self->links[seg->label], seg->id, 0);
 }
 
+static inline avl_tree_t *
+msp_get_segment_population(msp_t *self, segment_t *u)
+{
+    return &self->populations[u->population_id].ancestors[u->label];
+}
+
 static inline int MSP_WARN_UNUSED
 msp_insert_individual(msp_t *self, segment_t *u)
 {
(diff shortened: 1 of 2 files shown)


lib/msprime.c
```diff
@@ -674,6 +674,12 @@ msp_free_segment(msp_t *self, segment_t *seg)
     fenwick_set_value(&self->links[seg->label], seg->id, 0);
 }
 
+static inline avl_tree_t *
+msp_get_segment_population(msp_t *self, segment_t *u)
```
jeromekelleher: :+1:

DomNelson:
Sorry guys, getting to this. Lots on my plate at the moment.

jeromekelleher:
OK, going to merge this. Thanks @daniel-goldstein!

PR #906, opened 2020-02-21
Store C recombination map in model rates

### LINKED ISSUE #881: [C Code] Handling of rate conversions
In the C code we currently have a mix of methods for handling rates. Pre #862, all rates were converted from generations into model time when the model is set. Currently we don't do this for recombination, and the recombination 'masses' stored in the Fenwick tree are in generation rates, and we convert to model rate in the main loop [here](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L5649).

This is OK, but somewhat and a little inefficient maybe. It seems better to convert the recombination rates to model time along with all the other rates, but this would involve editing the recombination map, which is provided as input (and therefore we shouldn't modify it).

An easy fix would be to take a copy of the recombination map in the simulator, and to update the rates in that we we see fit. The extra memory and copy time will be negligible and it should keep things a bit simpler.

Unless some objects? @fbaumdicker, @petrelharp, @andrewkern, @daniel-goldstein?

daniel-goldstein:
Copy the given recombination map when allocating the C simulator. Also copy/free when switching models because the recombination map is stateful and would have to be re-initialized anyway.

Resolves #881.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -247,10 +247,11 @@ int
 msp_set_gene_conversion_rate(msp_t *self, double rate, double track_length)
 {
     int ret = 0;
+    double sequence_length = recomb_map_get_sequence_length(&self->recomb_map);
 
     /* if the rate is zero, we ignore the track length */
     if (rate > 0) {
-        if (track_length < 0 || track_length > self->recomb_map->sequence_length) {
+        if (track_length < 0 || track_length > sequence_length) {
             ret = MSP_ERR_BAD_PARAM_VALUE;
             goto out;
         }
@@ -420,15 +421,19 @@ msp_alloc(msp_t *self,
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
     }
+    ret = recomb_map_copy(&self->recomb_map, recomb_map);
+    if (ret != 0) {
+        goto out;
+    }
 
     /* Use the standard coalescent with coalescent time units by default. */
     self->model.type = -1;
     ret = msp_set_simulation_model_hudson(self, 0.25);
     assert(ret == 0);
     self->rng = rng;
-    self->recomb_map = recomb_map;
+
     self->tables = tables;
-    self->sequence_length = self->recomb_map->sequence_length;
+    self->sequence_length = recomb_map_get_sequence_length(&self->recomb_map);
 
     /* And rescale the rates */
     self->gene_conversion_rate = self->model.generation_rate_to_model_rate(
@@ -473,7 +478,7 @@ msp_alloc(msp_t *self,
             ret = msp_set_tsk_error(ret);
             goto out;
         }
-        if (self->recomb_map->sequence_length != self->tables->sequence_length) {
+        if (self->sequence_length != self->tables->sequence_length) {
             ret = MSP_ERR_INCOMPATIBLE_FROM_TS;
             goto out;
         }
@@ -598,6 +603,7 @@ msp_free(msp_t *self)
     /* free the object heaps */
     object_heap_free(&self->avl_node_heap);
     object_heap_free(&self->node_mapping_heap);
+    recomb_map_free(&self->recomb_map);
     if (self->from_ts != NULL) {
         tsk_treeseq_free(self->from_ts);
         free(self->from_ts);
@@ -768,16 +774,16 @@ msp_verify_segments(msp_t *self, bool verify_breakpoints)
                     assert(u->left < u->right);
                     assert(u->right <= self->sequence_length);
 
-                    l_mass = recomb_map_position_to_mass(self->recomb_map, u->left);
-                    r_mass = recomb_map_position_to_mass(self->recomb_map, u->right);
+                    l_mass = recomb_map_position_to_mass(&self->recomb_map, u->left);
+                    r_mass = recomb_map_position_to_mass(&self->recomb_map, u->right);
                     assert(u->left_mass == l_mass);
                     assert(u->right_mass == r_mass);
                     if (u->prev != NULL) {
                         s = recomb_map_mass_between(
-                                self->recomb_map, u->prev->right, u->right);
+                                &self->recomb_map, u->prev->right, u->right);
                     } else {
                         s = recomb_map_mass_between_left_exclusive(
-                                self->recomb_map, u->left, u->right);
+                                &self->recomb_map, u->left, u->right);
                     }
                     ss = fenwick_get_value(&self->links[k], u->id);
                     total_mass += ss;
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
@jeromekelleher Since the simulator struct now owns the recombination map, should it directly contain an instance of the recombination map instead of a pointer? Then we wouldn't have to malloc the map itself and I think that's more standard?

daniel-goldstein:
Need tests for new recomb map methods

jeromekelleher:
> @jeromekelleher Since the simulator struct now owns the recombination map, should it directly contain an instance of the recombination map instead of a pointer? Then we wouldn't have to malloc the map itself and I think that's more standard?

Yes, good call.

lib/msprime.c
```diff
@@ -406,6 +406,13 @@ msp_alloc_segment(msp_t *self, double left, double right,
     return seg;
 }
 
+static void
+msp_free_recomb_map(msp_t *self)
+{
+    recomb_map_free(self->recomb_map);
```
jeromekelleher: I think we can just have ``recomb_map_free(&self->recomb_map)`` in msp_free. Once the struct has been zero'd this is safe (and that's the first thing we do in msp_alloc).

lib/msprime.c
```diff
@@ -504,6 +520,9 @@ msp_alloc(msp_t *self,
     /* Set up pedigree */
     self->pedigree = NULL;
 out:
+    if (ret != 0 && self->recomb_map != NULL) {
```
jeromekelleher: This shouldn't be necessary - any cleanup gets done in msp_free.

lib/msprime.c
daniel-goldstein: Is it expected that if `msp_alloc` fails then a client should call `msp_free`?

lib/msprime.c
daniel-goldstein: Is it expected that `msp_free` should always be called should `msp_alloc` fail?

daniel-goldstein: That is currently not the case in the C unit tests.

jeromekelleher: > Is it expected that msp_free should always be called should msp_alloc fail

Yes, that's the convention we ended up with in tskit --- the cleanup is too messy otherwise. Where do we do otherwise in the C unit tests? (Might be quicker to call down to talk this one through)

lib/msprime.c
```diff
@@ -6149,6 +6157,77 @@ genic_selection_print_state(sweep_t *self, FILE *out)
...
+    recomb_map_t new_map;
+
+    rates = malloc(recomb_map_get_size(&self->recomb_map) * sizeof(double));
+    if (rates == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    recomb_map_get_rates(&self->recomb_map, rates);
+    for (i = 0; i < recomb_map_get_size(&self->recomb_map); i++) {
```
jeromekelleher: I think we can simplify this function a lot:
```c
 for (i = 0; i < recomb_map_get_size(&self->recomb_map); i++) {
     self->recomb_map.rates[i] = convert(&self->model, self->recomb_map.rates[i]);
}
```

If breaking the abstractions and reaching between objects is too nasty for you, then you could have a method like
```c
recomb_map_change_rates(recomb_map_t *self,  double convert (void *obj, double rate))
```
which would probably work.

Doing a bunch of mallocing seems overboard to me though.

daniel-goldstein: Ya that seems fine. Would also have to change the cumulative mass in the map too, but I think that's it

daniel-goldstein: would you also pass in the model as a parameter `obj` there?

jeromekelleher: Yeah, the idea is that we can use a generic pointer here rather than insisting it's of type model_t  or whatever. This is a slightly more general interface to make it easier to test and not quite so coupled to the rest of the code.

I *think* it'll work, but if not just use the actual type.

daniel-goldstein: That worked! I was trying to think of how to make a closure when I first tackled it because it felt weird to pass a model. This is neat.

lib/recomb_map.c
```diff
@@ -228,6 +241,16 @@ recomb_map_sample_poisson(recomb_map_t *self, gsl_rng *rng, double start)
...
+void
+recomb_map_change_rates(recomb_map_t *self,
+        double convert (void *obj, double rate), void *obj)
+{
+    size_t i;
+    for (i = 0; i < self->size; i++) {
+        self->rates[i] = convert(obj, self->rates[i]);
+    }
+    recomb_map_init_cumulative_recomb_mass(self);
+}
```
jeromekelleher: Nice!

lib/recomb_map.c
```diff
@@ -228,6 +241,16 @@ recomb_map_sample_poisson(recomb_map_t *self, gsl_rng *rng, double start)
     return recomb_map_shift_by_mass(self, left_bound, mass_to_next_recomb);
 }
 
+void
+recomb_map_change_rates(recomb_map_t *self,
```
jeromekelleher: I guess to make this as readable as we can, we should change this to ``recomb_map_convert_rates``. Then, in msprime.h we do 
```
typedef msp_convert_func (double convert (void *obj, double rate))
```
(Or whatever it is; I can never remember the syntax). Then we have 
```
void
￼recomb_map_change_rates(recomb_map_t *self, msp_convert_func convert, void *obj)
```

jeromekelleher:
Looks good thanks. Ready to merge after a quick rebase.

jeromekelleher:
Huh, that failure is annoying. I'll tack on a commit here to fix it and merge. Thanks @daniel-goldstein.

jeromekelleher:
Actually, no, maybe the test is right. We've changed the values within the Fenwick tree @daniel-goldstein, but I don't think we've updated the code in ``msp_run_sweep``. Don't we need to do some correction here? It's perfectly possible this was wrong before, and so I guess the right thing to do then is to unittest.skip the test and add an issue to the sweeps project for it.

daniel-goldstein:
Huh, I'll look at sweep to see if there's something obvious. Good thing that test went in!

lib/msprime.c
```diff
@@ -4267,7 +4275,8 @@ msp_run_sweep(msp_t *self)
             label = (label_id_t) j;
             recomb_mass = fenwick_get_total(&self->links[label]);
             sweep_pop_sizes[j] = avl_count(&self->populations[0].ancestors[label]);
-            rec_rates[j] = recomb_mass;
+            rec_rates[j] = self->model.model_rate_to_generation_rate(
```
daniel-goldstein: Looks like sweeps are done in generation time. Whether they should be I'm not sure but this line was a quick way of fixing the mismatch. Looks like we replaced one issue with another!

jeromekelleher: We definitely shouldn't be doing sweeps in generations, so I think making this change makes the code more wrong. I think the least-wrong thing to do is revert, skip the offending test (which is tuned in a parameter region when the code was more wrong), and open an issue to get the test into a sensible parameter region.

daniel-goldstein: Ok, I'll revert and skip.

Issue #902, opened 2020-02-18
Modernise RecombinationMap interface


jeromekelleher:
The RecombinationMap interface is clunky, using a lot of Java style 'getter' methods. Modernise this to use properties, where appropriate.

We also want to use numpy arrays natively in the low-level interface (we currently use lists). So, we should be able to use ``recomb_map.position`` to get the position array and ``recomb_map.rate``  to get the rate array (we should use the singular following the conventions in tskit).

jeromekelleher:
Add tests for this to the TestRecombinationMap class in tests/test_highlevel.py

jeromekelleher:
I tried to assign this to you @grahamgower, but you weren't on the list. Not sure what's up with GitHub.

hyanwong:
The 4 line numpy code that I posted in https://github.com/tskit-dev/msprime/issues/919#issue-584423306 and which is essentially used in the unit testing basically describes the behaviour that I would expect - being able to pass in a numpy array of physical positions and get out a numpy array of genetic positions. I suspect this is likely to be a common use case (e.g. I'm passing in all the site positions from a TS)

jeromekelleher:
See also #920 for proposal on more generic interface for rate maps. Pinging @daniel-goldstein for input here.

daniel-goldstein:
As a final product I think this most naturally fits as a method on interval rate maps that recombination map can directly wrap. Once we can access the underlying arrays as numpy arrays we have the following in an interval rate map:
```
def position_to_cumulative_rate(input_pos):
    return np.interp(input_pos, self.position, self.cumulative)
```

which takes care of both the single value and the array of values cases.

jeromekelleher:
Closing in favour of #1172

Issue #666, opened 2018-12-20
Convert _msprimemodule.c to use numpy C API where possible


jeromekelleher:
A lot of the code in _msprimemodule.c is concerned with parsing input. We can streamline this and make it much more efficient by using numpy instead. For example, functions like ``convert_integer_list``, ``convert_float_list`` etc can all be changed.  We can also probably update ``parse_samples`` to use numpy record array without losing backwards compatability.

The RecombinationMap class in particular should be updated, as this would be far simpler if we used numpy.

jeromekelleher:
See also #902

grahamgower:
You can assign this and #902 to me.

jeromekelleher:
Done! Thanks @grahamgower .

jeromekelleher:
I think this is basically done now. The one thing I spotted in the code was the ``migration_matrix`` argument to simulate could be changed to a 2D array, which would probably help us shed some code. Would you like to take this on @grahamgower. I'm happy to do it.

Should be done after #946

grahamgower:
I remember looking at it and thinking that the current `Simulator_parse_migration_matrix()` wouldn't be simplified much, so it wouldn't be worth the effort unless 2d numpy arrays were also returned from `Simulator_get_migration_matrix()`. But this will likely result in considerable breakage. I'll leave this one to you --- hopefully you've got a better handle on the scope of the breakage.

jeromekelleher:
Will do, ta.

PR #913, opened 2020-03-02
use numpy c api a bit more

grahamgower:
A few changes here related to #902 and #666.

The most common breakage will be in user code that assumes a list is returned by, e.g. RecombinationMap.get_positions() and then tries to append to the list.

It's not easy to hit all the error paths, so there might be ref leaks lurking here. And the numpy docs aren't always clear. I ran `stress_lowlevel.py` for a while and it reported thusly:
```
iter    tests   err     fail    skip    RSS     min     max     max@iter
46      290     0       0       1       96616   95316   96616   32
```

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -1939,6 +1939,18 @@ RecombinationMap_dealloc(RecombinationMap* self)
 }
 
 static int
+double_PyArray_converter(PyObject *in, PyObject **out)
+{
+    PyObject *ret = PyArray_FromAny(in, PyArray_DescrFromType(NPY_DOUBLE),
+            1, 1, NPY_ARRAY_CARRAY, NULL);
+    if (ret == NULL) {
+        return NPY_FAIL;
+    }
+    *out = ret;
+    return NPY_SUCCEED;
+}
+
+static int
 RecombinationMap_init(RecombinationMap *self, PyObject *args, PyObject *kwds)
 {
     int ret = -1;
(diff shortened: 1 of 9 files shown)


_msprimemodule.c
```diff
@@ -2109,9 +2095,19 @@ RecombinationMap_get_positions(RecombinationMap *self)
         handle_library_error(err);
         goto out;
     }
-    ret = convert_float_list(positions, size);
+    ret = PyArray_SimpleNewFromData(1, &size, NPY_FLOAT64, positions);
```
jeromekelleher: I think it would be simpler to do something like
```C
position_array =  PyArray_SimpleNewFromData(1, &size, NPY_FLOAT64);
if (position_array == NULL) { etc}
err = recomb_map_get_positions(self->recomb_map, PyArray_DATA(position_array));
```
Basically, we let numpy allocate the memory to store the array.

The capsule stuff is very clever, but this is simpler/more robust.

grahamgower:
I got rid of the capsule stuff. It's much simpler now, thanks.

I also tried using PyArrayConverter in PyArg_ParseTuple, but didn't have much luck. E.g. in RecombinationMap_init() I tried this:
```c
PyArg_ParseTupleAndKeywords(args, kwds, "O&O&p", kwlist,
            PyArray_Converter, &py_positions,
            PyArray_Converter, &py_rates,
            &discrete)
```
But the positions array can be converted to an array of ints, and there doesn't appear to be a way to ask for it to be NPY_DOUBLE, like when one converts using `PyArray_FromAny()`.

jeromekelleher:
Thanks very much @grahamgower. I'll take this one from here if you don't mind --- the numpy C API stuff is subtle, and I'll need to check the memory stuff carefully to make sure it's not introduced any leaks. I'll tack on a commit with any changes and comment if I spot an issues.

This is a big help, I really appreciate it!

jeromekelleher:
The idea for getting get_positions etc to return a list is so that we don't break existing code. The new numpy property-based API can then do things properly, while letting old code work as before.

grahamgower:
The updates look good to me.

The parse_samples change seemed dubious anyway, as the existing code is much simpler, so I'm happy you reverted that. Possibly it could be made to match current behaviour using numpy's iterator c-api. But using that efficiently and correctly for structured arrays seemed quite complicated and I didn't find any good examples to follow.

jeromekelleher:
Great, thanks @grahamgower, merged.

Issue #912, opened 2020-03-02
Full provenance for simulations


jeromekelleher:
msprime currently has half-baked support for provenance, in that only some parameters are serialised. Ideally, we want to be able to reproduce an identical tree sequence from the provenance information. This will involve serialising big things like recombination maps (sometimes) so we should also add a ``record_provenance=True`` argument, which will turn off provenance recording entirely if False. 

There's an argument for always writing the provenance, regardless (see this discussion: https://github.com/tskit-dev/pyslim/issues/67), but this seems heavy-handed in the case of msprime. Firstly, the vast majority of tree sequences generated in msprime are never stored, so provenance is irrelevant anyway. If we're doing millions of small simulations, this will be significant. Secondly,  I think we should take the Python philosophy of making it easy to write good programs rather than try to prevent people writing bad programs. There are good reasons for not wanting to write provenance info.

We'll also want a utility function ``parse_simulate_provenance(ts, index=None)`` function which will parse out the provenance from a tree sequence at a given index (defaulting to the last provenance) and return a dictionary of arguments. We should always have:
```python
ts1 = msprime.simulate(**args)
ts2 = msprime.simulate(**parse_simulate_provenance(ts1))
assert ts1.tables == ts2.tables # Well, not quite, see below
```

(We won't be able to assert exact equality, as the provenance time stamps will be different, but this is the idea anyway.)

jeromekelleher:
We'll also want a ``parse_mutate_provenance`` function which will do the same for the ``mutate`` function.

jeromekelleher:
The tricky bit here is dealing with replicates, and how we manage the random generator state. On the one hand, we want to be able to reproduce the a tree sequence that was produced during a ``num_replicates`` loop. On the other hand, saving the RNG state may be tricky and expensive to do in practise. (We'd have to use [this API](https://www.gnu.org/software/gsl/doc/html/rng.html#reading-and-writing-random-number-generator-state), which would be awkward.)

I think the thing to do is to have a ``replicate_index`` in the provenance, which stores the replicate number, making it at least possible to reproduce the simulation (even if we have to do all the other simulations first). I guess we could add a ``replicate_index`` option to the ``simulate`` function, which would only return the ``k``th replicate (and not an iterator over tree sequences). It's not ideal, but it would work.

I suppose we should **not** store ``num_replicates`` then within the provenance, since this is not relevant to how a particular tree sequence was generated.

jeromekelleher:
Closed in #914.

Issue #850, opened 2020-01-13
Review interface for pedigree sims


jeromekelleher:
#771 merged initial support for simulating through a pedigree, but there are some rough edges that will need to be cleared up before release:

1. Document the limitations (1 population, etc) and make sure that we raise appropriate errors rather than asserting out.
2. Review the approach for mapping input IDs to the output individual table. Currently we're encoding these as strings in the metadata, which isn't great. Ideally we'd want to populate the individual table with appropriate metadata in Python before the simulation.
3. Consider how we might extend the support for individuals in the DTWF, and at least make sure that we're forwards compatible with this.

jeromekelleher:
Moved this to a new milestone, as it doesn't look like the pedigree code will be ready for primetime at 1.0. There's not a huge amount to do, I think, but there's too may other things that need to be prioritised.

jeromekelleher:
This is out of date now.

PR #914, opened 2020-03-02
Complete provenance recording

benjeffery:
Issue #912 

I've started looking into this, the following `simulate` arguments use non-basic types that would need serialising:
- [x] recombination_map (`RecombinationMap`)
- [x] population_configurations (list of `PopulationConfiguration` or None.)
- [x] migration_matrix (list or numpy array)
- [x] demographic_events (list of subclasses of `DemographicEvent`)
- [x] samples (list of `Sample`)
- [x] from_ts (`TreeSequence`) 
- [x] model (string or `SimulationModel`)

Also the following two undocumented arguments have non-basic types:
- [x] pedigree (`Pedigree`)
- [x] ~~mutation_generator (`_msprime.MutationGenerator`)~~ Deprecated since 0.6.1 so don't include

Some of these classes do not support serialising by inspecting by `cls.__dict__`, for example `RecombinationMap` doesn't store it's arguments but passes them to the low-level instance. ~~My plan is to add `to_json` methods to these classes.~~ `asdict` methods as this allows easier recursive serialisation.

- [x] consider security implications of functional serialisation
- [x]  log errors in final provenance
- [x] `record_provenance` option to disable writing
- [x] `parse_simulate_provenance` to give dict of args from a provenance
- [x] replication index issues
- [x] `mutate` provenance
- [x] `parse_mutate_provenance`

### DIFF
--- a/msprime/provenance.py
+++ b/msprime/provenance.py
@@ -20,6 +20,14 @@
 Common provenance methods used to determine the state and versions
 of various dependencies and the OS.
 """
+import base64
+import importlib
+import json
+import marshal
+import logging
+import types
+
+import numpy
 import tskit
 
 import _msprime
@@ -30,6 +38,7 @@ try:
     __version__ = _version.version
 except ImportError:
     pass
+logger = logging.getLogger(__name__)
 
 
 def get_provenance_dict(parameters=None):
@@ -68,3 +77,129 @@ def get_environment():
     if _environment is None:
         _environment = _get_environment()
     return _environment
+
+
+class ProvenanceEncoderDecoder(json.JSONEncoder):
+    """
+    Extension of the `json` encoder that serializes arbitrary python class objects
+    by calling an `asdict` method on the object that should provide the arguments
+    needed to exactly recreate it. Special cases numpy arrays, classes and functions
+    """
+
+    def default(self, obj):
+        if isinstance(obj, types.FunctionType):
+            # Some provenance-relevant classes such as SimulationModelChange take
+            # functional arguments. Note that we have a tight definition of
+            # what we allow as a function here. We don't include builtins or
+            # callable classes, we also block functions that refer to globals
+            # or the outer scope for both complexity and security reasons.
+            # We can deserialize these later with
+            # FunctionType(fcode, fglobals, fname, fdefaults, fclosure)
+            if len(obj.__code__.co_names) > 0:
+                error = (f'Configuration function {obj.__code__.co_name} refers to'
+                         f' global variables {obj.__code__.co_names} which are not'
+                         f' currently serialized')
+                logger.warning(error)
+                return {'__error__': error}
+            if obj.__closure__ is not None and len(obj.__closure__) > 0:
+                error = (f'Configuration function {obj.__code__.co_name} refers to outer'
+                         f' scope variables which are not currently serialized')
+                logger.warning(error)
+                return {'__error__': error}
+            return {
+                '__function__': base64.b64encode(
+                    marshal.dumps(obj.__code__)).decode("utf-8"),
+                'defaults': obj.__defaults__
+            }
+
+        if isinstance(obj, tskit.TreeSequence):
+            # Store the tree sequence as the tables deserialised with
+            # tskit.TableCollection.fromdict(tables).tree_sequence()
+            return {'__ts__': obj.tables.asdict()}
+
+        elif isinstance(obj, numpy.ndarray):
+            # The most failsafe way would be be to `base64.b64encode(obj.tostring())`
+            # but as we expect only simple arrays we make this trade-off so that
+            # the array is human readable
+            return {
+                '__ndarray__': obj.tolist(),
+                'dtype': obj.dtype.str
+            }
+
+        elif isinstance(obj, numpy.number):
+            # Some numbers come through as numpy types
+            return {
+                '__npgeneric__': str(obj),
+                'dtype': obj.dtype.str,
+            }
+
+        try:
+            ret = obj.asdict()
+            cls = obj.__class__
+            ret['__class__'] = f'{cls.__module__}.{cls.__name__}'
+            return ret
+        except AttributeError:
+            raise TypeError(f'Object of type {obj.__class__.__name__} '
+                            f'is not JSON serializable, provide an `asdict` method '
+                            f'that gives the objects args')
+
+    @staticmethod
+    def decode(s):
+        def hook(obj):
+            if '__function__' in obj:
+                return types.FunctionType(
+                    marshal.loads(base64.b64decode(obj['__function__'])),
+                    {}, 'func', obj['defaults'], ())
+            elif '__ts__' in obj:
+                return tskit.TableCollection.fromdict(obj['__ts__']).tree_sequence()
+            elif '__ndarray__' in obj:
+                return numpy.asarray(obj['__ndarray__'], dtype=obj['dtype'])
+            elif '__npgeneric__' in obj:
+                return numpy.array([obj['__npgeneric__']]).astype(obj['dtype'])[0]
+            elif '__class__' in obj:
+                module, cls = obj['__class__'].rsplit('.', 1)
+                module = importlib.import_module(module)
+                del obj['__class__']
+                return getattr(module, cls)(**obj)
+            return obj
+        return json.JSONDecoder(object_hook=hook).decode(s)
+
+
+def parse_provenance(ts, index=None):
+    """
+    Convert the specified provenance to tuple of the command used and dict suitable
+    for calling the command again e.g `msprime.simulate(**parse_provenance(ts)[1])`
+    """
+    if index is None:
+        index = ts.num_provenances - 1
+    prov = ts.provenance(index).record
+    ret = ProvenanceEncoderDecoder.decode(prov)
+    if ret['software']['name'] != 'msprime':
+        raise ValueError(f'Only msprime provanances can be parsed,'
+                         f' found {ret["software"]["name"]}')
+    command = ret['parameters']['command']
+    del ret['parameters']['command']
+    return command, ret['parameters']
+
+
+def _human_readable_size(size, decimal_places=2):
+    for unit in ['B', 'KB', 'MB', 'GB', 'TB']:
+        if size < 1024.0:
+            break
+        size /= 1024.0
+    return f"{size:.{decimal_places}f}{unit}"
+
+
+def json_encode_provenance(provenance_dict, num_replicates=1):
+    """
+        Return a JSON representation of the provenance
+    """
+    prov = ProvenanceEncoderDecoder().encode(provenance_dict)
+    if len(prov) > 2_097_152:
+        logger.warning(f'The provenance information for the resulting tree sequence is'
+                       f' {_human_readable_size(len(prov))}.'
+                       f' This is nothing to worry about as provenance is a good thing'
+                       f' to have, but if you want to save this memory/storage space'
+                       f' you can disable provenance recording by setting'
+                       f' record_provenance=False')
+    return prov
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -158,10 +158,6 @@ def _replicate_generator(
     Generator function for the many-replicates case of the simulate
     function.
     """
-    # TODO We should encode the replicate index in here with the rest of the
-    # parameters. This will provide sufficient information to reproduce the
-    # simulation if necessary. Much simpler than encoding the details of
-    # the random number generator.
 
     encoded_provenance = None
     # The JSON is modified for each replicate to insert the replicate number.
@@ -357,6 +353,7 @@ def simulate(
         random_seed=None,
         mutation_generator=None,
         num_replicates=None,
+        replicate_index=None,
         from_ts=None,
         start_time=None,
         end_time=None,
@@ -438,6 +435,11 @@ def simulate(
         returned. If :obj:`num_replicates` is provided, the specified
         number of replicates is performed, and an iterator over the
         resulting :class:`tskit.TreeSequence` objects returned.
+    :param int replicate_index: Return only a specific tree
+        sequence from the set of replicates. This is used to recreate a specific tree
+        sequence from e.g. provenance. This argument only makes sense when used with
+        `random seed`, and is not compatible with `num_replicates`. Note also that
+        msprime will have to create and discard all the tree sequences up to this index.
     :param tskit.TreeSequence from_ts: If specified, initialise the simulation
         from the root segments of this tree sequence and return the
         completed tree sequence. Please see :ref:`here
@@ -490,9 +492,12 @@ def simulate(
     provenance_dict = None
     if record_provenance:
         argspec = inspect.getargvalues(inspect.currentframe())
+        # num_replicates is excluded as provenance is per replicate
+        # replicate index is excluded as it is inserted for each replicate
         parameters = {
             "command": "simulate",
-            **{arg: argspec.locals[arg] for arg in argspec.args}
+            **{arg: argspec.locals[arg] for arg in argspec.args
+                if arg not in ["num_replicates", "replicate_index"]}
         }
         parameters['random_seed'] = seed
         provenance_dict = provenance.get_provenance_dict(parameters)
@@ -545,9 +550,24 @@ def simulate(
                 "start_time. Please use msprime.mutate on the returned "
                 "tree sequence instead")
         mutation_generator = MutationGenerator(rng, mutation_rate)
-    if num_replicates is None:
-        return next(_replicate_generator(
-            sim, mutation_generator, 1, provenance_dict, end_time))
+
+    if replicate_index is not None and random_seed is None:
+        raise ValueError("Cannot specify replicate_index without random_seed as this "
+                         "has the same effect as not specifying replicate_index i.e. a "
+                         "random tree sequence")
+    if replicate_index is not None and num_replicates is not None:
+        raise ValueError("Cannot specify replicate_index with num_replicates as only "
+                         "the replicate_index specified will be returned.")
+    if num_replicates is None and replicate_index is None:
+        replicate_index = 0
+    if replicate_index is not None:
+        iterator = _replicate_generator(
+            sim, mutation_generator, replicate_index + 1, provenance_dict, end_time)
+        # Return the last element of the iterator
+        ts = next(iterator)
+        for ts in iterator:
+            continue
+        return ts
     else:
         return _replicate_generator(
             sim, mutation_generator, num_replicates, provenance_dict, end_time)
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -19,17 +19,18 @@
 """
 Module responsible for running simulations.
 """
+import bisect
 import collections
+import copy
 import gzip
+import inspect
 import json
+import logging
 import math
 import random
 import sys
 import os
 import warnings
-import copy
-import logging
-import bisect
 
 import tskit
 import numpy as np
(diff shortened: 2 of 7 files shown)


jeromekelleher:
Sounds like a plan @benjeffery. Quick thoughts:

- For ``from_ts`` I think we should use the ``TreeSequence.uuid`` property which I *think* is always defined. I guess the only other thing we could do is use the ``TreeSequence.asdict()`` method and coerce that into JSON. 

- We should be able to convert the ``Pedigree`` object to JSON easily enough (but see #850)

benjeffery:
Making progress.
Two questions I'm working on:

- Dealing with functional parameters such as those at `test_models.py:614`. I'm tempted to serialise those, you can't do this with `pickle` but you can with `dill`, as long as they aren't higher-order recursive functions. This would mean bringing in another dependency unless I extract the function serialiser from `dill` .

- For tree sequences are you talking about `file_uuid`? It seems to be `None`:
```
In [7]: msprime.simulate(10).file_uuid == None                                                                                                                                                                                                
Out[7]: True
```

jeromekelleher:
> Dealing with functional parameters such as those at test_models.py:614. I'm tempted to serialise those, you can't do this with pickle but you can with dill, as long as they aren't higher-order recursive functions. This would mean bringing in another dependency unless I extract the function serialiser from dill

Yuck, hadn't thought of that.  Not sure what to do there, possibly just print out a warning saying "incomplete provenance or something". I don't think this can be solved in general.

> For tree sequences are you(@jeromekelleher) talking about file_uuid? It seems to be None:

Yes I was. Oh well, looks like it's an attribute of the file so not stored here.

benjeffery:
> Yuck, hadn't thought of that. Not sure what to do there, possibly just print out a warning saying "incomplete provenance or something". I don't think this can be solved in general.

As a compromise I'm only serialising functions that don't refer to globals or the outer scope for now, although it would be possible to do more.

benjeffery:
Provenance can now be round-tripped successfully for `simulate`, rudimentary tests for this are in place, will add more for coverage next.

msprime/provenance.py
```diff
@@ -68,3 +76,108 @@ def get_environment():
     if _environment is None:
         _environment = _get_environment()
     return _environment
+
+
+class ProvenanceEncoderDecoder(json.JSONEncoder):
+    """
+      Extension of the `json` encoder that serializes arbitrary python class objects
```
jeromekelleher: Nitpick: we normally indent these like
```python
def something():
    """
    A docstring describing something
    """
```
happy to change this convention if you like (but it's a lot of code to update!)

tests/test_provenance.py
```diff
@@ -77,6 +80,14 @@ def decode(self, prov):
         ns = builder.build_classes()
         return ns.TskitProvenance.from_json(prov)
 
+    def verify_equal(self, ts1, ts2):
+        ts1_tables = ts1.dump_tables()
+        ts2_tables = ts2.dump_tables()
+        # We exclude provenance here due to timestamps
```
jeromekelleher: I prefer 
```python
ts1_tables.provenances.clear()
ts2_tables.provenances.clear()
self.assertEqual(ts1_tables, ts2_tables)
```
since this means that we don't need to know which tables are present (and maybe we'll add some more to tskit some day).

ts.dump_tables() will always return a copy of the tables so it's safe to do (ts.tables might return a read-only view on the underlying tables someday)

tests/test_provenance.py
```diff
@@ -86,13 +97,100 @@ def test_simulation(self):
...
+            return t + 0.01
+
+        def bad_func(t):
+            if msprime:
+                pass
+            return t
+
+        with self.assertLogs('msprime.provenance', level='WARNING') as cm:
+            ts = msprime.simulate(random_seed=seeds[0],
+                                  population_configurations=pop_configs,
```
jeromekelleher: I'm not a huge fan of visual indents, mainly because I think it makes the code layout more brittle. For example, here if I changed the variable name to ts1, I'd have to reindent this whole block. This exerts a pressure to *not* change the names of things, which means that code gets refactored less over time. Hanging indents are much more robust to code changes, so I'd do it like
```python
ts = msprime.simulate(
    random_seed=x, 
    population_configurations=pop_configs, 
   ....)
```

I agree it's a bit less pretty and not quite as easy for the eye to parse the structure though. Let's talk it through if you have strong opinions on indentation!

jeromekelleher:
Ah yes, I did have some input beyond niggling nitpicks! Is there a perf cost here over what we were already doing @benjeffery? There's two scenarios I have in mind:
```python
import msprime

for ts in msprime.simulate(10, num_replicates=10**7):
    pass
```
Is there an effect now of capturing the provenance?

Also, what about when we run a large genetic map? This is the easiest way to do it: 
```
$ stdpopsim HomSap -g HapMapIIGRCh37 -c chr22 -o tmp.trees
```
are we spending much time making JSON? Is the tree sequence file much bigger because of the provenance? (This doesn't matter so much, as we'll update stdpopsim to use ``record_provenance=False``)

I'm not sure what's going to happen with from_ts. It seems likely that provenance will be pretty large there, and we'll need to warn users about this. Let's not worry about that for now though.

msprime/simulations.py
```diff
@@ -460,13 +465,23 @@ def simulate(
         underlying object may be used for every TreeSequence
         returned which will most likely lead to unexpected behaviour.
     """
+    argspec = inspect.getargvalues(inspect.currentframe())
```
jeromekelleher: In the zillions of small replicates case, is it worthwhile skipping this step when record_provenance is False?

msprime/provenance.py
benjeffery: I prefer your way as it gives more room on the line - just have to change my editor setup!

Have you thought of using https://pypi.org/project/flake8-docstrings/ to enforce aspects of docstrings like this?

tests/test_provenance.py
benjeffery: I usually let the editor reformat the code, so hadn't thought of that. However it does make me realise that diffs are cleaner if the indent is not visual as the change is limited to only the relevant lines. I'll configure my editor to this style and look into getting the linter to enforce it.

msprime/simulations.py
benjeffery: I think this only gets called once for each call to `simulate`, not per replicate - but you're right there is no point calling it. Will move.

tests/test_provenance.py
benjeffery: :+1:

msprime/provenance.py
jeromekelleher: > Have you thought of using https://pypi.org/project/flake8-docstrings/ to enforce aspects of docstrings like this?

No, but we definitely should! (Across the board for tskit projects too, this would be great to have).

tests/test_provenance.py
jeromekelleher: Great, thanks. I'm not sure about getting the linter to enforce this though --- sometimes visual indents do work well and some regular contributors prefer it. We're already quite prescriptive about style and I don't want to push people away. Having said that, I would kind of like to just use [black](https://github.com/psf/black) and be done with it. Style discussions can become very unproductive...

benjeffery:
> Ah yes, I did have some input beyond niggling nitpicks! Is there a perf cost here over what we were already doing @benjeffery? There's two scenarios I have in mind:

For the simple case:
```python
for ts in msprime.simulate(10, num_replicates=10**5): 
    pass 
```
(You must have some sort of super-computer, 10**7 takes about 20min on my machine!)

master: `11.7 s ± 78.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
This PR: `11.8 s ± 31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

I think this is expected as the only cost here is a tiny bit more JSON in each provenance entry as the parameters are simple. 

> Also, what about when we run a large genetic map? This is the easiest way to do it:
> $ stdpopsim HomSap -g HapMapIIGRCh37 -c chr22 -o tmp.trees
> are we spending much time making JSON? Is the tree sequence file much bigger because of the provenance? (This doesn't matter so much, as we'll update stdpopsim to use `record_provenance=False`)

I used `stdpopsim HomSap -g HapMapII_GRCh37 -c chr22 -o tmp.trees 2000`
The resulting file is 37MB for master, 39MB for this PR. This tallies up with the JSON length which was measured at 1,290,049 chars.  ~49ms were spent encoding this JSON.

I haven't looked into how provenance is stored but the combination of ~1MB provenances and large numbers or replicates is problematic.

> I'm not sure what's going to happen with from_ts. It seems likely that provenance will be pretty large there, and we'll need to warn users about this. Let's not worry about that for now though.

One option is to detect a very large provenance as it is generated (or detect provenance size * # of replicates), spit out a warning or even drop the provenance.

jeromekelleher:
> One option is to detect a very large provenance as it is generated (or detect provenance size * # of replicates), spit out a warning or even drop the provenance.

That's a really good idea. I think if we detect a serialised provenance > 100K (say) we can raise a warning, telling the user that they should consider turning off provenance recording. This covers all bases nicely I think.

grahamgower:
Has there been any consideration given to having a container file/format/convention for a collection of tree sequence files? Then one full provenance table could be stored for n replicates, plus the n seeds (or other unique information).

jeromekelleher:
> Has there been any consideration given to having a container file/format/convention for a collection of tree sequence files? Then one full provenance table could be stored for n replicates, plus the n seeds (or other unique information).

No, not really. We've juggled using the idea for multiple chromosomes, but ended up deciding it wasn't worth the hassle. It seems unlikely many people are actually going to need to store lots of replicates, and if they do, that the provenance is going to be much of a burden.

There's no technical reason we couldn't though, we could imagine doing something like adding the tree sequence index as a prefix to all the column names in the underlying kastore as a way to do it. It'd be a **lot** of plumbing to update though.

jeromekelleher:
Looks like something funny's happened with your git IDs @benjeffery? Unless there really is two of you...

benjeffery:
> Looks like something funny's happened with your git IDs @benjeffery? Unless there really is two of you...

My evil twin strikes again! Set up the new machine with my BDI email, it's now added to my github account so should be all fixed.

benjeffery:
Assuming the CI is now happy (in progress) I think this PR is complete.

There is one aspect of testing that i want to flag up. It is hard to test that the subclasses of e.g. `DemographicEvent` are giving the correct return for `asdict`, as in the general case a class could have modified and then stored an argument, meaning the modified version gets stored in provenance. One solution to this would to implement `asdict` for every subclass even if it just calls the method on the superclass, then in the tests iterate over every subclass to check that it has this implementation. This seemed like overkill for a small concern.

I've set the provenance limit to 2MB total over all replicates, this seemed to cover most advanced single tree sequence use cases. Happy to change if anyone has a strong opinion.

msprime/provenance.py
```diff
@@ -20,6 +20,13 @@
 Common provenance methods used to determine the state and versions
 of various dependencies and the OS.
 """
+import base64
+import importlib
+import json
+import marshal
+import logging
+import types
```
jeromekelleher: Minor style point here. I like to split imports up into three groups: standard library, external imports and local imports. So, here it would be 
```python
import base64
￼import importlib
￼import json
￼import marshal
￼import logging
￼import types

import numpy
￼import tskit

import _msprime
```
This is a useful convention I think, as it lets you see in a glance what the module depends on.

msprime/provenance.py
```diff
@@ -68,3 +76,131 @@ def get_environment():
...
+        size /= 1024.0
+    return f"{size:.{decimal_places}f}{unit}"
+
+
+def json_encode_provenance(provenance_dict, num_replicates=1):
+    """
+        Return a JSON representation of the provenance
+    """
+    prov = ProvenanceEncoderDecoder().encode(provenance_dict)
+    if len(prov) * num_replicates > 2_097_152:
```
jeromekelleher: I don't think we need to worry about the number of replicates. If someone does 10**6 replicates, they're almost certainly not storing the actual files, so it won't matter (and this warning will worry users who haven't seen it before).

I think we just need to worry about the case in which a single rep goes over the threshold. (2M seems as good as any)

msprime/simulations.py
```diff
@@ -157,14 +158,23 @@ def _replicate_generator(
...
+    placeholder = '@@_MSPRIME_REPLICATE_INDEX_@@'
+    if provenance_dict is not None:
+        provenance_dict['parameters']['replicate_index'] = placeholder
+        encoded_provenance = provenance.json_encode_provenance(
+            provenance_dict, num_replicates)
+
     for j in range(num_replicates):
         sim.run(end_time)
-        tree_sequence = sim.get_tree_sequence(mutation_generator, provenance_record)
+        tree_sequence = sim.get_tree_sequence(
```
jeromekelleher: I'm not following the logic here - do we ever use the string "Provenance was disabled"? Can we pass ``None`` if not? I found myself having to think about what this does a bit too much, how about:
```python
provenance_record = None
if encoded_provenance is not None:
       provenance_record = encoded_provenance.replace(f'"{placeholder}"', str(j))
```
It's a bit more verbose, but clearer? I guess a rule of thumb for conditional assignments is "don't use them if you have to split lines"?

tests/test_provenance.py
```diff
@@ -124,3 +215,87 @@ def test_mutate_numpy(self):
...
+        pop_configs = [msprime.PopulationConfiguration(5) for _ in range(2)]
+        ts = msprime.simulate(population_configurations=pop_configs,
+                              migration_matrix=[[0, 1], [1, 0]],
+                              demographic_events=[msprime.SimulationModelChange(
+                                  time=lambda t: t+0.1),
+                              ])
+        command, parsed_prov = msprime.provenance.parse_provenance(ts, None)
+        recreated = getattr(msprime, command)(**parsed_prov)
+        self.verify_equal(ts, recreated)
+
```
jeromekelleher: Nice to split these up into discrete tests. First one is ``test_population_structure``, second is ``test_pedigree``, etc. Also nice to check a few more simple cases, like
```
def verify(ts):
    command, parsed_prov = msprime.provenance.parse_provenance(ts, None)
    recreated = getattr(msprime, command)(**parsed_prov)
    self.verify_equal(ts, recreated)


def test_sample_size(self):
    ts = msprime.simulate(10)
    self.verify(ts)

def test_random_seed(self):
     ts = msprime.simulate(10, random_seed=1234)
     self.verify(ts)
```

No need to be exhaustive, but good to capture a few of common usages here.

Also, good to test if we can round-round-trip --- we're not checking the provenance of the tree sequences we're generating from the parsed provenance. Can do this as a separate test I think.

tests/test_provenance.py
```diff
@@ -124,3 +215,87 @@ def test_mutate_numpy(self):
...
+        ts = msprime.simulate(10, random_seed=5)
+        from_ts = tsutil.decapitate(ts, ts.num_edges // 2)
+        start_time = from_ts.tables.nodes.time.max()
+        final_ts = msprime.simulate(
+            from_ts=from_ts, start_time=start_time, random_seed=2)
+        command, parsed_prov = msprime.provenance.parse_provenance(final_ts, None)
+        recreated = getattr(msprime, command)(**parsed_prov)
+        self.verify_equal(final_ts, recreated)
+
+    def test_mutate_round_trip(self):
```
jeromekelleher: Maybe better do this as it's own class? It could inherit the verify methods from a common superclass of TestSimulateRoundTrip.

tests/test_provenance.py
```diff
@@ -124,3 +215,87 @@ def test_mutate_numpy(self):
...
+        command, parsed_prov = msprime.provenance.parse_provenance(ts, None)
+        recreated = getattr(msprime, command)(**parsed_prov)
+        self.verify_equal(ts, recreated)
+
+        ts = msprime.simulate(10, random_seed=5)
+        from_ts = tsutil.decapitate(ts, ts.num_edges // 2)
+        start_time = from_ts.tables.nodes.time.max()
+        final_ts = msprime.simulate(
+            from_ts=from_ts, start_time=start_time, random_seed=2)
+        command, parsed_prov = msprime.provenance.parse_provenance(final_ts, None)
```
jeromekelleher: No point in specifying the provenance=None here, is there?

tests/test_provenance.py
```diff
@@ -124,3 +215,87 @@ def test_mutate_numpy(self):
...
+        # Most testing of parsing is in the round tripping below
+        with self.assertRaises(ValueError):
+            ts = msprime.simulate(5)
+            prov = json.loads(ts.provenance(0).record)
+            prov['software']['name'] = 'skynet'
+            tables = ts.tables
+            tables.provenances.clear()
+            tables.provenances.add_row(json.dumps(prov))
+            command, prov = msprime.provenance.parse_provenance(tables.tree_sequence())
+
```
jeromekelleher: We could use a test of the ``index`` argument here. Another test function would be most appropriate.

tests/test_provenance.py
```diff
@@ -124,3 +215,87 @@ def test_mutate_numpy(self):
         self.assertEqual(decoded.parameters.start_time, 0)
         self.assertEqual(decoded.parameters.end_time, 100)
         self.assertEqual(decoded.parameters.keep, False)
+
+
+class TestParseProvenance(unittest.TestCase):
+    def test_parse_provenance(self):
```
jeromekelleher: This is really ``test_parse_non_msprime`` I guess.

tests/test_provenance.py
```diff
@@ -86,13 +90,95 @@ def test_simulation(self):
...
+                             " ('msprime',) which are not currently serialized"},
+                'model': None, '__class__': 'msprime.simulations.SimulationModelChange'}
+        ),
+        self.assertDictEqual(
+            parameters.demographic_events[2],
+            {'time': {
+                '__error__': 'Configuration function <lambda> refers to outer scope'
+                             ' variables which are not currently serialized'},
+                'model': None, '__class__': 'msprime.simulations.SimulationModelChange'})
+
```
jeromekelleher: This is probably a different test case I guess?

msprime/provenance.py
benjeffery: Cool, we should use https://github.com/timothycrosley/isort to do this as part of https://github.com/tskit-dev/tskit/issues/470 (which I'm assuming once we're happy with we'll bring across to the other repos?)

msprime/provenance.py
benjeffery: I'm not so sure about this - even if they are not storing the files, it's still using up RAM. A simple, standard provenance is ~1500 chars, so the 10**6 replicate case would result in 1.5GB of provenance data in RAM without warning. 

I do get that anyone that uses a large number of replicates will always see this message though, which is not ideal. I propose we have 2M limit for single replicate and 500M total, and make the warning less scary by adding: "This is nothing to worry about unless you wish to save memory/storage"

msprime/simulations.py
benjeffery: I agree the logic could be clearer. 
The intent here was that upon opening a file and looking at the provenance you clearly saw that it was disabled, rather than just getting "None", but I'm in danger of overthinking! I'll put it back to None.

tests/test_provenance.py
benjeffery: Nice catch - the round-round-trip found an edge case with functional arguments.

tests/test_provenance.py
benjeffery: Yep, cleaner.

tests/test_provenance.py
benjeffery: Yeah, that line is gone now after factoring out `verify`.

tests/test_provenance.py
benjeffery: Cool.

benjeffery:
Thanks for the review @jeromekelleher! I've addressed all the points - just the warning message that I had some more thought on above. 

One aspect that came out of the test refactor was provenance in the `from_ts` case. Previously the provenance of the `from_ts` tree sequence was pre-pended to the output tree sequence. This didn't seem to make sense if we are storing the tree sequence in provenance, so I changed it to clear those provenances.

msprime/provenance.py
jeromekelleher: Great idea! Yes, I'm imagining us getting the tskit repo fully up to date with new linting stuff before porting across to msprime, kastore, tsinfer etc.

msprime/provenance.py
jeromekelleher: The standard approach for doing lots of replicates is 
```python
for ts in msprime.simulate(10, num_replicates=10**6):
    # do something with ts
```
So, only one tree sequence (and therefore provenance) is in memory at any one time. Unless you store the replicates in a list (which is a terrible idea anyway) the product of num_reps x provenance size isn't going to matter.

Unless I'm missing something here?

jeromekelleher:
LGTM, thanks @benjeffery! I think there's just the issue about the size across replicates to resolve and we're good to go.

msprime/provenance.py
benjeffery: picard_facepalm.gif 
Of course!

benjeffery:
Ok, I've fixed my howling error! Should be good to go.

jeromekelleher:
Excellent! OK, let's squash it back to 3 or 4 commits (the last few seem like they don't need to be in the history) and merge.

benjeffery:
Done! When I was going through the history it boiled down to two distinct changes - adding the provenance mechanisms, and the new `replicate_index` argument to `simulate`.

jeromekelleher:
Great stuff, thanks @benjeffery!

benjeffery:
:tada: Thanks for all the great comments and help.

PR #917, opened 2020-03-10
avoid merge_ancestors with less than 3 people

daniel-goldstein:
DTWF does a k-way merge for all individuals assigned the same parent each generation. When there are only two individuals, `merge_two_ancestors` is preferable as it has to do less book-keeping than a k-way merge. Worse, when there's only one individual it reconstructs the full segment chain for that individual instead of just doing nothing. There is also no work to be done when a parent has 0 offspring. This change avoids the call to `merge_ancestors` in these special cases.

This also changes slightly how parents are selected in `algorithms.py` because `np.random.choice(cur_inds)` was expanding the range `cur_inds` to a list of values which was very slow. This problem only existed in python.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -718,6 +718,15 @@ msp_remove_individual(msp_t *self, segment_t *u)
     msp_free_avl_node(self, node);
 }
 
+static void
+msp_remove_individuals_from_population(msp_t *self, avl_tree_t *Q)
+{
+    avl_node_t *node;
+    for (node = Q->head; node != NULL; node = node->next) {
+        msp_remove_individual(self, (segment_t *) node->item);
+    }
+}
+
 /* Returns true if the specified breakpoint exists */
 static bool
 msp_has_breakpoint(msp_t *self, double x)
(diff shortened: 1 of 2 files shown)


daniel-goldstein:
As a rough benchmark, here's the DTWF runtime in seconds vs sample size on 1Mb with r = 1e-8 and Ne = 10_000.

Seems like the population size increases so rapidly that the starting population size doesn't matter too much, but the runtime reduces from an average of about ~72s to ~12s.
dtwf_perf

lib/msprime.c
```diff
@@ -3925,11 +3946,21 @@ msp_dtwf_generation(msp_t *self)
...
-                    msp_remove_individual(self, (segment_t *) node->item);
+                segments_to_merge = avl_count(&Q[i]);
+                if (segments_to_merge == 1) {
+                    msp_priority_queue_pop(self, &Q[i]);
+                } else if (segments_to_merge >= 2) {
+                    msp_remove_individuals_from_population(self, &Q[i]);
+                    if (segments_to_merge == 2) {
+                        ret = msp_merge_two_ancestors(self,
+                                (population_id_t) j, label,
+                                msp_priority_queue_pop(self, &Q[i]),
```
jeromekelleher: The order in which arguments are evaluated [isn't defined in C](https://stackoverflow.com/questions/376278/parameter-evaluation-order-before-a-function-calling-in-c), which could lead to non-determinism in the output. Let's assign the result of pop to local variables before calling the func.

jeromekelleher:
Also:

> Seems like the population size increases so rapidly that the starting population size doesn't matter too much, but the runtime reduces from an average of about ~72s to ~12s.

Woohoo, well done! @DomNelson, you'll be interested in this!

lib/msprime.c
daniel-goldstein: 😮 ok, sounds good.

jeromekelleher:
Looks good thanks, just needs a quick rebase.

PR #920, opened 2020-03-23
Interval map

jeromekelleher:
WIP. Based #711 first.


jeromekelleher:
From @petrelharp:
> And: Yes!! Please make values one shorter!!! That eliminates uncertainty about which number goes with which interval, a serious source of bugs.

Will do.

jeromekelleher:
Comment on pr #711 
https://github.com/tskit-dev/msprime/pull/711#discussion_r397148226

jeromekelleher:
Also https://github.com/tskit-dev/msprime/pull/711#discussion_r397150271

jeromekelleher:
This can be pushed out until after Blackening the code. The diffs will be easy to fix, but I don't want to get into changing the C code until we have the code formatting stuff in place.

jeromekelleher:
Note to self: make sure interval maps are pickleable as well.

jeromekelleher:
This is mostly working now, but still a fair bit to be done. It's getting rid of a lot of cruft, though, and opening up the way for lots of new stuff.

- Change the Mutgen low-level Python class into a function. Gets rid of some boilerplate.
- Update the RecombinationMap class to contain the required functionality, but porting newer stuff into the RateMap class instead.
- Add some more functionality to the RateMap class and test.
- Either port some of the unit tests that we have for testing rate map translation functionality into C (and drop all the low-level python boiler plate) or reinstate the _msprime.RateMap class and push those tests through.

jeromekelleher:
This ended up being a bit of monster, but I think it's good enough to merge now. If anyone wants to have a look over, that'd be much appreciated (but don't feel like you need to go through all of it!).  What's been done:

1. The RecombinationMap class is dead, and has been replaced by a RateMap. This is a generic stuff-that-varies-along-the-genome container. At the Python layer, I've put in a skeleton implementation of the RateMap class (which I'll open some issues for filling out once this is merged). At the C layer, I've gotten rid of the recombination_map_t and similarly moved the functionality into the ``rate_map_t``. The big change is that discreteness is handled in the simulator now, not the map. @daniel-goldstein, it'd be great if you could have a look at this bit and make sure I haven't dropped any clangers. At the Python-C layer, I've dropped a whole bunch of cruft with custom classes and used a pair of numpy arrays to pass RateMaps around instead, which is *way* better.
2. I consolidated some of the mutation generation code, to get rid of the Python-C MutationGenerator object. This was largely pointless, as we were redoing all the allocations each time we used it anyway. This drops a whole load of cruft too.
3. I refactored the Simulator API in the high-level Python layer so that it's easier to see what's going on.

Together, hopefully this will make the various next bits we need to do for 1.0 a lot easier. Getting rid of the specialness of the RecombinationMap though was a critical step, and opens up all sorts of possibilities.

Pinging @grahamgower, @petrelharp, @benjeffery and @daniel-goldstein for review here please, as I've hacked at code you've edited recently here. Don't feel like you need to do major review here, though. A quick skim through, pausing at the bits you're familiar with would be great.

jeromekelleher:
Grr, still 1% off on the Python coverage. Will need to fix that and rebase/squash too.

lib/rate_map.c
```diff
@@ -0,0 +1,200 @@
...
+            }
+            if ((!isfinite(rate[j])) || rate[j] < 0) {
+                ret = MSP_ERR_BAD_RATE_VALUE;
+                goto out;
+            }
+            self->rate[j] = rate[j];
+            sum += (position[j + 1] - position[j]) * rate[j];
+        }
+    }
+    /* We should never touch this, but we initialise it for neatness. */
```
petrelharp, line 87: Is "neatness" so that an off-by-one error when indexing `rate` will give us an easier-to-debug error than if it was just off the end of the array?

jeromekelleher: Yeah,  it's mainly so we don't have to put in a special case for the ``print_state`` loop. Arguably we'd be better off leaving it uninitialised so that valgrind would catch any accesses off the end of the array, though.

jeromekelleher: Happy to change it.

petrelharp: Well, if it's easier to find bugs if it's shorter, and the issue is only in `print_state`, then I vote to make it shorter. If `rate` isn't shorter by one then you have to remember where the spurious value is - is it the first or the last element?

jeromekelleher: OK, will do.

jeromekelleher: Done in ce85e75

jeromekelleher:
Any chance you could take a look at this today @benjeffery? It's a big diff and we want to get it so we're not blocking people.

jeromekelleher:
Excellent! Merging.

daniel-goldstein:
Sorry for the late look at it but looks good! Impressive how little it took to move all the recombination-specific stuff out of the map and into the sim, RateMap is much simpler!

PR #926, opened 2020-03-24
put back provenance in from_ts

### LINKED ISSUE #924: simulate with `from_ts` removes previous provenance entries
This bug was introduced in c2bf47a2ac59f9b8281647e63d4a8414786d4431. MWE:
```
ts = msprime.simulate(10, end_time=0.2)
rts = msprime.simulate(from_ts=ts)
print("Has ", rts.num_provenances, "provenances")
```
This prints `Has  1 provenances`, but should print `Has  2 provenances`.

petrelharp:
This fixes #924... but may well mess up some other stuff.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -815,8 +815,6 @@ class Simulator(object):
             self.recombination_map.get_sequence_length())
         if self.from_ts is not None:
             from_ts_tables = self.from_ts.tables.asdict()
-            # Clear the provenance as it's included in the new ts's provenance
-            from_ts_tables['provenances'] = tskit.TableCollection(1).provenances.asdict()
             self.ll_tables.fromdict(from_ts_tables)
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM, what do you think @benjeffery? Maintaining compatability with existing code and workflows is the most important thing. Also, in retrospect, I agree that keeping the provenance entries as a record of how the tree sequence was created is the right thing to do. I should have spotted this at the time, my bad.

jeromekelleher:
On the grounds that this fixes a known bug, I say we merge this and follow up elsewhere to think about the provenance on from_ts.

@petrelharp, can you do a quick rebasearoo please?

petrelharp:
`git rebasearoo`-'ed

petrelharp:
Thanks!

Issue #928, opened 2020-03-25
Update the ``from_ts`` provenance


jeromekelleher:
As discussed over in #924, keeping the full tree sequence that's provided as input to msprime 
 with ``from_ts`` is probably not a great idea.

We can probably do something equivalent that's more elegant though by storing the offsets of the nodes and edges of the input tables. The original tree sequence should then be recoverable.

benjeffery:
If `from_ts`'s provenance is stored in `provenances[i-1]` then `from_ts` should be recoverable from that? Is that good enough?

petrelharp:
I think that 'simulate from' should add one additional entry to the provenance table, like other operations. That's not what you're suggesting, is it?

jeromekelleher:
I'll come up with a proposal this morning and report back - easiest way to see if what I have in mind will work.

petrelharp:
It'd be nice to get this fixed up soon, because people using the SLiM engine in stdpopsim are going to end up with giant tree sequences. For instance, in this fairly small example, it looks like the provenance is ~90% of the file size (if the numbers are accurate?):
```
import stdpopsim
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("chr22", length_multiplier=0.1)
# default is a flat genetic map
model = species.get_demographic_model("Africa_1T12")
samples = model.get_samples(200)
engine = stdpopsim.get_engine("slim")
ts = engine.simulate(model, contig, samples, slim_scaling_factor=10)
# The provenance information for the resulting tree sequence is 44.22MB. This is nothing to worry about as provenance is a good thing to have, but if you want to save this memory/storage space you can disable provenance recording by setting record_provenance=False
ts.dump("test.trees")
```
and then
```
$ ls -lth test.trees 
-rw-r--r-- 1 peter peter 48M Apr  2 20:49 test.trees
```

jeromekelleher:
Fair enough. @benjeffery, I guess the simplest thing to do for now is just remove the ts from the provenance, and skip any round-tripping tests? We can follow up on the right way to say "current_ts" at some other point.

benjeffery:
@jeromekelleher Yep, makes sense - will do a PR later.

PR #933, opened 2020-03-26
Store from_ts as offsets in provenance.

jeromekelleher:
Proposal for #928. All the tests pass, so we can round trip successfully. It's not 100% robust because you could have called ``mutate`` on the tree sequence output by ``simulate`` afterwards and the extra sites and mutations (which won't just be appended but will also be sorted) will be in there.

Anyway, though, I wonder if this is missing the point. The idea of provenances is that if we sequentially apply the provenance commands we should end up with the same tree sequence at the end. So, the first provenance will just have the simulate arguments encoded. The second provenance will have extra arguments encoded but will assume that the tree seq is present.

I'll open another PR with a different proposal.

### DIFF
--- a/msprime/provenance.py
+++ b/msprime/provenance.py
@@ -113,9 +113,14 @@ class ProvenanceEncoderDecoder(json.JSONEncoder):
             }
 
         if isinstance(obj, tskit.TreeSequence):
-            # Store the tree sequence as the tables deserialised with
-            # tskit.TableCollection.fromdict(tables).tree_sequence()
-            return {'__ts__': obj.tables.asdict()}
+            # This is a special case in which we know that the input tree sequence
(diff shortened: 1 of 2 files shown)


benjeffery:
> Anyway, though, I wonder if this is missing the point. The idea of provenances is that if we sequentially apply the provenance commands we should end up with the same tree sequence at the end. So, the first provenance will just have the simulate arguments encoded. The second provenance will have extra arguments encoded but will assume that the tree seq is present.

This is what I was trying to say in https://github.com/tskit-dev/msprime/issues/928#issuecomment-604161283

jeromekelleher:
> This is what I was trying to say in #928 (comment)

Ah, now I follow. Yes, I think that's the way to make this solvable.

PR #934, opened 2020-03-26
Initial proposal for 'rerun' function.

jeromekelleher:
This is an alternative proposal for dealing with provenance. I think fundamentally this is the right way to do things, but there's details to be worked out.

The idea is that we should guarantee that we can reproduce a tree sequence **only** if the same sequence of operations is applied.  We use the ``__current_ts__`` as a marker to say, "apply the function to the current state of the tree sequence`` as you are running through the provenances, one-by-one.

The cool thing here is that we don't even need to do this within msprime! If there was a standard way of decoding provenance (i.e., we associated a schema with it) and we provided the map of command->callable as input, we could do this with tskit.

Anyway, we don't need to do this now, but we should decide on the protocol used to denote the current state of the tree sequence.

### DIFF
--- a/msprime/provenance.py
+++ b/msprime/provenance.py
@@ -113,9 +113,7 @@ class ProvenanceEncoderDecoder(json.JSONEncoder):
             }
 
         if isinstance(obj, tskit.TreeSequence):
-            # Store the tree sequence as the tables deserialised with
-            # tskit.TableCollection.fromdict(tables).tree_sequence()
-            return {'__ts__': obj.tables.asdict()}
+            return "__current_ts__"
 
         elif isinstance(obj, numpy.ndarray):
             # The most failsafe way would be be to `base64.b64encode(obj.tostring())`
@@ -150,8 +148,6 @@ class ProvenanceEncoderDecoder(json.JSONEncoder):
                 return types.FunctionType(
                     marshal.loads(base64.b64decode(obj['__function__'])),
                     {}, 'func', obj['defaults'], ())
-            elif '__ts__' in obj:
-                return tskit.TableCollection.fromdict(obj['__ts__']).tree_sequence()
             elif '__ndarray__' in obj:
                 return numpy.asarray(obj['__ndarray__'], dtype=obj['dtype'])
             elif '__npgeneric__' in obj:
@@ -165,21 +161,25 @@ class ProvenanceEncoderDecoder(json.JSONEncoder):
         return json.JSONDecoder(object_hook=hook).decode(s)
 
 
-def parse_provenance(ts, index=None):
+def parse_provenance(provenance, current_ts):
     """
     Convert the specified provenance to tuple of the command used and dict suitable
     for calling the command again e.g `msprime.simulate(**parse_provenance(ts)[1])`
     """
-    if index is None:
-        index = ts.num_provenances - 1
-    prov = ts.provenance(index).record
-    ret = ProvenanceEncoderDecoder.decode(prov)
+    ret = ProvenanceEncoderDecoder.decode(provenance.record)
     if ret['software']['name'] != 'msprime':
         raise ValueError(f'Only msprime provanances can be parsed,'
                          f' found {ret["software"]["name"]}')
-    command = ret['parameters']['command']
-    del ret['parameters']['command']
-    return command, ret['parameters']
+    parameters = ret['parameters']
+    command = parameters.pop('command')
+    updated_parameters = {}
+    for key, value in parameters.items():
+        # TODO This is potentially unsafe. What if we had another argument to
+        # the function that happened to be "__current_ts__"?
+        if value == "__current_ts__":
+            value = current_ts
+        updated_parameters[key] = value
+    return command, updated_parameters
 
 
 def _human_readable_size(size, decimal_places=2):
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -2477,3 +2477,20 @@ class DemographyDebugger(object):
         Returns the number of epochs defined by the demographic model.
         """
         return len(self.epochs)
+
+
+def rerun(ts):
+    """
+    Reruns the simulations encoded in the provenance of the specified tree
+    sequence.
+    """
+    # TODO We can imagine passing in extra software/cmd arguments which map to a
+    # callable.
+    # TODO figure out how to workaround the circular import problems. Possibly
+    # this should just be in it's own module or something.
+    cmd_map = {"simulate": simulate}  # , "mutate": mutations.mutate}
+    current_ts = None
+    for prov in ts.provenances():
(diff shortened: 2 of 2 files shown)


jeromekelleher:
See also #928 and #933.

jeromekelleher:
An example:
```python
from_ts = msprime.simulate(
    5, end_time=1, random_seed=5, mutation_rate=5)
start_time = from_ts.tables.nodes.time.max()
final_ts = msprime.simulate(
    from_ts=from_ts, start_time=start_time, random_seed=2)

other_ts = msprime.rerun(final_ts)

t1 = final_ts.dump_tables()
t2 = other_ts.dump_tables()
t1.provenances.clear()
t2.provenances.clear()

assert t1 == t2

for prov in final_ts.provenances():

    decoded = json.loads(prov.record)
    pprint.pprint(decoded)
```

gives:
```json
 'parameters': {'Ne': 1,
                'command': 'simulate',
                'demographic_events': [],
                'end_time': None,
                'from_ts': '__current_ts__',
                'gene_conversion_rate': None,
                'gene_conversion_track_length': None,
                'length': None,
                'migration_matrix': None,
                'model': None,
                'mutation_generator': None,
                'mutation_rate': None,
                'num_labels': None,
                'pedigree': None,
                'population_configurations': None,
                'random_seed': 2,
                'recombination_map': None,
                'recombination_rate': None,
                'record_full_arg': False,
                'record_migrations': False,
                'record_provenance': True,
                'replicate_index': 0,
                'sample_size': None,
                'samples': None,
                'start_time': 1.0},
``

**Comment by petrelharp** [comment]
Nice, I like it.

**Comment by jeromekelleher** [comment]
If getting tskit to decode the provenance automatically turns out to be hard (which is probably is, since msprime wants msprime class instances as input), we can always as ask the source library to do this via a callable - we're already asking it for a callable for run methods, so asking it for a ``decode_provenance`` method seems reasonable. We just need to document clearly the responsibilities.

File: msprime/provenance.py [hunk 1]
```diff
@@ -113,9 +113,7 @@ def default(self, obj):
             }
 
         if isinstance(obj, tskit.TreeSequence):
-            # Store the tree sequence as the tables deserialised with
-            # tskit.TableCollection.fromdict(tables).tree_sequence()
-            return {'__ts__': obj.tables.asdict()}
+            return "__current_ts__"
```
benjeffery: I think `{'__constant__': '__current_ts__'}` would be better here like the other "interpretable" parts, this way it doesn't get mixed up with strings.

msprime/provenance.py
```diff
@@ -150,8 +148,6 @@ def hook(obj):
                 return types.FunctionType(
                     marshal.loads(base64.b64decode(obj['__function__'])),
                     {}, 'func', obj['defaults'], ())
-            elif '__ts__' in obj:
-                return tskit.TableCollection.fromdict(obj['__ts__']).tree_sequence()
```
benjeffery: Similarly here instead of removing replace with something like:
```python
elif '__constant__' in obj:
    return {
        '__current_ts__': tskit.CURRENT_TREE_SEQUENCE
    }[obj['__constant__']
```

msprime/provenance.py
```diff
@@ -165,21 +161,25 @@ def hook(obj):
...
-    command = ret['parameters']['command']
-    del ret['parameters']['command']
-    return command, ret['parameters']
+    parameters = ret['parameters']
+    command = parameters.pop('command')
+    updated_parameters = {}
+    for key, value in parameters.items():
+        # TODO This is potentially unsafe. What if we had another argument to
+        # the function that happened to be "__current_ts__"?
+        if value == "__current_ts__":
```
benjeffery: `if value == tskit.CURRENT_TREE_SEQUENCE`

msprime/simulations.py
```diff
@@ -2477,3 +2477,20 @@ def num_epochs(self):
...
+def rerun(ts):
+    """
+    Reruns the simulations encoded in the provenance of the specified tree
+    sequence.
+    """
+    # TODO We can imagine passing in extra software/cmd arguments which map to a
+    # callable.
+    # TODO figure out how to workaround the circular import problems. Possibly
+    # this should just be in it's own module or something.
+    cmd_map = {"simulate": simulate}  # , "mutate": mutations.mutate}
```
benjeffery: I would suggest that in provenances `command` becomes fully qualified i.e msprime.simulate. Then it can be dynamically imported as long as the pip package is in the users env.

benjeffery:
> If getting tskit to decode the provenance automatically turns out to be hard (which is probably is, since msprime wants msprime class instances as input)

As long as the class names are fully qualified then it should all work, just need the pip package in the env. Note that the recreation command could check the version numbers match up to the provenance and throw a warning on mismatch.

jeromekelleher:
OK, great, thanks @benjeffery. Let's have a chat about this soon to hash out the details.

jeromekelleher:
> As long as the class names are fully qualified then it should all work, just need the pip package in the env. Note that the recreation command could check the version numbers match up to the provenance and throw a warning on mismatch.

I guess we'd also want to have a protocol where each module that supports this would implement ``decode_provenance``? I'm not sure the command needs to be fully qualified, since we have the ``software.name`` key above?

We don't want to make this **too** tied down to Python either. Anyway, looking forward to chatting about the details.

petrelharp:
I just want to check that you all don't go too far down a rabbit hole here - this is a neat problem, and a nice property, but one I don't think is going to be used much in practice.

benjeffery:
> I just want to check that you all don't go too far down a rabbit hole here - this is a neat problem, and a nice property, but one I don't think is going to be used much in practice.

I agree re-running won't be used much - maybe for when you want to see what difference code changes have made. In my mind it's value comes from using it to prove that nothing is missing from the provenance.

jeromekelleher:
Totally agree with Ben here - it won't get used much in practise now, but if means that we get provenance properly sorted for good, it's very much worth it.

jeromekelleher:
Ben and I just discussed this and we think there's a really neat way to do it. Modules that support tskit should provide a function ``parse_provenance`` which take the ``current_ts`` and a provenance record as parameters. This function would then return a tuple ``(callable, args)``, such that running ``callable(**args)`` reproduces the tree sequence described by the provenance, given current_ts. This makes the ``rerun()`` function trivial.

jeromekelleher:
Oh yeah, the one other thing we'd need is to add a ``python_module`` section somewhere in the provenance so that we can load the right module.

petrelharp:
I'll have to think about how to do this for eg `pyslim.simplify`, which is only a thin layer around simplify, which records it's own provenance.

jeromekelleher:
closing this due to inactivity. I'll keep the branch alive in case we're ever interested.

PR #935, opened 2020-03-26
Update verification.py

TPPSellinger:
I added a test to check that for Xi coalescent we have similar distribution for recombination and mutation event (when both have same rate).

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -2081,7 +2081,9 @@ class SimulationVerifier(object):
 
     def compare_xi_dirac_sfs(self, sample_size, psi, c, sfs, num_replicates=1000):
         """
-        Runs simulations of the xi dirac model and compares to the expected SFS.
+        Runs simulations of the xi dirac model and compares to the expected normalized SFS.
+        It compares the observed E[Bi]/E[B] vs the expected  E[Bi]/E[B], where Bi is the branch length having i leaves
+        and B is the total branch length 
         """
         print("running SFS for", sample_size, psi, c)
         reps = msprime.simulate(
@@ -2090,6 +2092,7 @@ class SimulationVerifier(object):
 
         data = collections.defaultdict(list)
         tbl_sum = [0] * (sample_size - 1)
+        tot_bl_sum = [0] 
         for j, ts in enumerate(reps):
             for tree in ts.trees():
                 tot_bl = 0.0
@@ -2101,39 +2104,58 @@ class SimulationVerifier(object):
                         tot_bl = tot_bl + tree.branch_length(node)
 
                 for xi in range(sample_size - 1):
-                    rescaled_x = tbl[xi]/tot_bl
-                    data["total_branch_length"].append(rescaled_x)
+                    rescaled_x = tbl[xi]
+                    data["total_branch_length"].append(rescaled_x/tot_bl)
                     tbl_sum[xi] = tbl_sum[xi] + rescaled_x
+                tot_bl_sum[0] = tot_bl_sum[0] + tot_bl
                 data["num_leaves"].extend(range(1, sample_size))
 
         basedir = os.path.join("tmp__NOBACKUP__", "xi_dirac_expected_sfs")
         if not os.path.exists(basedir):
             os.mkdir(basedir)
-        f = os.path.join(basedir, "n={}_psi={}.png".format(sample_size, psi))
+        f = os.path.join(basedir, "n={}_psi={}_c={}.png".format(sample_size, psi, c))
 
         ax = sns.violinplot(
             data=data, x="num_leaves", y="total_branch_length", color="grey")
         ax.set_xlabel("num leaves")
         l1 = ax.plot(np.arange(sample_size - 1), sfs[::], "--", linewidth=3)
         l2 = ax.plot(
-            np.arange(sample_size - 1), [x/num_replicates for x in tbl_sum],
+            np.arange(sample_size - 1), [(x/num_replicates)/(tot_bl_sum[0]/num_replicates) for x in tbl_sum],
             "--", linewidth=3)
         ax.legend((l1[0], l2[0]), ("Expected", "Observed"))
         pyplot.savefig(f, dpi=72)
         pyplot.close('all')
 
     def run_xi_dirac_expected_sfs(self):
+        """
+        For c=1000 we compare to the expected normalized SFS for c tending to infinity (to check convergence).  
+        """
         self.compare_xi_dirac_sfs(
-            num_replicates=5000, sample_size=3, psi=0.01, c=1, sfs=[0.666667, 0.333333])
+            num_replicates=50000, sample_size=10, psi=0.1, c=10000, sfs=[0.36939374, 0.17057448, 0.11408360, 0.08571572, 0.06874076, 0.05749423, 0.04958115, 0.04390987, 0.04050644])
         self.compare_xi_dirac_sfs(
-            num_replicates=5000, sample_size=3, psi=0.99, c=1,
-            sfs=[0.6722604, 0.3277396])
+            num_replicates=50000, sample_size=10, psi=0.3, c=10000, sfs=[0.39876239, 0.15840021, 0.10834860, 0.08165271, 0.06562863, 0.05508280, 0.04777344, 0.04280604, 0.04154517])
         self.compare_xi_dirac_sfs(
-            num_replicates=5000, sample_size=4, psi=0.01, c=1,
-            sfs=[0.5457826, 0.2728913, 0.1813261])
+            num_replicates=50000, sample_size=10, psi=0.5, c=10000, sfs=[0.42603419, 0.14512841, 0.10505636, 0.07956441, 0.06368639, 0.05328134, 0.04595869, 0.04078814, 0.04050205])
         self.compare_xi_dirac_sfs(
-            num_replicates=5000, sample_size=4, psi=0.99, c=1,
-            sfs=[0.5611642, 0.2747103, 0.1641255])
+            num_replicates=50000, sample_size=10, psi=0.9, c=10000, sfs=[0.47543921, 0.11338801, 0.10691661, 0.08342993, 0.06358921, 0.05162311, 0.04334855, 0.03416865, 0.02809671])
+
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.1, c=1, sfs=[0.35352303, 0.17672997, 0.11781921, 0.08836481, 0.07069227, 0.05891075, 0.05049574, 0.04418514, 0.03927908])
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.3, c=1, sfs=[0.35430737, 0.17650201, 0.11762438, 0.08822363, 0.07058696, 0.05883259, 0.05044232, 0.04416277, 0.03931799])
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.5, c=1, sfs=[0.35655911, 0.17596878, 0.11711820, 0.08785514, 0.07030139, 0.05860142, 0.05025410, 0.04402755, 0.03931431])
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.9, c=1, sfs=[0.36443828, 0.17490683, 0.11614708, 0.08717119, 0.06965759, 0.05790491, 0.04939935, 0.04279132, 0.03758346])
+
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.1, c=10, sfs=[0.35385062, 0.17661522, 0.11773706, 0.08830646, 0.07064941, 0.05887993, 0.05047626, 0.04418035, 0.03930470])
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.3, c=10, sfs=[0.36053858, 0.17456975, 0.11610005, 0.08713599, 0.06977685, 0.05822906, 0.05002797, 0.04398723, 0.03963453])
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.5, c=10, sfs=[0.37556917, 0.17015781, 0.11285655, 0.08495119, 0.06808802, 0.05683977, 0.04886055, 0.04309158, 0.03958537])
+        self.compare_xi_dirac_sfs(
+            num_replicates=50000, sample_size=10, psi=0.9, c=10, sfs=[0.41154361, 0.15908770, 0.10852899, 0.08341563, 0.06647774, 0.05471783, 0.04592602, 0.03818041, 0.03212207])
 
         # MORE, NEED TO CHECK THESE VALUES
 
--- a/verification.py
+++ b/verification.py
@@ -2249,17 +2249,13 @@ class SimulationVerifier(object):
                 empirical_rho.append(ts.num_breakpoints)
         empirical_rho.sort()
         empirical_theta.sort()
-        pyplot.plot(empirical_rho, empirical_theta)
-        pyplot.plot(empirical_theta, empirical_theta, color='red')  
-        pyplot.xlabel('breakpoint number')
-        pyplot.ylabel('segregating site number')
-        pyplot.axis('square')
+        empirical_rho=np.array(empirical_rho)
+        empirical_theta=np.array(empirical_theta)
+        plot_qq(empirical_theta,empirical_rho)
         path = os.path.join(basedir, f"{name}_growth={growth_rate}_rec_check.png")
         print("Writing", path)
         pyplot.savefig(path)
-        pyplot.close('all')	
-	
-
+        pyplot.close('all')
 
     def run_hudson_breakpoints(self):
         basedir = "tmp__NOBACKUP__/hudson_breakpoints"
--- a/verification.py
+++ b/verification.py
@@ -2220,6 +2220,47 @@ class SimulationVerifier(object):
         pyplot.savefig(path)
         pyplot.close('all')
 
+    def verify_recombination(
+            self, basedir, name, sample_size, Ne, r, L, model, growth_rate=0):
+        """
+        Verifies that the number of recombination equals the number of mutation.
+        """
+        if not os.path.exists(basedir):
+            os.mkdir(basedir)
+        empirical_theta = []
+        empirical_rho = []
+        for i in range(1,200):
+                ts = msprime.simulate(
+                    Ne=Ne, recombination_rate=r, mutation_rate=r ,length=L,
+                    population_configurations=[
+                    msprime.PopulationConfiguration(
+                    sample_size=sample_size, initial_size=Ne,
+                    growth_rate=growth_rate)],
+                    model=model)
+                empirical_theta.append(ts.get_num_sites())
+                ts = msprime.simulator_factory(
+                    Ne=Ne, recombination_rate=r,length=L,
+            	    population_configurations=[
+                    msprime.PopulationConfiguration(
+                    sample_size=sample_size, initial_size=Ne,
+                    growth_rate=growth_rate)],
+            	    model=model)
+                ts.runs()
+                empirical_rho.append(ts.num_breakpoints)
+        empirical_rho.sort()
+        empirical_theta.sort()
+        pyplot.plot(empirical_rho, empirical_theta)
+        pyplot.plot(empirical_theta, empirical_theta, color='red')  
+        pyplot.xlabel('breakpoint number')
+        pyplot.ylabel('segregating site number')
+        pyplot.axis('square')
+        path = os.path.join(basedir, f"{name}_growth={growth_rate}_rec_check.png")
+        print("Writing", path)
+        pyplot.savefig(path)
+        pyplot.close('all')	
+	
+
+
     def run_hudson_breakpoints(self):
         basedir = "tmp__NOBACKUP__/hudson_breakpoints"
         self.verify_breakpoint_distribution(
@@ -2251,11 +2292,11 @@ class SimulationVerifier(object):
     def run_xi_dirac_breakpoints(self):
         basedir = "tmp__NOBACKUP__/xi_dirac_breakpoints"
         for psi in [0.1, 0.3, 0.6, 0.9]:
-            for c in [0.9, 0.5]:
+            for c in [10,100]:
                 self.verify_breakpoint_distribution(
                     basedir, f"n=100_psi={psi}_c={c}",
                     sample_size=100, Ne=10**4, r=1e-8,
-                    L=10**6, model=msprime.DiracCoalescent(psi=psi, c=c))
+                    L=10**7, model=msprime.DiracCoalescent(psi=psi, c=c))
                 # Add a growth rate with a higher recombination rate so
                 # we still get decent numbers of trees
                 self.verify_breakpoint_distribution(
@@ -2263,6 +2304,37 @@ class SimulationVerifier(object):
                     sample_size=100, Ne=10**4, r=1e-7,
                     L=10**6, model=msprime.DiracCoalescent(psi=psi, c=c),
                     growth_rate=0.05)
+
+    def run_xi_beta_recombinations(self):
+        basedir = "tmp__NOBACKUP__/xi_beta_recombinations"
+        for alpha in [1.1, 1.3, 1.6, 1.9]:
+            self.verify_recombination(
+                basedir, f"n=100_alpha={alpha}", sample_size=100, Ne=10**4, r=1e-8,
+                L=10**6, model=msprime.BetaCoalescent(alpha=alpha))
+
+
+    def run_xi_dirac_recombinations(self):
+        basedir = "tmp__NOBACKUP__/xi_dirac_recombinations"
+        for psi in [0.01, 0.1, 0.3, 0.6, 0.9]:
+            for c in [0.01, 1, 10, 100]:
+                self.verify_recombination(
+                    basedir, f"n=100_psi={psi}_c={c}",
+                    sample_size=100, Ne=10**4, r=1e-8,
+                    L=10**6, model=msprime.DiracCoalescent(psi=psi, c=c))
+
+    def run_Hudson_recombinations(self):
+        basedir = "tmp__NOBACKUP__/hudson_recombinations"
+        self.verify_recombination(
+            basedir, f"n=10_hudson",
+            sample_size=10, Ne=10**4, r=1e-9,
+            L=10**7, model="hudson")
+
+    def run_DTWF_recombinations(self):
+        basedir = "tmp__NOBACKUP__/DTWF_recombinations"
+        self.verify_recombination(
+            basedir, f"n=4_dtwf",
+            sample_size=4, Ne=10**4, r=1e-9,
+            L=10**7, model="dtwf")
 
     def run_cont_discrete_comparison(self, key, model,
                                      discrete_recomb_map,
@@ -2361,17 +2433,41 @@ class SimulationVerifier(object):
         """
         self._instances["hudson_breakpoints"] = self.run_hudson_breakpoints
 
+    def add_hudson_recombination(self):
+        """
+        Adds a check for xi_beta recombination breakpoints
+        """
+        self._instances["hudson_recombinations"] = self.run_Hudson_recombinations
+
+    def add_dtwf_recombination(self):
+        """
+        Adds a check for xi_beta recombination breakpoints
+        """
+        self._instances["dtwf_recombinations"] = self.run_DTWF_recombinations
+
     def add_xi_beta_breakpoints(self):
         """
         Adds a check for xi_beta recombination breakpoints
         """
         self._instances["xi_beta_breakpoints"] = self.run_xi_beta_breakpoints
 
+    def add_xi_beta_recombination(self):
+        """
+        Adds a check for xi_beta recombination breakpoints
+        """
+        self._instances["xi_beta_recombinations"] = self.run_xi_beta_recombinations
+
     def add_xi_dirac_breakpoints(self):
         """
         Adds a check for xi_dirac recombination breakpoints
         """
         self._instances["xi_dirac_breakpoints"] = self.run_xi_dirac_breakpoints
+
+    def add_xi_dirac_recombination(self):
+        """
+        Adds a check for xi_dirac recombination breakpoints
+        """
+        self._instances["xi_dirac_recombinations"] = self.run_xi_dirac_recombinations
 
     def run_xi_beta_expected_sfs(self):
         self.compare_xi_beta_sfs(
@@ -2571,6 +2667,8 @@ class SimulationVerifier(object):
 
 
 def run_tests(args):
+
+
     # random.seed(2)
     verifier = SimulationVerifier("tmp__NOBACKUP__")
 
@@ -2847,9 +2945,16 @@ def run_tests(args):
             num_populations=1, num_replicates=200, num_demographic_events=8)
 
     verifier.add_hudson_breakpoints()
-    # Check Xi coalesesent recombination breakpoint distributions
+    # Check Xi coalescent recombination breakpoint distributions
+
     verifier.add_xi_beta_breakpoints()
     verifier.add_xi_dirac_breakpoints()
+
+    # Check mutation and recombination have same distribution when mutation equals recombination rate
+    verifier.add_xi_beta_recombination()
+    verifier.add_xi_dirac_recombination()
+    verifier.add_hudson_recombination()
+    #verifier.add_dtwf_recombination()
 
     # DTWF checks against SLiM
     # TODO: Add back multi-pop tests of DTWF vs. SLiM when full diploid
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -5777,7 +5777,7 @@ compute_dirac_coalescence_rate(unsigned int num_ancestors, double psi, double c)
     assert(psi > 0);
     assert(psi < 1);
     assert(c >= 0);
-    ret = b * (b - 1) / 2 + c;
+    ret = 2*( b * (b - 1) / 2 + c);
     return ret;
 }
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -3218,9 +3218,12 @@ msp_reset_from_samples(msp_t *self)
     node_id_t u;
     tsk_id_t tsk_ind;
 
-    tsk_table_collection_clear(self->tables);
+    tsk_population_table_clear(&self->tables->populations);
+    tsk_edge_table_clear(&self->tables->edges);
+    tsk_node_table_clear(&self->tables->nodes);
+    tsk_migration_table_clear(&self->tables->migrations);
 
-    self->tables->sequence_length = recomb_map_get_sequence_length(&self->recomb_map);
+    self->tables->sequence_length = self->recomb_map.sequence_length;
     for (j = 0; j < self->num_populations; j++) {
         ret = tsk_population_table_add_row(&self->tables->populations, NULL, 0);
         if (ret < 0) {
@@ -5725,26 +5728,27 @@ msp_std_common_ancestor_event(msp_t *self, population_id_t population_id,
 static double
 dirac_model_time_to_generations(simulation_model_t *model, double t)
 {
-    return 4 * model->reference_size * t;
+    return 4 * gsl_pow_2(model->reference_size) * t / (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) ;
 }
 
 static double
 dirac_generations_to_model_time(simulation_model_t *model, double g)
 {
-    return g /(4 * model->reference_size);
+
+    return g * (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) /(4 * gsl_pow_2(model->reference_size));
 }
 
 static double
 dirac_generation_rate_to_model_rate(simulation_model_t *model, double rate)
 {
-    return rate * 4 * model->reference_size;
+     return rate * 4 * model->reference_size / (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi)));
+
 }
 
 static double
 dirac_model_rate_to_generation_rate(simulation_model_t *model, double rate)
 {
-    // This works for comparing dirac kingman case ... but why 4Ne^2
-    return rate / ( 4 * gsl_pow_2(model->reference_size));
+    return rate *(1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) / ( 4 * model->reference_size);
 }
 
 double
@@ -5766,8 +5770,6 @@ compute_falling_factorial_log(unsigned int m)
 double
 compute_dirac_coalescence_rate(unsigned int num_ancestors, double psi, double c)
 {
-    unsigned int l, m;
-    double r[5], r_max;
     double ret = 0;
     double b = num_ancestors;
 
@@ -5775,31 +5777,7 @@ compute_dirac_coalescence_rate(unsigned int num_ancestors, double psi, double c)
     assert(psi > 0);
     assert(psi < 1);
     assert(c >= 0);
-
-    m = GSL_MIN(num_ancestors, 4);
-    /* An underflow error occurs because of the large exponent (b-l). We use the
-     * LSE trick to approximate this calculation. For details, see at
-     * https://en.wikipedia.org/wiki/LogSumExp
-     */
-    r[0] = b * gsl_sf_log(1 - psi);
-    r_max = r[0];
-    for (l = 1; l <= m; l++){
-        r[l] = gsl_sf_lnchoose(num_ancestors, l)
-                + compute_falling_factorial_log(l)
-                + l * gsl_sf_log(psi / 4.0)
-                + (b - l) * gsl_sf_log(1 - psi);
-        r_max = GSL_MAX(r_max, r[l]);
-    }
-
-    for (l = 0; l <= m; l++)  {
-        ret += exp(r[l] - r_max);
-    }
-    ret = exp(r_max + log(ret));
-
-    ret = 1 - ret;
-    ret *= c;
-    ret += b * (b - 1) / 2;
-
+    ret = b * (b - 1) / 2 + c;
     return ret;
 }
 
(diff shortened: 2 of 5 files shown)


JereKoskela:
Thanks for putting this together @TPPSellinger. One thing I notice is that `dirac_generation_rate_to_model_rate` and `dirac_model_rate_to_generation_rate` in `/msprime/lib/msprime.c` are not yet inverses of each other. That might explain why you're getting odd amounts of recombination. As a first attempt, try fixing `dirac_generation_rate_to_model_rate` so that it is scaling by `4 N^2`; if the plots are still off, then we should look at something more involved.

@jeromekelleher, while I'm thinking of scalings, where are mutation rates scaled? We need to make sure that they're scaled like recombination rates (i.e. by `4 N`), not like generations (i.e. `4 N^2`).

jeromekelleher:
> @jeromekelleher, while I'm thinking of scalings, where are mutation rates scaled? We need to make sure that they're scaled like recombination rates (i.e. by 4 N), not like generations (i.e. 4 N^2).

Mutation rates aren't scaled - the tree branch lengths are in generations and the mutation rate is defined per generation. Is this a problem?

JereKoskela:
> > @jeromekelleher, while I'm thinking of scalings, where are mutation rates scaled? We need to make sure that they're scaled like recombination rates (i.e. by 4 N), not like generations (i.e. 4 N^2).
> 
> Mutation rates aren't scaled - the tree branch lengths are in generations and the mutation rate is defined per generation. Is this a problem?

No, I don't think it's a problem. There is some arbitraryness in defining what constitutes a generation in the Moran model so we need to document our choices clearly, but that was always necessary anyway. I just can't get away from thinking in coalescent time units.

TPPSellinger:
Actually, I obtained those results in both cases, when they are (if what i did was correct) and when they are not inverse function. I even tried when we scale by and N and not N² (but I always have the problem when psi is small). And if it was linked to the inverse function shouldn't the bias be stronger when psi is big, since  we are "further" away  from a Kingman ? Technically, when psi tends to 0, we should "fall back" on a Kingman ?  But when c tends to  0,  we have a good fit. Maybe I am forgetting something linked to psi in the scaling. I will check this in more details tomorrow.

JereKoskela:
Interesting, thanks for the further update @TPPSellinger . I had a quick look through the note @eldonb circulated, and it seems to me like the time scaling should be (1 + c psi^2) / (4 Ne^2). The plots you circulated seemed to suggest that large values of c might be a problem, and that missing factor could account for them.

By the way, am I right in thinking you're using Ne = 100? One other useful diagnostic would be to try Ne = 1, so that all rates are already specified on the coalescent timescale. That will wipe away some of the potential sources of error in the scalings, and hopefully help us zoom in on the exact problem.

TPPSellinger:
I agree with you. The scaling given by Bjarki makes perfect sense! I tried to plug it in the scaling but didn't change the result. I must be plugging it in the wrong place. Here is what I did (it's a bit brute force) : 
~~~~
static double
dirac_model_time_to_generations(simulation_model_t *model, double t)
{
    return 4 * gsl_pow_2(model->reference_size) * t  / (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi)))  ;
}

static double
dirac_generations_to_model_time(simulation_model_t *model, double g)
{
    return g * (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi)))  /(4 * gsl_pow_2(model->reference_size));

}

static double
dirac_generation_rate_to_model_rate(simulation_model_t *model, double rate)
{
 
    return rate * 4 * gsl_pow_2(model->reference_size)/(1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) ;
}

static double
dirac_model_rate_to_generation_rate(simulation_model_t *model, double rate)
{

    return rate * (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi)))  / ( 4 * gsl_pow_2(model->reference_size)* model->params.dirac_coalescent.psi);
}
~~~~

JereKoskela:
The `dirac_generation_rate_to_model_rate` and `dirac_model_rate_to_generation_rate` should only scale like `N`, not `N^2`, but that wouldn't cause this problem by itself.

Sorry if this is an obvious question, but are you rebuilding the package every time you change things? Instructions for recompiling the C code are here: https://msprime.readthedocs.io/en/latest/development.html#c-library, and the python module can be rebuilt locally by calling `make` in the root directory: https://msprime.readthedocs.io/en/latest/development.html#quickstart. I remember it took me quite some time to get to grips with the full compilation pipeline when I started making changes.

TPPSellinger:
Yes, after each modification I rebuild it by doing : 
~~~~
make
~~~~
I think it works because results changes every time. But nothing of what I tried seems to work : /

TPPSellinger:
What actually is this generation rate ? And why should we multiply it by the reference size ?  Intuitively, I would divide the generation by the  reference size to get the rate, no ?

JereKoskela:
> What actually is this generation rate ? And why should we multiply it by the reference size ? Intuitively, I would divide the generation by the reference size to get the rate, no ?

`generation_rate` is a rate expressed in units of generations, before time-scaling. `model_rate` is the same rate per unit of coalescent time. `generation_rate` is multiplied by `reference_size` (or appropriate polynomial thereof) because that is how many generations there are in a unit of coalescent time.

TPPSellinger:
I tried many different scaling, but none of them worked. Could you forward me the notes of Bjarki on the subject ? There must be something I am missing or not understanding.

verification.py
```diff
@@ -2264,6 +2305,37 @@ def run_xi_dirac_breakpoints(self):
...
+                    L=10**6, model=msprime.DiracCoalescent(psi=psi, c=c))
+
+    def run_Hudson_recombinations(self):
+        basedir = "tmp__NOBACKUP__/hudson_recombinations"
+        self.verify_recombination(
+            basedir, f"n=10_hudson",
+            sample_size=10, Ne=10**4, r=1e-9,
+            L=10**7, model="hudson")
+
+    def run_DTWF_recombinations(self):
```
jeromekelleher: Can we remove this function please @TPPSellinger? We're not going to be able to run it, so no point in keeping it in there.

TPPSellinger: Yes sure!  You can remove it. I was only using it because it was when I ran it that the error occured.

JereKoskela:
think I've made a start on why this isn't behaving as expected. The common ancestor rate we compute is the rate until a merger event (whether isolated or multiple, or binary or larger), but the way we actually simulate a multiple merger events has a positive probability of not merging anything. Thus, mergers are happening at a slower rate than they should be, and the problem is worse for large `c` (more multiple merger events) and small `psi` (more likely that a multiple merger event doesn't result in a coalescence).

The odd thing is that I _think_ this should be a problem for our Beta-coalescent too, and I'm not yet sure whether I'm missing something, or whether our Beta tests just didn't pick it up. I'll keep digging tomorrow.

TPPSellinger:
May be we didn't check for value extrem enough. I will check fot  1.01. Then problem may then arise.

TPPSellinger:
I checked for alpha value up to 1.0001 and not problem has arisen. Mutation and recombination still seem to share same distribution. But there is something I don't understand.  I imagine that mutation are put after the whole genealogy has been built. If this is true, how is it possible to observe more recombination than mutation ? I mean, even if trees are longer because  of the probability of no merging, we should simply have more mutation than expected and not more recombination than mutation? May be the scaling is correct, as well as the recombination process, but it's the way we put mutation for the dirac coalescent that is wrong? "Increasing" the mutation rate by (1+ c psi²) might solve the problem. If this solves the problem, we then simply check the SFS?

jeromekelleher:
Mutations are just thrown down on the trees afterwards in units of generations. This is completely independent of how the trees are generated.

TPPSellinger:
So what Jere said, is that because coalescent/merger/collision rate is slower, we have extra recombination ? But then, genealogy should have longer branch length, thus also more mutation ? Or both would be affected in different way ?

jeromekelleher:
I don't know to be honest @TPPSellinger, but mutation and recombination here are entirely separate, and mutations happen afterwards on the generated trees. Let's see what @JereKoskela thinks.

JereKoskela:
I think I've got it! The problem is not what I alluded to in my earlier post. That is accounted for by rejection sampling, where we undo multiple merger events that end up merging no lineages. That's good news, since it means there's no reason to doubt the correctness of the Beta-coalescent.

The mistake is on line 5836 of `msprime.c`, which currently reads:

`    ancestors = &self->populations[pop_id].ancestors[label];`
`    if (gsl_rng_uniform(self->rng) < (1 / (1.0 + self->model.params.dirac_coalescent.c))) {`

Instead, that should be

`    ancestors = &self->populations[pop_id].ancestors[label];`
`    n = avl_count(ancestors);`
`    if (gsl_rng_uniform(self->rng) < (gsl_sf_choose(n, 2) / (gsl_sf_choose(n, 2) + self->model.params.dirac_coalescent.c))) {`

In addition, there is an extra factor of two on line 5821 of `msprime.c` which should not be there.

@TPPSellinger, could you make those two changes and rerun your analysis please? I've done a cursory check on my end and get a plot that looks good, but a second pair of eyes would be helpful given how fiddly this has been.

TPPSellinger:
I'm on  it !

TPPSellinger:
The problem seems to be solved!!! However, results display high variance (which is to be expected from the dirac coalescent). I will increase number of simulation and sequence length to be sure that the problem is solved.

TPPSellinger:
I don't know if the problem is solved. But the results are definitely better. it seems that for big value of c, we still have too many recombination (but for small value of c everything looks fine).
n=100_psi=0 5_c=100_growth=0_rec_check

 
Do you have similar results for psi=0.5 and  c=100  (strangely when psi=0.9 and c =100 results are quite good)? I did the changes to asked and here is my scaling : 
~~~~

static double
dirac_model_time_to_generations(simulation_model_t *model, double t)
{
    return 4 * gsl_pow_2(model->reference_size) * t / (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) ;
}

static double
dirac_generations_to_model_time(simulation_model_t *model, double g)
{

    return g * (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) /(4 * gsl_pow_2(model->reference_size));
}

static double
dirac_generation_rate_to_model_rate(simulation_model_t *model, double rate)
{
     return rate * 4 * model->reference_size / (1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi)));

}

static double
dirac_model_rate_to_generation_rate(simulation_model_t *model, double rate)
{
    return rate *(1+ (model->params.dirac_coalescent.c * gsl_pow_2(model->params.dirac_coalescent.psi))) / ( 4 * model->reference_size);
}

~~~~ 
What do you think of the results ?

JereKoskela:
> What do you think of the results ?

I get a similar result. I think this is where we need a hack to address the fact that parameters and time scale differently under the Moran model (i.e. `N^2` for time and `N` for rates). Try to do your comparison with `recombination_rate=r`, and `mutation_rate=r/Ne`. Results with that regime look much better on my machine, and correctly adapt to the fact that if we were simulating mutations on the coalescent time scale, we would scale the mutation rate in the same way as the recombination rate.

This is something we have to make very clear when documenting the Dirac coalescent, because I suspect it will catch people out.

TPPSellinger:
Are you sure of this ? I don't have good result with this. Shouldn't it be r=mu/Ne ?

JereKoskela:
> Are you sure of this ? I don't have good result with this. Shouldn't it be r=mu/Ne ?

Branch lengths are proportional to `N^2` generations, so if we simulate mutations at unit generation rate, then there will also be `O(N^2)` mutations. We need that to go down to `O(N)` to match the scaling of the recombination rate.

You're right though - further investigation showed that my suggestion results in a similar picture to yours, but with an excess of mutations rather than recombinations, i.e. the blue line hops on the other side of the red.

TPPSellinger:
We should check for different value of psi and c (with lots of repetitions). May be the deviation between the red and the blue line could help us understand what we are doing wrong. We could also try different values of  Ne, if deviation increases with Ne, would that mean the problem is linked to the unit of generation rate ?

JereKoskela:
Actually, I think this might be working. I made one more change to the `compute_dirac_coalescence_rate()` function, which was to add

`ret *= 2 / (1 + c * psi * psi / 4);`

as the last line of code before `return ret;`. That ensures that the pair coalescence rate between two lineages is 2 (in units of coalescent time), which matches the way that the standard Hudson ARG is implemented in ``msprime``. I then ran the same test on the Hudson model, as well as on a few different Dirac coalescents. 

Here is the Hudson picture:

test_hudson

And here is a Dirac picture with `c = 0`, so that in fact it is just a needlessly complicated way to simulate the Hudson model:

test_dirac_c=0_psi=0 01

Finally, here is a more extreme example with `c = 10` and `psi = 0.5`:

test_dirac_c=10_psi=0 5

A few other tests were similar.

All three show an excess of mutations, which worried me until I remembered that the tree sequence data structure does not store all recombinations - only those which actually change the ancestry of genetic material! On the other hand, all mutations are stored, so we would expect to see an excess of mutations. The `record_full_arg` setting can be used to rectify that in the Hudson case, but it isn't available for multiple merger coalescents, so we can't use it to help here.

Incidentally, I think this apparent excess of mutations doesn't show up with the Beta coalescent because both mutations and recombinations are quite rare there, unless you use an astronomical effective population size. A small number of recombinations and a decent sample size won't result in silent recombinations, which is why the straight-line fit was good for that model.

I'm not sure what exactly the upshot is. Assuming my reasoning is right, then the output of this test appears to be good, but it is much harder to determine correctness exactly since we can't really quantify how many excess mutations there should be. At least we did find some problems with the implementation as a result of the bug chase.

verification.py
```diff
@@ -2220,6 +2220,47 @@ def verify_breakpoint_distribution(
...
+            	    population_configurations=[
+                    msprime.PopulationConfiguration(
+                    sample_size=sample_size, initial_size=Ne,
+                    growth_rate=growth_rate)],
+            	    model=model)
+                ts.runs()
+                empirical_rho.append(ts.num_breakpoints)
+        empirical_rho.sort()
+        empirical_theta.sort()
+        pyplot.plot(empirical_rho, empirical_theta)
```
jeromekelleher, line 2252: We should use a qqplot here rather than the line plot. See the other examples on how to do this.

verification.py
```diff
@@ -2220,6 +2220,47 @@ def verify_breakpoint_distribution(
...
+            self, basedir, name, sample_size, Ne, r, L, model, growth_rate=0):
+        """
+        Verifies that the number of recombination equals the number of mutation.
+        """
+        if not os.path.exists(basedir):
+            os.mkdir(basedir)
+        empirical_theta = []
+        empirical_rho = []
+        for i in range(1,200):
+                ts = msprime.simulate(
```
jeromekelleher, line 2233: If we use ``msprime.simulator_factory()`` to get a simulator, we should be able to use the num_breakpoints from there.

jeromekelleher:
You're right @JereKoskela, but we have ways of getting at the actual number of recombinations from the simulator itself (@TPPSellinger and I have talked about this elsewhere), so this is fixable.

verification.py
TPPSellinger: I edited the code to do a real qqplot. Can you see it on the  pull request ?

TPPSellinger:
I think the correction of @JereKoskela  helped solving the problem but there might be a small problem left. Here is the dirac coalescent rate modification : 
~~~~
    ret += b * (b - 1) / 2;
    ret *= 2 / (1 + c * psi * psi / 4);
    return ret;
~~~~ 
All results are good, but when psi is small and c is big, there is still a small scaling issue. 

n=100_psi=0 01_c=100_growth=0_rec_check

I think we are very close to definitely solving the problem.

TPPSellinger:
On the other  hand when I do : 
~~~~
    ret *= c;
    ret += b * (b - 1) / 2;
    ret *= 1 / (1 + c * psi * psi / 4);
~~~~
I seem to have a better fit (psi=0.01 and c =100 ):  
n=100_psi=0 01_c=100_growth=0_rec_check

JereKoskela:
> On the other hand when I do :
> 
> ```
>     ret *= c;
>     ret += b * (b - 1) / 2;
>     ret *= 1 / (1 + c * psi * psi / 4);
> ```

That multiplication by `c` should definitely be there. Could you push your version of `msprime.c` into this PR as well so that its easy to see the whole implementation in one place?

TPPSellinger:
Oh sorry, the  multiplication by c was also there before. I didn't copy it. Here is the full function: 

~~~~
compute_dirac_coalescence_rate(unsigned int num_ancestors, double psi, double c)
{
    unsigned int l, m;
    double r[5], r_max;
    double ret = 0;
    double b = num_ancestors;

    assert(b > 0);
    assert(psi > 0);
    assert(psi < 1);
    assert(c >= 0);

    m = GSL_MIN(num_ancestors, 4);
    /* An underflow error occurs because of the large exponent (b-l). We use the
     * LSE trick to approximate this calculation. For details, see at
     * https://en.wikipedia.org/wiki/LogSumExp
     */
    r[0] = b * gsl_sf_log(1 - psi);
    r_max = r[0];
    for (l = 1; l <= m; l++){
        r[l] = gsl_sf_lnchoose(num_ancestors, l)
                + compute_falling_factorial_log(l)
                + l * gsl_sf_log(psi / 4.0)
                + (b - l) * gsl_sf_log(1 - psi);
        r_max = GSL_MAX(r_max, r[l]);
    }

    for (l = 0; l <= m; l++)  {
        ret += exp(r[l] - r_max);
    }
    ret = exp(r_max + log(ret));

    ret = 1 - ret;
    ret *= c;
    ret += b * (b - 1) / 2;
    ret *= 1 / (1 + c * psi * psi / 4);
    return ret;
}
~~~~
Should I still do the PR?

JereKoskela:
Sorry, I didn't spot the other difference in your two posts. I thought the multiplication by `c` was what you wanted to highlight. Now I see the factor of two has disappeared as well.

I put the factor of two in to match the implementation of the Hudson ARG in my tests, but in my opinion the Dirac coalescent is more natural without it. That looks like a successful test to me. You should still push in your `msprime.c` as well, because we'll need to include the fixes that have been implemented along the way.

TPPSellinger:
I don't find the fit perfect yet. I will keep searching for some issue, and do the PR tomorrow morning if this is okay with you ?

JereKoskela:
> I don't find the fit perfect yet. I will keep searching for some issue, and do the PR tomorrow morning if this is okay with you ?

That's fine, no rush on my account.

TPPSellinger:
I think I solved it. I used the original scaling given in Birkner et al 2013. I don't know why results look better.   Here is the plot  (c=100 and psi=0.01). All the other plots looks fine to me. 

n=100_psi=0 01_c=100_growth=0_rec_check
Here is the calculation of the coalesent rate. 

~~~~
double
compute_dirac_coalescence_rate(unsigned int num_ancestors, double psi, double c)
{
    unsigned int l, m;
    double r[5], r_max;
    double ret = 0;
    double b = num_ancestors;

    assert(b > 0);
    assert(psi > 0);
    assert(psi < 1);
    assert(c >= 0);

    m = GSL_MIN(num_ancestors, 4);
    /* An underflow error occurs because of the large exponent (b-l). We use the
     * LSE trick to approximate this calculation. For details, see at
     * https://en.wikipedia.org/wiki/LogSumExp
     */
    r[0] = b * gsl_sf_log(1 - psi);
    r_max = r[0];
    for (l = 1; l <= m; l++){
        r[l] = gsl_sf_lnchoose(num_ancestors, l)
                + compute_falling_factorial_log(l)
                + l * gsl_sf_log(psi / 4.0)
                + (b - l) * gsl_sf_log(1 - psi);
        r_max = GSL_MAX(r_max, r[l]);
    }

    for (l = 0; l <= m; l++)  {
        ret += exp(r[l] - r_max);
    }
    ret = exp(r_max + log(ret));

    ret = 1 - ret;
    ret *= c;
    ret += b * (b - 1) / 2;
    ret *= 4 / (4 + c * psi * psi );
    return ret;
}
~~~~

I will try more extreme values to make sure the fit stays good.

TPPSellinger:
Unfortunately the problem is not solved. Here are the results for psi=0.001 and  c=1000. 

n=100_psi=0 001_c=1000_growth=0_rec_check

However, we must be very close to solving the problem.

JereKoskela:
Hold on, I'm still not sure this is actually a valid test. Trapped non-ancestral material will contribute a positive recombination rate to the ARG, but will not generate visible mutations. Thus, we might now expect an excess of recombinations, which is what we're seeing in the latest plots.

@jeromekelleher, is there an easy way to apply the SMC constraint to a multiple merger coalescent? That would fix this issue, but is probably not worth it if we'd have to build the functionality first.

TPPSellinger:
Good point! I didn't think about that.

jeromekelleher:
> @jeromekelleher, is there an easy way to apply the SMC constraint to a multiple merger coalescent? That would fix this issue, but is probably not worth it if we'd have to build the functionality first.

Should be straightforward enough, but I don't think that'll solve anything. What does it look like for Hudson? We're only counting recombinations in ancestral material with ``sim.num_breakpoints``.

TPPSellinger:
Here are the Hudson results : 
n=100_hudson_growth=0_rec_check

JereKoskela:
> > @jeromekelleher, is there an easy way to apply the SMC constraint to a multiple merger coalescent? That would fix this issue, but is probably not worth it if we'd have to build the functionality first.
> 
> Should be straightforward enough, but I don't think that'll solve anything. What does it look like for Hudson? We're only counting recombinations in ancestral material with `sim.num_breakpoints`.

Ah, if we're counting break points, as opposed to recombination events, then indeed the problem doesn't arise. I should really look up what this `simulator_factory` does... That means that the excess of recombinations in @TPPSellinger's plot remains unexplained.

TPPSellinger:
I have similar results with a recombination and mutation rate 10 times smaller (but sequences 10 times bigger). So I don't think it's a finite site hypothesis violation problem.

JereKoskela:
I spotted another issue, and it's in the same place as the first one: the probabilities for choosing between binary and multiple merger events in `msp_common_ancestor_event()`.

The rate of binary mergers is `n * (n - 1) / 2`, and the rate of potential multiple merger events is `c`. However, of those potential multiple mergers, some will not result in any mergers. The rate of actual, effective multiple mergers is the more complicated sum expression we compute as part of  `compute_dirac_coalescence_rate()`. The rates of the two kinds of competing events in `msp_dirac_common_ancestor_event()` should therefore also be `n * (n - 1) / 2`, and the rate of effective multiple mergers, not `n * (n - 1) / 2` and `c`.

The simplest way to fix this is to make `compute_dirac_coalescence_rate()` return just

`ret = b * (b - 1) / 2 + c`

without any of the complicated stuff, because then the competing rates will be correct. We then also need to delete the 

`if (max_pot_size < 2) {ret = 1;}`

at the bottom of `msp_dirac_common_ancestor_event()`, because now we need to keep the time increments even when nobody coalesces.

I've tested that on a few cases including the problematic one above, and it gives a remarkably good qqplot in all of my cases. @TPPSellinger, could you try it out on your setup?

TPPSellinger:
Unfortunately I don't have acces to my computer anymore. I will do this tomorrow morning. If I also obtained remarkably good results I will also do the PR.

TPPSellinger:
Thank you @JereKoskela ! After intensive testing I didn't find any problems. I think the problem is finally solved. I will now try to do the PR.

Should we do a quick test on the SFS by proof of principle or we now consider the multiple merger simulator working ?

JereKoskela:
> Should we do a quick test on the SFS by proof of principle or we now consider the multiple merger simulator working ?

Yes, let's check the SFS as well. For one thing, that will make sure we haven't broken anything with the tweaks we've made here.

TPPSellinger:
Is there a test already implemented somewhere ?  I tried the one implemented in verification.py (verifier.add_xi_dirac_expected_sfs()). I don't know if we can trust it. I was running it after every "surgery" of the dirac coalescent and they always gave me the same results.  Should we implement a new one ?

JereKoskela:
Do the results look good? If yes, then we should check the hard-coded values because Joe has left a comment to say that that needs to be done, but I think we can trust the test provided they're right.

@eldonb has implemented a method for computing expected normalised SFSs for Xi-coalescents here: http://page.math.tu-berlin.de/~eldon/programs.html. @TPPSellinger, could you run that and verify that our analytical SFS expectations in verification.py are right please?

TPPSellinger:
Thanks for the link  @JereKoskela ! I will try to check the SFS but because of schedule I might only do it this afternoon or tomorrow morning.

I tried to do the PR but I think my version of msprime was one or two weeks old. @jeromekelleher  is it possible for you to keep the changes only related to the dirac coalescent or should I rebase and do the PR again ?

TPPSellinger:
> Do the results look good? If yes, then we should check the hard-coded values because Joe has left a comment to say that that needs to be done, but I think we can trust the test provided they're right.

Yes they look good but they always had :/  . That's why I am not so confident about them and another test would help me to be truly convinced.

JereKoskela:
I think everything we've done thus far has affected all branches equally, so would not have shown up in the normalised SFS. If you have the time and energy, you could certainly put in more parameter combinations, or come up with something new if you prefer.

TPPSellinger:
I have one last small question. Is it normal that when we have almost nor multiple merger, c=0.1 and psi=0.01 we seem to have 2 as much mutation than the hudson model (with same mutation and recombination rate of course) ?

JereKoskela:
Yes, I think that's the rationale for the mysterious factor of two that we removed at one stage.

The coalescence rate in the Hudson implementation in `msprime` is `n(n - 1)`, not `n(n - 1) / 2`. So our imlementation is correct either way, but it's probably best to keep the factor of two, match the Hudson model, and eventually, document that we've done that.

TPPSellinger:
Okay  I will modify the PR to ass this 2 in order to match the Hudson's model. I have one last issue. Here are the results for the beta coalescent (alpha=1.95) : 
n=100_alpha=1 95_growth=0_rec_check


Why do we have so few mutations with the beta coalescent ? 
Here are the parameters I used : 

~~~~
def run_xi_beta_recombinations(self):
        basedir = "tmp__NOBACKUP__/xi_beta_recombinations"
        for alpha in [1.95]:
            self.verify_recombination(
                basedir, f"n=100_alpha={alpha}", sample_size=100, Ne=10000, r=1e-8,
                L=10**6, model=msprime.BetaCoalescent(alpha=alpha))


    def run_xi_dirac_recombinations(self):
        basedir = "tmp__NOBACKUP__/xi_dirac_recombinations"
        for psi in [0.001]:
            for c in [0.01]:
                self.verify_recombination(
                    basedir, f"n=100_psi={psi}_c={c}",
                    sample_size=100, Ne=10000, r=1e-8,
                    L=10**6, model=msprime.DiracCoalescent(psi=psi, c=c))

    def run_Hudson_recombinations(self):
        basedir = "tmp__NOBACKUP__/hudson_recombinations"
        self.verify_recombination(
            basedir, f"n=100_hudson",
            sample_size=100, Ne=10000, r=1e-8,
            L=10**6, model="hudson")

~~~~

JereKoskela:
> Why do we have so few mutations with the beta coalescent ?

The Beta-coalescent implementation follows the Scweinsberg model, which is known to predict implausibly short timescales. Typically, effective population sizes need to be orders of magnitude (often *many* orders of magnitude) larger than census population sizes in order to see a reasonable amount of diversity.

TPPSellinger:
Okay ! So I only need to build a small SFS check in the verification.py for the dirac coalescent based on Bjarki's code and our job is done (if results are good) ?

JereKoskela:
> Okay ! So I only need to build a small SFS check in the verification.py for the dirac coalescent based on Bjarki's code and our job is done (if results are good) ?

It would be helpful if you could double-check the SFS for the Beta-coalescent at the same time. I think Bajrki's code handles both kinds of Xi-coalescent. Since you've already done your qqplot test for the Beta-coalescent, we will then have solid verification of both multiple merger models.

TPPSellinger:
Okay !  I will do this.

jeromekelleher:
Just FYI - I can sort out the merge conflict here if you're not comfortable with doing that @TPPSellinger, just let me know. Great work both!

TPPSellinger:
Thank you @jeromekelleher ! I don't really know how to do it. So if it doesn't take too much of your time I would appreciate. Otherwise I can also try to solve the conflict.

jeromekelleher:
Well it's simplest if I do this when you're finished with your updates. So, just let me know when you're ready to merge this in and I'll fix it up then.

TPPSellinger:
@JereKoskela  have you been using Bjarki's c code ?

TPPSellinger:
I was wondering if the two functions compare_xi_beta_sfs and compare_xi_dirac_sfs in verification.py have been checked ?  Whatever the model I use or the parameter values, I seem to always have the same observed SFS (which I don't think is possible).

Here are some example of results : 
dirac,psi=0.1,c=1
n=10_psi=0 1
dirac , psi=0.9 , c=1
n=10_psi=0 9
beta, alpha=1.1
n=10_alpha=1 1

I will try to re-implement those functions and see if it solves the problem.

JereKoskela:
I don't know to what extent they've been validated, but you're right that something looks off. Well spotted! I might be imagining things but I think your third picture (the Beta one) is different to the other two, which do look far too similar.

TPPSellinger:
True, the last one is slightly different. But I would expect way more difference. Here is the beta 1.9 : 
n=10_alpha=1 9
I am currently trying to understand what the code is doing. I suspect the code to not correctly handle multiple merger events,  explaining why they all look like a Wright Fisher model.

jeromekelleher:
> I was wondering if the two functions compare_xi_beta_sfs and compare_xi_dirac_sfs in verification.py have been checked ? Whatever the model I use or the parameter values, I seem to always have the same observed SFS (which I don't think is possible).

No, they probably weren't checked very thoroughly (or at all). Thanks for digging in to this @TPPSellinger!

TPPSellinger:
I think I found the problem (for the dirac coalescent at least). If I simulate data with c=100 and not 1, the fit is perfect (cf below). This mean that the observed one is with c=100 and the expected one with c=1. I calculated the expected SFS with Bjarki's code, and according to him c is fixed to 1 (by default and cannot be changed for the moment). @JereKoskela is the c in msprime equivalent to the c in his work ?  I will send an email to Bjarki to ask him a second time to make sure c is fixed to 1 in his code. 

 psi=0.9 
n=10_psi=0 9
psi=0.5
n=10_psi=0 5
psi=0.3
n=10_psi=0 3
psi=0.1
n=10_psi=0 1

JereKoskela:
The `c` in msprime should be equivalent to the `c` in Bjarki's code, but in msprime it definitely does not have a fixed, hard-coded value. It is possible (likely, even) that the expected SFSs that were hard-coded into these function calls in `verification.py` were calculated with Bjarki's code, so if the issue of the fixed `c` was not spotted before, then perhaps it could explain why the fits are wrong.

TPPSellinger:
> > I was wondering if the two functions compare_xi_beta_sfs and compare_xi_dirac_sfs in verification.py have been checked ? Whatever the model I use or the parameter values, I seem to always have the same observed SFS (which I don't think is possible).
> 
> No, they probably weren't checked very thoroughly (or at all). Thanks for digging in to this @TPPSellinger!

I went through them very carefully, checking line by line for any potential mistake. I think  they are good.

TPPSellinger:
> The `c` in msprime should be equivalent to the `c` in Bjarki's code, but in msprime it definitely does not have a fixed, hard-coded value. It is possible (likely, even) that the expected SFSs that were hard-coded into these function calls in `verification.py` were calculated with Bjarki's code, so if the issue of the fixed `c` was not spotted before, then perhaps it could explain why the fits are wrong.

I don't really understand. I hardcoded those values from Barki's code (where c is normally equal to 1).  Then I calculate the expected SFS, setting c=1  (since I cannot calculate the expected SFS for c other than 1). Those are the first results I displayed (where the fit is bad). Then, if I set c=100 for the observed SFS. When c=100, it seems to fit to the expected SFS I calculated with Bjarki's code (where c=1). 

I think we should trust msprime because when c increases, we tend to the expected value. I check for higher of values c (up to 1000) the fit is still perfect. I would guess that if Bjarki's code doesn't have an input for c would be because it calculates some kind of limit (which we tend to when c increases).

JereKoskela:
Ah, I think I know what's going on. I _think_ that code was written for the benefit of [this](https://reader.elsevier.com/reader/sd/pii/S0040580916300053?token=8B43C73A39BEF02CAADEEA47B0D3EDFE3C7FEDA4866E8627FF0171084018323DA88F69830CCC15E627568832A35D9CED) paper. Figure 3 deals with the Dirac coalescent, which for them has no Kingman component - essentially c = infinity. I think you might well be right with your limiting reasoning - that's encouraging! Do the Beta-coalescents give a good fit? Those are definitely parametrised the same way as they are in `msprime`.

TPPSellinger:
Results for the Beta are good but not perfect. Thing is that with msprime we have an extra parameter (the truncation point). I don't really know to which value I should set it and how much "deviation" is to be expected because I this.

JereKoskela:
Under the current implementation, `truncation_point = 1` is the standard Beta-Xi-coalescent. That should match Bjarki's code, though you should expect very high variance.

TPPSellinger:
I think  we have a problem with the beta coalescent. I set truncation point to 1. But results makes no sense.  According to the expected SFS, the weighet of num_leave=1 should increase with alpha diminishing (when more multiple collisions are expected). This results is similar in the dirac. Hoiwever, if we check the observed one, the weight of num_leave=1 increases with alpha (cf below). I don't think those results can be accounted to high variance since I set the number of replicates to 50 000.

alpha=1.1
n=10_alpha=1 1
alpha=1.3
n=10_alpha=1 3
alpha=1.5
n=10_alpha=1 5
alpha=1.9
n=10_alpha=1 9

 I find those results very troublesome. Here are the observed value (not easy to read on the graphs):
SFS for alpha= 1.1
[0.34242653231014397, 0.18799971598754583, 0.12816882500379684, 0.09668755561734022, 0.07450433020775742, 0.05984360633138009, 0.04812747311006567, 0.0378626459642046, 0.024379315467768177]
SFS for alpha=  1.3
[0.34764082516028727, 0.18669803626029272, 0.12641624190821105, 0.09395345825189876, 0.07360450896100897, 0.05873682571060857, 0.04782505090729617, 0.038353361519183546, 0.026771691321211363]
SFS for alpha=  1.5
[0.3556175569154827, 0.18553191107362424, 0.12517773482556221, 0.09131451186199692, 0.07086960311106953, 0.05692222252270775, 0.047382460835732915, 0.038420891202967, 0.028763107650859373]
SFS for alpha=  1.9
[0.3742328708682222, 0.18051073366987624, 0.119284070674048, 0.08609720713277744, 0.06772013078293714, 0.053824775222386885, 0.046474137557685015, 0.03830718705181216, 0.033548887040255125]

JereKoskela:
I would try a few other random seeds, and a somewhat higher replicate number, if that's feasible. The expected vectors look plausible, and the observed vectors are good matches at alpha = 1.9, and worse at alpha = 1.1. We expect variance to decrease with alpha, so it could still be that that's all that is going on.

If the observations don't fluctuate between above and below the expected SFS with changing random seed, and improve with replicate numbers, then I'm sure you're right and something else is the matter.

TPPSellinger:
I am afraid results didn't change. I set the number of replicates to 10⁵ and changed random seed. It's strange, we have more of a U shaped observed SFS with alpha increasing. Have you observed this when previously working with the beta coalescent ?  Here are the results:
alpha=1.9

n=10_alpha=1 9

alpha=1.5
n=10_alpha=1 5
alpha=1.3
n=10_alpha=1 3
alpha=1.1
n=10_alpha=1 1

Here are the observed values : 
 
SFS for alpha=1.1
[0.34500625434951115, 0.18756264726270644, 0.12699734743435856, 0.09576015699157647, 0.075212464816101, 0.05953118616929164, 0.0474370585045796, 0.03764537079622483, 0.024847513675650724]
SFS for alpha=1.3
[0.34816245955317193, 0.1869371611261868, 0.12684893277826045, 0.09395685341838994, 0.07309197483352368, 0.05853759589169457, 0.047549487780003175, 0.03823366710094917, 0.02668186751781788]
SFS for alpha=1.5
[0.3566942874852179, 0.1856909827642993, 0.12339644798690337, 0.09069913990769628, 0.071547881672636, 0.05668891820756783, 0.046805469788929, 0.03905507088808089, 0.02942180129867061]
SFS for alpha=1.9
[0.3730727168743765, 0.18207084405507953, 0.11803728032719094, 0.0864220022518443, 0.06735779478238861, 0.05477541414667995, 0.045577438607925236, 0.03921207327231111, 0.033474435682205164]

PS: I think we can consider the dirac coalescent okay for the moment. When c tends toward 0 we have a kingman SFS and when c increases it tends to the limits SFS for c= infinity. I like this behaviour. I will try to set a check for intermediate values of c but I am optimistic.

TPPSellinger:
I quickly checked the beta coalescent rate calculations  in msprime.c and two pieces of equation don't seem intuitive to me. In the function beta_integrand  we have : 
~~~~
exponent = (-1 - alpha) * log_x + (alpha - 1) * log_1_minus_x
                + log(4) - gsl_sf_lnbeta(2 - alpha, alpha);
    ret = exp(exponent) - exp(exponent + r_max + log(ret));
~~~~
1) I don't understand where the :  -1 - alpha comes from. Intuitively I would expect  (alpha - 1) or (1-alpha)
2) I don't understand why we don't have the number of ancestors in gsl_sf_lnbeta(2 - alpha, alpha). In my mind alpha,  the number of ancestors and the  number of individual coalescing should be parameter of the Beta function.

JereKoskela:
> 1. I don't understand where the :  -1 - alpha comes from. Intuitively I would expect  (alpha - 1) or (1-alpha)

A Beta(a, b) density has kernel `x^(a - 1) * (1 - x)^(b - 1)`. Substitute in Beta(2 - a, a), and you find `x^(-1 - a) * (1 - x)^(a - 1)`. This is equation (31) in Bjarki's notes.

> 2. I don't understand why we don't have the number of ancestors in gsl_sf_lnbeta(2 - alpha, alpha). In my mind alpha,  the number of ancestors and the  number of individual coalescing should be parameter of the Beta function.

The number of ancestors enters through the computation of `r[l]`s a few lines earlier. It would appear as a parameter like you describe if we were to actually compute the integral analytically rather than numerically, but that is a rather horrible task here.

I have an idea for what might be going on, which is that I don't think we're drawing our per-lineage merger probability from exactly the correct distribution. We're drawing it from Lambda, when it should be drawn from a density proportional to `beta_integrand`. I dealt with a similar issue with Beta-Xi-Sim using rejection sampling (see the appendix [here](https://www.degruyter.com/view/journals/sagmb/17/3/article-20170011.xml)), but that suffers from numerical instability for large samples. I'll try to verify that that's actually the case, and think about how to solve it if it is, but won't get to it today. I'm not 100% sure of this yet, so it's worth thinking about other possibilities too.

TPPSellinger:
Okay! I have a test idea test that could help us point the problem in the distribution of multiple collision. I will try to implement it and  display results thursday afternoon.

TPPSellinger:
Thanks to Bjarki's code I finished testing the dirac coalescent. Results are good. Here there are : 
c=1 , psi=0.1
n=10_psi=0 1_c=1
c=1 , psi=0.3
n=10_psi=0 3_c=1
c=1 , psi=0.5
n=10_psi=0 5_c=1
c=1 , psi=0.9
n=10_psi=0 9_c=1
c=10 , psi=0.1
n=10_psi=0 1_c=10
c=10 , psi=0.3
n=10_psi=0 3_c=10
c=10 , psi=0.5
n=10_psi=0 5_c=10
c=10 , psi=0.9
n=10_psi=0 9_c=10

We could test for c=100, but I think we can trust the dirac. I modified the PR to add to test where c is finite and add extra comment to explain what the test does. However I don't think results for the beta coalescent are perfect enough yet.  @jeromekelleher, I think  you can do the merger of the verification.py and everything related to the dirac coalescent. @JereKoskela  and I will keep working on the beta coalescent. I think it's easier if we open a new PR when we fixed everything with the beta coalescent.

jeromekelleher:
Excellent, thanks @TPPSellinger and @JereKoskela! So is it OK if I tidy this PR up and merge then?

JereKoskela:
> Excellent, thanks @TPPSellinger and @JereKoskela! So is it OK if I tidy this PR up and merge then?

Yes, that would be great. Let us know if you spot anything unclear. There is a problem lurking somewhere in the Beta-coalescent implementation which is causing it to simulate considerably more multiple mergers than it should, but we'll deal with that in a separate PR.

In somewhat related news, I've worked out how to implement the Beta-coalescent without numerical integration (modulo the above bug). I'll contribute that too in a future update, once the problem is fixed.

jeromekelleher:
> In somewhat related news, I've worked out how to implement the Beta-coalescent without numerical integration (modulo the above bug). I'll contribute that too in a future update, once the problem is fixed.

Fantastic news! I look forward to seeing how you managed that. I'll clean this up and merge so, I'll ping you/open some issues if I spot problems.

jeromekelleher:
I've had a go at tidying this up @TPPSellinger, but I'm afraid it's in a bit of a mess. It looks like you copied some old files that you had lying around into the commit rather than updating the current state of the code, and  there's therefore a large amount of changes that are deleting the recent updates. I'm therefore a bit stuck, because I don't know what you intended to add and not and I can't bring this up to date.

I think the only thing we can do now is to make a new branch, and a new PR from that:
```
git fetch upstream
git checkout upstream/master
git checkout -b dirac-fixes
```
This gives you a fresh branch "dirac-fixes", which you now want to update with the changes that you  made before. The simplest way of doing this is to go through the diff here on github (use the [files changed](https://github.com/tskit-dev/msprime/pull/935/files) tab above), and copy the changes that you wanted to make into the new branch. Leave out all the stuff like changing copyright headers, etc. Once this is done, you can then commit, push to github and open a new PR.

Does that sound OK?

TPPSellinger:
Yes I understand Jerome. I apologize for this. I noticed that while doing the PR. I will make a new PR tomorrow morning with only all the required changes for the dirac coalescent and verification.py.

jeromekelleher:
No worries, thanks for all the work @TPPSellinger. There's a steep learning curve for git, and a lot of ways to go wrong! Once you get into a good workflow though it all just works most of the time.

jeromekelleher:
Close in favour of #953

PR #939, opened 2020-03-30
Reject zero mass_to_next_recomb in poisson sampling for dtwf

### LINKED ISSUE #938: Assertion failure in DTWF with high recombination
When we run this simulation:
```python

seed = 1
ts = msprime.simulate(
    100, Ne=10**4, recombination_rate=1e-4, length=10**8,
    model="dtwf", random_seed=seed)
```
we get
```
python3: lib/msprime.c:1581: msp_dtwf_generate_breakpoint: Assertion `k > start' failed.
Aborted (core dumped)
```

Would you mind looking into this please @daniel-goldstein? 

Moved from #936.

daniel-goldstein:
In sampling breakpoints for DTWF, we use `gsl_ran_exponential` to get the recombination mass from `start` to step forward, sampled from a Poisson distribution. `gsl_ran_exponential` returns values >= 0, though we cannot have two breakpoints in the same location so we want values strictly greater than 0. Sure enough, the day that it would return 0 came at last in #938. I verified that `mass_to_next_recomb == 0` on the iteration that the assert `k > start` failed.

To address this, just reject values that == 0 and resample. Manual test of the sim in #938 no longer fails the assert.

Resolves #938

### DIFF
--- a/lib/recomb_map.c
+++ b/lib/recomb_map.c
@@ -210,7 +210,9 @@ recomb_map_sample_poisson(recomb_map_t *self, gsl_rng *rng, double start)
     double left_bound, mass_to_next_recomb;
 
     left_bound = self->discrete ? start + 1 : start;
-    mass_to_next_recomb = gsl_ran_exponential(rng, 1.0);
+    do {
+        mass_to_next_recomb = gsl_ran_exponential(rng, 1.0);
+    } while (mass_to_next_recomb == 0.0);
 
     return recomb_map_shift_by_mass(self, left_bound, mass_to_next_recomb);
 }


daniel-goldstein:
@jeromekelleher Does rejection sampling sound like a fine solution here?

lib/recomb_map.c
```diff
@@ -210,7 +210,9 @@ recomb_map_sample_poisson(recomb_map_t *self, gsl_rng *rng, double start)
     double left_bound, mass_to_next_recomb;
 
     left_bound = self->discrete ? start + 1 : start;
-    mass_to_next_recomb = gsl_ran_exponential(rng, 1.0);
+    do {
+        mass_to_next_recomb = gsl_ran_exponential(rng, 1.0);
+    } while (mass_to_next_recomb == 0);
```
jeromekelleher: I can't find out whether 0 is guaranteed to be the same as 0.0 - can you change to 0.0 for peace of mind please? (Or educate me!)

jeromekelleher:
> @jeromekelleher Does rejection sampling sound like a fine solution here?

Rejection sampling is the right approach here, it's what we do in practise in the other situations where these 1 in a billion corner cases. Writing stochastic simulators tends to give you a keen appreciate for Murphy's Law!

lib/recomb_map.c
daniel-goldstein: I also can't find out through a quick search, which is scary, so I'll just change it 😅

PR #940, opened 2020-03-31
Bring back genetic and physical space conversion in Recombination Map

### LINKED ISSUE #919: Reimplement recombination_map.physical_to_genetic()
It would be useful to be able to provide a list of physical positions to a genetic map and obtain a list of genetic positions. In my particular case, I want the "recombination distances" between a set of physical loci to help guide inference, which I can get by using `np.diff` once I've got the genetic coords from the physical ones.

There is already a method [`RecombinationMap.physical_to_genetic()`](https://github.com/tskit-dev/msprime/blob/ed4879ad923eeb0119752f51e5f96c5e9cb9f1d2/msprime/simulations.py#L1120) but it raises a "Genetic space is no longer supported" error. I suggest we reinstate this using something like the following code:

```
def physical_to_genetic(recombination_map, input_physical_positions):
    map_pos = recombination_map.get_positions()
    map_rates = recombination_map.get_rates()
    map_genetic_positions = np.insert(np.cumsum(np.diff(map_pos) * map_rates[:-1]), 0, 0)
    return np.interp(input_physical_positions, map_pos, map_genetic_positions)
```

daniel-goldstein:
These changes reinstate the conversion methods between genetic and physical coordinates on the Python `RecombinationMap`. Genetic coordinate is synonymous with the cumulative rate of recombination from the start of the sequence to the corresponding physical location. Methods to do this conversion already existed in C.

This definition differs _slightly_ from the previous semantics because there is no longer the separate notion of genetic loci as a parameter to the map. Given the above definition, and the fact that genetic loci no longer exist, the method for computing the discrete genetic coordinate doesn't seem to make much sense and remains deprecated.

 The tests are those that existed before [this PR](https://github.com/tskit-dev/msprime/commit/6c08fcf2dd01c04f85391323a89fe9dd1bca303c), save for some clean up to remove varying number of loci and the complications that result. They are also tested against the proposed implementation of `physical_to_genetic` in #919 

Resolves #919 

cc @jeromekelleher @hyanwong

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -2194,6 +2194,50 @@ out:
 }
 
 static PyObject *
+RecombinationMap_position_to_mass(RecombinationMap *self, PyObject *args)
+{
+    PyObject *ret = NULL;
+    double position;
+
+    if (RecombinationMap_check_recomb_map(self) != 0) {
+        goto out;
+    }
+    if (!PyArg_ParseTuple(args, "d", &position)) {
+        goto out;
+    }
+    if (position < 0) {
+        PyErr_SetString(PyExc_ValueError, "Position must be >= 0");
+        goto out;
+    }
+
+    ret = Py_BuildValue("d", recomb_map_position_to_mass(self->recomb_map, position));
+out:
+    return ret;
+}
+
+static PyObject *
+RecombinationMap_mass_to_position(RecombinationMap *self, PyObject *args)
+{
+    PyObject *ret = NULL;
+    double mass;
+
+    if (RecombinationMap_check_recomb_map(self) != 0) {
+        goto out;
+    }
+    if (!PyArg_ParseTuple(args, "d", &mass)) {
+        goto out;
+    }
+    if (mass < 0) {
+        PyErr_SetString(PyExc_ValueError, "Recombination mass must be >= 0");
+        goto out;
+    }
+
+    ret = Py_BuildValue("d", recomb_map_mass_to_position(self->recomb_map, mass));
+out:
+    return ret;
+}
+
+static PyObject *
 RecombinationMap_get_total_recombination_rate(RecombinationMap *self)
 {
     PyObject *ret = NULL;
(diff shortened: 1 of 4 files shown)


hyanwong:
LGTM. Nice to test against my python implementation, thanks!

_msprimemodule.c
```diff
@@ -2193,6 +2193,42 @@ RecombinationMap_init(RecombinationMap *self, PyObject *args, PyObject *kwds)
...
+static PyObject *
+RecombinationMap_position_to_mass(RecombinationMap *self, PyObject *py_position)
+{
+    PyObject *ret = NULL;
+    double position;
+
+    if (RecombinationMap_check_recomb_map(self) != 0) {
+        goto out;
+    }
+    if (!PyArg_ParseTuple(py_position, "|d", &position)) {
```
jeromekelleher, line 2205: The "|d" makes position an optional argument, which we don't want here. Just get rid of the |

_msprimemodule.c
```diff
@@ -2193,6 +2193,42 @@ RecombinationMap_init(RecombinationMap *self, PyObject *args, PyObject *kwds)
     return ret;
 }
 
+static PyObject *
+RecombinationMap_position_to_mass(RecombinationMap *self, PyObject *py_position)
```
jeromekelleher, line 2197: The standard convention here is the make the second parameter ``*args``, since this is actually the args Python tuple, not a single Python object.

daniel-goldstein:
@jeromekelleher Ready for another check

jeromekelleher:
Perfect! Just needs a quick rebase.

petrelharp:
Sorry, a bit late to this (great and very useful!) party, but maybe calling it something more descriptive than "recombination mass" would make it less ambiguous what those functions do? Like maybe "expected number of recombinations"?  Or, just "genetic map position", since that's what the genetic map means? Why not `physical_to_genetic` instead of `position_to_mass`?

Thanks for the work here, though!!

jeromekelleher:
These are just the low-level methods which mirror the C code @petrelharp; the idea here was to reproduce the old API for compatability as simply as we could. We can (and should!) totally add a nicer high-level API on top now.

petrelharp:
Ok, sounds good.

hyanwong:
I just tried `chr20_map.physical_to_genetic(np.array([array, of, positions]))` and it came back with

```
  File "/home/yan/.local/lib/python3.7/site-packages/msprime/simulations.py", line 875, in physical_to_genetic
    return self._ll_recombination_map.physical_to_genetic(physical_x)
TypeError: only size-1 arrays can be converted to Python scalars
```

I was expecting to be able to pass in a numpy array of positions here?

daniel-goldstein:
Sorry about that you'll have to map the function across your array for now. The old API that we need to keep accepts just single value. A nice change to achieve the higher-level numpy-like API would be to accept either a scalar or array and do the appropriate thing.

jeromekelleher:
Let's take this discussion over to #902. Can you write out what you'd like to see there please @hyanwong? Also good to get your thoughts there @petrelharp.

PR #947, opened 2020-04-07
Speeding up dtwf sims with large Ne

### LINKED ISSUE #878: [C code] DTWF performance
Run a performance analysis on the DTWF for a large simulation and see where the time is being spent. Since it seems like we're not using the Fenwick tree at all within DTWF simulations, it may be worthwhile 'turning off' Fenwick tree updates during a DTWF sim (which we should be able to do easily enough now, with the new abstractions). 

If we do, we can update the Fenwick tree after we finish the DTWF phase so that the simulation state is correct (we want to be able to transition into a Hudson simulation if necessary).

ivan-krukov:
Replaced a large (Ne) array of pointers of parental segments with an AVL tree to store the offspring for each parent. This results in considerable speedups for the simulations with large Ne. Fixes #878.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -3882,11 +3882,47 @@ out:
     return ret;
 }
 
-/* List structure for collecting segments by parent */
-typedef struct _segment_list_t {
+// AVL tree item and linked list to store a sparse record of parents
+typedef struct _parent_record_t {
+    struct _parent_record_t *next;
+    uint32_t parent;
     avl_node_t *node;
-    struct _segment_list_t *next;
-} segment_list_t;
+} parent_record_t;
+
+static parent_record_t*
+parent_record_alloc(avl_node_t* node, uint32_t parent)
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Sorry @ivan-krukov, this seems to have slipped under the radar somehow. Would it be possible to bring this up to date please? We have some benchmarks now, so it should be easy to quantify the improvements.

jeromekelleher:
@ivan-krukov do you think we can bring this up to date and get ready for merging? The merge window for msprime 1.0 is quickly approaching, and we'd need to have this sorted out in the next week if it's going to get in for the release. It would be a shame for it to miss out, since it's a significant perf improvement.

jeromekelleher:
Closing this as it's a long way behind the head of main. It would be nice to incorporate this idea back into the DTWF simulation at some point, as it should have a significant impact.

PR #948, opened 2020-04-10
Store a placeholder for tree sequences in provenance

### LINKED ISSUE #928: Update the ``from_ts`` provenance

benjeffery:
Fixes #928 

This is based on #934 in that it puts an object into provenance to represent the "current" tree sequence, which `parse_provenance` then replaces. This will work with any future tskit provenance work as discussed in #934.

### DIFF
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -162,8 +162,7 @@ def mutate(
     argspec = inspect.getargvalues(inspect.currentframe())
     parameters = {
         "command": "mutate",
-        **{arg: argspec.locals[arg] for arg in argspec.args
-           if arg not in ["tree_sequence"]}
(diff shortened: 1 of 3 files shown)


petrelharp:
This looks promising! I'm curious to see how it works.

jeromekelleher:
Can we rebase this please?

benjeffery:
@jeromekelleher rebased!

petrelharp:
Nice. This general idea is pretty amazing.

Issue #942, opened 2020-04-01
[Sweeps] Assertion failures


jeromekelleher:
This code fragment leads to an assertion failure:
```python
import msprime
import numpy as np

refsize = 1e4
seqlen = 1e4
mu = 2.5e-4
rho = 2.5e-4

mod = msprime.SweepGenicSelection(
    reference_size=refsize,
    position = np.floor(seqlen/2),
    start_frequency=1.0 / (2 * refsize),
    end_frequency=1.0 - (1.0 / (2 * refsize)),
    alpha=1000.0,
    dt=1.0/(40 * refsize)
)
ts = msprime.simulate(
    10,
    model=mod,
    length=seqlen,
    recombination_rate=rho,
    mutation_rate=mu,
    random_seed=1,
    # Change to Hudson after sweep finishes
    demographic_events=[msprime.SimulationModelChange()])
```

We get 
```
python3: lib/msprime.c:2127: msp_recombination_event: Assertion `x != NULL' failed.
```

I also got 
```
python3: lib/msprime.c:4299: msp_change_label: Assertion `node != NULL' failed.
```
on a different seed. Looks like some basic bug in model changing.

jeromekelleher:
I think there's two things happening here. The first is that we've got **crazy** high levels of recombination during the sweep phase, so we must be missing a scaling factor here somewhere. We're getting lots of *very* short segments.

We shouldn't bomb out on an assertion even in this case though, so I'd like to get to the bottom of what's happening here and return a proper error. After doing a bit of digging, it looks to me like we're hitting some limit of resolution on the ``msp_init_segments_and_compute_breakpoint`` which we should try to detect and understand.

Would you mind taking a look @daniel-goldstein? Also, if you're in there, I think the rejection sampling should be done with a simpler do-while loop. I'm not convinced the current recursive method is doing what it should.

daniel-goldstein:
The resolution problem was definitely a top-level issue, but a symptom of these segments getting very short. The rejection sample condition tested `k == y->left` and we were getting `k` that printed the same but was slightly smaller. It is probably safer to reject if `k <= y->left && y->prev == NULL` (in addition to using a do-while).

I haven't yet identified the deeper problem of why we're getting such short segments. However, I am suspicious of the code in `msp_run_sweep` that deals with recombination mass.
https://github.com/tskit-dev/msprime/blob/c9a27e9f825921dfe586810736af01dc7813833b/lib/msprime.c#L4392-L4419

[...]

PR #951, opened 2020-04-13
Safely account for double inequality when calculating breakpoints

daniel-goldstein:
In the rejection sampling of breakpoints, this change:
- Uses a do-while instead of recursion
- Uses an inequality instead of `==` for the rejection case because `==` was sometimes failing due to floating point precision.

A more detailed description of the problem where this arose is discussed in #942. This PR only addresses the assertion error in the issue and not any additional problems concerning rates.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2076,12 +2076,9 @@ msp_init_segments_and_compute_breakpoint(msp_t *self, label_id_t label, segment_
     t = fenwick_get_cumulative_sum(tree, y->id);
     x = y->prev;
 
-    k = recomb_map_mass_to_position(&self->recomb_map, y->right_mass - (t - h));
-
-    /* Reject if fallen directly on y left when not allowed to */
-    if (k == y->left && y->prev == NULL) {
-      return msp_init_segments_and_compute_breakpoint(self, label, x_ret, y_ret);
-    }
+    do {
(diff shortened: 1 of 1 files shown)


petrelharp:
So, sometimes you'd have `k<left`? Makes sense, and good solution.

daniel-goldstein:
Thanks, @petrelharp! And yup, that's exactly what happened.

lib/msprime.c
```diff
@@ -2076,12 +2076,9 @@ msp_init_segments_and_compute_breakpoint(msp_t *self, label_id_t label, segment_
...
 
-    k = recomb_map_mass_to_position(&self->recomb_map, y->right_mass - (t - h));
-
-    /* Reject if fallen directly on y left when not allowed to */
-    if (k == y->left && y->prev == NULL) {
-      return msp_init_segments_and_compute_breakpoint(self, label, x_ret, y_ret);
-    }
+    do {
+        k = recomb_map_mass_to_position(&self->recomb_map, y->right_mass - (t - h));
+    } while ((y->left >= k) && y->prev == NULL);
```
jeromekelleher, line 2081: Do we need the first set of parens here? I don't think there's any ambiguity (>= should bind as tightly as ==, right?), and it's confusing to have parens around the first clause but not the second

PR #957, opened 2020-04-19
Multiple merger

JereKoskela:
This is a seemingly corrected implementation of the Beta-Xi-coalescent, in that all of the tests we've added look good. @TPPSellinger, could you rerun all of your tests on this at your end to double-check I haven't missed anything?

I've also cleaned up the Dirac-Xi-coalescent a little bit. The latest implementation is quite a bit simpler than the original, so I got rid of a few separate functions that had turned into simple, one-line calculations.

@jeromekelleher, there is a TODO note on line 3735 of msprime.c explaining that we disallow store_full_arg tracking with multiple mergers because events get cancelled. That no longer happens in either the Beta or Dirac cases. Do we need to change something in the code to allow store_full_arg, or is it just a case of deleting the comment? Fairly big blocks of code have changed, so let me know if you have any questions or comments too, of course.

Ping @eldonb as well to keep you in the loop.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -28,7 +28,6 @@
 #include <gsl/gsl_cdf.h>
 #include <gsl/gsl_statistics_int.h>
 #include <gsl/gsl_sf.h>
-#include <gsl/gsl_integration.h>
 
 #include "util.h"
 #include "avl.h"
@@ -78,10 +77,11 @@ ran_inc_beta_its(gsl_rng *r, double a, double b, double upper_bound)
  * the upper bound used with ran_inc_beta_its().
  */
 static double
-ran_inc_beta(gsl_rng *r, double a, double b, double x, double upper_bound)
+ran_inc_beta(gsl_rng *r, double a, double b, double x)
 {
-    if (upper_bound < 0.1) {
-        return ran_inc_beta_its(r, a, b, upper_bound);
+    double ub = gsl_sf_beta_inc(a, b, x);
+    if (ub < 0.1) {
+        return ran_inc_beta_its(r, a, b, ub);
     } else {
         return ran_inc_beta_rej(r, a, b, x);
     }
@@ -3732,10 +3732,6 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
                 ret = self->common_ancestor_event(self, ca_pop_id, label);
                 if (ret == 1) {
                     /* The CA event has signalled that this event should be rejected */
-                    /* TODO things are more complicated in the store_full_arg case
-                     * because we will have stored a node and bunch of edges recording the
-                     * ARG. For now we disallow ARG recording for multiple merger
-                     * coalescents */
                     self->time -= t_wait;
                     ret = 0;
                 }
@@ -4494,9 +4490,11 @@ msp_run(msp_t *self, double max_time, unsigned long max_events)
     if (self->store_full_arg && ! (
             self->model.type == MSP_MODEL_HUDSON
             || self->model.type == MSP_MODEL_SMC
-            || self->model.type == MSP_MODEL_SMC_PRIME)) {
+            || self->model.type == MSP_MODEL_SMC_PRIME
+            || self->model.type == MSP_MODEL_BETA
+            || self->model.type == MSP_MODEL_DIRAC)) {
         /* We currently only support the full ARG recording on the standard
-         * coalescent. */
+         * coalescent, SMC, or multiple merger coalescents. */
         ret = MSP_ERR_UNSUPPORTED_OPERATION;
         goto out;
     }
@@ -5751,50 +5749,14 @@ dirac_model_rate_to_generation_rate(simulation_model_t *model, double rate)
     return rate * x / ( 4 * model->reference_size);
 }
 
-double
-compute_falling_factorial_log(unsigned int m)
-{
-    unsigned int l = 0;
-    double ret = 1.0;
-    while (l < m){
-        l++;
-        ret *= (4.0 - l + 1.0);
-    }
-    return gsl_sf_log(ret);
-}
-
-/* This calculates the rate given by Eq (2) in the notes
- */
-double
-compute_dirac_coalescence_rate(unsigned int num_ancestors, double psi, double c)
-{
-    double ret = 0;
-    double b = num_ancestors;
-
-    assert(b > 0);
-    assert(psi > 0);
-    assert(psi < 1);
-    assert(c >= 0);
-    ret = 2*(b * (b - 1) / 2 + c);
-    return ret;
-}
-
-static double
-msp_dirac_compute_coalescence_rate(msp_t *self, unsigned int num_ancestors)
-{
-    double psi = self->model.params.dirac_coalescent.psi;
-    double c = self->model.params.dirac_coalescent.c;
-
-    return compute_dirac_coalescence_rate(num_ancestors, psi, c);
-}
-
 static double
 msp_dirac_get_common_ancestor_waiting_time(msp_t *self, population_id_t pop_id,
         label_id_t label)
 {
     population_t *pop = &self->populations[pop_id];
     unsigned int n = (unsigned int) avl_count(&pop->ancestors[label]);
-    double lambda = msp_dirac_compute_coalescence_rate(self, n);
+    double c = self->model.params.dirac_coalescent.c;
+    double lambda = 2 * (gsl_sf_choose(n, 2) + c);
 
     return msp_get_common_ancestor_waiting_time_from_rate(self, pop, lambda);
 }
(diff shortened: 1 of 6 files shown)


TPPSellinger:
Sorry for the delay. Great Work @JereKoskela ! All the expected SFS I checked work display perfect fit (beta and dirac). Same for recombination graphics. I will test more extreme cases and increase number of repetitions to be sure. Should We add a test in verification.py for sample of size 3 (to check probability of multiple merger) ?

lib/msprime.c
```diff
@@ -5881,159 +5856,61 @@ beta_generation_rate_to_model_rate(simulation_model_t *model, double rate)
...
-
-    ret = msp_beta_compute_coalescence_rate(self, n, &lambda);
-    if (ret != 0) {
-        assert(ret < 0);
-        /* An error occured, and we signal this back using a negative waiting time */
-        result = ret;
-    } else {
-        result = msp_get_common_ancestor_waiting_time_from_rate(self, pop, lambda);
-    }
+    /* Factor of 4 because only 1/4 of binary events result in a merger due to
```
jeromekelleher, line 5868: Nice, thanks for the comment!

lib/msprime.c
```diff
@@ -5881,159 +5856,61 @@ beta_generation_rate_to_model_rate(simulation_model_t *model, double rate)
...
      * lineages get assigned to, where all lineages in a given pot are merged into
      * a common ancestor.
      */
-    k = gsl_ran_binomial(self->rng, x, avl_count(ancestors));
-    for (i = 0; i < k; i++){
-        j = (uint32_t) gsl_rng_uniform_int(self->rng, avl_count(ancestors));
-        node = avl_at(ancestors, j);
-        assert(node != NULL);
+    for (i = 0; i < 4; i++) {
+        pot_size = gsl_ran_binomial(self->rng, 1.0 / (4 - (int)i),
```
jeromekelleher, line 5891: Simpler/less likely to have obscure problems: ``1.0 / (4.0 - i)``. The ``i`` should be automatically cast to double then.

lib/msprime.c
```diff
@@ -6568,25 +6444,10 @@ msp_set_simulation_model_beta(msp_t *self, double reference_size, double alpha,
     self->model.params.beta_coalescent.scalar = scalar;
```
jeromekelleher, line 6444: I know this is left over from before, but ``scalar`` isn't the most descriptive of names. Is there much point in actually storing this I wonder?

jeromekelleher:
Looks like there' a few Python tests that need to be adapted/removed also now.

jeromekelleher:
Can you remove the gsl integration headers now also please? These will be in a few places.

JereKoskela:
> Sorry for the delay. Great Work @JereKoskela ! All the expected SFS I checked work and display perfect fit (beta and dirac). Same for recombination graphics. I will test more extreme cases and increase number of repetitions to be sure. Should We add a test in verification.py for sample of size 3 (to check probability of multiple merger) ?

I don't think that's necessary. It was helpful for pin-pointing the problem earlier, but if the triple merger rate was wrong, it would show up on the existing SFS tests.

JereKoskela:
> Regarding the full ARG, I think it should just work if we turn off the checks, but we'd need to write a few tests to be sure.

Something akin to `run_arg_recording` in verification.py for the Hudson model? If you think that would suffice, then I can put it together fairly easily.

lib/msprime.c
JereKoskela: Thanks! The rules around automatic casting give me endless headaches, especially when dealing with unsigned quantities.

jeromekelleher:
> Something akin to run_arg_recording in verification.py for the Hudson model? If you think that would suffice, then I can put it together fairly easily.

Yes, and maybe a couple of simple unit tests at the C and Python level just to make sure nothing blows up when we do this.

lib/msprime.c
JereKoskela: I've dropped it from the model, and instead compute it on the fly in each of the rescaling functions. They've become a little longer as a result, but are ultimately still pretty simple.

jeromekelleher:
Can you ping me when this is ready for review please @JereKoskela?

JereKoskela:
> Can you ping me when this is ready for review please @JereKoskela?

Absolutely. It should all be there now, but I was waiting for the automated checks to finish before pinging you in case I messed anything up while tweaking tests. I'll ping you once that's done @jeromekelleher .

JereKoskela:
> Yes, and maybe a couple of simple unit tests at the C and Python level just to make sure nothing blows up when we do this.

Ok @jeromekelleher, this is looking good to me now. I've put the Beta and Dirac full-arg implementation through the same sanity checks in the C unit testing as we conduct for multiple merger models without full-arg. I've also done all of the other tidying that you pointed out above. Let me know if anything else looks in need of attention.

I've added the `<gsl/gsl_minmax.h>` header to `mutgen.c` because the compiler started complaining about it after I removed the integration headers. `gsl_integration.h` contains the `gsl_minmax.h` header, so presumably that was the only route through which `mutgen.c` saw it previously.

I'll fix the overlong python line and squash some of the commits once we're happy to merge.

lib/msprime.c
```diff
@@ -5863,177 +5824,106 @@ msp_dirac_common_ancestor_event(msp_t *self, population_id_t pop_id, label_id_t
...
 beta_model_time_to_generations(simulation_model_t *model, double t)
 {
-    return 4 * model->params.beta_coalescent.scalar * t;
+    double alpha = model->params.beta_coalescent.alpha;
+    double truncation_point = model->params.beta_coalescent.truncation_point;
+    double reference_size = model->reference_size;
+    double m = 2.0 + exp(alpha * log(2) + (1 - alpha) * log(3) - log(alpha - 1));
+    double timescale = exp(log(alpha) - alpha * log(m)
+        - (alpha - 1) * log(reference_size))
+        * gsl_sf_beta_inc(2 - alpha, alpha, truncation_point);
```
jeromekelleher, line 5833: There's a fair bit of repeated code here. How about 
```c
static double 
beta_model_compute_timescale(simulation_model_t *model)
{   
    double alpha = model->params.beta_coalescent.alpha;
￼    double truncation_point = model->params.beta_coalescent.truncation_point;
￼    double reference_size = model->reference_size;
￼    double m = 2.0 + exp(alpha * log(2) + (1 - alpha) * log(3) - log(alpha - 1));
￼    double timescale = exp(log(alpha) - alpha * log(m)
￼        - (alpha - 1) * log(reference_size))
￼        * gsl_sf_beta_inc(2 - alpha, alpha, truncation_point);
    return timescale;
}

static double
￼beta_model_time_to_generations(simulation_model_t *model, double t)
￼{
    return  4 * t * beta_model_compute_timescale(model);
}
```

lib/msprime.c
```diff
@@ -6044,75 +5934,73 @@ static int MSP_WARN_UNUSED
...
-                 self->model.params.beta_coalescent.acceptance_rate);
-
-    ret = msp_multi_merger_common_ancestor_event(self, beta_x, ancestors, Q);
-    if (ret < 0) {
-        goto out;
+                 self->model.params.beta_coalescent.truncation_point);
+
+    /* We calculate the probability of accepting the event */
+    if (beta_x > 1e-9) {
+        u = (n - 1) * log(1 - beta_x)
```
jeromekelleher, line 5953: We'll need to make sure that the C tests are hitting both of these paths. Might take some parameter tweaking.

JereKoskela: I've changed one of the tests C test cases to have `alpha` near 1, and the other near 2. They will go down the two different paths with rare exceptions.

lib/msprime.c
JereKoskela: Good idea. Added.

JereKoskela:
Another update ready for you @jeromekelleher. I've covered the two points you raised above, and hopefully all the lines are short enough this time too...

lib/msprime.c
jeromekelleher: Yep, both cases are covered. Great.

jeromekelleher:
Looks great! Ready to merge after a rebase and squash.

TPPSellinger:
I did further checks  in more extreme cases with high number of repetitions. I did not find any issues.

jeromekelleher:
Would you mind doing the rebase and squash please @JereKoskela? I'd like to get this merged while it's fresh.

JereKoskela:
> Would you mind doing the rebase and squash please @JereKoskela? I'd like to get this merged while it's fresh.

On it as we speak by happy accident @jeromekelleher. Following the guide in stdpopsim is throwing a conflict error which I'm currently investigating. "merge conflict in lib/msprime.c". I have no idea what it's talking about, but should be able to get it figured out today.

JereKoskela:
Ok @jeromekelleher, unless I've broken something on the way, this should now be rebased, squashed, and ready for merging.

jeromekelleher:
Merged! Great work, thanks @JereKoskela!

PR #966, opened 2020-04-21
Multiple merger documentation

JereKoskela:
I'm assuming that the autoclasses will pull the model docstrings through to readthedocs. If not, then this probably needs more work. Either way, let me know what you think @jeromekelleher.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -284,6 +284,25 @@ Coalescent and approximations
 
 .. autoclass:: msprime.SmcPrimeApproxCoalescent
 
+.. _sec_api_simulation_models_multiple_mergers:
+
++++++++++++++++++++++++++++
+Multiple merger coalescents
++++++++++++++++++++++++++++
+
+Some evolutionary scenarios, such as a skewed offspring distribution
+combined with a type III survivorship curve, range expansion, and
+rapid adaptation, can predict diploid genealogies with up to four
+simultaneous multiple mergers. Msprime provides the option to simulate
+from two classes of such genealogical processes.
+
+See :ref:`sec_tutorial_multiple_mergers` for examples of how to use these
+multiple merger coalescents.
+
+.. autoclass:: msprime.BetaCoalescent
+
+.. autoclass:: msprime.DiracCoalescent
+
 +++++++++++++++++++++++++++
 Discrete time Wright-Fisher
 +++++++++++++++++++++++++++
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1731,8 +1731,8 @@ Running this code we get::
     ┃ ┃ ┃ ┏━╋━┓ ┃ ┃ ┃ ┃
     8 9 2 0 4 6 1 3 5 7
 
-The specified value of :math:`\\alpha = 1.001` corresponds to a heavily skewed
-offspring distribution. Values closer to :math:`\\alpha = 2` result in trees
+The specified value of :math:`\alpha = 1.001` corresponds to a heavily skewed
+offspring distribution. Values closer to :math:`\alpha = 2` result in trees
 whose distribution is closer to that of the standard coalescent, often featuring
 no multiple mergers for small sample sizes:
 
@@ -1767,17 +1767,17 @@ Running this code we get::
     ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓
     5 2 3 7 8 0 6 9 1 4
 
-The timescale of the Beta-Xi-coalescent depends nonlinearly on both `alpha`
-and the effective population size `Ne` as detailed in
-:ref:`_sec_api_simulation_models_multiple_mergers`. For a fixed :math:`\\alpha`,
-a unit of coalescent time is proportional to :math:`Ne^{\\alpha - 1}` generations,
-albeit with a complicated constant of proportionality that depends on :math:`\\alpha`.
-The dependence on :math:`\\alpha` for fixed :math:`Ne` is not monotone.
+The timescale of the Beta-Xi-coalescent depends nonlinearly on both :math:`\alpha`
+and the effective population size :math:`N_e` as detailed in
+:ref:`sec_api_simulation_models_multiple_mergers`. For a fixed :math:`\alpha`,
+a unit of coalescent time is proportional to :math:`N_e^{\alpha - 1}` generations,
+albeit with a complicated constant of proportionality that depends on :math:`\alpha`.
+The dependence on :math:`\alpha` for fixed :math:`N_e` is not monotone.
 Since two lineages merge in 0.5 units of coalescent time on average regadless of
-:math:`Ne` and :math:`alpha`, the time to their most recent common ancestor depends 
-on both of these paramenters when measured in generations.
+:math:`N_e` and :math:`\alpha`, the time to their most recent common ancestor depends
+on both of these parameters when measured in generations.
 
-To illustrate, for :math:`alpha` close to 2 the relationship between effective
+To illustrate, for :math:`\alpha` close to 2 the relationship between effective
 population size and number of generations is almost linear:
 
 .. code-block:: python
@@ -1801,7 +1801,7 @@ which results in::
     1.5705367504768712
     99.09416974894381
 
-For :math:`alpha` close to 1 the effective population size has little effect:
+For :math:`\alpha` close to 1 the effective population size has little effect:
 
 .. code-block:: python
 
@@ -1837,28 +1837,26 @@ The Dirac-Xi-coalescent is simulated similarly:
 
 which gives::
 
-         15
-      ┏━━━┻━━┓
-      ┃     14
-      ┃   ┏━━┻━━┓
-     12   ┃    13
-    ┏━╋━┓ ┃ ┏━┳━╋━━━┓
-    ┃ ┃ ┃ ┃ ┃ ┃ ┃  11
-    ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┏━┻┓
-    ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ 10
-    ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓
-    1 4 6 8 0 7 9 2 3 5
+           14
+       ┏━━━━┻━━━┓
+       ┃       13
+       ┃      ┏━┻━━┓
+      12      ┃    ┃
+    ┏━┳┻━━┓   ┃    ┃
+    ┃ ┃  11   ┃   10
+    ┃ ┃ ┏━╋━┓ ┃ ┏━┳┻┳━┓
+    2 6 3 4 8 1 0 5 7 9
 
 Larger values of the parameter :math:`c > 0` result in more frequent multiple
-mergers, while larger values of :math:`0 < psi <= 1` result in multiple mergers
+mergers, while larger values of :math:`0 < \psi \leq 1` result in multiple mergers
 with more participating lineages. Setting either parameter to 0 would correspond
 to the standard coalescent.
 
 The Dirac-Xi-coalescent is obtained as the infinite population scaling limit of
 Moran models, and therefore coalescent time is measured in units of
-:math:`Ne^2 (1 + c \\psi^2)` generations. However, under a Moran model, the
+:math:`N_e^2` generations. However, under a Moran model, the
 population-rescaled recombination rate is still obtained from the per-generation
-recombination probability by rescaling with :math:`Ne (1 + c \\psi^2)`.
+recombination probability by rescaling with :math:`N_e`.
 The overall effect is that coalescent branch lengths scale with the square of the
 effective population size in units of generations, and thus so do other quantities
 which depend on branch lengths such as the number of mutations, while the number of
@@ -1876,18 +1874,18 @@ recombinations scales linearly.
         sites = 0
         for t in ts:
             trees = trees + t.num_trees
-            sites = trees + t.num_sites
+            sites = sites + t.num_sites
         print(sites / trees)
 
 Running this code results in::
 
-    21.93897882938979
+    1241.8241770462635
 
-which is larger than 1, though less than :math:`Ne = 1000` because not every
-recombination event results in a new tree.
+which is larger than :math:`N_e = 1000` because not every recombination
+results in a new tree.
 
 This behaviour is a consequence of the time and parameter scalings under the Moran
 model, as well as the fact that ``msprime`` simulates recombinations on the coalescent
 time scale, but mutations on trees in units of generations. The rates of mutations
 and recombinations can be made commensurate by either dividing the mutation rate by
-:math:`Ne`, or multiplying the recombination rate by :math:`Ne`.
+:math:`N_e`, or multiplying the recombination rate by :math:`N_e`.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1572,7 +1572,8 @@ example:
 
     def full_arg_example():
         ts = msprime.simulate(
-            sample_size=5, recombination_rate=0.1, record_full_arg=True, random_seed=42)
+            sample_size=5, recombination_rate=0.1,
+            record_full_arg=True, random_seed=42)
         print(ts.tables.nodes)
         print()
         for tree in ts.trees():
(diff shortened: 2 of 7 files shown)


msprime/simulations.py
```diff
@@ -1929,12 +1929,49 @@ def get_ll_representation(self):
 
 
 class BetaCoalescent(ParametricSimulationModel):
-    # TODO document.
+    """
+    A diploid Xi-coalescent with up to four simultaneous multiple mergers and
+    crossover recombination. The string ``"beta"`` can be used to refer to this
```
jeromekelleher, line 1934: I don't think we can use "beta" because this is a parametric model. The strings like "smc" are only available as a shortcut for non parametetric models.

msprime/simulations.py
```diff
@@ -1929,12 +1929,49 @@ def get_ll_representation(self):
...
+    model.
+
+    There are two main differences between the Beta-Xi-coalescent and the
+    standard coalescent. Firstly, the number of lineages that take part in each
+    common ancestor event is random, with distribution determined by moments of
+    the Beta(`2 - alpha`, `alpha`)-distribution. All participating lineages are
```
jeromekelleher, line 1940: Maybe use ``:math:`` here and elsewhere in these descriptions? It works pretty well (see the "simulation model" section for examples of it being used.) https://www.sphinx-doc.org/en/1.0/ext/math.html

msprime/simulations.py
```diff
@@ -1929,12 +1929,49 @@ def get_ll_representation(self):
...
+    Secondly, the timescale predicted by the Beta-Xi-coalescent is proportional
+    to `Ne^(alpha - 1)` generations. Note that it depends on `alpha`, and can
+    be dramatically shorter than the timescale of the standard coalescent.
+    Thus, effective population sizes must often be many orders of magnitude
+    larger than census population sizes. The per-generation recombination rate
+    is rescaled similarly to obtain the population-rescaled recombination rate.
+
+    For details on the distribution of the number of lineages taking part
+    in each common ancestor event, and on the associated timescale, see
+    Schweinsberg (2003) at
```
jeromekelleher, line 1954: A bit nicer 
```
see `Schweinsberg (2003) <https://www.sciencedirect.com/science/article/pii/S0304414903000280>`_
```

msprime/simulations.py
```diff
@@ -1929,12 +1929,49 @@ def get_ll_representation(self):
...
+    The Beta-Xi-coalescent has two parameters. `alpha` determines the
+    degree of skewness in the offspring distribution, and must satisfy
+    `1 < alpha < 2`. `truncation_point` determines the maximum fraction
+    of the population replaced by offspring in one reproduction event.
+    It defaults to one for a standard Beta-Xi-coalescent, and must
+    satisfy `0 < truncation_point <= 1`. When `truncation_point < 1`,
+    then the distribution of the number of lineages participating in
+    a common ancestor event is determined by moments of the
+    Beta(`2 - alpha`, `alpha`)-distribution conditional on not
+    exceeding `truncation_point`.
```
jeromekelleher, line 1971: Great! I think we can refactor this last paragraph a little to follow sphinx conventions. something like
```

:param float alpha:  Determines the ￼degree of skewness in the offspring distribution, 
     and must satisfy   `1 < alpha < 2`. 
:param float truncation_point: <stuff you said about truncation point>
```

msprime/simulations.py
```diff
@@ -1929,12 +1929,49 @@ def get_ll_representation(self):
...
+    """
     name = "beta"
 
-    # Must have 1 < alpha < 2 and 0 < truncation_point <= 1.
```
jeromekelleher, line 1935: It would be nice to preserve this information: telling the user when it's equivalant to Hudson is useful, and if we still have numerical instability issues we should warn about them.

msprime/simulations.py
```diff
@@ -1929,12 +1929,49 @@ def get_ll_representation(self):
...
-    # alpha = 2 and truncation_point = 1 is equivalent to the Kingman coalescent,
-    # but numerical instability occurs with alpha > 2.0 - 0.0005.
     def __init__(self, reference_size=1, alpha=1.5, truncation_point=None):
```
jeromekelleher, line 1975: We should think carefully about these defaults now. I'm inclined to not set any, other than reference_size=1 to avoid mistakes. Unless you think these are really well chosen defaults @JereKoskela, let's follow this up in an issue.

msprime/simulations.py
```diff
@@ -1944,11 +1981,32 @@ def __init__(self, reference_size=1, alpha=1.5, truncation_point=None):
 
 
 class DiracCoalescent(ParametricSimulationModel):
-    # TODO document
+    """
+    A diploid Xi-coalescent with up to four simultaneous multiple mergers and
+    crossover recombination. The string ``"dirac"`` can be used to refer to this
+    model.
+
+    The Dirac-Xi-coalescent is an implementation of the model of
```
jeromekelleher, line 1989: Basically the same comments as above apply here.

msprime/simulations.py
```diff
@@ -1944,11 +1981,32 @@ def __init__(self, reference_size=1, alpha=1.5, truncation_point=None):
...
+    all participating lineages are randomly split into four groups,
+    corresponding to the four parental chromosomes present in a diploid,
+    bi-parental reproduction event, and the lineages within each group merge
+    simultaneously.
+
+    Note that the Dirac-Xi-coalescent is obtained as the scaling limit of
+    Moran models, rather than Wright-Fisher models. As a consequence, one
+    unit of coalescent time is proportional to `Ne^2` generations, rather
+    than `Ne` generations as in the standard coalescent. Recombination and
+    other rates are still scaled by `Ne`.
```
jeromekelleher, line 2005: This is going to confuse people (certainly confuses me!). Perhaps we should point people to the tutorial to show an example of why this is different and what to expect?

This probably warrants a formal warning as well. If you do this
```
.. warning:
    Note that the Dirac-Xi-coalescent is obtained as the scaling limit of
￼    Moran models, rather than Wright-Fisher models. As a consequence, one
￼    unit of coalescent time is proportional to `Ne^2` generations, rather
￼    than `Ne` generations as in the standard coalescent. Recombination and
￼    other rates are still scaled by `Ne`.

```
it should render it in a warning box.

docs/api.rst
```diff
@@ -259,6 +259,35 @@ Coalescent and approximations
...
+Multiple merger coalescents
++++++++++++++++++++++++++++
+
+Some evolutionary scenarios, such as a skewed offspring distribution
+combined with a type III survivorship curve, range expansion, and
+rapid adaptation, can predict diploid genealogies with up to four
+simultaneous multiple mergers. Msprime provides the option to simulate
+from two such genealogical processes.
+
+To use the Beta-Xi-coalescent, set the flag
```
jeromekelleher, line 272: I don't think the examples are terribly helpful here. Probably better to just point the reader to the tutorial and do all of the examples there. 
```
See :ref:`sec_tutorial_multiple_mergers` for examples of how to use these multiple merger coalescents.

File: docs/tutorial.rst [hunk 10]
```diff
@@ -1484,3 +1484,107 @@ Running this code we get::
...
+offspring numbers among individuals in the population is highly skewed.
+Provided are the classes of Beta-Xi-coalescents and Dirac-Xi-coalescents,
+whose use is illustrated below, first for the Beta-Xi-coalescent.
+
+.. code-block:: python
+
+    def beta_multiple_merger_example():
+        ts = msprime.simulate(
+            sample_size=10,
+            recombination_rate=350,
```
jeromekelleher, line 1505: Let's not bother with recombination for the initial examples, it's just obscuring the important things.

docs/tutorial.rst
```diff
@@ -1484,3 +1484,107 @@ Running this code we get::
...
+
++++++++++++++++++++++++++++
+Multiple merger coalescents
++++++++++++++++++++++++++++
+
+``msprime`` can simulate several multiple merger coalescent models,
+in which any number of lineages can coalesce in up to four simultaneous
+mergers. These can often be appropriate models when the distribution of
+offspring numbers among individuals in the population is highly skewed.
+Provided are the classes of Beta-Xi-coalescents and Dirac-Xi-coalescents,
```
jeromekelleher, line 1497: I think we need a take it a little bit more slowly here. We're not giving a lot of insight into what the parameters are, etc. How about we do a beta coalescent simulation, and show how changing the alpha parameter produces (a) slightly non-binary trees and (b) very very non-binary trees? This will give an intuition for what's going on.

Also, it's good to link directly to the docs using ``:class:`.BetaCoalescent```, etc when referring to them here.

msprime/simulations.py
JereKoskela: I picked this up from the `name = "beta"` line under the model class in simulations.py, which was already there before I started. I did wonder how that could work. I guess we should give the whole ``msprime.BetaCoalescent(alpha, truncation_point)`` string instead?

msprime/simulations.py
JereKoskela: The instability was due to the numerical integrator. It is definitely gone. Preserving the line about equivalence to Kingman is a good point.

msprime/simulations.py
jeromekelleher: We should just remove support for creating these classes from strings - it's not helpful I think. We can follow up in an issue, no need to solve the problem here.

benjeffery: +1 for configuration classes over strings

msprime/simulations.py
JereKoskela: There is no obvious default for `alpha`, or for `psi` and `c` in the Dirac case. I'm happy to leave those without defaults. In contrast, I think `truncation_point = 1` is a good default, because it corresponds to the usual Beta-coalescent. It is almost certainly going to be what everyone who doesn't know to worry about it will want.

jeromekelleher:
I'm going to mute this thread @JereKoskela, can you ping me when it's ready please?

JereKoskela:
> I'm going to mute this thread @JereKoskela, can you ping me when it's ready please?

Sure, I'll let you know once I've pushed another draft.

JereKoskela:
Ping @jeromekelleher for another draft.

I see I've created conflicts in verification.py by changing some of the multiple merger statistical tests. I noticed that an `Ne` argument in `msprime.simulate()` did not feed through to a `model=msprime.DiracCoalescent(psi, c)` specification unless it was also included as a `reference_size` argument within the model class, i.e. `model=msprime.DiracCoalescent(reference_size=Ne, psi, c)`. I'm not sure whether that's intentional, but it meant that the tests weren't doing exactly what they were supposed to be.

I also found one more problem with the scalings of the Beta and Dirac coalescents in `msprime.c`: some factors were the reciprocals of what they were meant to be. I've fixed those here too.

The actual docs are more extensive, and I've tried to make more use of Sphinx functionality too. I've not been able to work out how to preview what I've written, so let me know if something looks wrong or out of place. I'm happy for you to tweak these directly too if that's easier in places. For example, I didn't find a way to break long displayed math or hyperlinks across multiple lines, which is why CircleCI is complaining.

msprime/simulations.py
```diff
@@ -1929,26 +1929,117 @@ def get_ll_representation(self):
...
+    .. math::
+        m = 2 + (2 / 3)^{\\alpha} 3 / (\\alpha - 1).
+
+    Note that the time scale depends both on the effective population size :math:`Ne`
+    and :math:`alpha`, and can be dramatically shorter than the timescale of the 
+    standard coalescent. Thus, effective population sizes must often be many orders
+    of magnitude larger than census population sizes. The per-generation recombination
+    rate is rescaled similarly to obtain the population-rescaled recombination rate.
+
+    See `Schweinsberg (2003) <https://www.sciencedirect.com/science/article/pii/S0304414903000280>`_
```
jeromekelleher, line 1978: Can split this like 
```
See `Schweinsberg (2003) 
<https://www.sciencedirect.com/science/article/pii/S0304414903000280>`_
for 
```

msprime/simulations.py
```diff
@@ -1929,26 +1929,117 @@ def get_ll_representation(self):
...
+
+    There are two main differences between the Beta-Xi-coalescent and the
+    standard coalescent. Firstly, the number of lineages that take part in each
+    common ancestor event is random, with distribution determined by moments of
+    the :math:`Beta(2 - \\alpha, \\alpha)`-distribution. In particular, when there
+    are :math:`n` lineages, each set of :math:`k <= n` of them participates in a
+    common ancestor event at rate
+
+    .. math::
+        8 \\int_0^1 x^{k - \\alpha - 1} (1 - x)^{n - k + \\alpha - 1} dx / B(2 - \\alpha, \\alpha),
```
jeromekelleher, line 1944: Should be able to split this across multiple lines?

jeromekelleher:
> I see I've created conflicts in verification.py by changing some of the multiple merger statistical tests. 

Probably your conflicts have happened because we've just changed over to using [black](https://github.com/psf/black) to format the code. Sorry for the hassle. The simplest thing to do is install the pre-commits, then this will get run each time before you commit:

```
python3 -m pip install requirements/development.txt
pre-commit install
```

The documentation for this is being written...

> I noticed that an Ne argument in msprime.simulate() did not feed through to a model=msprime.DiracCoalescent(psi, c) specification unless it was also included as a reference_size argument within the model class, i.e. model=msprime.DiracCoalescent(reference_size=Ne, psi, c). I'm not sure whether that's intentional, but it meant that the tests weren't doing exactly what they were supposed to be.

That sounds like a bug to me. Perhaps the thing to do is open an issue to track it (or, you can just fix it here, if you prefer).

> The actual docs are more extensive, and I've tried to make more use of Sphinx functionality too. I've not been able to work out how to preview what I've written, 

Once you've installed all the development requirements, you should be able to make the docs like
```
cd docs
make 
```
The html will be in the ``_build`` directory.

JereKoskela:
Thanks @jeromekelleher, that's very instructive! I'll tidy this up, hopefully as early as tomorrow, and push through one more update.

JereKoskela:
Another update for you @jeromekelleher. I managed to get pre-commit to run, but don't really understand it. I'm also struggling to make sense of the automated test output this time. If I'm reading it right, at least some of the complaints seem to be about code I didn't touch. Is this something you can either tidy up easily, or try to explain to the computer-illiterate in the absence of documentation?

Regarding the update itself, I'm sorry for packaging more things in, but I realised upon tidying up the docs that we had made the Dirac scaling more complicated than it needed to be. The variable powers of `Ne` need to be there (as per the tutorial), but there's no need to involve other parameters in the scalings. So I've stripped those out, and amended the docs accordingly. Now the scaling is slightly simpler, and in line with what appears in the Blath et al. (2013) paper.

I also took one more look at the C tests, cleaned out some leftover commented lines from the time when store-full-arg could not yet be run for multiple merger coalescents, and activated the test to check reciprocality of the scalings.

jeromekelleher:
The changes look good to me @JereKoskela, thanks for going through this so thoroughly (having to explain stuff to other people does help though, doesn't it?!).

You've been a bit unlucky in that your changes are happening while we're making large but semantically neutral changes to the code. It's probably simplest if I update this PR and get it ready to merge. Mind if I jump in here and close it out?

JereKoskela:
> You've been a bit unlucky in that your changes are happening while we're making large but semantically neutral changes to the code. It's probably simplest if I update this PR and get it ready to merge. Mind if I jump in here and close it out?

That would be very helpful @jeromekelleher. Thanks!

jeromekelleher:
OK, that should do it @JereKoskela. I've fixed the problem you pointed out about the parametric models not inheriting Ne from msprime.simulate() - good spot, thanks. This shouldn't break any downstream, I think?

JereKoskela:
> OK, that should do it @JereKoskela. I've fixed the problem you pointed out about the parametric models not inheriting Ne from msprime.simulate() - good spot, thanks. This shouldn't break any downstream, I think?

I think you know how the plumbing works better than I do @jeromekelleher, but I don't think it should. The only way something would break would be if we had entered inconsistent `Ne` values through msprime.simulate() and the model class, and that hasn't happened anywhere.

Looks good to merge to me. Thanks for the help!

jeromekelleher:
Done! And that is the Xi coalescents project (https://github.com/tskit-dev/msprime/projects/1) officially closed: hooray! Congratulations, and thanks to @shajoezhu, @JereKoskela and @TPPSellinger for the great work on this!

shajoezhu:
Many thanks guys! @JereKoskela @jeromekelleher @TPPSellinger This is amazing work! I had nightmares with the scaling! This is absolutely worth a celebration! Virtual beers and shots!

Issue #958, opened 2020-04-20
Change ``parse_species_tree`` to return Demography object.


jeromekelleher:
Currently parse_species_tree returns a tuple ``(pop_configs, demographic_events)``. The idiomatic way of calling this is 

```python
pop_configs, demographic_events =parse_species_tree()
```

This is fine, but it's possible that we'll want to return more stuff from parse_species_tree at some point, and then we'll be in a tricky situation. It seems possible that we'll return a migration matrix as well, for example.

To make this more extensible, we could turn the configuation as a namedtuple, such that the idiomatic methods is 
```python
config = parse_species_tree(...)
ts = msprime.simulate(
     samples=# something
     population_configurations=config.population_configurations,
     demographic_events=config.demographic_events)
```

This can be made even easier, with
```python
config = parse_species_tree(...)
ts = msprime.simulate(
     samples=# something
     **config._asdict())
```

The changes needed are pretty small.

Any thoughts @mmatschiner?

benjeffery:
One thing to consider (that you mentioned on a call @jeromekelleher) is to use `attr`:
```python
@attr.s(slots=True, frozen=True, kw_only=True, auto_attribs=True)
    class _ParseSpeciesTreeReturn:
        population_configurations: List[msprime.PopulationConfiguration]
        demographic_events: List[msprime.DempgraphicEvent]
```
This gives you an immutable class that can only be init' d by keyword (to avoid ordering errors).

jeromekelleher:
Yes, good idea @benjeffery. The downside is we don't currently depend on ``attrs``, so we'd need to pull in a dependency (but a good dependency, which we should probably start using to make the code cleaner anyway). The upside (from a user perspective) is we can use  ``config.asdict()`` rather than ``config._asdict()`` (which I know I'd forget and have to keep looking up).

Do we get some documentation for free with your nice type-hints @benjeffery?

benjeffery:
I wanted to talk `attr` over anyway as I'm using in in the metadata branch in `tskit`. I haven't looked into the docs yet, especially for this specific case with `attr` (was planning to as part of the metadata work) but I'm hoping that we can get the docs to use the type hints wherever we use them.

jeromekelleher:
OK, let's assume we're using ``attrs`` so. If we're already pulling it in for tskit, then there's no reason not to use it here.

mmatschiner:
[...]

PR #955, opened 2020-04-18
Species tree docs

mmatschiner:
Rebased following description by @jeromekelleher

PR #967, opened 2020-04-21
Updated species tree docs

mmatschiner:
I guess I made it a bit more complicated than necessary by merging branches master and species-tree-docs locally, being at the impression that suggested changes are not just on the documentation. This addresses most comments to PR #955 . Issue #958 is not addressed yet.


jeromekelleher:
Great, thanks for the updates @mmatschiner, let's merge this ASAP and we can follow up with any other changes we need.

I think it might be easier to merge this on your other PR though, as you're back on your master branch here rather than using a feature branch. I think this should work:

```
git checkout species-tree-docs
git cherry-pick 680b61c
git push origin species-tree-docs
```

that should put the changes you've made here onto PR #955, which we can then merge.

jeromekelleher:
Closed in favour of #955.

PR #969, opened 2020-04-22
Function in DemographyDebugger to check where lineages are possible

apragsdale:
This function in the demography debugger returns a epoch times and indicators for whether lineages are possible in each population from the population configurations. There are some simple examples in the tests for this function, and I've tried to make it play nicely with ancient samples.

If no samples are given, it assumes we are sampling from each population at time zero. Otherwise, samples must be given as a list of `msprime.Sample` objects, which have times and populations.

cc: @grahamgower, @jeromekelleher

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -2206,6 +2206,8 @@ class DemographyDebugger:
                 population_configurations, saved_sample_sizes
             ):
                 pop_config.sample_size = sample_size
+        self.migration_matrix = migration_matrix
+        self.population_configurations = population_configurations
 
     def _make_epochs(self, simulator, demographic_events):
         self.epochs = []
@@ -2345,6 +2347,159 @@ class DemographyDebugger:
             N, _ = self._pop_size_and_migration_at_t(t)
             N_t[j] = N
         return N_t
+
+    def lineage_probabilities(self, steps, sample_time=0):
+        """
+        Returns an array such that P[j, a, b] is the probability that a lineage that
+        started in population a at time sample_time is in population b at time steps[j]
+        ago.
+
+        This function reports sampling probabilities _before_ mass migration events
+        at a step time, if a mass migration event occurs at one of those times.
+        Migrations will then effect the next time step.
+
+        :param list steps: A list of times to compute probabilities.
+        :param sample_time: The time of sampling of the lineage. For any times in steps
+            that are more recent than sample_time, the probability of finding the
+            lineage in any population is zero.
+        :return: An array of dimension len(steps) by num pops by num_pops.
+        """
+        num_pops = self.num_populations
+        # P[i, j] will be the probability that a lineage that started in i is now in j
+        P = np.eye(num_pops)
+
+        # epochs are defined by mass migration events or changes to population sizes
+        # or migration rates, so we add the epoch interval times to the steps that we
+        # need to account for
+        epoch_breaks = [t for t in self.epoch_times if t not in steps]
+        all_steps = np.concatenate([steps, epoch_breaks])
+
+        sampling = []
+        if sample_time not in all_steps:
+            sampling.append(sample_time)
+        all_steps = np.concatenate((all_steps, sampling))
+
+        ix = np.argsort(all_steps)
+        all_steps = all_steps[ix]
+        # keep track of the steps to report in P_out
+        keep_steps = np.concatenate(
+            [
+                np.repeat(True, len(steps)),
+                np.repeat(False, len(epoch_breaks)),
+                np.repeat(False, len(sampling)),
+            ]
+        )[ix]
+
+        assert len(np.unique(all_steps)) == len(all_steps)
+        assert np.all(steps == all_steps[keep_steps])
+        P_out = np.zeros((len(all_steps), num_pops, num_pops))
+
+        first_step = 0
+        while all_steps[first_step] < sample_time:
+            first_step += 1
+
+        P_out[first_step] = P
+
+        # get ordered mass migration events
+        mass_migration_objects = []
+        mass_migration_times = []
+        for demo in self.demographic_events:
+            if demo.type == "mass_migration":
+                mass_migration_objects.append(demo)
+                mass_migration_times.append(demo.time)
+
+        for jj in range(first_step, len(all_steps) - 1):
+            t_j = all_steps[jj]
+
+            # apply any mass migration events to P
+            # so if we sample at this time, we do no account for the instantaneous
+            # mass migration events that occur at the same time. that will show up
+            # at the next step
+            if t_j > sample_time:
+                for mass_mig_t, mass_mig_e in zip(
+                    mass_migration_times, mass_migration_objects
+                ):
+                    if mass_mig_t == t_j:
+                        S = np.eye(num_pops, num_pops)
+                        S[mass_mig_e.source, mass_mig_e.dest] = mass_mig_e.proportion
+                        S[mass_mig_e.source, mass_mig_e.source] = (
+                            1 - mass_mig_e.proportion
+                        )
+                        P = np.matmul(P, S)
+
+            # get continuous migration matrix over next interval
+            _, M = self._pop_size_and_migration_at_t(t_j)
+            dt = all_steps[jj + 1] - all_steps[jj]
+            dM = np.diag([sum(s) for s in M])
+            # advance to next interval time (dt) taking into account continuous mig
+            P = P.dot(_matrix_exponential(dt * (M - dM)))
+            P_out[jj + 1] = P
+
+        return P_out[keep_steps]
+
+    def possible_lineage_locations(self, samples=None):
+        """
+        Given the sampling configuration, this function determines when lineages are
+        possibly found within each population over epochs defined by demographic events
+        and sampling times. If no sampling configuration is given, we assume we sample
+        lineages from every population at time zero. The samples are specified by a list
+        of msprime Sample objects, so that possible ancient samples may be accounted for.
+
+        :param list samples: A list of msprime Sample objects, which specify their
+            populations and times.
+        :return: Returns a dictionary with epoch intervals as keys whose values are a
+            list with length equal to the number of populations with True and False
+            indicating which populations could possibly contain lineages over that
+            epoch. The epoch intervals are given by tuples: (epoch start, epoch end).
+            The first epoch necessarily starts at time 0, and the final epoch has end
+            time of infinity.
+        """
+        # get configuration of sampling times from samples ({time:[pops_sampled_from]})
+        if samples is None:
+            sampling_times = {0: [i for i in range(self.num_populations)]}
+        else:
+            sampling_times = collections.defaultdict(list)
+            for sample in samples:
+                sampling_times[sample.time].append(sample.population)
+            for t in sampling_times.keys():
+                sampling_times[t] = list(set(sampling_times[t]))
+
+        all_steps = sorted(
+            list(set([t for t in self.epoch_times] + list(sampling_times.keys())))
+        )
+
+        epochs = [(x, y) for x, y in zip(all_steps[:-1], all_steps[1:])]
+        epochs.append((all_steps[-1], np.inf))
+
+        # need to go a bit beyond last step and into the final epoch that extends to inf
+        all_steps.append(all_steps[-1] + 1)
+
+        indicators = {e: np.zeros(self.num_populations, dtype=bool) for e in epochs}
+        for sample_time, demes in sampling_times.items():
+            P_out = self.lineage_probabilities(all_steps, sample_time=sample_time)
+            for epoch, P in zip(epochs, P_out[1:]):
+                if epoch[1] <= sample_time:
+                    # samples shouldn't affect the epoch previous to the sampling time
+                    continue
+                for deme in demes:
+                    indicators[epoch][P[deme] > 0] = True
+
+        # join epochs if adjacent epochs have same set of possible live populations
+        combined_indicators = {}
+        skip = 0
+        for ii, (epoch, inds) in enumerate(indicators.items()):
+            if skip > 0:
+                skip -= 1
+                continue
+            this_epoch = epoch
+            while ii + skip + 1 < len(epochs) and np.all(
+                indicators[epochs[ii + 1 + skip]] == inds
+            ):
+                this_epoch = (this_epoch[0], epochs[ii + 1 + skip][1])
+                skip += 1
+            combined_indicators[this_epoch] = inds
+
+        return combined_indicators
 
     def mean_coalescence_time(
         self, num_samples, min_pop_size=1, steps=None, rtol=0.005, max_iter=12
--- a/tests/test_demography.py
+++ b/tests/test_demography.py
@@ -2813,3 +2813,306 @@ class TestCensusEvent(unittest.TestCase):
             demographic_events=[msprime.CensusEvent(time=2000)],
         )
         self.assertEqual(ts.tables.nodes, tsc.tables.nodes)
+
+
+class TestPossibleLineages(unittest.TestCase):
+    """
+    Tests for checking where lineages are possible within the demography debugger.
+    """
+
+    def test_possible_lineages_no_migration(self):
+        samples = [
+            msprime.Sample(time=0, population=0),
+            msprime.Sample(time=0, population=1),
+            msprime.Sample(time=0, population=2),
+            msprime.Sample(time=0, population=3),
+        ]
+        dem_events = [
+            msprime.MassMigration(time=50, source=3, destination=2),
+            msprime.MassMigration(time=100, source=2, destination=1),
+            msprime.MassMigration(time=150, source=1, destination=0),
+        ]
+        pop_config = [
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+        ]
+        dd = msprime.DemographyDebugger(
+            demographic_events=dem_events, population_configurations=pop_config
+        )
+        lineages = dd.possible_lineage_locations(samples=samples)
+        self.assertTrue(len(lineages) == 4)
+        self.assertTrue(np.all(lineages[(0, 50)] == [1, 1, 1, 1]))
+        self.assertTrue(np.all(lineages[(50, 100)] == [1, 1, 1, 0]))
+        self.assertTrue(np.all(lineages[(100, 150)] == [1, 1, 0, 0]))
+        self.assertTrue(np.all(lineages[(150, np.inf)] == [1, 0, 0, 0]))
+
+    def test_possible_lineages_with_migration(self):
+        # draw sample from first population, which has migrants from second
+        samples = [msprime.Sample(time=0, population=0)]
+        mig_mat = [[0, 0.1], [0.1, 0]]
+        dem_events = [msprime.MassMigration(time=50, source=1, destination=0)]
+        pop_config = [
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+        ]
+        dd = msprime.DemographyDebugger(
+            demographic_events=dem_events,
+            population_configurations=pop_config,
+            migration_matrix=mig_mat,
+        )
+        lineages = dd.possible_lineage_locations(samples=samples)
+        self.assertTrue(len(lineages) == 1)
+        self.assertTrue(np.all(lineages[(0, np.inf)] == [1, 1]))
+
+    def test_possible_lineages_ancient_samples(self):
+        samples = [
+            msprime.Sample(time=0, population=0),
+            msprime.Sample(time=0, population=1),
+            msprime.Sample(time=100, population=1),
+        ]
+        dem_events = [
+            msprime.MassMigration(time=50, source=1, destination=0),
+            msprime.MassMigration(time=150, source=1, destination=0),
+        ]
+        pop_config = [
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+        ]
+        dd = msprime.DemographyDebugger(
+            demographic_events=dem_events, population_configurations=pop_config
+        )
+        lineages = dd.possible_lineage_locations(samples=samples)
+        self.assertTrue(len(lineages) == 4)
+        self.assertTrue(np.all(lineages[(0, 50)] == [1, 1]))
+        self.assertTrue(np.all(lineages[(50, 100)] == [1, 0]))
+        self.assertTrue(np.all(lineages[(100, 150)] == [1, 1]))
+        self.assertTrue(np.all(lineages[(150, np.inf)] == [1, 0]))
+
+    def test_possible_lineages_complex_history(self):
+        samples = [
+            msprime.Sample(time=0, population=0),
+            msprime.Sample(time=0, population=2),
+            msprime.Sample(time=0, population=3),
+            msprime.Sample(time=500, population=4),
+        ]
+        mig_mat = [
+            [0, 0, 0, 0, 0],
+            [0, 0, 1e-5, 0, 0],
+            [0, 0, 0, 0, 0],
+            [0, 0, 0, 0, 0],
+            [0, 0, 0, 0, 0],
+        ]
+        dem_events = [
+            msprime.MassMigration(time=100, source=0, destination=1, proportion=0.1),
+            msprime.MassMigration(time=200, source=3, destination=2),
+            msprime.MigrationRateChange(time=200, rate=0),
+            msprime.MassMigration(time=300, source=1, destination=0),
+            msprime.MassMigration(time=400, source=2, destination=4, proportion=0.1),
+            msprime.MassMigration(time=600, source=2, destination=0),
+            msprime.MassMigration(time=700, source=4, destination=0),
+        ]
+        pop_config = [
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+            msprime.PopulationConfiguration(initial_size=100),
+        ]
+        dd = msprime.DemographyDebugger(
+            demographic_events=dem_events,
+            population_configurations=pop_config,
+            migration_matrix=mig_mat,
+        )
+        lineages = dd.possible_lineage_locations(samples=samples)
+        self.assertTrue(len(lineages) == 7)
+        self.assertTrue(np.all(lineages[(0, 100)] == [1, 0, 1, 1, 0]))
+        self.assertTrue(np.all(lineages[(100, 200)] == [1, 1, 1, 1, 0]))
+        self.assertTrue(np.all(lineages[(200, 300)] == [1, 1, 1, 0, 0]))
+        self.assertTrue(np.all(lineages[(300, 400)] == [1, 0, 1, 0, 0]))
+        self.assertTrue(np.all(lineages[(400, 600)] == [1, 0, 1, 0, 1]))
+        self.assertTrue(np.all(lineages[(600, 700)] == [1, 0, 0, 0, 1]))
+        self.assertTrue(np.all(lineages[(700, np.inf)] == [1, 0, 0, 0, 0]))
+
+
+class TestLineageProbabilities(unittest.TestCase):
+    """
+    Tests for checking where lineages are possible within the demography debugger.
+    """
+
+    def two_pop_example(self, a, b):
+        # Just migration between two populations:
+        #  the transition probability matrix for the lineage random walk is:
+        def f(t):
+            return np.array(
+                [
+                    [
+                        np.exp(-(a + b) * t) + (1 - np.exp(-(a + b) * t)) * b / (a + b),
+                        (1 - np.exp(-(a + b) * t)) * a / (a + b),
+                    ],
+                    [
+                        (1 - np.exp(-(a + b) * t)) * b / (a + b),
+                        np.exp(-(a + b) * t) + (1 - np.exp(-(a + b) * t)) * a / (a + b),
+                    ],
+                ]
+            )
+
+        dd = msprime.DemographyDebugger(
+            population_configurations=[
+                msprime.PopulationConfiguration(initial_size=1),
+                msprime.PopulationConfiguration(initial_size=1),
+            ],
+            migration_matrix=[[0, a], [b, 0]],
+        )
+        return dd, f
+
+    def verify_simulation(self, dd):
+        samples = [
+            msprime.Sample(pop, time)
+            for pop in range(dd.num_populations)
+            for time in dd.epoch_times
+        ]
+        reps = msprime.simulate(
+            samples=samples,
+            population_configurations=dd.population_configurations,
+            migration_matrix=dd.migration_matrix,
+            demographic_events=dd.demographic_events,
+            end_time=max(dd.epoch_times + 1),
+            num_replicates=100,
+        )
+        for ts in reps:
+            t = ts.first()
+            for u in ts.samples():
+                p = u
+                lineage = []
+                while p != tskit.NULL:
+                    lineage.append(ts.node(p))
+                    p = t.parent(p)
+                probs = dd.lineage_probabilities(
+                    [n.time for n in lineage], sample_time=lineage[0].time
+                )
+                locs = dd.possible_lineage_locations(
+                    samples=[
+                        msprime.Sample(time=n.time, population=n.population)
+                        for n in lineage
+                    ]
+                )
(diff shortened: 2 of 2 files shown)


msprime/simulations.py
```diff
@@ -2178,6 +2178,95 @@ def population_size_trajectory(self, steps):
...
+        and sampling times. If no sampling configuration is given, we assume we sample
+        lineages from each population at time zero. The samples are specified by a list
+        of msprime Sample objects, so that possible ancient samples may be accounted for.
+
+        :param list samples: A list of msprime Sample objects, which specify their
+            populations and times.
+        :return: Returns a list of times defining the epoch intervals and a list of
+            lists, one for each epoch, with lengths equal to the number of populations
+            and containing 1s or 0s for whether lineages could be found in that
+            population.
```
grahamgower, line 2194: It seems this function returns a dictionary of lists?

grahamgower:
This is very cool, thanks @apragsdale!

I sketched out a more complicated example to play with it a bit. This is the kind of scenario I envision will be common in stdpopsim. Please forgive my terrible drawing skills (I did it with a mouse in `xournalpp`)! Time is on the left, red dots are samples, and the possible lineage vectors are on the right.
demog1

And here it is as an additional unit test.
```
diff --git a/tests/test_demography.py b/tests/test_demography.py
index d0623ce..7614c4f 100644
--- a/tests/test_demography.py
+++ b/tests/test_demography.py
@@ -2675,3 +2675,40 @@ class TestEpochLineages(unittest.TestCase):
         self.assertTrue(np.all(lineages[50] == [1, 0]))
         self.assertTrue(np.all(lineages[100] == [1, 1]))
         self.assertTrue(np.all(lineages[150] == [1, 0]))
+
+    def test_complex_history(self):
+        samples = [msprime.Sample(time=0, population=0),
+                   msprime.Sample(time=0, population=2),
+                   msprime.Sample(time=0, population=3),
+                   msprime.Sample(time=500, population=4)]
+        mig_mat = [[0, 0, 0, 0, 0],
+                   [0, 0, 0, 0, 0],
+                   [0, 1e-5, 0, 0, 0],
+                   [0, 0, 0, 0, 0],
+                   [0, 0, 0, 0, 0]]
+        dem_events = [msprime.MassMigration(
+                          time=100, source=1, destination=0, proportion=0.1),
+                      msprime.MassMigration(time=200, source=3, destination=2),
+                      msprime.MigrationRateChange(time=200, rate=0),
+                      msprime.MassMigration(time=300, source=1, destination=0),
+                      msprime.MassMigration(
+                          time=400, source=4, destination=2, proportion=0.1),
+                      msprime.MassMigration(time=600, source=2, destination=0),
+                      msprime.MassMigration(time=700, source=4, destination=0),
+                    ]
+        pop_config = [msprime.PopulationConfiguration(initial_size=100),
+                      msprime.PopulationConfiguration(initial_size=100),
+                      msprime.PopulationConfiguration(initial_size=100),
+                      msprime.PopulationConfiguration(initial_size=100),
+                      msprime.PopulationConfiguration(initial_size=100)]
+        dd = msprime.DemographyDebugger(demographic_events=dem_events,
+                                        population_configurations=pop_config)
+        lineages = dd.possible_lineages(samples=samples)
+        self.assertTrue(len(lineages) == 7)
+        self.assertTrue(np.all(lineages[0] == [1, 0, 1, 1, 0]))
+        self.assertTrue(np.all(lineages[100] == [1, 1, 1, 1, 0]))
+        self.assertTrue(np.all(lineages[200] == [1, 1, 1, 0, 0]))
+        self.assertTrue(np.all(lineages[300] == [1, 0, 1, 0, 0]))
+        self.assertTrue(np.all(lineages[400] == [1, 0, 1, 0, 1]))
+        self.assertTrue(np.all(lineages[600] == [1, 0, 0, 0, 1]))
+        self.assertTrue(np.all(lineages[700] == [1, 0, 0, 0, 0]))
```

But it currently fails. The returned lineages are instead:
```
for k, v in lineages.items():
            print(k, v)
0 [1 0 1 1 0]
200 [1 0 1 0 0]
500 [1 0 1 0 1]
600 [1 0 0 0 1]
700 [1 0 0 0 0]
```

grahamgower:
Oh, whoops. The migration matrix needs to be added to the `msprime.DemographyDebugger()` call in my example. The output is the same.

msprime/simulations.py
apragsdale: You're right.. I had changed the output format and forgot to update the docstring. Will fix that.

apragsdale:
That's a good example to test on. I realize I hadn't handled the case of mass migration without 100% movement of lineages, so that will need to be fixed. Now I'm a bit confused about the migration matrix though, since I had assumed the transpose of the migration matrix in your example would take lineages from 2 to 1 (backward in time). I need to double check that. (The docs aren't very clear on how the migration matrix is defined, if M[i,j] is the prob that a lineage in deme i is a migrant from j, and if that "migration" is looking forward to backward.)

In either case though, I think in your drawing we should expect the epoch [0,100) to have lineage states [1,1,1,1,0], since lineages from deme 2 could immediately move to deme 1, unless I'm interpreting your arrows in the reverse direction?

apragsdale:
In the drawing, and in the `demographic_events`, the mass migration at time 400 has source=4 and dest=2, but there wouldn't be any lineages in deme 4 at that time. So I think you still wouldn't see a 1 for deme 4 until the sampling event at time 500. Is that right?

grahamgower:
> Now I'm a bit confused about the migration matrix though, since I had assumed the transpose of the migration matrix in your example would take lineages from 2 to 1 (backward in time). I need to double check that. (The docs aren't very clear on how the migration matrix is defined, if M[i,j] is the prob that a lineage in deme i is a migrant from j, and if that "migration" is looking forward to backward.)

I'm pretty sure I have matrix-index dyslexia, so it could well the transpose.

jeromekelleher:
This looks great to me @apragsdale! I've asked @petrelharp for a review here as he understands this stuff a lot better than me.

jeromekelleher:
ps. I've put your name in the author list for the paper @apragsdale : https://github.com/tskit-dev/msprime-1.0-paper/issues/2

grahamgower:
> In either case though, I think in your drawing we should expect the epoch [0,100) to have lineage states [1,1,1,1,0], since lineages from deme 2 could immediately move to deme 1, unless I'm interpreting your arrows in the reverse direction?

Yes, but we're not sampling from deme 1. So no genealogies in the sampled individuals could come from deme 1 in epoch [0, 100). The arrows I drew are the directions of migration. Thinking backwards in time, the ancestry of sampled individuals can only flow along the reverse direction of the arrows.

> In the drawing, and in the `demographic_events`, the mass migration at time 400 has source=4 and dest=2, but there wouldn't be any lineages in deme 4 at that time. So I think you still wouldn't see a 1 for deme 4 until the sampling event at time 500. Is that right?

I think a lineage needs to exist in deme 4, because migrants can arrive from there in the ancestry of sampled lineages.

jeromekelleher:
ps. Love the drawing skillz @grahamgower!

grahamgower:
@apragsdale, I've drawn a purple rectangle around the branches which I think should be "pruned", because they are unnecessary computational burden for forwards-time simulation. My expectation is that the pruning could be done by looking at zeros in the lineage vectors.

demog2

grahamgower:
Hey @apragsdale, I'm really sorry for the confusion, I definitely didn't code up the same thing that I drew! Here is the corrected code. In particular, the mass migration events with proportion=0.1 had source and dest switched.

```
        mig_mat = [[0, 0, 0, 0, 0],
                   [0, 0, 1e-5, 0, 0],
                   [0, 0, 0, 0, 0],
                   [0, 0, 0, 0, 0],
                   [0, 0, 0, 0, 0]]
        dem_events = [msprime.MassMigration(
                          time=100, source=0, destination=1, proportion=0.1),
                      msprime.MassMigration(time=200, source=3, destination=2),
                      msprime.MigrationRateChange(time=200, rate=0),
                      msprime.MassMigration(time=300, source=1, destination=0),
                      msprime.MassMigration(
                          time=400, source=2, destination=4, proportion=0.1),
                      msprime.MassMigration(time=600, source=2, destination=0),
                      msprime.MassMigration(time=700, source=4, destination=0),
                      ]
        pop_config = [msprime.PopulationConfiguration(initial_size=100),
                      msprime.PopulationConfiguration(initial_size=100),
                      msprime.PopulationConfiguration(initial_size=100),
                      msprime.PopulationConfiguration(initial_size=100),
                      msprime.PopulationConfiguration(initial_size=100)]
        dd = msprime.DemographyDebugger(demographic_events=dem_events,
                                        population_configurations=pop_config,
                                        migration_matrix=mig_mat)
```

apragsdale:
Ok @grahamgower, that makes more sense to me now! I have a hard time keeping migration arrow directions and the migration matrix vs transpose straight, and I think I even have papers where I've defined them in both ways. Doesn't help with my or others' confusion. I'm definitely not one to lecture about clear migration notation :)

I'll update your test and make sure everything comes out as expected now.



> This looks great to me @apragsdale! I've asked @petrelharp for a review here as he understands this stuff a lot better than me.
>
> We might do it here, or in a follow-up, but it would be nice to have a little bit of a tutorial on this to illustrate what it's doing and how it works. Nothing extensive, just a simple example.

That's a good idea -- I'll take a look at the docs and see where to fit in a very short tutorial on this.




> ps. I've put your name in the author list for the paper @apragsdale : [tskit-dev/msprime-1.0-paper#2](https://github.com/tskit-dev/msprime-1.0-paper/issues/2)

Thanks @jeromekelleher, that's really kind. I've only really contributed small bits and bobs here or there, but I'm happy to contribute where I can.

apragsdale:
Following up on the migration rate/direction... I just did a quick test with two populations to confirm that m[i][j] is the probability that a lineage in deme i has its parent in deme j. So in your drawing, if the arrows are the direction of movement of migrants forward in time (as opposed to the movement of lineages from one deme to another backward in time), then the migration matrix in your updated test code is correct. In that case, I want to take the transpose when constructing my "reachable" matrix, since we want to know where ancestral lineages could possibly end up through those migration events.

I pushed that change, and maybe @petrelharp or @jeromekelleher can confirm about the migration matrix directions.

petrelharp:
I'll have a look at this and the code a bit later on, but I'll bet you've got it worked out. I'd really like to rewrite the docs without using the phrase "backwards in time" which is never helpful...

grahamgower:
So this code has got me to a neat solution to a vaguely-related problem, and I was wondering if you'd be opposed to instead putting in a generalisation of the `possible_lineages()` method? (`reachable_through_migration()` is pure genius btw! I modified it slightly below)

Something like these two functions, appropriately altered to be methods of DemographyDebugger.
```
def reachable_through_migration(mig_mat, lineages):
    # get connected populations via migration
    k = len(lineages)
    M = np.eye(k) + lineages + mig_mat
    reachable = (np.linalg.matrix_power(M, k) > 0) * 1
    return (reachable.dot(lineages) > 0) * 1

def reachability_matrix(ddb):
    """
    A lineage-reachability matrix generator. One matrix is yielded for each
    epoch in the demography debugger `ddb`.
    """
    lineages = np.eye(ddb.num_populations, dtype='i1')
    for epoch in ddb.epochs:
        for de in epoch.demographic_events:
            if de.type == "mass_migration":
                for lineage in lineages:
                    if lineage[de.source]:
                        lineage[de.dest] = 1
                        if de.proportion == 1:
                            lineage[de.source] = 0
        lineages = reachable_through_migration(epoch.migration_matrix, lineages)
        yield lineages.copy()
```

The vaguely-related problem I wanted to solve is below. I imagine using it to get the split time, or most recent time of migration, for two populations in an arbitrary stdpopsim model.
```
def min_coalescence_time(ddb, samples):
    """
    Minimum coalescent time of `samples`.
    """
    coal_time = None
    for epoch, lineages in zip(ddb.epochs, reachability_matrix(ddb)):
        samples_now = [sample for sample in samples if sample.time < epoch.end_time]
        intersect = np.ones(ddb.num_populations)
        for sample in samples_now:
            np.logical_and(lineages[sample.population], intersect, out=intersect)
        if any(intersect):
            coal_time = max(epoch.start_time, max(sample.time for sample in samples))
            break
    return coal_time
```

I didn't yet figure out how to write `possible_lineages()` using `reachability_matrix()` above, but here is the analogous code.
```
def possible_lineages(ddb, samples=None):
    """
    Generates a vector of reachable lineage in each epoch for the given samples.
    """
    k = ddb.num_populations
    if samples is None:
        samples = [msprime.Sample(time=0, population=p) for p in range(k)]
    lineages = np.zeros(shape=(k, k), dtype='i1')
    for epoch in ddb.epochs:
        for sample in samples:
            if epoch.start_time <= sample.time < epoch.end_time:
                lineages[sample.population, sample.population] = 1
        for de in epoch.demographic_events:
            if de.type == "mass_migration":
                for lineage in lineages:
                    if lineage[de.source]:
                        lineage[de.dest] = 1
                        if de.proportion == 1:
                            lineage[de.source] = 0
        lineages = reachable_through_migration(epoch.migration_matrix, lineages)
        reachable = np.zeros(len(lineages), dtype='i1')
        for sample in samples:
            np.logical_or(reachable, lineages[sample.population], out=reachable)
        yield reachable
```

Thoughts?

petrelharp:
Note - we had to work out this sort of thing in writing [`coalescence_rate_trajectory( )`](https://github.com/tskit-dev/msprime/blob/master/msprime/simulations.py#L2252). I guess you're doing the one-lineage case of what we did there (the two-lineage case).

petrelharp:
Well, let me rephrase what I said above. I don't want to suggest that you do everything totally differently, but it would help me figure out where you all are at if you tell me what you think. The way I would have done this is to step back through time updating the matrix `P`, for which `P[i,j]` is the probability that a lineage that was initially in population `i` is currently in population `j`. This is a lot easier than the coalescence case because it doesn't depend on population size; so for instance you update `P` by right-multiplying it by `expm(dt * M)`, where `M` is derived from the migration matrix, and `dt` is the length of the epoch.  Strictly speaking it's more than you need, but you could extract other information from it.

If the answer is "ok, sure, but look we have simpler code working that does what we want already" then let me know and I'll review it.

apragsdale:
I think that's essentially what I've done here, but instead of checking the probability of finding a lineage that started in a given deme in another, I've just recorded indicators. The initial purpose of this function was to be able to set population sizes to zero if there is no ancestral material possible in a given deme over some epoch, so 0s and 1s let you just multiply that array against the population sizes. But I agree that there is more information you could get in a more general framework here.

Maybe @grahamgower could comment on what would be most useful - from his last comment, it looks like generalization would be useful, so I'll think about that a bit.

The other thing I was trying to be careful with is ancient samples, which may not align with demographic event times. Most of the function deals with checking sampling times and splitting epochs accordingly if needed.

grahamgower:
The probability matrix version sounds interesting, but the indicator version is definitely simpler. I see some code resembling the `expm(dt *M)` suggestion in `_calculate_coalescence_rate_trajectory()`, but my maths is weak, so I'd need to stare at this function for a very long time to figure out what's going on.

With regard to generalisation, I can't think of use cases other than the two already mentioned (pruning non-ancestral parts of the demography; and finding the minimum TMRCA of a set of samples).

The epoch splitting for ancient samples does seem tricky. I wonder if this can be simplified though? I'll poke it a bit more and see if anything falls out.

grahamgower:
> With regard to generalisation, I can't think of use cases other than the two already mentioned

Although, maybe it will help with https://github.com/apragsdale/demography/pull/44?

grahamgower:
I've taken another look @apragsdale, and I think that your `possible_lineages()` method should be kept as it is, without futher generalisation. It's pretty easy to copy and paste small chunks of code for other use cases if they arise.

petrelharp:
I've had a quick go at simplifying the coalescence rate code. This (or something very close) should be right:
```
def calculate_lineage_probabilities(self, steps):
        """
        Returns an array such that P[j, a, b] is the probability that a lineage that started in population a
        is in population b at time steps[j] ago.
        """
        num_pops = self.num_populations
        # P[i, j] will be the probability that a lineage that started in i is now in j
        P = np.zeros([num_pops, num_pops])
        Identity = np.eye(num_pops)
        for x in range(num_pops):
                P[x, x] = 1
        # add epoch breaks if not there already but remember which steps they are
        epoch_breaks = list(
            set([0.0] + [t for t in self.epoch_times if t not in steps])
        )
        steps_b = np.concatenate([steps, epoch_breaks])
        ix = np.argsort(steps_b)
        steps_b = steps_b[ix]
        keep_steps = np.concatenate(
            [np.repeat(True, len(steps)), np.repeat(False, len(epoch_breaks))]
        )[ix]
        assert np.all(steps == steps_b[keep_steps])
        mass_migration_objects = []
        mass_migration_times = []
        for demo in self.demographic_events:
            if type(demo) == MassMigration:
                mass_migration_objects.append(demo)
                mass_migration_times.append(demo.time)
        num_steps = len(steps_b)
        P_out = np.zeros((num_steps, num_pops, num_pops))
        P_out[0, :, :] = P
        # recall that steps_b[0] = 0.0
        for j in range(num_steps - 1):
            time = steps_b[j]
            dt = steps_b[j + 1] - steps_b[j]
            _, M = self._pop_size_and_migration_at_t(time)
            dM = np.diag([sum(s) for s in M])
            if time in mass_migration_times:
                idx = mass_migration_times.index(time)
                a = mass_migration_objects[idx].source
                b = mass_migration_objects[idx].dest
                p = mass_migration_objects[idx].proportion
                S = np.eye(num_pops, num_pops)
                S[a, b] = p
                S[a, a] = 1 - p
                P = np.matmul(P, S)
            G = M - dM
            P = np.matmul(P, _matrix_exponential(dt * G))
            P_out[j, :, :] = P
        return P[keep_steps, :, :]
```

I think what you're doing (indicators) is very nice; I guess I'm pushing to do this more general thing so that if we want to do it in the future we don't have to write more code: doing the general computation is just as easy, so we save on code complexity.

grahamgower:
That's awesome @petrelharp! [Here](https://github.com/tskit-dev/msprime/files/4531086/lineage_probs.py.txt) is a standalone script that includes your function to compare it with the analogous indicator version on the model I drew above. I just had to `s/P/P_out/` on the return value. The two versions don't quite do the same thing, so the comparison isn't perfect. But casting probabilities to indicators suggests it's working! There does seem to be a numerical oddity giving small negative numbers for some probabilities.

Unfortunately though, I reckon it will be quite nasty, or impossible, to write a wrapper of these functions that works with ancient samples (either the indicator or probabilitiy version). The ancient samples need to be considered inside this function, which does make it a lot more messy. And the `min_coalescent_time()` application I showed above doesn't work quite right with ancient samples, requiring yet a different perversion of the "nice" code.

petrelharp:
> ancient samples

Ah-ha - right; to get this information, we could change this so that `P[i, j]` is the probability that "sampling group `i`" is currently in population `j`, where a "sampling group" is a collection of samples taken at the same time+place combination. This will make `P` rectangular instead of square, but it is updated in just the same way, except that `P[i,j] = 0` until the time that sampling group `i` was sampled, in the case of ancient samples. I could make this change (and help figure out the discrepancy you observed?).

But - let me know what you all think we should do?

apragsdale:
I have no problem with rewriting some code to make it more general, and I don't have a particular attachment to what I've written up so far (it didn't take all that long). If you think that generalization will be useful in other scenarios, and I can see that it would be, especially considering "sampling groups" separately. If that's the consensus, I'm happy to rework some things here.

I've been a bit slow this last month, with both replying and pushing code and projects forward, but let me know what you think is the best thing to do here and I'll get to working on it.

petrelharp:
Let me know how I can help!

apragsdale:
Ok, curious to know what you all think about this. I added the function that @petrelharp suggested to get lineage probabilities at give time points, which I think works fine, but it doesn't always give the same answer as the function that computes indicators over epochs.

This discrepancy can be seen in the simple two population test with symmetric migration, where there should be lineages in both populations even though we sample only from one. But to pick up that migration moves lineages to the non-sampled population, we would need to look farther into the past using the migration matrix to see that. At the time of sampling (time 0), there are no lineages in second population, but there could be for any epsilon time farther into the past.

Using that DemographyDebugger object from `test_possible_lineages_with_migration(self)`, this can be seen by running `dd.calculate_lineage_probabilities(steps=[0, .01])`.

msprime/simulations.py
```diff
@@ -2254,6 +2254,237 @@ def population_size_trajectory(self, steps):
...
+
+        sample_sets = {}
+        if samples is None:
+            sample_sets[0] = list(range(num_pops))
+        else:
+            for sample in samples:
+                sample_sets.setdefault(sample.time, [0 for _ in range(num_pops)])
+                sample_sets[sample.time][sample.population] = 1
+
+        steps = [e.start_time for e in self.epochs]
```
grahamgower, line 2462: How about adding an epsilon to each start_time here, where `epsilon = np.finfo(float).eps`?

grahamgower:
> Ok, curious to know what you all think about this. I added the function that @petrelharp suggested to get lineage probabilities at give time points, which I think works fine, but it doesn't always give the same answer as the function that computes indicators over epochs.
> 
> This discrepancy can be seen in the simple two population test with symmetric migration, where there should be lineages in both populations even though we sample only from one. But to pick up that migration moves lineages to the non-sampled population, we would need to look farther into the past using the migration matrix to see that. At the time of sampling (time 0), there are no lineages in second population, but there could be for any epsilon time farther into the past.

Right. I think a reasonable approach is to just add some epsilon to the times. I don't know how useful it is to consider the time=0 (exactly) case? Similarly for ancient samples, we're probably more interested in the possible lineages at `time=ancient_sample.time + epsilon`, rather than `ancient_sample.time`.

apragsdale:
I think that will end up being an issue, because we'd then have our epochs offset by epsilon and not aligned with the epochs from the demographic events or exact sampling times. The original goal was to get the epochs over which lineages can exist in each population, so even though one of those keys in the returned dictionary is `0`, that just means the epoch starting at time 0, or the epoch starting at some other sampling time.

In the end it looks like adding epsilon does end up being an issue, and it's not a very clean solution, because we'd then have our epochs offset by epsilon and not aligned with the epochs from the demographic events or exact sampling times. My feeling right now is to use the original function that has been more tested and looks to be working, though is specialized, and then also keep in the general function for computing probabilities of finding lineages in a each population at a given time. The function now might not use it, but like Peter said, it could come in handy for something else in the future.

apragsdale:
Just pushed changes to reflect that last comment, and I think as is this is the simplest solution for the time being. If/when it's good to go, I can squash my commits.

But I'm not really sure why the build is failing now, since it passes `flake8` on my machine.. do I need to reformat all of it that is being suggested when I check out the details of the failed builds @jeromekelleher

jeromekelleher:
We've just added [black](https://github.com/psf/black) formatting to the lint @apragsdale, sorry for the confusion. The simplest thing to do is install pre-commit and run it:
```
python3 -m pip install -r requirements/development.txt
pre-commit install
```

Now Black should be run each time you make a commit. (The dev docs explaining all this are being written)

apragsdale:
Ok, that makes sense, and no problem. The automatic formatting looks like a nice feature.

I think I need some extra tests of the `calculate_lineage_probabilities` function, but I want to see what people think about including it here first.

petrelharp:
Looks very nice - you have done what I was going to suggest instead of adding epsilon.

I also really like the `sample_time` argument. That looks quite useful. I don't think it's quite right yet though - if the sampling time comes halfway through an epoch then I don't think the migration during that epoch happens? I could be missing something, but it doesn't look like `sample_time` is tested, either.

We should probably pull in the explicit tests from `TestDemographyTrajectories` to get some easy analytic expectations to compare to (see eg `test_two_pop`).

petrelharp:
> I think I need some extra tests of the calculate_lineage_probabilities function, but I want to see what people think about including it here first.

Would we implement `possible lineages` in terms of `calculate_lineage_probabilities` (once they agree, of course)?

grahamgower:
> 
> Would we implement `possible lineages` in terms of `calculate_lineage_probabilities` (once they agree, of course)?

Wouldn't the `sample_time` need to be per-population in `calculate_lineage_probabilities` to do that?

jeromekelleher:
Just a quick note here @apragsdale - I think you need to update your local git config to use the same email address as your github account. Otherwise your commits won't be accounted in the contributor history.

petrelharp:
> Wouldn't the sample_time need to be per-population in calculate_lineage_probabilities to do that?

No, but we would need to run `calculate_lineage_probabilities` once per unique sampling time.

apragsdale:
I'm having a bit of trouble getting a simple function for lineage indicators using the `calculate_lineage_probabilites` function. I've been taking the same strategy that you suggest @petrelharp (one call per sampling time), but then I'm getting a little stumped by continuous migration over the epoch. I think I need to check probabilities at the _end_ of epochs instead of the beginning.

This brings up another choice I need to make, about the order of events and what we expect the behavior of that function to be. As the function is written now, if I want the lineage probabilities at some time step, I apply the mass migration event before reporting the lineage probabilities, but I'm not sure this is the right strategy. Take the following example:

Sample deme 0 at time 0, but also at time 0 there is a mass migration event that takes all lineages from deme 0 to deme 1. Then the epoch from time 0 onward would have lineage probability of 1 in deme 1. Should `P_out` at time zero be `[[1, 0], [0, 1]]` or `[[0, 1], [0, 1]]`?

I think I want the first of those, and then check epoch lineage probabilities at the end of epochs before applying events that affect the next epoch.

apragsdale:
I think I've got a reasonably simple implementation to get indicators using the lineage probability function now.

petrelharp:
Hello! I've just had a bit of a mix-up around branches and repos and things but I think it's all sorted, and I've made a PR against your branch with some additional tests.  I was motivated to put these in because of this good question:

> Sample deme 0 at time 0, but also at time 0 there is a mass migration event that takes all lineages from deme 0 to deme 1. Then the epoch from time 0 onward would have lineage probability of 1 in deme 1. Should P_out at time zero be [[1, 0], [0, 1]] or [[0, 1], [0, 1]]?

My answer was: I don't know! It depends what the simulator does, right? We should be reporting something that matches what people observe in simulations? 
- I don't think we can simulate for zero time (but if we could, probably we'd just get back all the samples being in their initial demes, thus not affected by the mass migration?)
- Similarly, for an ancient sample at time t, we can ask where its lineage was at time t, and my guess (that is I think right?) is that it'll be in the population it started out in
So, in the tests I added to the PR, I'm checking that actual lineages are always in places where they have positive probability of being according to calculate_lineage_probabilities.  This fails for examples with mass migrations, I think because we've got the wrong choice here.

Also, I added a comparison to the analytical answer for a two-pop-with-migration example, that passes.

What do you think? Merge in my PR if you like it.

petrelharp:
Also: how about `DemographyDebugger.lineage_probabilities( )` instead of `calculate_`?

apragsdale:
Ok! No problem about the git mixups -- I've also been wrestling with some git rebasing for a few days now. It keeps overwriting my changes on some branches. Can't quite figure it out...

I agree that it looks like I chose the wrong one of those options, that mass migration events don't move lineages if they were sampled at the same time. Or rather, I don't think one is wrong and the other is right, just that it's a modeling choice that matters for this question. The tests that you wrote are a really nice way of checking that. Thank you!

I think I just need an if-statement to only apply the mass migration event if it has time greater than the sampling time. That should be an easy fix and should pass the test. I'll check that soon.

petrelharp:
LMK when it's ready to look at.

apragsdale:
I think it is ready to look at, although at this point I've still left in my original implementation and tests for that function to compare with the more recent function. It should be removed before merging.

msprime/simulations.py
```diff
@@ -2355,6 +2357,263 @@ def population_size_trajectory(self, steps):
...
+                mass_migration_times.append(demo.time)
+
+        for jj in range(first_step, len(all_steps) - 1):
+            t_j = all_steps[jj]
+
+            # apply any mass migration events to P
+            # so if we sample at this time, we do no account for the instantaneous
+            # mass migration events that occur at the same time. that will show up
+            # at the next step
+            if t_j > sample_time:
```
petrelharp, line 2535: Isn't this always true since we started at `first_step`?

tests/test_demography.py
```diff
@@ -2797,3 +2797,326 @@ def test_no_census_nodes_above_root_nodes(self):
...
+
+        lineages2 = dd.indicators_from_probabilities(samples=samples)
+        self.assertTrue(len(lineages2) == 4)
+        self.assertTrue(np.all(lineages2[(0, 50)] == [1, 1, 1, 1]))
+        self.assertTrue(np.all(lineages2[(50, 100)] == [1, 1, 1, 0]))
+        self.assertTrue(np.all(lineages2[(100, 150)] == [1, 1, 0, 0]))
+        self.assertTrue(np.all(lineages2[(150, np.inf)] == [1, 0, 0, 0]))
+
+    def test_possible_lineages_with_migration(self):
+        # draw sample from one population, but others connection through migration
```
petrelharp, line 2843: I don't quite understand this comment?

tests/test_demography.py
```diff
@@ -2797,3 +2797,326 @@ def test_no_census_nodes_above_root_nodes(self):
...
+                lineage = []
+                while p != tskit.NULL:
+                    lineage.append(ts.node(p))
+                    p = t.parent(p)
+                probs = dd.lineage_probabilities(
+                    [n.time for n in lineage], sample_time=lineage[0].time
+                )
+                pop = lineage[0].population
+                for j, n in enumerate(lineage):
+                    self.assertGreater(probs[j, pop, n.population], 0.0)
```
petrelharp, line 3010: Should also test `possible_lineages` here

petrelharp, line 3010: Should also test possible_lineages here

msprime/simulations.py
```diff
@@ -2355,6 +2357,263 @@ def population_size_trajectory(self, steps):
...
+            that are more recent than sample_time, the probability of finding the
+            lineage in any population is zero.
+        :return: An array of dimension len(steps) by num pops by num_pops.
+        """
+        num_pops = self.num_populations
+        # P[i, j] will be the probability that a lineage that started in i is now in j
+        P = np.zeros([num_pops, num_pops])
+
+        for ii in range(num_pops):
+            P[ii, ii] = 1
```
grahamgower, line 2485: Use `P = np.eye(num_pops)` instead?

msprime/simulations.py
apragsdale: I want to exclude the migration events that occur at the same time as the sampling time.

apragsdale:
Thanks for the comments. I do tend to over-comment my code, comments mostly leftover placeholders from first writing the functions, so much of it can be removed. I'll tidy up the functions and tests and take care of your suggestions.

petrelharp:
Let us know when we should look at it again? It's probably time to take out the extra implementation?

jeromekelleher:
> It's probably time to take out the extra implementation?

if we have a second implementation, maybe put this in as a part of the tests?

apragsdale:
Ack, I had already deleted the extra implementation in the last push (and had squashed those commits), and I'm not having much luck recovering the intermediate commits. (Maybe best not to squash commits until ready to merge?). Otherwise, I think things are ready for another review.

msprime/simulations.py
```diff
@@ -2214,6 +2214,8 @@ def __init__(
                 population_configurations, saved_sample_sizes
             ):
                 pop_config.sample_size = sample_size
+        self.migration_matrix = migration_matrix
+        self.population_configurations = population_configurations
```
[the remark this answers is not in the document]
jeromekelleher: LGTM, makes sense to store these here (so long as we don't modify them).

jeromekelleher:
> Ack, I had already deleted the extra implementation in the last push (and had squashed those commits), and I'm not having much luck recovering the intermediate commits. (Maybe best not to squash commits until ready to merge?). 

Oh well, not to worry. It is better to just keep adding commits until ready for the final merge and then rebase, mostly for this reason.

petrelharp:
>  I'm not having much luck recovering the intermediate commits. 

I've just put my version of history to a branch on my repo: https://github.com/petrelharp/msprime/tree/unsquashed_where_are_lineages if you want to pull it back out... up to you, really.

jeromekelleher:
How's this looking @apragsdale? It'd be nice to get this in soon as there's a bit of refactoring going on.

apragsdale:
I think it's looking fine here, and we got all the subtleties of the behavior worked out. Happy to wrap this up if it looks good to you.

jeromekelleher:
OK, let's merge then. Can you do a final rebase please?

apragsdale:
Ok, looks like the rebase worked, so it should be ready to go.

jeromekelleher:
Merged! Thanks @apragsdale, this is a great feature. I think it would be good to highlight the new additions to the DemographyDebugger in the msprime 1 paper, so I've added an issue to track this: https://github.com/tskit-dev/msprime-1.0-paper/issues/16

Would you be up for writing a paragraph or two about this @apragsdale, while it's fresh in your head ?

PR #970, opened 2020-04-23
Streamline species tree parsing

jeromekelleher:
This does two important things

1) Starts the initial use of attrs. Since we're going to be depending on this upstream in tskit soon, there's no point in not using it and it makes the type of data classes we commonly use *so* much more streamlined.

2) Adds a new DemographySpecification class, which is returned by the species tree parsers. This is nice as we don't have to unpack tuples, and gives us neat ways to take samples from populations. We've ended up doing something similar in stdpopsim, so I think it's worth digging in a bit more here to give support for it in msprime.

We'd need to do a bit more to support sampling times, but I thought I'd open this PR initially to get some feedback. I think we'd probably want to add an optional "name" argument to the PopulationConfiguration object, which would then match naturally to the kwargs magic.

@grahamgower, any thoughts here? This would simplify our lives a bit in stdpopsim, right?

Thoughts appreciated!

(I accidentally ran black on the code as this is automatic for me now, so ignore all the formatting noise. We'll hopefully have run black on all the code soon, so this diff will disappear.)

The important bit is this:
```python
        species_tree = "(((human:5.6,chimpanzee:5.6):3.0,gorilla:8.6):9.4,orangutan:18.0)"
        spec = msprime.parse_species_tree(
            species_tree, branch_length_units="myr", Ne=10000, generation_time=20)

        # Take one sample from each population
        ts = msprime.simulate(samples=spec.sample(1, 1, 1, 1), **spec.asdict())
        self.assertEqual(ts.num_trees, 1)
        self.assertEqual(ts.num_samples, 4)
        self.assertEqual(ts.num_populations, 4)
        for j, u in enumerate(ts.samples()):
            self.assertEqual(ts.node(u).population, j)

        # Use the population names to get the samples
        samples = spec.sample(human=4, gorilla=2)
        ts = msprime.simulate(samples=samples, **spec.asdict())
        self.assertEqual(ts.num_trees, 1)
        self.assertEqual(ts.num_samples, 6)
        for j, u in enumerate(ts.samples()):
            pop = 0 if j < 4 else 2
            self.assertEqual(ts.node(u).population, pop)

        # Order of keywords is respected, 
        samples = spec.sample(gorilla=2, human=4)
        ts = msprime.simulate(samples=samples, **spec.asdict())
        self.assertEqual(ts.num_trees, 1)
        self.assertEqual(ts.num_samples, 6)
        for j, u in enumerate(ts.samples()):
            pop = 2 if j < 2 else 0
            self.assertEqual(ts.node(u).population, pop)
```


grahamgower:
Very nice! It looks like this will make speccing small models very neat and tidy! I wonder, is it reasonable to start constructing a model by using a newick string, and then putting non-tree bits in afterwards (e.g. just apending to the `demographic_events` and resorting)? Are there any plans to have a `migration_matrix` in `class DemographySpecification`?

The dict reusage for kwargs makes things super clean too. I do worry slightly that the logical endpoint of this looks something like matplotlib, where you have to traverse a labyrinth of documentation to figure out which function "claims" a certain kwarg key, only to find you just misspelled it. We'll need to be careful of this in stdpopsim, particularly if some kwarg keys are passed transparently into msprime.

jeromekelleher:
> Very nice! It looks like this will make speccing small models very neat and tidy! I wonder, is it reasonable to start constructing a model by using a newick string, and then putting non-tree bits in afterwards (e.g. just apending to the demographic_events and resorting)?

Sure, no reason not to I guess. 

> Are there any plans to have a migration_matrix in class DemographySpecification?

That's what got me thinking about it in the first place - we'll surely want to set up migration matrices with these inputs at some point (see the parse_starbeast function as well), so we want to make it extensible.

> The dict reusage for kwargs makes things super clean too. I do worry slightly that the logical endpoint of this looks something like matplotlib, where you have to traverse a labyrinth of documentation to figure out which function "claims" a certain kwarg key, only to find you just misspelled it. We'll need to be careful of this in stdpopsim, particularly if some kwarg keys are passed transparently into msprime.

Yeah, it's definitely something to be used sparingly. I'm imaging us using the DemographySpecification in stdpopsim, using this to generate samples (once we've figured out an appropriate API for specifying sampling times), and then passing the spec and samples to ``engine.simulate()``. So, we shouldn't be doing any more ``**kwargs`` funkiness than we are already.

jeromekelleher:
Updated to use #1025. Ready to go after that's merged, I think.

jeromekelleher:
Closes #958

PR #978, opened 2020-04-24
Refactor migration matrix handling

### LINKED ISSUE #666: Convert _msprimemodule.c to use numpy C API where possible

jeromekelleher:
This refactors the handling of migration matrices to natively use numpy and to avoid linearising coordinates within it. This is a nice update, and will clear the way for some planned simplifications and refactoring of the demography definition code.

**HOWEVER**, as I was going through the process of changing the old ``matrix_index`` used to describe the bit of the migration matrix we change on a MigrationRateChangeEvent to (source, dest), I hit a worrying inconsistency. In the mspms code, it turns out that we use (dest, source) instead and I'm worried that there's a long-standing bug in the Python API where I've flipped the direction of migration. It would be surprising if such a fundamental issue had slipped through all this time, but migration has been a persistent source of confusion.

@petrelharp, any chance you can take a look through here and see what you think?

I'm sure that mspms is correct, because there's loads of statistical tests checking all sorts of migration patterns (probably because I was worried about just this sort of thing).

Closes #666

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -28,6 +28,7 @@ import tempfile
 import unittest
 
 import newick
+import numpy as np
 import tskit
 
 import msprime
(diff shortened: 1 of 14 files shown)


jeromekelleher:
So, ms says:

> The  elements  of  the  migration  matrix are Mij= 4N0mij, i,j= 1,...,npop, where, mij is the fraction of subpopulation which is made up of migrants from subpopulation j each generation.

so that's *forwards* in time, i=dest, j = source.

Msprime model definition docs say:

> Each element of the matrix M[j, k] defines the fraction of population j that consists of migrants from population k in each generation.

So that's the same thing, except we've factored out the 4N. Our M is therefore Hudson's m. 

Then,  **backwards in time** time as the algorithm is working, we get: https://github.com/tskit-dev/msprime/blob/39cd6c973b038fd173b763c2cd363557db00a314/lib/msprime.c#L3641-L3660

So we're moving lineages from population j (source) to k (dest), in the backwards time sense, which is consistent with the how we use source and dest in [MassMigration](https://msprime.readthedocs.io/en/latest/api.html#msprime.MassMigration).

OK, I think I've convinced myself that it's correct, I've just swapped the time direction of source/dest in the mspms code. Phew. 

God*damn* this stuff is confusing. Is there any way we can get better terminology for this? I'm in the middle of a refactor of the Demography code (some nice stuff in the pipeline, I think), and it would be really nice to reduce the confusion (and my stress levels!) here.

petrelharp:
I'll look through this carefully in a bit. Maybe this is just unavoidably confusing, but I think some of the confusion here is the terms "source" and "destination". I also think that using the phrases "backwards in time" and "forwards in time" usually leads to confusing descriptions.

petrelharp:
I searched for everywhere it says `migration` in the docs. Thoughts:

1. [Here](https://msprime.readthedocs.io/en/stable/api.html?highlight=migration#msprime.MigrationRateChange) we should say exactly what the `matrix_index` means, maybe:

> A matrix index of (j, k) sets the fraction of population j that is composed of immigrants from population k in each generation.

2.  I'd rephrase this bit:
> A mass migration event in which some fraction of the population in one deme simultaneously move to another deme, viewed backwards in time. Each lineage currently present in the source population moves to the destination population (backwards in time) with probability equal to proportion.

to say

> A mass migration event in which some fraction of the population in the "source" deme is composed of migrants from the "destination" deme. Following the coalescent process back through time, each lineage currently present in the source population moves to the destination population with probability equal to proportion.

3. The description [here](https://msprime.readthedocs.io/en/stable/api.html?highlight=migration#simulation-model) (fourth paragraph) is good.
Maybe this is the right place to say that 

> Note that migration rates are specified in a way that makes sense for the coalescence process, so that the `(j, k)`th entry of the migration matrix, `M[i,j]`, gives the rate at which an ancestral lineage moves from population `j` to population `k` as one follows it back through time. In terms of the demographic process of the populations, if a total of `n` migrants move from population `k` to population `j` per generation, and population `j` has a total of `N[j]` individuals, then the migration matrix has `M[j,k] = n/N[j]`. This differs from the migration matrix one usually uses in population demography: if `m[k,j]` is the proportion of individuals (in the usual sense; **not** lineages) in population `k` that move to population `j` per generation, then translating this proportion of population `k` to a proportion of population `j`, we have `M[j,k] = m[k,j] * N[k] / N[j]`.

tests/test_cli.py
```diff
@@ -827,6 +836,7 @@ def check(args, results):
             [(2.2, 1, None), (3.3, 5.5, (2, 0))],
         )
 
+    @unittest.skip("Eek! Migration indexes!")
```
petrelharp, line 839: note this is still being skipped

jeromekelleher:
Thanks @petrelharp, this is really helpful. I'll work through this and update a bit later. A quick question first though: are you OK with changing the API for MigrationRateChange from ``matrix_index`` to ``source`` and ``dest``? Since source and dest have the same sense as MassMigration, this seems more helpful than ``matrix_index``, and it does simplify other code.

petrelharp:
> are you OK with changing the API for MigrationRateChange

That makes sense, yes.

petrelharp:
And - I don't think me reading over the code will convince me of the correctness, as it's all very confusing, as you say. But, a few simple tests should do it: is there currently a test that has a migration matrix like `[[0, 1][0, 0]]` and checks which population they all end up in? (all the roots should be in population 1, not population 0). Doing this for a migration matrix, a migration matrix change, and a mass migration should verify that things are going the right way.

jeromekelleher:
OK, thanks @petrelharp. I'm sure there must be a number of tests that do this, I just panicked a bit (I need a break I think). I'll have a good check for these tests when I finish this up.

jeromekelleher:
This is ready to go I think, I've incorporated your documentation changes @petrelharp.  I'll follow up with an issue to make sure that we review the documentation to make sure it makes sense.

I can confirm that the rates are all going in the correct direction - my panic attack was induced by some poor choice of variable names in mspms, which decided to look at source and dest in the opposite direction to the rest of the code.

PR #980, opened 2020-04-27
Format C code with clang-format

benjeffery:
I've used the same configuration as tskit, excluded `avl`.


jeromekelleher:
Great, let's get #966 merged first and then re-apply this (it's making some changes to the C)

PR #989, opened 2020-05-04
Refactor demography

jeromekelleher:
Add attrs as a dependency, and refactor some of the simple data classes that we have to use it.

This is to improve code readability, maintainability inspectability (automatic reprs, etc).

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -31,6 +31,7 @@ import os
 import sys
 import warnings
 
+import attr
 import numpy as np
 import tskit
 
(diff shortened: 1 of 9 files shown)


jeromekelleher:
I think this is ready to go in. The idea is to port over as much of the simple data classes that we use in the public API for msprime as possible to use attrs. This is mainly to take advantage of the nice reprs, equality operations etc that we get for free with attrs.

I've done the minimum here to port the existing code to using attrs, as I didn't want to change too much at once. We can probably start bringing in some things like type and bounds checking using attrs here afterwards, which might give earlier errors to users (which would be helpful, probably).

jeromekelleher:
I don't think there's anything controversial here other than using ``attrs``, which we've already agreed on, so I'm going to merge this. #990 will make it clearer where I'm aiming to go on this, so let's focus review effort there.

PR #1009, opened 2020-05-11
generalize mutation models

petrelharp:
Here's how I'd like to make mutation models more flexible, as outlined in #1006. Mutation models would only need to have the 'pick root allele' and 'transition allele' methods; what we've done so far is the special case of MutationMatrixModels. I've coded the (small) changes up in python, with the SLiM mutation model.

I've also made the check that we aren't sticking a mutation above an existing one optional (in python only, again).

How's this look?

Note: with this in place, it'll be easy to do models of microsats, etcetera.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -80,6 +80,11 @@ typedef struct {
 
 typedef struct {
     PyObject_HEAD
+    mutation_model_t *mutation_model;
+} InfiniteAllelesMutationModel;
+
+typedef struct {
+    PyObject_HEAD
     mutgen_t *mutgen;
     RandomGenerator *random_generator;
     IntervalMap *rate_map;
@@ -2041,27 +2046,16 @@ static PyGetSetDef MatrixMutationModel_getsetters[] = {
     {NULL}  /* Sentinel */
 };
 
-static PyMemberDef MatrixMutationModel_members[] = {
-    {NULL}  /* Sentinel */
-};
-
-static PyMethodDef MatrixMutationModel_methods[] = {
-    {NULL}  /* Sentinel */
-};
-
 static PyTypeObject MatrixMutationModelType = {
     .tp_name = "_msprime.MatrixMutationModel",
     .tp_basicsize = sizeof(MatrixMutationModel),
     .tp_dealloc = (destructor)MatrixMutationModel_dealloc,
     .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
     .tp_doc = "MatrixMutationModel objects",
-    .tp_methods = MatrixMutationModel_methods,
-    .tp_members = MatrixMutationModel_members,
     .tp_getset = MatrixMutationModel_getsetters,
     .tp_init = (initproc)MatrixMutationModel_init,
     .tp_new = PyType_GenericNew,
-}
-;
+};
 
 /*===================================================================
  * Slim mutation model
@@ -2147,24 +2141,109 @@ static PyGetSetDef SlimMutationModel_getsetters[] = {
     {NULL}  /* Sentinel */
 };
 
-static PyMemberDef SlimMutationModel_members[] = {
-    {NULL}  /* Sentinel */
-};
-
-static PyMethodDef SlimMutationModel_methods[] = {
-    {NULL}  /* Sentinel */
-};
-
 static PyTypeObject SlimMutationModelType = {
     .tp_name = "_msprime.SlimMutationModel",
     .tp_basicsize = sizeof(SlimMutationModel),
     .tp_dealloc = (destructor)SlimMutationModel_dealloc,
     .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
     .tp_doc = "SlimMutationModel objects",
-    .tp_methods = SlimMutationModel_methods,
-    .tp_members = SlimMutationModel_members,
     .tp_getset = SlimMutationModel_getsetters,
     .tp_init = (initproc)SlimMutationModel_init,
+    .tp_new = PyType_GenericNew,
+};
+
+/*===================================================================
+ * Infinite alleles mutation model
+ *===================================================================
+ */
+
+static int
+InfiniteAllelesMutationModel_check_state(InfiniteAllelesMutationModel *self)
+{
+    int ret = 0;
+    if (self->mutation_model == NULL) {
+        PyErr_SetString(PyExc_SystemError, "SlimMutationModel not initialised");
+        ret = -1;
+    }
+    return ret;
+}
+
+static void
+InfiniteAllelesMutationModel_dealloc(InfiniteAllelesMutationModel* self)
+{
+    if (self->mutation_model != NULL) {
+        mutation_model_free(self->mutation_model);
+        PyMem_Free(self->mutation_model);
+        self->mutation_model = NULL;
+    }
+    Py_TYPE(self)->tp_free((PyObject*)self);
+}
+
+static int
+InfiniteAllelesMutationModel_init(InfiniteAllelesMutationModel *self,
+        PyObject *args, PyObject *kwds)
+{
+    int ret = -1;
+    int err;
+    static char *kwlist[] = {"start_allele", NULL};
+    unsigned long long start_allele = 0;
+
+    self->mutation_model = NULL;
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|K", kwlist, &start_allele)) {
+        goto out;
+    }
+
+    /* Note: it's important we zero out mutation_model here because
+     * we can error before we can mutation_model_alloc, leaving the
+     * object in an uninitialised state */
+    self->mutation_model = PyMem_Calloc(1, sizeof(*self->mutation_model));
+    if (self->mutation_model == NULL) {
+        PyErr_NoMemory();
+        goto out;
+    }
+    err = infinite_alleles_mutation_model_alloc(self->mutation_model,
+        (uint64_t) start_allele, 0);
+    if (err != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = 0;
+out:
+    return ret;
+}
+
+static PyObject *
+InfiniteAllelesMutationModel_get_start_allele(InfiniteAllelesMutationModel *self,
+        void *closure)
+{
+    infinite_alleles_t *params = &self->mutation_model->params.infinite_alleles;
+    return Py_BuildValue("K", (unsigned long long) params->start_allele);
+}
+
+static PyObject *
+InfiniteAllelesMutationModel_get_next_allele(InfiniteAllelesMutationModel *self,
+        void *closure)
+{
+    infinite_alleles_t *params = &self->mutation_model->params.infinite_alleles;
+    return Py_BuildValue("K", (unsigned long long) params->next_allele);
+}
+
+static PyGetSetDef InfiniteAllelesMutationModel_getsetters[] = {
+    {"start_allele", (getter) InfiniteAllelesMutationModel_get_start_allele, NULL,
+        "Returns the initial allele"},
+    {"next_allele", (getter) InfiniteAllelesMutationModel_get_next_allele, NULL,
+        "Return the next allele"},
+    {NULL}  /* Sentinel */
+};
+
+static PyTypeObject InfiniteAllelesMutationModelType = {
+    .tp_name = "_msprime.InfiniteAllelesMutationModel",
+    .tp_basicsize = sizeof(InfiniteAllelesMutationModel),
+    .tp_dealloc = (destructor)InfiniteAllelesMutationModel_dealloc,
+    .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+    .tp_doc = "InfiniteAllelesMutationModel objects",
+    .tp_getset = InfiniteAllelesMutationModel_getsetters,
+    .tp_init = (initproc)InfiniteAllelesMutationModel_init,
     .tp_new = PyType_GenericNew,
 };
 
@@ -2212,6 +2291,7 @@ MutationGenerator_init(MutationGenerator *self, PyObject *args, PyObject *kwds)
     PyObject *py_model = NULL;
     MatrixMutationModel *matrix_mutation_model = NULL;
     SlimMutationModel *slim_mutation_model = NULL;
+    InfiniteAllelesMutationModel *infinite_alleles_model = NULL;
     mutation_model_t *model = NULL;
 
     self->mutgen = NULL;
@@ -2246,10 +2326,17 @@ MutationGenerator_init(MutationGenerator *self, PyObject *args, PyObject *kwds)
             goto out;
         }
         model = slim_mutation_model->mutation_model;
+    } else if (PyObject_TypeCheck(py_model, &InfiniteAllelesMutationModelType)) {
+        infinite_alleles_model = (InfiniteAllelesMutationModel *) py_model;
+        if (InfiniteAllelesMutationModel_check_state(
+                    infinite_alleles_model) != 0) {
+            goto out;
+        }
+        model = infinite_alleles_model->mutation_model;
     } else {
         PyErr_SetString(PyExc_TypeError,
-            "model must be an instance of MatrixMutationModel "
-            "or SlimMutationModel");
+            "model must be an instance of MatrixMutationModel, "
+            "SlimMutationModel or InfiniteAllelesMutationModel.");
         goto out;
     }
     self->model = py_model;
@@ -4540,6 +4627,15 @@ PyInit__msprime(void)
     Py_INCREF(&SlimMutationModelType);
     PyModule_AddObject(module, "SlimMutationModel",
             (PyObject *) &SlimMutationModelType);
+
+    /* InfiniteAllelesMutationModel type */
+    InfiniteAllelesMutationModelType.tp_base = &BaseMutationModelType;
+    if (PyType_Ready(&InfiniteAllelesMutationModelType) < 0) {
+        return NULL;
+    }
+    Py_INCREF(&InfiniteAllelesMutationModelType);
+    PyModule_AddObject(module, "InfiniteAllelesMutationModel",
+            (PyObject *) &InfiniteAllelesMutationModelType);
 
     /* Errors and constants */
     MsprimeInputError = PyErr_NewException("_msprime.InputError", NULL, NULL);
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -61,6 +61,13 @@ typedef struct {
     interval_map_t *interval_map;
 } IntervalMap;
 
+/* TODO we should refactor some of the code for dealing with the
+ * mutation_model in this base class (which currently does nothing).
+ */
+typedef struct {
+    PyObject_HEAD
+} BaseMutationModel;
+
 typedef struct {
     PyObject_HEAD
     mutation_model_t *mutation_model;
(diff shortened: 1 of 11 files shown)


tests/test_mutations.py
```diff
@@ -27,11 +27,11 @@
 import attr
 import numpy as np
 import scipy.stats as stats
-import tskit
 
 import _msprime
 import msprime
 import tests.wright_fisher as wf
+import tskit
```
jeromekelleher, line 34: tskit is third party - these are *local* imports.

petrelharp:
Huh - the pre-commit failed because for some reason I had a `tskit/` directory in my repo, with only a `__pycache__` in it, and so `pre-commit` thought that tskit was a local repo so wanted to reorder imports.

petrelharp:
> modulo the text processing, which will be tedious

Can we use `strcmp` and related tools?

jeromekelleher:
> Can/should we use strcmp and related tools?

Sure, whatever is simplest. We'll just have to look very hard at it to make sure that everything is correct. Memory management for variable length strings like this is painful.

petrelharp:
This is not yet working (segfaults), but input on the general approach would be good. Not sure at all this is the right way to go.

petrelharp:
Exciting update! This passes tests (er, CI notwithstanding??)!  I'll stop force-pushing.

So, the idea is that a mutation model has to define two method: `root_allele( )` and `transition_allele( )`. They each return ints (which is necessary so they can error). And, currently, they provide output by being passed in a `**char` which then gets its contents copied over into the correct place.  For the mutation matrix model case, wehre the possible alleles are fixed, this is natural. But, for the next (SLiM) case, each allele will be new, and so these methods will have to allocate some memory. We can just malloc, but then that's lots of little mallocs. Or, we could give the slim mutation model its own blkalloc?  Or, a totally different approach?

lib/msprime.h
```diff
@@ -318,25 +318,25 @@ typedef struct demographic_event_t_t {
...
-typedef struct mutation_model_t {
-    int (*root_allele)(void *params, gsl_rng *rng, char **allelep);
-    int (*transition_allele)(void *params, gsl_rng *rng,
-        char *parent_allele, tsk_size_t parent_allele_length, char **allelep);
-    void *params;
+typedef struct _mutation_model_t {
+    union {
+        mutation_matrix_t mutation_matrix;
+        /* Other known mutation models */
+    } params;
```
petrelharp, line 332: oh, nice!

petrelharp:
Oh, gee, that's much more elegant, good.

Thinking ahead to the SLiM version, I wonder if we want to do some of this slightly differently. Right now, `transition` gives a pointer to the allele's pre-existing location in memory, which gets copied into `mut->derived_state`. For SLiM mutations, we'll have to create the allele somewhere just for it to be immediately copied over to `mut->derived_state`. This is an extra alloc and copy that would be unnecessary if `transition` was in charge of setting `mut->derived_state` directly. The main advantage to the current set-up is that `transition` won't need to alloc... but, for SLiM mutations it'll have to alloc anyhow.  So, I'm thinking that we should make `transition` (and `choose_root`) in charge of allocating memory too, maybe by adding a `blkalloc` attribute to `mutation_model_t`?  Or by sharing with mutgen, somehow?

A related issue is that for the SLiM generator we'll want to add metadata to the new mutations also. So, maybe it makes sense for `transition` to just get a pointer to the mutation object, and `choose_root` a pointer to the site object, and that they are responsible for updating appropriately (including setting `keep`, for mutations).

I'm happy to have a go at implementing this, but some input would be very helpful first.

jeromekelleher:
That's a good idea @petrelharp - maybe the two key functions should get the site_t and mutation_t objects as parameters, which they are responsible for updating. The ``mutation_model_t`` could have a blkalloc that's for doing the various mallocs needed for the Site and Mutation information.

petrelharp:
Ok - I've done a refactor here, so that now the signatures are
```
int mutation_model_choose_root_state(mutation_model_t *self,
        gsl_rng *rng, tsk_site_t *site);
int mutation_model_transition(mutation_model_t *model, gsl_rng *rng,
        const char *parent_allele, tsk_size_t parent_allele_length,
        tsk_mutation_t *mutation);
```
Two things:

1.  `copy_string` doesn't work because I don't understand `blkalloc`: replacing the `blkalloc` with a `malloc` fixes it; something to do with the block size? Help?

2. Previously, we had mutgen's `mutation_t` mostly like `tsk_mutation_t` but with some extra stuff (eg `new` and `keep`). But then we had to pass in a `mutation_t` to `mutation_model_transition`, and so `mutation_t` had to be public, which seemed confusing. So, I changed `mutation_t` to carry a pointer to a `tsk_mutation_t`, which is the thing passed in to `mutation_model_transition`. This means that `mutation_model_transition` can't set the `keep` parameter directly, but this seems OK.

jeromekelleher:
It's not immediately clear to me what the problem with the block_allocator is @petrelharp, but I'm having a hard time getting things straight in my head.

I think it would be a lot simpler if we just forget about the existence of tsk_site_t and tsk_mutation_t entirely. I see no problem in defining ``msp_site_t`` and ``msp_mutation_t`` in ``msprime.h``. Sure, they're quite like the tsk_ counterparts, but so what? We're not going to be changing ``tsk_site_t``, e.g., ever if we can help it so I don't think it's a maintenance burden having two similar types defined across the different projects. 

I can go ahead and make these changes and (hopefully) simplify the memory management if you like?

petrelharp:
Ok! That'd be great. Go for it.

petrelharp:
Well:mMy concern was that the extra stuff in `mutation_t` doesn't feel like we want to commit it to the public API, as opposed to `tsk_mutation_t`.

jeromekelleher:
> Well: my concern was that the extra stuff in mutation_t doesn't feel like we want to commit it to the public API, as opposed to tsk_mutation_t.

The msprime C API isn't public anyway, we can do whatever we want.

jeromekelleher:
I've tidied things up @petrelharp and made it work again. The memory management is much simpler now, and I think the algorithm flow is more straightforward. It has turned out to be quite time consuming, though.

However, I'll have to rewrite the Python C API side of things completely to make the model polymorphism work, and unfortunately I don't have time for this. I'm not quite sure how this will work, so it'd be better to leave it with me. I'll try and get some time to implement the SLiM model in C and update here when I get a chance.

petrelharp:
Ah, ok, that's simpler, and more memory efficient! Nice.

How about I take a first pass at the SLiM model? It's pretty clear in my head what it should do, and if I malloc instead of blkalloc'ing, then I'm sure I can make it work.

jeromekelleher:
That'd be really helpful, thanks. Good test coverage on the C code would be super helpful too. I can wrangle the top-to-bottom structure then so we also work in Python.

petrelharp:
Ok - here's a go at the SLiM mutator. Note that I've had to also pass in the parental metadata, since we need to append to that.

FYI, the model here is that each new mutation gets a unique ID (assigned sequentially starting from a number that's passed in), which is appended (as bytes) to the previous allele to make the derived state; and, a corresponding metadata entry (which just describes the mutation type, and so will always be the same) is appended to the metadata.  There's a pretty good C test that's checking that's exactly what's happening.

lib/mutgen.c
```diff
@@ -292,6 +294,101 @@ mutation_matrix_free(mutation_model_t *self)
...
 
+/***********************
+ * SLiM mutation model */
+
+/* from slim_globals.h */
+typedef int64_t slim_mutationid_t;		// identifiers for mutations, which require 64 bits since there can be so many
+
+/* Copy of MutationMetadataRec in slim_sim.h: */
+typedef struct __attribute__((__packed__)) {
+    int32_t mutation_type_id_; // 4 bytes (int32_t): the id of the mutation type the mutation belongs to
```
jeromekelleher, line 305: We can get rid of the trailing underscores here I'd say, might as well use the local coding conventions. A bit more details on the provenance would be help here I think - what date/what git hash were these structs copied at?

jeromekelleher:
Looks good @petrelharp. We'll need to change to useing a blkalloc though, as it's currently leaking memory.

petrelharp:
> Looks good @petrelharp. We'll need to change to useing a blkalloc though, as it's currently leaking memory.

I did?

lib/mutgen.c
petrelharp: I'll put in more details to the reference. But, I guess I vote to keep them exactly matching, as a reminder that we're copying this from elsewhere (so it's not to be changed).

jeromekelleher: OK, that sounds good.

lib/mutgen.c
```diff
@@ -187,93 +184,348 @@ mutation_model_check_validity(mutation_model_t *self)
...
+        int64_t next_mutation_id, size_t block_size)
+{
+    int ret = 0;
+    slim_mutator_t *params = &self->params.slim_mutator;
+    slim_mutation_metadata_t *slim_md = NULL;
+
+    memset(self, 0, sizeof(mutation_model_t));
+    
+    /* In practice this is the minimum we can support */
+    block_size = GSL_MAX(block_size, 128);
```
jeromekelleher, line 463: Where does the 128 come from here? That seem very small to me. Unless you've got a good reason, I would say the best behaviour here is to have a default of say 4K, which we use if the input block size is 0. We don't ever set these in practise, so I'm not sure there's much point in having the parameter.

lib/mutgen.c
```diff
@@ -187,93 +184,348 @@ mutation_model_check_validity(mutation_model_t *self)
...
+    ret = tsk_blkalloc_init(&params->allocator, block_size);
+    if (ret != 0) {
+        ret = msp_set_tsk_error(ret);
+        goto out;
+    }
+
+    slim_md = malloc(sizeof(*slim_md));
```
jeromekelleher, line 470: I'm not sure there's much point in storing this in the struct - since it's a fixed size, it would be simpler/more obvious what's going on if we alloced it statically in the function that's using it.

jeromekelleher:
I'm not sure why this isn't getting full CI checks  - maybe worth a rebase here @petrelharp?

lib/mutgen.c
petrelharp: er, wherever I copied the blkalloc from? sorry, I forgot to track down what this means (and what it should be)

petrelharp:
Ok, I've removed the slim_metadata property and put it in the transition method; and also rebased, doing a bit of squashing as well.  I'm still getting seemingly unrelated errors from the local tests, strangely.

lib/mutgen.c
jeromekelleher: Just put in a fixed size of 4K here. There's no point in having this knob to tune, it only makes a difference for really big stuff.

petrelharp: I've set it to 8192, following what's in `mutgen_alloc` and `mutgen_init_allocator`.

petrelharp:
The windows build was failing because Windows doesn't do packed; there are some other possible fixes [here](https://stackoverflow.com/questions/1537964/visual-c-equivalent-of-gccs-attribute-packed), but I've just implemented a lower-tech solution.

petrelharp:
I'm going to leave this alone now - want to have another look, @jeromekelleher? (and, thanks!!)

petrelharp:
Screenshot from 2020-06-08 12-15-18
<relaxed sigh>

tests/test_mutations.py
```diff
@@ -1087,6 +1087,60 @@ def test_arbitrary_model(self):
...
+
+    def validate_slim_mutations(self, ts, increasing=True):
+        # slim alleles should be unique comma-separated lists of integers
+        # and ancestral states the empty string
+        # if `increasing` then each one should have one more integer than the one before
+        alleles = {}
+        for s in ts.sites():
+            self.assertEqual(s.ancestral_state, "")
+            for m in s.mutations:
+                a = tuple(m.derived_state.split(","))
```
jeromekelleher, line 1103: This is different to the C generator now though, isn't it? There we're using little-endian 8 byte values and here we're using comma delimited text. Which one does SLiM do?

jeromekelleher:
Looks good to me @petrelharp. The low-tech solution to packing bytes looks just right. I am confused about the exact format here, as we seem to be doing different things in the C and Python versions. But, I guess the next step is to setup the linkage between C and Python. Or, we can tidy this much up and merge first, I don't mind.

tests/test_mutations.py
petrelharp: Whoops, yes. SLiM does the [concatenated bytes thing](https://github.com/MesserLab/SLiM/blob/1c85d00a0417416431ca69f2cb68a1bcc134ec3d/core/slim_sim.cpp#L5137), not text. I was waiting to compare these until we've hooked the slim generator up to python.

petrelharp:
> I guess the next step is to setup the linkage between C and Python. 

It'd be nice to do this before merging so we can properly test it (including I'll check if SLiM can read the mutations we've produced). Before you said you wanted to set up the linkage; is that still true?

jeromekelleher:
> It'd be nice to do this before merging so we can properly test it (including I'll check if SLiM can read the mutations we've produced). Before you said you wanted to set up the linkage; is that still true?

I can do this, unless you're super keen. Will I jump in now?

jeromekelleher:
How about I do some coding here if you do some reviewing on #1025 @petrelharp?

jeromekelleher:
I've added the basic plumbing here @petrelharp, should all basically work now. Good bit of work left in tidying everything up and making tests, but you should be able to test it against SLiM now anyway.

_msprimemodule.c
```diff
@@ -4682,13 +4838,23 @@ PyInit__msprime(void)
...
+    /* SlimMutationModel type */
+    SlimMutationModelType.tp_new = PyType_GenericNew;
+    if (PyType_Ready(&SlimMutationModelType) < 0) {
         return NULL;
     }
-    Py_INCREF(&MutationModelType);
-    PyModule_AddObject(module, "MutationModel", (PyObject *) &MutationModelType);
+    Py_INCREF(&SlimMutationModelType);
+    PyModule_AddObject(module, "SlimMutationModel",
+            (PyObject *) &SlimMutationModelType);
```
petrelharp, line 4857: whoo, yep, I'm glad you did this bit!

lib/mutgen.c
```diff
@@ -491,31 +490,35 @@ mutation_matrix_model_alloc(mutation_model_t *self, size_t num_alleles, char **a
 }
 
 int MSP_WARN_UNUSED
-slim_mutator_alloc(mutation_model_t *self, int32_t mutation_type_id,
+slim_mutation_model_alloc(mutation_model_t *self, int32_t mutation_type_id,
```
petrelharp, line 493: good resolution to the inconsistent naming problem!

petrelharp:
Gah! I've just been around in about five circles until I rediscovered/remembered that (a) SLiM keeps track of derived states as bytes in its *internal tables*, but since these have to be ascii for tskit, converts them to comma-separated strings on writing them out.

jeromekelleher:
> Gah! I've just been around in about five circles until I rediscovered/remembered that (a) SLiM keeps track of derived states as bytes in its internal tables, but since these have to be ascii for tskit, converts them to comma-separated strings on writing them out.

Ah, well that's a relief in a way. I was getting worried that we'd have to do some hacks to support arbitrary bytes in the ancestral/derived states.

petrelharp:
> Ah, well that's a relief in a way. I was getting worried that we'd have to do some hacks to support arbitrary bytes in the ancestral/derived states.

Our former selves decided not to do this, I think.

jeromekelleher:
i've updated this @petrelharp, and it should be nearly ready now. I'll add an InfiniteSitesModel too though, before merging.

jeromekelleher:
This is ready for merging I think @petrelharp - can you have a quick look over please? There's still some loose ends re the API, but these can be tidied up with #1070.

_msprimemodule.c
```diff
@@ -2112,18 +2314,41 @@ MutationGenerator_init(MutationGenerator *self, PyObject *args, PyObject *kwds)
...
+        infinite_alleles_model = (InfiniteAllelesMutationModel *) py_model;
+        if (InfiniteAllelesMutationModel_check_state(
+                    infinite_alleles_model) != 0) {
+            goto out;
+        }
+        model = infinite_alleles_model->mutation_model;
+    } else {
+        PyErr_SetString(PyExc_TypeError,
+            "model must be an instance of MatrixMutationModel "
+            "or SlimMutationModel");
```
petrelharp, line 2339: or InfiniteAllelesMutationModel

lib/mutgen.c
```diff
@@ -187,93 +184,494 @@ mutation_model_check_validity(mutation_model_t *self)
...
+    slim_mutator_t *params = &self->params.slim_mutator;
+    char *buff = NULL;
+    int len;
+    /* The maximum number of digits for a signed 64 bit integer (including
+     * the leading "-") */
+    const size_t max_digits = 20;
+    /* We allow for a potential comma to separate previous elements in the
+     * list. We allow for a possible comma to separate the previous element
+     * in the list, as well as a NULL byte added by snprintf. We don't bother
+     * trying to alloc the exact number of bytes needed. */
```
petrelharp, line 399: duplicated sentence in the comment. (also, good call here)

lib/mutgen.c
```diff
@@ -187,93 +184,494 @@ mutation_model_check_validity(mutation_model_t *self)
...
+infinite_alleles_print_state(mutation_model_t *self, FILE *out)
+{
+    infinite_alleles_t params = self->params.infinite_alleles;
+    fprintf(out, "infinite alleles:\n");
+    fprintf(out, "\tstart_allele:%" PRIu64 "\n", params.start_allele);
+}
+
+/* The maximum number of digits for an unsigned 64 bit integer is 20
+ * and one byte for the NULL terminator. */
+#define MAX_UINT_BUFF_SIZE 21
```
petrelharp, line 464: we could use this above (`max_digits` in slim_transition_allele) as well

lib/mutgen.c
```diff
@@ -187,93 +184,494 @@ mutation_model_check_validity(mutation_model_t *self)
...
+
+    buff = tsk_blkalloc_get(&params->allocator, len);
+    if (buff == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    memcpy(buff, tmp_buff, len);
+    params->next_allele++;
+    *dest = buff;
+    *dest_length = (tsk_size_t) len;
```
petrelharp, line 495: `len` was already cast to `tsk_size_t` on assignment - but, why not just use `len` instead of `err` throughout, as in `slim_transition_allele`?

lib/mutgen.c
```diff
@@ -483,81 +880,104 @@ mutgen_add_site(mutgen_t *self, double position, bool new, const char *ancestral
...
     return ret;
 }
 
-static int MSP_WARN_UNUSED
-mutgen_add_mutation(mutgen_t *self, site_t *site, tsk_id_t id, node_id_t node,
-    double time, bool new, const char *derived_state, tsk_size_t derived_state_length,
-    const char *metadata, tsk_size_t metadata_length)
+static int
+mutgen_add_mutation(
+    mutgen_t *self, site_t *site, node_id_t node, double time, mutation_t **new_mutation)
```
petrelharp, line 928: good call

petrelharp:
Looks great! Awesome!  A few minor suggestions, but it looks good!  I should check that things can be loaded correctly into SLiM - will report back.

lib/mutgen.c
jeromekelleher: Good spot on the cast. We're passing ``len`` directly to blk_alloc_get as well as assigning it, so we'd have to cast it twice if we used the return value from sprintf. I thought this was the simplest and clearest. I've changed ``err`` to ``num_digits`` now though, as err was a bit misleading.

lib/mutgen.c
jeromekelleher: I thought about this, but it's a bit awkward as we'd have to have another implementation of make_allele as we're using uint64 in one and int64 in the other. The reason for having ``uint64`` here, btw, is so that we have well defined overflow semantics. Unsigned ints are guaranteed to wrap around to zero, whereas signed ints are undefined. Having overflow seemed simpler than having to test for it, but we could totally do it same way as the slim model too.

Anyway, it didn't seem worth complicating the code more, as the slim code is pretty water tight and reasonably simple now. The extra few bytes spent per allele is trivial.

jeromekelleher:
Great! I'm going to merge this now @petrelharp - we can follow up with updates if we need them, but let's get this big diff merged.

PR #1014, opened 2020-05-11
Initial infrastructure for mutation model verification

jeromekelleher:
A reasonable approach to verifying mutation models would seem to be to compare with the output of [Seq-Gen](https://snoweye.github.io/phyclust/document/Seq-Gen.v.1.3.2/Seq-Gen.Manual.html) which is the standard software for phylogenetic models. This adds Seq-Gen to the local list of programs that we use for statistical comparisons and some basic plumbing for setting up verification tests.

We probably won't get this fully done here as I don't have time to work on this any more, and I'm hoping that Gertjan can help fill this out. So, this PR should be seen as an initial framework rather than something finished.

Does anyone disagree with the basic approach of comparing with Seq-Gen?


petrelharp:
Sounds good. I see that seq-gen doesn't do microsats, so if we implement a microsat model we may want to pull in something different for that.

Note that `tests/test_mutgen.py` already does some statistical testing, and could easily do more, e.g. checking for the right stationary distribution; these ought to be just as good, really, as comparing to another implementation.

jeromekelleher:
> Note that tests/test_mutgen.py already does some statistical testing, and could easily do more, e.g. checking for the right stationary distribution; these ought to be just as good, really, as comparing to another implementation.

A bit of redundancy is no harm here I think - always good to be able to point to another program that's widely used as a believable verification.

jeromekelleher:
OK, just chatted with @GertjanBisschop and we're going to merge this much to give him a base to work on. I'll follow up with some more issues to track what needs doing from here.

PR #1015, opened 2020-05-11
Get the species_tree example working

hyanwong:
Needed to remove a reference to a non-existent variable, and remove/explain the samples param


jeromekelleher:
Thanks for the fix!

Issue #985, opened 2020-05-03
Demographic events not interacting with multiple mergers


JereKoskela:
This code implements a demographic event in the Beta coalescent, and behaves strangely:
```
import msprime

n = 158
a = 1.001
gen_time = 4
N0 = 1e5
N1 = 1e4
t1 = 150 / gen_time

ts = msprime.simulate(length=2e7,
    recombination_rate=1e-8,
    mutation_rate=1e-7,
    Ne=N0,
    model=msprime.BetaCoalescent(alpha=a),
    population_configurations = [
        msprime.PopulationConfiguration(
            sample_size=n,
            initial_size=N0
        )
    ],
    demographic_events = [
        msprime.PopulationParametersChange(
            time = t1,
            initial_size = N1
        )
    ],
    num_replicates=1,
    random_seed=1)
```
Varying the value of `N0` between `N0 = 1e5`, `N0 = 1e4`, and `N0 = 1e3` results in respective runtimes of around 1 second, 4 seconds, and more than several minutes. Smaller effective population sizes should result in faster mergers, so the changes are in the wrong direction. What's more, then Beta coalescent timescale is proportional to `Ne^(alpha - 1)`, so these changes in `Ne` should hardly change anything because `(1e5)^0.001` and `(1e3)^0.001` are essentially identical.

JereKoskela:
Come to think of it, we should probably turn off variable population size for multiple merger coalescents if that's easy to do, at least for the time being, because the interaction of variable population size and multiple mergers is not necessarily simple.

For example, a Beta-Xi-coalescent with parameter `1 < alpha < 2`, per-generation growth rate `g > 0` and time scaling `c` will experience population growth with rate `c g (alpha - 1)` on the coalescent timescale, not `c g` like a Wright-Fisher model would. The derivation of the rate is laid out in section 3 here: https://www.sciencedirect.com/science/article/abs/pii/S0025556418303523.

Migration rates between populations behave like they do in the Wright-Fisher case, even if migration rates change, so multiple populations are ok, but instantaneous population size changes will also interact with the `alpha` parameter.

I'm not immediately sure whether the Dirac-Xi-coalescent is similarly awkward.

jeromekelleher:
[...]

Issue #766, opened 2019-05-31
Clear up confusion about Ne


jeromekelleher:
An early version of #754 had some documentation trying to clear up what Ne means for the various models, and how it interacts with the individual population sizes. This [thread](https://github.com/tskit-dev/msprime/pull/754#discussion_r287438943) shows a basic disagreement.

molpopgen:
I was just wrestling with this bit of the docs for that thing I'm writing. Happy to help rework the docs--ping me if you want input.

jeromekelleher:
That would be very helpful, thanks @molpopgen.

jeromekelleher:
Pushing this back to the next milestone, as we want to get a bugfix release out now ASAP.

Issue #557, opened 2018-07-10
Add ploidy argument to simulate and allocate individuals


jeromekelleher:
It's becoming increasingly clear that the current approach of sort-of being diploid isn't great. Perhaps we should add a ``ploidy`` argument to ``simulate`` so that we can try to fix this. Here's one possible way forward.

- Add an argument ``ploidy`` which defaults to None. If it is None, have the same behaviour as now: time is scaled as 4Ne and no individuals are allocated. 
- If ``ploidy`` is not None, scale time appropriately and allocate individuals sequentially among the samples. The total number of samples must be divisible by ``ploidy``, or an error is raised.

Hopefully this will keep backwards compatibility so that all existing code will still work as expected, while also providing a less confusing path forward. It would be nice to set ``ploidy`` to 2 by default, but I don't see how we can deal with allocating individuals in the case where the number of samples is odd. 

Any thoughts @petrelharp, @molpopgen? 

See also #548, #529

molpopgen:
For odd-numbered samples:

The final nodes are random w.r.to genealogical relationship.  In ms/msprime, one make 'diploids' simply by pairing up adjacent rows in the output matrix.  Thus, for odd-numbered sample sizes and even ploidy, you could simulate one extra lineage that is not output?  However, this scheme means that the error condition you outline above is undesirable.

It may be useful to avoid setting the ploidy to two by default.  As-is, it is straightforward for a user to simulate haploids, etc., with the current machinery by simply scaling their input parameters appropriately.

petrelharp:
Partly, I think this is an unavoidably confusing issue. We can't completely remove the burden of thinking through things carefully from the user.  

A `ploidy` argument does seem like a good way to ask msprime to give you diploids, etcetera. But, with the `ploidy` argument, it would be natural if everything referring to samples refered to polyploid samples (thus avoiding the odd-numbers problem).  This would be a major change, which we probably want to avoid.

I don't understand what you mean by "time scaling" - which time scaling? Do you mean that `ploidy` is a factor that the provided Ne is multiplied by?

jeromekelleher:
> A ploidy argument does seem like a good way to ask msprime to give you diploids, etcetera. But, with the ploidy argument, it would be natural if everything referring to samples refered to polyploid samples (thus avoiding the odd-numbers problem). This would be a major change, which we probably want to avoid.

I agree it would have been better for samples to be polyploid, but it's too late to change that now I think.

[...]

PR #1028, opened 2020-05-13
Nuke model.reference_size from orbit

jeromekelleher:
This started as an attempt to clear up the confusion about what Ne/model.reference_size actually means (see #766). However, while doing this I realised that it doesn't actually do anything at all for most models, if we set the population size appropriately. (The models where it does do something  is possibly a reflection of another bug, #985)

I'll dig in more and report back here on this thread. I'm not sure whether this is an insidious bug or an opportunity to greatly simplify this aspect of msprime.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,13 @@
 - Add replicate_index to simulate, allowing output of a single tree sequence
   from a set of replicates. (:user:`benjeffery` :pr:`914`).
 
+**Breaking changes**:
+
+- The class form for specifying models (e.g., ``msprime.StandardCoalescent()``)
+  no longer take a ``reference_size`` argument.
+  (TODO add paper trail for this)
+
+
 ********************
 [0.7.4] - 2019-12-05
 ********************
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -2862,8 +2862,7 @@ out:
 }
 
 static int
-Simulator_parse_sweep_genic_selection_model(Simulator *self, PyObject *py_model,
-        double reference_size)
+Simulator_parse_sweep_genic_selection_model(Simulator *self, PyObject *py_model)
 {
     int ret = -1;
     int err;
@@ -2900,7 +2899,7 @@ Simulator_parse_sweep_genic_selection_model(Simulator *self, PyObject *py_model,
     }
     dt = PyFloat_AsDouble(value);
 
-    err = msp_set_simulation_model_sweep_genic_selection(self->sim, reference_size,
+    err = msp_set_simulation_model_sweep_genic_selection(self->sim,
             position, start_frequency, end_frequency, alpha, dt);
     if (err != 0) {
         handle_input_error("sweep genic selection", err);
@@ -2928,7 +2927,7 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
     PyObject *value;
     int is_hudson, is_dtwf, is_smc, is_smc_prime, is_dirac, is_beta, is_sweep_genic_selection;
     int is_wf_ped;
-    double reference_size, psi, c, alpha, truncation_point;
+    double psi, c, alpha, truncation_point;
 
     if (Simulator_check_sim(self) != 0) {
         goto out;
@@ -2966,15 +2965,6 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
         goto out;
     }
 
-    value = get_dict_number(py_model, "reference_size");
-    if (value == NULL) {
-        goto out;
-    }
-    reference_size = PyFloat_AsDouble(value);
-    if (reference_size <= 0) {
-        PyErr_SetString(PyExc_ValueError, "population size must be >= 0");
-        goto out;
-    }
     py_name = get_dict_value(py_model, "name");
     if (py_name == NULL) {
         goto out;
@@ -2988,7 +2978,7 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
         goto out;
     }
     if (is_hudson) {
-        err = msp_set_simulation_model_hudson(self->sim, reference_size);
+        err = msp_set_simulation_model_hudson(self->sim);
     }
 
     is_dtwf = PyObject_RichCompareBool(py_name, dtwf_s, Py_EQ);
@@ -2996,14 +2986,14 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
         goto out;
     }
     if (is_dtwf) {
-        err = msp_set_simulation_model_dtwf(self->sim, reference_size);
+        err = msp_set_simulation_model_dtwf(self->sim);
     }
     is_wf_ped = PyObject_RichCompareBool(py_name, wf_ped_s, Py_EQ);
     if (is_wf_ped == -1) {
         goto out;
     }
     if (is_wf_ped) {
-        err = msp_set_simulation_model_wf_ped(self->sim, reference_size);
+        err = msp_set_simulation_model_wf_ped(self->sim);
     }
 
     is_smc = PyObject_RichCompareBool(py_name, smc_s, Py_EQ);
@@ -3011,7 +3001,7 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
         goto out;
     }
     if (is_smc) {
-        err = msp_set_simulation_model_smc(self->sim, reference_size);
+        err = msp_set_simulation_model_smc(self->sim);
     }
 
     is_smc_prime = PyObject_RichCompareBool(py_name, smc_prime_s, Py_EQ);
@@ -3019,7 +3009,7 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
         goto out;
     }
     if (is_smc_prime) {
-        err = msp_set_simulation_model_smc_prime(self->sim, reference_size);
+        err = msp_set_simulation_model_smc_prime(self->sim);
     }
 
     is_dirac = PyObject_RichCompareBool(py_name, dirac_s, Py_EQ);
@@ -3045,7 +3035,7 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
             PyErr_SetString(PyExc_ValueError, "c >= 0");
             goto out;
         }
-        err = msp_set_simulation_model_dirac(self->sim, reference_size, psi, c);
+        err = msp_set_simulation_model_dirac(self->sim, psi, c);
     }
 
     is_beta = PyObject_RichCompareBool(py_name, beta_s, Py_EQ);
@@ -3064,7 +3054,7 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
         }
         truncation_point = PyFloat_AsDouble(value);
         /* TODO range checking on alpha and truncation_point */
-        err = msp_set_simulation_model_beta(self->sim, reference_size,
+        err = msp_set_simulation_model_beta(self->sim,
                 alpha, truncation_point);
     }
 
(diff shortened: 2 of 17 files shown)


petrelharp:
This is definately confusing.

jeromekelleher:
Here's how it works.

- Internally, population sizes are stored as *relative* to the model->reference_size, e.g. [here](https://github.com/tskit-dev/msprime/blob/44a1abb74777a8e60f9912835c9301f64a7d5a64/lib/msprime.c#L319)

- We then use this relative size to compute coalescence rates [here](https://github.com/tskit-dev/msprime/blob/44a1abb74777a8e60f9912835c9301f64a7d5a64/lib/msprime.c#L3501)

- When we create a [new node](https://github.com/tskit-dev/msprime/blob/44a1abb74777a8e60f9912835c9301f64a7d5a64/lib/msprime.c#L1243), we rescale time from model time back to generations using [the model](https://github.com/tskit-dev/msprime/blob/44a1abb74777a8e60f9912835c9301f64a7d5a64/lib/msprime.c#L5612)

So, if there's no recombination or any other processes happening, it'll all cancel out. However, if we do have recombination things are different, because the **rates** are all scaled by the model, which has nothing to do with the sizes of the individual populations.

So, it's confusing, but expected.

jeromekelleher:
Rightyho, well this is a bit depressing. It turns out that ``SimulationModel.reference_size`` does **nothing at all** for any of the simple coalescent models. I've just put in a quick hack to make sure that it's always set to 1 for the low-level code, and sure enough (basically) all the tests pass and stats checks seem to go OK too.

So, pros:
- Hooray! We can ditch the **unbelievably annoying and confusing** concept of "model time" entirely. It all just works out, if you set the population sizes.
- We're coming up to 1.0, so now's the time to make changes to simplify things and clarify the semantics.

Cons:
- It's going to be a complicated cleanup getting rid of model time, time/rate rescaling etc. 

@JereKoskela, what does this mean for the multiple merger models? It's not so clear cut here I think that the scaling to generations just happens automatically? Any thoughts much appreciated...

JereKoskela:
> @JereKoskela, what does this mean for the multiple merger models? It's not so clear cut here I think that the scaling to generations just happens automatically? Any thoughts much appreciated...

I'm going to need a little more detail before I can start answering that question, I think. If I call
```
ts = msprime.simulate(sample_size=2, Ne=N, random_seed=1)
ts.first().tmrca(0,1)
```
with `N = 1` and `N = 10`, I get two different TMRCAs that differ by a factor of ten, as I should. So it can't be that the reference size does literally nothing. Is the point that the following simulations are equivalent when the per-generation recombination probability is `r` and there are two populations with sizes `N_A` and `N_B`?

1. A simulation with recombination rate `r`, merger rates `1 / N_A` and `1 / N_B`, and no rescaling of time.
2. A simulation with recombination rate `N r`, merger rates `1 / (N_A / N)`, and `1 / (N_B / N)` and a rescaling of time by `N`.

If so, then similar things will be true for multiple merger models, albeit with slightly more complicated formulas as is already evident from the existing conversion functions between times and rates in generations and "coalescent time".

jeromekelleher:
Sorry @JereKoskela, I didn't give much detail here! The issue is that if we run this simulation, the value of Ne is irrelevant:
```
ts =  msprime.simulate(
                Ne=10**6,
                population_configurations=[
                    msprime.PopulationConfiguration(sample_size=10, initial_size=100)
                ],
                random_seed=123,
            )
```
This is because the actual time scaling is done by the ``initial_size`` argument for the population. All we do with the model is divide the rates by 10^6 on the way in, and then multiply the times by 10^6 on the way out.

In the example that you give, because there are no population configurations specified, by default we allocate 1 in which the ``initial_size=Ne``; that's why we get the right scaling of times. All the generations-to-model and model-to-generations transformations just cancel out (and always have done).

JereKoskela:
Thanks @jeromekelleher. This is undoubtedly what is behind #985 in some form. While the relationships between parameters and simulation rates is more complicated, the multiple merger coalescents are also fully specified by all the individual population sizes and per-generation probabilities, growth rates, etc. There is no need for an exogenous reference population size.

jeromekelleher:
OK, thanks @JereKoskela. I'll try pulling out ``reference_size`` and seeing how much I break...

jeromekelleher:
OK, it turns out that whole model time concept was bullshit, and I have confused myself and a bunch of other people by ending up having some *very* convoluted logic. The good news is that I'm sure that we can sort this out for the Hudson model and (with a bit of help) I think we can make it actually make sense. It was all correct before, so there's no actual bugs involved here.

For Hudson, we can get rid of all model time scaling/unscaling if we get an extra factor of 4 into the population size/coalescent waiting time calculation. I think what had happened before is that all of the ``reference_size`` values ended up cancelling each other out, but there was an initial factor of 4 which got applied during the initial scaling, but didn't get unscaled (somehow).

There's two issues that need to be resolved before I can push through the final changes, and document what Ne actually means:

1. What should we do with the Dirac and Beta models? My hope is that we can just get rid of the time rescaling stuff entirely. @JereKoskela, what do you think? The effect of Ne (in the single population case) is to set the Population->initial_size value, which should automatically lead to time being scaled in the right way. Is this what will happen in Dirac and Beta?
2. We need to account for the extra factor of 4 in the coalescent time calculation. I think this is probably due to two different factors of 2. Maybe one is missing [here](https://github.com/tskit-dev/msprime/blob/cf5258239f7b4642653f6db0e00d8349a5722226/lib/msprime.c#L5643) I think (the rate should be n choose 2, right?), and the other 2 is the extra diploid factor. Do you agree with this @JereKoskela?

So, perhaps we could use this as an opportunity to add the "ploidy" argument (#557)?

JereKoskela:
> 1. What should we do with the Dirac and Beta models? My hope is that we can just get rid of the time rescaling stuff entirely. @JereKoskela, what do you think? The effect of Ne (in the single population case) is to set the Population->initial_size value, which should automatically lead to time being scaled in the right way. Is this what will happen in Dirac and Beta?

In the standard coalescent, you can either simulate mergers at 1 per pair and rescale time by `4N`, or you can simulate mergers at rate `1 / (4N)` per pair and rescale time by `4N`. In the same way, a multiple merger coalescent can either undergo potential merger events at rate 1 per pair (with some extra computation to decide when they involve multiple lineages) when time is rescaled by the model-specific rescaling `f(N)`, or undergo potential merger events at rate `1 / f(N)` without time rescaling. Either way, we can't get rid of a model-specific rescaling function, though it does not have to apply to time directly.

When there are multiple populations, in the standard coalescent it makes sense to take `N = total initial population size`, whereupon it happens that the right merger rates are just reciprocals of relative population sizes. Taking `N = total initial population size` also works for multiple merger coalescents (so we don't need an exogenous `reference_size` variable), though the per-population merger rates are more complicated functions of both `N` and the local population size.

I suggest you finish up the Hudson model, and then I can take a look and replicate the same change to multiple mergers. A separate, free `model->reference_size` parameter is definitely not needed provided all population sizes are specified.

> 2. We need to account for the extra factor of 4 in the coalescent time calculation. I think this is probably due to two different factors of 2. Maybe one is missing [here](https://github.com/tskit-dev/msprime/blob/cf5258239f7b4642653f6db0e00d8349a5722226/lib/msprime.c#L5643) I think (the rate should be n choose 2, right?), and the other 2 is the extra diploid factor. Do you agree with this @JereKoskela?

That missing factor of 2 always confused me. I initially assumed that the idea was to simulate mergers at rate `n(n - 1)/2` and recombination at rate `n(4Nr)/2` by simulating mergers at rate `n(n - 1)`, recombination at rate `n(4Nr)`, and multiplying the waiting time by 2, and rescaling time by `4N` at the end. Everything else is as expected, but that multiplication of waiting times by 2 never happens.

If we want to simulate a population with `2N` diploids, and minimise rescaling and transforming of parameters, then I think the simplest thing would be to set the merger rate at `(n choose 2) / (4 N)` and leave the recombination rate at `nr`. Simulating `2N` haploids would change those to `(n choose 2) / (2 N)` and `nr`.

> So, perhaps we could use this as an opportunity to add the "ploidy" argument (#557)?

This will also be slightly more involved in multiple merger coalescents, but is perfectly doable, and in fact @eldonb and I had already talked about trying to add it there.

jeromekelleher:
Thanks @JereKoskela, this is very helpful. What's not clear to me is where the time rescaling will need to happen for the multiple merger coalescents, and I don't want to yank out all the existing infrastructure (which is unnecessary for Hudson and DTWF) until we figure this out.

Would it be sufficient to rescale the current simulation time based on the current state (individual population sizes, etc) when we store time values in the node table [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L1243)? 

But then we have the confusion about the units for input recombination/gc/migration etc rates.

JereKoskela:
> Would it be sufficient to rescale the current simulation time based on the current state (individual population sizes, etc) when we store time values in the node table [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L1243)?
>
> But then we have the confusion about the units for input recombination/gc/migration etc rates.

In the simplest case we can just rescale the common ancestor event rate right at the start. Then we could keep node times and other rates in units of generations throughout, and require minimal infrastructure. A population size change would result in a change of merger rate through a model-specific function of the current and new population sizes.

I'm not quite sure whether that's essentially what you have in mind for the Hudson model @jeromekelleher. But in either case, there should be little no no extra infrastructure needed for multiple mergers, beyond the sorts of things we have already implemented in the BetaCoalescent and DiracCoalescent model classes.

jeromekelleher:
Great, thanks @JereKoskela. I'll have a go at setting this up and ping you.

jeromekelleher:
This is ready for review now I think. I looks like I grotesquely over-complicated things and all the time-changing infrastructure was unnecessary. Sorry about that everyone.

I've pulled all this mess out, and it looks like things are working as they were. I haven't gone through the statistical tests with a fine-tooth comb yet, but they look OK to me.

@JereKoskela, @TPPSellinger, the only substantive change to the tests that I made is in ``run_xi_dirac_recombinations``, where the mutation rate was divided by Ne (see the very bottom of the diff). I removed this division, and it all seems to work now. All the complicated time changing stuff I made you do was unnecessary :facepalm: 

There'll be a few cleanup issues to follow up on (like, let's make the Hudson case make sense), but I suggest we merge this and create issues to track those.

petrelharp:
> This is ready for review now I think. It looks like I grotesquely over-complicated things and all the time-changing infrastructure was unnecessary. Sorry about that everyone.

A case of something that simplifies writing the math papers but doesn't make applying them easier!

JereKoskela:
> @JereKoskela, @TPPSellinger, the only substantive change to the tests that I made is in `run_xi_dirac_recombinations`, where the mutation rate was divided by Ne (see the very bottom of the diff). I removed this division, and it all seems to work now. All the complicated time changing stuff I made you do was unnecessary 🤦

I'll have to look over this in more detail another day in the not-too-distant future, but at first glance, I don't think the complicated stuff will have been unnecessary in the end. Some of it will have to go back, albeit in a slightly different form. It will be an easy fix though.

There is still a model-specific scale going on implicitly in line 3502 of msprime.c, which says that waiting times for common ancestor events are proportional to population sizes. That's not true for multiple merger coalescents, so instead, it will have to be something like
```
ret = 4 * dirac_model_time_to_generations(pop->initial_size) * u;
```
and analogously for the Beta coalescent.

That will also be the reason why the `run_xi_dirac_recombinations` test now look like it works without dividing by `Ne`, even though a similar tweak will probably have to go back in the end.

I think it would be fine to get this merged, have the multiple merger models be flagged as broken again, and fixed in a separate issue. I don't expect it will be hard to do because it's just reimplementing what we already had in a slightly different way. We will probably fix #985 as a by-product while we're at it.

jeromekelleher:
OK, great, thanks @JereKoskela. I'll merge this so and open some issues to track the remaining problems.

PR #1032, opened 2020-05-16
Compare msprime against seq-gen #1018

### LINKED ISSUE #1018: Complete Seq-Gen comparison and validate JC69 model
#1014 added some initial infrastructure for comparing with SeqGen. Finish this infrastructure work so that we can compare the output of msprime and SeqGen, and validate JC69 model.

GertjanBisschop:
closes #1018 . Completes initial code for comparing with Seq-Gen for JC69. Will be extended to accommodate for the mutation models that will be implemented later on.

### DIFF
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -29,6 +29,8 @@ import _msprime
 from . import core
 from . import provenance
 
+_ACGT_ALLELES = [b"A", b"C", b"G", b"T"]
+
 
 class MutationModel(_msprime.MutationModel):
     """
@@ -83,11 +85,105 @@ class JukesCantor(MutationModel):
     """
 
     def __init__(self):
-        alleles = [b"A", b"C", b"T", b"G"]
+        alleles = _ACGT_ALLELES
         root_distribution = [0.25, 0.25, 0.25, 0.25]
         transition_matrix = np.zeros((4, 4))
         transition_matrix[:] = 1 / 3
         np.fill_diagonal(transition_matrix, 0)
+        super().__init__(alleles, root_distribution, transition_matrix)
+
+
+class HKY(MutationModel):
+    """
+    The Hasegawa, Kishino and Yano mutation model (Hasegawa et al. 1985).
+
+    .. todo: documentation
+    """
+
+    def __init__(self, kappa=1.0, equilibrium_frequencies=None, root_distribution=None):
+        alleles = _ACGT_ALLELES
+        if equilibrium_frequencies is None:
+            equilibrium_frequencies = np.array(
+                [0.25, 0.25, 0.25, 0.25], dtype="float64"
+            )
+        if root_distribution is None:
+            root_distribution = equilibrium_frequencies.copy()
+
+        transition_matrix = np.full((4, 4), 1.0)
+        np.fill_diagonal(transition_matrix, 0.0)
+        # positions in transition matrix for transversions
+        transition_pos = ((0, 1, 2, 3), (2, 3, 0, 1))
+        transition_matrix[transition_pos] = kappa
+        transition_matrix *= equilibrium_frequencies
+        row_sums = transition_matrix.sum(axis=1, dtype="float64")
+        transition_matrix /= max(row_sums)
+        np.fill_diagonal(transition_matrix, 1.0 - row_sums / max(row_sums))
+
+        super().__init__(alleles, root_distribution, transition_matrix)
+
+
+class F84(MutationModel):
+    """
+    The F84 mutation model (Felsenstein and Churchill, 1996).
+    2 types of events
+    I: no change/transition
+    II: no change/transition/transversion
+    .. todo: documentation
+    """
+
+    def __init__(self, kappa=1.0, equilibrium_frequencies=None, root_distribution=None):
+        alleles = _ACGT_ALLELES
+        if equilibrium_frequencies is None:
+            equilibrium_frequencies = [0.25, 0.25, 0.25, 0.25]
+        if root_distribution is None:
+            root_distribution = equilibrium_frequencies.copy()
+
+        transition_matrix = np.full((4, 4), 1.0, dtype="float64")
+        np.fill_diagonal(transition_matrix, 0.0)
+        # positions in transition matrix for transversions
+        transition_pos_AG = ((0, 2), (2, 0))
+        transition_pos_CT = ((1, 3), (3, 1))
+        p_AG = equilibrium_frequencies[0] + equilibrium_frequencies[2]
+        p_CT = equilibrium_frequencies[1] + equilibrium_frequencies[3]
+        gamma = kappa - 1
+        transition_matrix[transition_pos_AG] = np.float64(1 + gamma / p_AG)
+        transition_matrix[transition_pos_CT] = np.float64(1 + gamma / p_CT)
+        transition_matrix *= equilibrium_frequencies
+        row_sums = transition_matrix.sum(axis=1)
+        transition_matrix = transition_matrix / max(row_sums)
+        row_sums = transition_matrix.sum(axis=1, dtype="float64")
+        np.fill_diagonal(transition_matrix, 1.0 - row_sums)
+
+        super().__init__(alleles, root_distribution, transition_matrix)
+
+
+class GTR(MutationModel):
+    """
+    The  Generalised time-reversible mutation model (Tavaré et al. 1986).
+    .. todo: documentation
+    """
+
+    def __init__(
+        self, relative_rates, equilibrium_frequencies=None, root_distribution=None
+    ):
+        alleles = _ACGT_ALLELES
+        assert len(relative_rates) == 6
+        if equilibrium_frequencies is None:
+            equilibrium_frequencies = [0.25, 0.25, 0.25, 0.25]
+        if root_distribution is None:
+            root_distribution = equilibrium_frequencies
+
+        transition_matrix = np.zeros((4, 4))
+        # relative_rates: [A->C, A->G,A->T,C->G,C->T,G->T]
+        tri_upper = np.triu_indices_from(transition_matrix, k=1)
+        transition_matrix[tri_upper] = relative_rates
+        transition_matrix += transition_matrix.T
+        transition_matrix *= equilibrium_frequencies
+        row_sums = transition_matrix.sum(axis=1)
+        transition_matrix = transition_matrix / max(row_sums)
+        row_sums = transition_matrix.sum(axis=1, dtype="float64")
+        np.fill_diagonal(transition_matrix, 1.0 - row_sums)
+
         super().__init__(alleles, root_distribution, transition_matrix)
 
 
--- a/tests/test_mutations.py
+++ b/tests/test_mutations.py
@@ -93,7 +93,7 @@ class TestMutationModel(unittest.TestCase):
         self.assertEqual((num_alleles, num_alleles), model.transition_matrix.shape)
         self.assertEqual(sum(model.root_distribution), 1.0)
         for j in range(num_alleles):
-            self.assertEqual(sum(model.transition_matrix[j]), 1.0)
+            self.assertTrue(np.allclose(sum(model.transition_matrix[j]), 1.0))
         s = model.__str__()
         self.assertIsInstance(s, str)
         self.assertTrue(s.startswith("Mutation model with alleles"))
@@ -879,7 +879,7 @@ class StatisticalTestMixin:
 
     p_threshold = 0.01
 
-    def chisquare(self, observed, expected):
+    def chisquare(self, observed, expected, p_th=p_threshold):
         obs = observed.reshape((np.product(observed.shape)),)
         exp = expected.reshape((np.product(observed.shape)),)
         not_zeros = exp > 0
@@ -889,7 +889,7 @@ class StatisticalTestMixin:
         )
         if sum(not_zeros) > 1:
             chisq = stats.chisquare(obs[not_zeros], exp[not_zeros])
-            self.assertGreater(chisq.pvalue, self.p_threshold)
+            self.assertGreater(chisq.pvalue, p_th)
 
     def bonferroni(self, pvals):
         self.assertGreater(min(pvals), self.p_threshold / len(pvals))
@@ -910,7 +910,7 @@ class TestMutationStatistics(unittest.TestCase, StatisticalTestMixin):
                 ts = msprime.mutate(
                     ots, random_seed=6, rate=rate, model=model, discrete=discrete
                 )
-                self.verify_model_ts(ts, model, discrete, verify_roots)
+                self.verify_model_ts_general(ts, model, discrete, verify_roots, rate)
 
     def verify_model_ts(self, ts, model, discrete, verify_roots):
         alleles = [a.decode() for a in model.alleles]
@@ -935,6 +935,61 @@ class TestMutationStatistics(unittest.TestCase, StatisticalTestMixin):
             p[j] = 0
             p /= sum(p)
             self.chisquare(row, sum(row) * p)
+
+    def verify_model_ts_general(self, ts, model, discrete, verify_roots, mutation_rate):
+        alleles = [a.decode() for a in model.alleles]
+        num_alleles = len(alleles)
+        roots = np.zeros((num_alleles,))
+
+        observed_transitions = np.zeros((num_alleles, num_alleles))
+        for mut in ts.mutations():
+            if mut.parent == tskit.NULL:
+                pa = ts.site(mut.site).ancestral_state
+            else:
+                pa = ts.mutation(mut.parent).derived_state
+            da = mut.derived_state
+            observed_transitions[alleles.index(pa), alleles.index(da)] += 1
+        pth = 0.05 / num_alleles
+        for j, (row, p) in enumerate(
+            zip(observed_transitions, model.transition_matrix)
+        ):
+            p[j] = 0
+            p /= sum(p)
+            self.chisquare(row, sum(row) * p, pth)
+
+        if verify_roots:
+            for site in ts.sites():
+                aa = site.ancestral_state
+                roots[alleles.index(aa)] += 1
+
+            expected_ancestral_states = np.zeros(num_alleles)
+            change_probs = model.transition_matrix.sum(axis=1) - np.diag(
+                model.transition_matrix
+            )
+            for t in ts.trees():
+                if discrete:
+                    t_span = np.ceil(t.interval[1] - np.ceil(t.interval[0]))
+                    expected_ancestral_states += (
+                        model.root_distribution
+                        * t_span
+                        * (
+                            1
+                            - np.exp(
+                                -mutation_rate * t.total_branch_length * change_probs
+                            )
+                        )
+                    )
+                else:
+                    t_span = t.span
+                    expected_ancestral_states += (
+                        model.root_distribution
+                        * mutation_rate
+                        * t.total_branch_length
+                        * t_span
+                        * change_probs
+                    )
+
+            self.chisquare(roots, expected_ancestral_states)
 
     def verify_mutation_rates(self, model):
         # this test only works if the probability of dropping a mutation
@@ -981,6 +1036,31 @@ class TestMutationStatistics(unittest.TestCase, StatisticalTestMixin):
         model = msprime.JukesCantor()
         self.verify_model(model)
         self.verify_mutation_rates(model)
+
+    def test_HKY(self):
+        equilibrium_frequencies = [0.3, 0.2, 0.3, 0.2]
+        kappa = 0.75
+        model = msprime.HKY(
+            kappa=kappa, equilibrium_frequencies=equilibrium_frequencies
+        )
+        self.verify_model(model)
+
+    def test_F84(self):
+        equilibrium_frequencies = [0.4, 0.1, 0.1, 0.4]
+        kappa = 0.75
+        model = msprime.F84(
+            kappa=kappa, equilibrium_frequencies=equilibrium_frequencies
+        )
+        self.verify_model(model)
+
+    def test_GTR(self):
+        relative_rates = [0.2, 0.1, 0.7, 0.5, 0.3, 0.4]
+        equilibrium_frequencies = [0.3, 0.4, 0.2, 0.1]
+        model = msprime.GTR(
+            relative_rates=relative_rates,
+            equilibrium_frequencies=equilibrium_frequencies,
+        )
+        self.verify_model(model)
 
     def test_arbitrary_model(self):
         model = msprime.MutationModel(
--- a/tests/test_mutations.py
+++ b/tests/test_mutations.py
@@ -172,6 +172,21 @@ class TestMutationModel(unittest.TestCase):
         self.validate_model(model)
         self.validate_stationary(model)
 
+    def test_HKY_default(self):
+        model = msprime.HKY()
+        self.validate_model(model)
+        self.validate_stationary(model)
+
+    def test_F84_default(self):
+        model = msprime.F84()
+        self.validate_model(model)
+        self.validate_stationary(model)
+
+    def test_GTR_default(self):
+        model = msprime.GTR([1 / 6] * 6)
+        self.validate_model(model)
+        self.validate_stationary(model)
+
     def test_new_model(self):
         alleles = [b"Alligator", b"Camel", b"Goat"]
         root_distribution = [0.5, 0.25, 0.25]
--- a/verification.py
+++ b/verification.py
@@ -5,6 +5,7 @@ known statistical results and benchmark programs such as Hudson's ms.
 import argparse
 import ast
 import collections
+import itertools
 import math
 import os
 import random
@@ -13,20 +14,24 @@ import sys
 import tempfile
 import time
 
+import allel
 import dendropy
 import matplotlib
 import numpy as np
 import pandas as pd
+import pyvolve
 import scipy.special
 import scipy.stats
 import seaborn as sns
 import tqdm
+import tskit
 from matplotlib import lines as mlines
 from matplotlib import pyplot
 
 import _msprime
 import msprime
 import msprime.cli as cli
+from msprime.demography import _matrix_exponential
 
 # Force matplotlib to not use any Xwindows backend.
 # Note this must be done before importing statsmodels.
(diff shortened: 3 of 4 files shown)


jeromekelleher:
This is great, thanks @GertjanBisschop!

The main thing you need to do to keep the CI services happy is to let Black format your code. The [tskit docs](https://tskit.readthedocs.io/en/latest/development.html#code-style) explain the setup and how to use pre-commit to set things up for you.

Can you paste in an example of the plots here?

How sensitive do you think these statistics are to picking up problems? Say if we change the mutation rate a small bit, or use a slightly different mutation model, how well would they pick these things up? Are there other sequence comparison stats we could include?

GertjanBisschop:
These are qq-plots for sequence diversity and watterson's theta respectively for a sample of 10, Ne=1e4, mutation rate=1e7, sequence length=1000bp, num_replicates=1000.
seq_gen_comparison_JC69_pi
seq_gen_comparison_JC69_watt

These statistics are quite sensitive to changes in mutation rate. (see images underneath, Seg-Gen specified with mutation rate of 1.25e-7 rather than 1e-7). I will check the sensitivity to the specified model once I have implemented another model. It seems to me I will only be able to pick up differences as soon as the chance of hitting the same site twice becomes high enough. Otherwise I might have to resort to using different statistics outside of scikit-allel (e.g. probability of seeing a particular nucleotide among the SNPs). 

seq_gen_comparison_JC69_pi
seq_gen_comparison_JC69_watt

jeromekelleher:
Excellent! Wrt to the sensitivity, an easy check to try now would be to specify the wrong model for Seq-Gen and see if we do pick that up.

That's a good point about hitting the same base - maybe we should do this for three different sequence lengths (say, 10, 100, 1000) so we should get absurdly many mutations per site on one extreme, and often just one mutation per site at another.

petrelharp:
Gee, this is very nice. What's your plan for counting mutations by state? Ideally, we'd compare things by ancestral/derived state, eg theta for A to T mutations, etc?

GertjanBisschop:
@petrelharp Yes! I've been thinking about that today. As well as getting out the actual sample sequences I can also retrieve the ancestral sequence with Seq-Gen. We can even get the sequences for all internal nodes. The same is true for variants simulated with msprime. I guess there are two possibilities here then. Either we use all information (= the sequences at the internal nodes) and determine the rate of transitions, explicitely taking into account the possibility of multiple hits on the same sites. Or we only compare ancestral sequences to the sequences at the tips. If the mutation rate is not particularly high then this won't be all that different from the first possibility. And this involves less parsing and bookkeeping I guess. 
I would love to here your thoughts on this.

Secondly, in [mutations.py](msprime/mutations.py), alleles are listed as alleles = ["A", "C", "T", "G"] but there is also a remark stating '#ACGT order'. Does the transition matrix assume a particular order (being ACGT ) or does the order depend on the order in the alleles list?

jeromekelleher:
> Secondly, in mutations.py, alleles are listed as alleles = ["A", "C", "T", "G"] but there is also a remark stating '#ACGT order'. Does the transition matrix assume a particular order (being ACGT ) or does the order depend on the order in the alleles list?

I'm guessing this is a typo, and we should have the alleles in ACGT order. The ordering matters all right, i.e., transition_matrix[0, 1] is the rate of mutation from ``alleles[0]`` to ``alleles[1]``. The underlying code doesn't know anything about the actual alleles, it just works with indexes.

petrelharp:
Well, maybe the easiest way to compare is with a chi-squared test for the table of (ancestral, derived) states across all the branches. This'll make it easy to compare to the theoretical expectation as well, computed from the matrix exponential of the mutation rate matrix. I thought about this sort of thing in these [statistical tests](https://github.com/tskit-dev/msprime/blob/cf5258239f7b4642653f6db0e00d8349a5722226/tests/test_mutations.py#L905).  How's that sound?

GertjanBisschop:
Thanks, I didn't realize this was implemented. Wonderful. So we can compare against theoretical expectations. As well as implement a similar comparison to Seq-Gen. (I think we are (almost) on the same page now).

petrelharp:
> Thanks, I didn't realize this was implemented. 

Sorry, I thought you knew about that.

jeromekelleher:
My bad @GertjanBisschop, @petrelharp, I should have pointed out the existing stats tests to Gertjan (although, independence of these tests and the statistical checks is nice). I'd rather keep the verification tests to visual plots rather than test/fail status too.

BTW, if you need a matrix exponential function @GertjanBisschop, we have one in the demography module.

GertjanBisschop:
I have been testing both the Jukes-Cantor as well as the HKY model. So we're making progress. I have been using some of @petrelharp 's code to parse trough the mutations of of each simulated tree. The transition matrix built in this way is more accurate then the one I can make from the seq-gen generated sequences as I can't telll whether more than one mutation has occurred on a particular site along a branch connecting two nodes.

This is a qqplot for the sum of chisquare statistics for each matrix row, for the HKY model for a transition to transversion ratio of 1. Note the outlier, to get rid of that I will reconstruct the sequences at the interal nodes such that everything is perfectly comparable with the seq-gen implementation. This means 'collapsing' mutations on top of each other, so ignoring all mutations happening between the state of the parent node and the derived state of the next node. I don't think any code exists to do that, right?

seq_gen_comparison_HKY_transition_matrix

Also, thank god for well-document software. I spent quite some time trying to figure out in what order Seq-Gen returns sequences for internal nodes.

petrelharp:
Gee, this would be easier if we had a method to get genotypes at internal nodes.

jeromekelleher:
> Gee, this would be easier if we had a method to get genotypes at internal nodes, https://GitHub.com/tskit-dev/tskit/issues/605

We can do this now:
```python
ts = msprime.simulate(5, mutation_rate=0.5, random_seed=2)

for var in ts.variants(
        samples=np.arange(ts.num_nodes).astype(np.int32),
        impute_missing_data=True):
    print(var.genotypes)
```
Here ``var.genotypes[u]`` is the state for node ID u.

GertjanBisschop:
I have run quite a few checks now, most come up as expected. However when testing `msprime.HKY(ts_tv_ratio=0.75, root_distribution=[0.1, 0.4, 0.15, 0.35])` the qq-plot for the sum of chisquare statistics (one sum per matrix) is off (first plot). I could not quite figure out what was wrong, so I also benchmarked all tests against pyvolve. Although quite a bit slower, here I do get the expected result. So, @petrelharp would you mind taking a look at my implementation of the mutation models? I want to make sure I have done this properly before I continue testing. I have also run tests using your test_mutations.py script, and here I get an error for one of the F84 tests I implemented. So there is probably still something wrong with that one.

Secondly, the tests that are in place check whether all row sums of the transition matrix equal one. However, do to floating point precision this is sometimes not the case. What can I do to overcome this (or is it just a case of adapting the test to `abs(1-transition_matrix.sum(axis=1))<0.00001`?).

@jeromekelleher I have now done the pre-commit checks, all passed but I still see three tests have failed. I will check how I can fix this.
seq_gen_comparison_HKY_transition_matrix
pyvolve_comparison_HKY_transition_matrix

petrelharp:
> What can I do to overcome this (or is it just a case of adapting the test to abs(1-transition_matrix.sum(axis=1))<0.00001?).

I think using [`np.allclose()`](https://numpy.org/doc/1.18/reference/generated/numpy.allclose.html) might be the way to go.

msprime/mutations.py
```diff
@@ -83,14 +83,95 @@ class JukesCantor(MutationModel):
     """
 
     def __init__(self):
-        alleles = [b"A", b"C", b"T", b"G"]
+        alleles = [b"A", b"C", b"G", b"T"]
```
petrelharp, line 86: Gee, we should have these defined in just one place, to avoid errors like this?  At least, just in this file as `_ACGT_ALLELES`?

msprime/mutations.py
```diff
@@ -83,14 +83,95 @@ class JukesCantor(MutationModel):
...
+
+        transition_matrix = np.full((4, 4), 1.0)
+        np.fill_diagonal(transition_matrix, 0.0)
+        # positions in transition matrix for transversions
+        transition_pos = ((0, 1, 2, 3), (2, 3, 0, 1))
+        kappa = 2 * ts_tv_ratio
+        transition_matrix[transition_pos] = kappa
+        transition_matrix *= root_distribution
+        row_sums = transition_matrix.sum(axis=1, dtype="float64")
+        transition_matrix = transition_matrix / row_sums[:, np.newaxis]
```
petrelharp, line 115: Ah-ha. This is an infinitesimal rate matrix, not a discrete transition matrix, so normalizing the rows separately changes the model. We are parameterizing things so that the the overall rate matrix, denoted `Q` [here](https://en.wikipedia.org/wiki/Models_of_DNA_evolution#HKY85_model_(Hasegawa,_Kishino_and_Yano_1985)) is defined so that
```
Q[i,j] = mutation_rate * transition_matrix[i,j]
```
for `i != j`. Under the hood, we lay down  mutations at rate `mutation_rate` without looking at the state of each site, then use the corresponding row of the transition_matrix to pick the next state. This means that the `Q[i,i]` entries will correspond to "mutations" that we don't actually include since they don't change the state; so our implementation will be most efficient if these are as small as possible - meaning the smallest one should be zero.  So, let's see, I'd do it like this:
```
         transition_matrix /= max(row_sums)
         row_sums = transition_matrix.sum(axis=1, dtype="float64")
         np.fill_diagonal(transition_matrix, 1.0 - row_sums)
```
... and then we can say in the documentation that the mutation model will be be the HKY rates such that (a) the maximum mutation rate is `mutation_rate` and the stationary distribution is `root_distribution`. (oh wait, I have a thought about the parameterization though; I"ll put it in the discussion).

petrelharp:
I see that you're setting the root distribution and the stationary distribution equal. I agree this is usually what people should do, but they can be different. I'd propose that the HKY model has five parameters that specify the transition matrix: kappa, pi_A, pi_C, pi_G, and pi_T, and then also a root distribution. The root distribituion should default to pi, and the documentation should say something strong about how the user should not specify the root distribution unless they really know what they're doing.

verification.py
```diff
@@ -445,14 +682,122 @@ def _run_seq_gen_msprime_comparison(self, model, length=100):
...
+            transitions_ts = np.zeros((num_alleles, num_alleles))
+            for site in sequences[0]:
+                roots_d_ts[alleles.index(site)] += 1
+            transitions_ts = self._transitions(sequences, triplets, alleles)
+            transition_matrix = model_dict[model]["model_id"].transition_matrix
+            tm_chisq_ts = self._transition_matrix_chi_sq(
+                transitions_ts, transition_matrix
+            )
+            exp_roots_d_ts = (
+                np.sum(roots_d_ts) * model_dict[model]["model_id"].root_distribution
```
petrelharp, line 773: This is only the expectation if the chance of not mutating (the diagonal) is the same for all alleles, since sites with no mutations are not added.

petrelharp: The actual probability that a partticular location has ancestral allele k is `mutation_rate * root_distribution[k] * (1 - exp(-transition_matrix[k,k] * total_branch_length))`. We could get the expected number of ancestral alleles by computing this iterating over the trees, like (untested):
```
expected_ancestral_states = np.zeros(num_alleles)
for t in ts.trees():
  expected_ancestral_states += mutation_rate * root_distribution * (1 - np.exp(-np.diag(transition_matrix)*t.total_branch_length))
```

petrelharp:
I haven't figured out yet why you're using [triplets](https://github.com/tskit-dev/msprime/pull/1032/files#diff-99dcd80775a2e22814e7cb431ecfc7daR635) to count transitions? Why is this?

petrelharp:
For computing the expected number of transitions I was imagining something like this:
```
Q = transition_matrix.copy()
np.fill_diagonal(Q, 0.0)
row_sums = Q.sum(axis=1)
np.fill_diagonal(Q, -1 * row_sums)
observed = np.zeros((num_alleles, num_alleles))
expected = np.zeros((num_alleles, num_alleles))
for v in ts.variants(samples=range(ts.num_nodes):
  t = ts.at(v.position)
  for n in t.nodes():
    pn = t.parent(n)
    if pn != tskit.NULL:
      pa = v.genotypes[pn]
      da = v.genotypes[n]
      observed[pa, da] += 1
      expected += expm(t.branch_length(n) * mutation_rate * Q)
```
where `expm()` is the matrix exponential.

msprime/mutations.py
GertjanBisschop: OK, thanks for taking the time to lay this out so clearly. 
Now this part of the [verify_model_ts](https://github.com/tskit-dev/msprime/blob/aa2eca95bfff8a19c41bb0db5e1f077503a158de/tests/test_mutations.py#L915)
`for j, (row, p) in enumerate(zip(transitions, model.transition_matrix)):
            p[j] = 0
            p /= sum(p)
            self.chisquare(row, sum(row) * p)`
makes total sense.

GertjanBisschop:
> I haven't figured out yet why you're using [triplets](https://github.com/tskit-dev/msprime/pull/1032/files#diff-99dcd80775a2e22814e7cb431ecfc7daR635) to count transitions? Why is this?

I am using this as a way to link parent-nodes to children-nodes (triplet = parent: child1 and child 2). But just iterating over edges without building the triplets dict would do the same. So overkill.

verification.py
GertjanBisschop: @petrelharp , a few more questions. I now realize I went head in first without thinking enough about whether I understood the concepts of the root distribution and the transition matrix enough (in terms of how they are being used by the machinery). Sorry about that. I was assuming the root distribution is the frequency distribution from which ancestral state is drawn as soon as a site gets hit by the first mutation. And that from then on, everything is determined by the transition matrix. 

So your code `exp_roots = num_muts * model.root_distribution` in the test_mutations.py modulee made total sense to me. This is what we expect to see in terms of ancestral states for the root of the tree. 

Now in the code you are proposing here, you calculate the probability of no mutation happening on a certain site for each of the samples, right? And I assume that `transition_matrix[k,k]` means `-row_sum(transition matrix)`, so the total probability of leaving a certain state? (Otherwise this is 0 for JukesCantor)

If root distribution is indeed what I think it is, doesn't it suffice to calculate the expectation as you were doing before in your test code?

GertjanBisschop:
Then to get all other definitions clear. pi_A, pi_C, pi_G, pi_T in the HKY-model then represent the stationary distribution or the equilibrium frequencies as referred to in Seq-Gen right? Can we give it a name (as for the root distribution) so we can specify it as a list?

verification.py
petrelharp: Whoops - I made a mistake above. Instead of the diagonal of the transition matrix I should have put the sum of the offdiagonal terms. Also, I had the mutation rate in the wrong place. I think it should be this:
```
expected_ancestral_states = np.zeros(num_alleles)
change_probs = transition_matrix.sum(axis=1) - np.diag(transition_matrix)
for t in ts.trees():
  expected_ancestral_states += root_distribution * (1 - np.exp(-mutation_rate * change_probs*t.total_branch_length))
```
... but, I see that I should draft up some documentation on how the mutation matrix models work. I'll do that now, which will hopefully answer the other questions.

petrelharp:
> Then to get all other definitions clear. pi_A, pi_C, pi_G, pi_T in the HKY-model then represent the stationary distribution or the equilibrium frequencies as referred to in Seq-Gen right? Can we give it a name (as for the root distribution) so we can specify it as a list?

Yes - either `stationary_distribution` or `equilibrium_frequencies` sound good!

petrelharp:
See #1045 - and please feel free to improve what I"ve started there.

GertjanBisschop:
I have implemented all comments and necessary changes. This is now the HKY mutation model code:
```
class HKY(MutationModel):
    """
    The Hasegawa, Kishino and Yano mutation model (Hasegawa et al. 1985).

    .. todo: documentation
    """

    def __init__(self, kappa=1.0, equilibrium_frequencies=None, root_distribution=None):
        alleles = _ACGT_ALLELES
        if equilibrium_frequencies is None:
            equilibrium_frequencies = [0.25, 0.25, 0.25, 0.25]
        if root_distribution is None:
            root_distribution = equilibrium_frequencies

        transition_matrix = np.full((4, 4), 1.0)
        np.fill_diagonal(transition_matrix, 0.0)
        # positions in transition matrix for transversions
        transition_pos = ((0, 1, 2, 3), (2, 3, 0, 1))
        transition_matrix[transition_pos] = kappa
        transition_matrix *= equilibrium_frequencies
        row_sums = transition_matrix.sum(axis=1, dtype="float64")
        transition_matrix /= max(row_sums)
        np.fill_diagonal(transition_matrix, 1.0 - row_sums/max(row_sums))

```
However, when running a first test, the diversity compared to Seq-Gen is off. This is not the case for JukesCantor. Still something wrong?

petrelharp:
> However, when running a first test, the diversity compared to Seq-Gen is off. This is not the case for JukesCantor. Still something wrong?

Are you sure they are parameterized in the same way? Like, what does Seq-Gen's "mutation rate" mean?

GertjanBisschop:
Seq-Gen scales all branch lengths in order to make them equal the expected number of substitutions per site for each branch. So for any branch of length t, the expected number of mutations per site is `t*mutation_rate`. The scaling factor should be the same mutation rate as specified for `msprime.mutate()`

petrelharp:
Ok, but what is the "expected number of substitutions per site", since this depends on the nucleotide?

petrelharp:
I see that the documentation does not explain. Maybe it's best to implement the comparisons to theory first?

GertjanBisschop:
yeah I guess so, documentation is quite sparse. Reading the original paper now.

petrelharp:
Speaking of the theoretical checks, would you like me to put those into `tests/test_mutations.py`, @GertjanBisschop? Or, do you want to do it? Above we've described the theoretical expectations for the number (not just proportion) of segregating sites per ancestral allele and the number of transitions along each branch; this is more general than the checks I've got there now, which only work if the probability of mutation doesn't depend on the nucleotide.

GertjanBisschop:
Yes, I can do that. I've been thinking of asking you that. Do you agree it would be best, for example for the number of transitions along each branch to do a bunch of replicates along the same tree and compare those to the expectation for that particular tree?

petrelharp:
This is probably what you're doing, but I'd iterate over the trees and add up expected numbers for each.

GertjanBisschop:
The test code is in place. There is one issue. The test for the root distribution (number of segregating sites per ancestral allel) errors if discrete=False (so infinite sites). I am not seeing why this expectation should be somewhat different than for the finite sites (the correction we discussed earlier).

I've been looking into the Seq-Gen mutation rate issue. The mutation rate we have to pass is the average mutation rate (across the rates for each nucleotide). Then diversity checks out. My intuition tells me we therefore can't simply compare the observed transition counts from Seq-Gen to the expectated transition probabilities as calculated in test_mutations.py. This would mean that we have to calculate expected transition probabilities for Seq-Gen as well... It this the best way to compare seq-gen to msprime? I think I get Jerome's point about the independence of statistical test and the seq-gen comparison. But maybe I am missing something crucial. 

I have attached a pdf with qqplots (chisquare) of comparing msprime against the theoretical expectations (transitions along branches and distribution of ancestral alleles). 500 runs each.

[test_mutations.pdf](https://github.com/tskit-dev/msprime/files/4704561/test_mutations.pdf)

petrelharp:
>  The test for the root distribution (number of segregating sites per ancestral allel) errors if discrete=False (so infinite sites). I am not seeing why this expectation should be somewhat different than for the finite sites (the correction we discussed earlier).

Ah-ha. The calculation we did for finite sites was "there will be Poisson(rate * total length) mutations at this site; what's the chance none of them mutate out of the ancestral state", and then summed across the span of the tree. But for infinite sites, the calculation we want to do is "there will be *one* mutation at this site; what's the chance it doesn't mutate out of the ancestral state", but then a Poisson(rate * total length *span) number of these across the tree.   So, the correct comparison for infinite sites should be to
```
 expected_ancestral_states += root_distribution * rate * t.total_branch_length * (1 - change_probs) * (t.span[1] - t.span[0])
```
(gee, I see I missed the `span` contribution above for the finite sites bit - did you get it in?).

petrelharp:
> The mutation rate we have to pass is the average mutation rate (across the rates for each nucleotide). Then diversity checks out. My intuition tells me we therefore can't simply compare the observed transition counts from Seq-Gen to the expectated transition probabilities as calculated in test_mutations.py.

Gee, I think if we pass mutation rates correctly, then they should be simulating the same model, and so the transition counts should agree? Oh, but: I think the goal here is to compare observed counts in msprime to observed counts in Seq-Gen; without bringing in the theoretical expectations: the point here is not to check Seq-Gen but to check if msprime and Seq-Gen agree.  (also, simpler that way, no?)  Was this what you meant?

petrelharp:
The comparison pdf looks good?

verification.py
petrelharp: ... and fixing this up to account for span:
```
expected_ancestral_states = np.zeros(num_alleles)
change_probs = transition_matrix.sum(axis=1) - np.diag(transition_matrix)

for t in ts.trees():
  if discrete:
    t_span = np.ceil(t.span[1]) - np.ceil(t.span[0])
  else:
    t_span = t.span[1] - t.span[0]
  expected_ancestral_states += root_distribution * t_span * (1 - np.exp(-mutation_rate * change_probs*t.total_branch_length))
```

petrelharp:
> The test code is in place.

I can't find it - have you pushed the changes?

GertjanBisschop:
Sorry didn't push the changes. Should have mentioned that.

GertjanBisschop:
The code for comparison against seq-gen and pyvolve is now in place. For all tests I have done so far, there is a consistent pattern: using the mean mutation rate across all nucleotides, qqplots for pi, total number of muations and the distribution of ancestral allele frequencies line up nicely. 

The transition probabilities along the branches are evaluated by storing the chi-square statistic of evaluating the observed number of transitions from allele i to any other allele against the expectations. The qqplots for these (with i either A, C, G or T) are plotted below both for the comparison against seq-gen and pyvolve. 

Comparison for HKY model:
seq_gen_comparison_tm_HKY

pyvolve_comparison_tm_HKY

I've pushed the code, but I am basically using the same code for pyvolve and seq-gen. I don't know yet how to explain the consistently bigger deviation for seq-gen.

tests/test_mutations.py
```diff
@@ -936,6 +937,91 @@ def verify_model_ts(self, ts, model, discrete, verify_roots):
...
+                t = ts.at(v.position)
+                for n in t.nodes():
+                    pn = t.parent(n)
+                    if pn != tskit.NULL:
+                        pa = v.alleles[v.genotypes[pn]]
+                        da = v.alleles[v.genotypes[n]]
+                        observed_transitions[alleles.index(pa), alleles.index(da)] += 1
+                        expected += _matrix_exponential(
+                            t.branch_length(n) * mutation_rate * Q
+                        )
```
petrelharp, line 963: whoops - this is the expected number *conditioned* on the parent of the branch, so I think it should be
```
j = alleles.index(pa)
expected[, :] += _matrix_exponential(
    t.branch_length(n) * mutation_rate * Q)[j, :]
```

tests/test_mutations.py
```diff
@@ -936,6 +937,91 @@ def verify_model_ts(self, ts, model, discrete, verify_roots):
...
+
+            np.fill_diagonal(observed_transitions, 0.0)
+            np.fill_diagonal(expected, 0.0)
```
petrelharp, line 966: let's not zero these out - if we've got it right the diagonals should match also

tests/test_mutations.py
```diff
@@ -936,6 +937,91 @@ def verify_model_ts(self, ts, model, discrete, verify_roots):
...
+
+            for j, (row, p) in enumerate(zip(observed_transitions, expected)):
+                p[j] = 0
+                p /= sum(p)
```
petrelharp, line 970: also, we should not normalize by the sum - the raw numbers should match

tests/test_mutations.py
```diff
@@ -936,6 +937,91 @@ def verify_model_ts(self, ts, model, discrete, verify_roots):
...
+                self.chisquare(row, sum(row) * p)
+
+        else:
+            observed_transitions = np.zeros((num_alleles, num_alleles))
+            for mut in ts.mutations():
+                if mut.parent == tskit.NULL:
```
petrelharp, line 976: if we're not doing discrete mutations, the mutation parents should always be NULL?

tests/test_mutations.py
petrelharp: Oh, wait - this is a bit trickier, because the observed numbers are conditioned on having at least one mutation on the tree. Hm.  I don't think what you had is right, because since *all* rows of exp(t mu Q) are included, that's essentially assuming the uniform distribution over alleles at the head of every branch.  But, maybe I'm missing something? I'm having a hard time seeing what the right way to compute this is.

petrelharp:
I haven't figured out what's going on with seq-gen, but I did realize that I had something wrong in the expected numbers of transitions. Sorry about that! The basic problem was that observed counts only include transitions at sites with a mutation, so the corresponding expected numbers should be conditioned on a mutation-that-changes-the-allele happening somewhere on that tree. I haven't yet figured out the right way to do this. How's your Markov chain theory, @GertjanBisschop? I can figure out a good comparison (maybe it isn't the per-branch transition counts), I'm sure, or leave it to you if you like?

jeromekelleher:
This is looking great @GertjanBisschop! One quick point though: isn't there a systematic difference in the QQplots above for C and T in SeqGen?

GertjanBisschop:
> One quick point though: isn't there a systematic difference in the QQplots above for C and T in SeqGen?

Yes there is. The equilibrium frequencies for both tests were `[0.2, 0.3, 0.1, 0.4]`, so both C and T had higher equilibrium frequencies.

jeromekelleher:
> Yes there is. The equilibrium frequencies for both tests were [0.2, 0.3, 0.1, 0.4], so both C and T had higher equilibrium frequencies.

Is this expected, or something we need to explain? Or, is that what you meant here:

> I don't know yet how to explain the consistently bigger deviation for seq-gen.

GertjanBisschop:
> > Yes there is. The equilibrium frequencies for both tests were [0.2, 0.3, 0.1, 0.4], so both C and T had higher equilibrium frequencies.
> 
> Is this expected, or something we need to explain? Or, is that what you meant here:
> 
> > I don't know yet how to explain the consistently bigger deviation for seq-gen.

I definitely did not expect this. And yes, that was part of what I meant with my comment. I guess we'll first have to come up with a bullet-proof comparison.

GertjanBisschop:
@petrelharp, I think you are right about the conditioning. I had implemented it exactly like you did here, at some point but second-guessed myself. I will think about it this afternoon, but it would be reassuring knowing that this is also on the back of your mind. I do not consider my Markov chain theory to be up there ...

petrelharp:
Right now the code is counting transitions across branches for the discrete case, and transitions from parent to child mutations for the non-discrete case (even though in that case there aren't parent mutations). I think that the easiest thing to do is just to count parent-to-child mutations in both cases? In which case - I think - we can just do the same check as what is implemented for both? I would need to think carefully.

petrelharp:
My next step will probably be to do some coding. Let me know when I should take a look at it (so we're not doing this at the same time).

Re: seq-gen - it may be useful to write to Rambaut to check if we're interpreting parameters right?

GertjanBisschop:
> My next step will probably be to do some coding. Let me know when I should take a look at it (so we're not doing this at the same time).
> 
Good point!
> Re: seq-gen - it may be useful to write to Rambaut to check if we're interpreting parameters right?
I've contacted Rambaut, but he is to busy with his covid-19 work to answer any emails.

jeromekelleher:
> I've contacted Rambaut, but he is to busy with his covid-19 work to answer any emails.

There must be other people out there besides Andrew who know about Seq-Gen, right? It's been around for donkey's years.

jeromekelleher:
FYI @GertjanBisschop, we've just merge a big refactor of the verification.py code in #1035. This is going to cause horrendous conflicts with your code. The best thing to do is not worry about it for now, and *don't* try to rebase. Let me know when you think you're ready to merge and I'll make the changes needed for the new code layout.

Sorry for the awkwardness here.

GertjanBisschop:
I've pushed the new verification code and have included a pdf with the results comparing msprime against seq-gen. You will see on the graphs for the transition probabilities along branches that the fit is not always perfect. To get an idea of the variance along the perfect 45 degree line I have compared the outcome of msprime.mutate against itself by sprinkling mutations twice along the same tree (with a different seed). This shows that our results are good. (And model misspecification looks completely different.) Do tell me what you guys think! How do I 'package' the results for the comparison? Or is the code that generates similar figures ok?

[seq_gen_comparison.pdf](https://github.com/tskit-dev/msprime/files/4734991/seq_gen_comparison.pdf)

jeromekelleher:
This is *excellent* @GertjanBisschop, the results look really strong to me! With QQplots for this sort of thing a bit of wavering either side of the 45 degree line is to be expected. It's consistent, regular deviations that you look for, and there's none here.

In the interest of merging this ASAP, would you mind taking out the Amino acid models for now? We can add these in a separate follow-up PR. Once you've take these out, I'll update the PR to resolve the conflicts and merge it (it's best if I do it, as there's another refactor of verification.py going on at the moment, and I want to arrange the code in a way that'll minimise conflicts with that PR).

@petrelharp, do you agree this much is looking good and solid now?

tests/test_mutations.py
```diff
@@ -1022,6 +1024,63 @@ def verify_model_ts_general(self, ts, model, discrete, verify_roots, mutation_ra
...
+                pa = ts.mutation(mut.parent).derived_state
+            da = mut.derived_state
+            observed_transitions[alleles.index(pa), alleles.index(da)] += 1
+        pth = 0.05 / num_alleles
+        for j, (row, p) in enumerate(
+            zip(observed_transitions, model.transition_matrix)
+        ):
+            p[j] = 0
+            p /= sum(p)
+            self.chisquare(row, sum(row) * p, pth)
```
petrelharp, line 1046: ok, good, I agree with this

tests/test_mutations.py
```diff
@@ -1022,6 +1024,63 @@ def verify_model_ts_general(self, ts, model, discrete, verify_roots, mutation_ra
...
+                        * t_span
+                        * (
+                            1
+                            - np.exp(
+                                -mutation_rate * t.total_branch_length * change_probs
+                            )
+                        )
+                    )
+                else:
+                    t_span = t.span
```
petrelharp, line 1071: whoops this is unused here

tests/test_mutations.py
```diff
@@ -1022,6 +1024,63 @@ def verify_model_ts_general(self, ts, model, discrete, verify_roots, mutation_ra
...
+                    expected_ancestral_states += (
+                        model.root_distribution
+                        * mutation_rate
+                        * t.total_branch_length
+                        * t.span
+                        * change_probs
+                    )
```
petrelharp, line 1078: yep, good!

tests/test_mutations.py
```diff
@@ -1022,6 +1024,63 @@ def verify_model_ts_general(self, ts, model, discrete, verify_roots, mutation_ra
...
+
+            expected_ancestral_states /= expected_ancestral_states.sum()
+            expected_roots = expected_ancestral_states * ts.num_sites
+            self.chisquare(roots, expected_roots)
```
petrelharp, line 1082: I think that you should just do `self.chisquare(roots, expected_ancestral_states)`, without renormalizing it here?

tests/test_mutations.py
```diff
@@ -937,7 +937,9 @@ def verify_model_ts(self, ts, model, discrete, verify_roots):
             p /= sum(p)
             self.chisquare(row, sum(row) * p)
 
-    def verify_model_ts_general(self, ts, model, discrete, verify_roots, mutation_rate):
+    def verify_model_ts_general_old(
```
petrelharp, line 940: we can get rid of this one, now?

petrelharp:
> To get an idea of the variance along the perfect 45 degree line I have compared the outcome of msprime.mutate against itself by sprinkling mutations twice along the same tree (with a different seed).

Very nice. And yes, looks good!  A minor comment or two in the testing code.

jeromekelleher:
@GertjanBisschop, have Peter's comments been addressed? If so, I'll bring this up to date and merge it.

GertjanBisschop:
> @GertjanBisschop, have Peter's comments been addressed? 

Yes, all comments have been addressed.

>If so, I'll bring this up to date and merge it.

Thanks!

jeromekelleher:
Great, thanks @GertjanBisschop. I'll bring this up to date and merge ASAP.

jeromekelleher:
I've updated the PR here to use the newer framework for defining stats checks, and I also refactored things a bit while I was in there to reduce code duplication. Things could still be factored a little bit better (for example, we're computing the expectations over and over again, I think, but they're always the same) but this is a huge step forward. Thanks again for contributing this @GertjanBisschop !

The merge was hideous (I'm glad I didn't subject anyone else to it!), so I'm not 100% confident that everything has come through cleanly. Would you mind taking a quick look through please @GertjanBisschop to see if I've left anything out?

verification.py
```diff
@@ -3916,6 +3921,499 @@ def add_random_instance(
         self.add_ms_instance(key, cmd)
 
 
+class MutationTest(SimulationVerifier):
+    def _transition_matrix_chi_sq(self, transitions, transition_matrix):
+        tm_chisq = []
+        for row, p in zip(transitions, transition_matrix):
+            not_zeros = row > 0
```
GertjanBisschop, line 3928: This is a mistake from my side. Should be `not_zeros = p > 0`. Given the mutation rates we are using, this does not change much.

verification.py
```diff
@@ -3916,6 +3921,499 @@ def add_random_instance(
...
+            t = ts_mutated.first()
+
+            # expected number of ancestral alleles for sites
+            expected_ancestral_states_ts = np.zeros(num_alleles)
+            change_probs = transition_matrix.sum(axis=1) - np.diag(transition_matrix)
+            expected_ancestral_states_ts += root_distribution * (
+                1 - np.exp(-mutation_rate * t.total_branch_length * change_probs)
+            )
+            expected_ancestral_states_ts /= expected_ancestral_states_ts.sum()
+            expected_num_ancestral_states_ts = expected_ancestral_states_ts * num_sites
```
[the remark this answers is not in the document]
petrelharp: hm, yes, let's fix this

GertjanBisschop:
> I've updated the PR here to use the newer framework for defining stats checks, and I also refactored things a bit while I was in there to reduce code duplication. Things could still be factored a little bit better (for example, we're computing the expectations over and over again, I think, but they're always the same) but this is a huge step forward. Thanks again for contributing this @GertjanBisschop !
> 
> The merge was hideous (I'm glad I didn't subject anyone else to it!), so I'm not 100% confident that everything has come through cleanly. Would you mind taking a quick look through please @GertjanBisschop to see if I've left anything out?

Thanks Jerome for reducing duplication. Everything seems to be there. I spotted two minor mistakes of mine. Can I correct those myself?

It is true that we are now calculating the matrixexponential too often. It would suffice to do that only once for each branch and store that result. Should I do this in a seperate PR afterwards?

petrelharp:
> It is true that we are now calculating the matrixexponential too often. It would suffice to do that only once for each branch and store that result. Should I do this in a seperate PR afterwards?

Meh - does the test take a long time to run? There's no need to add complexity just to be efficient in testing code.

GertjanBisschop:
> > It is true that we are now calculating the matrixexponential too often. It would suffice to do that only once for each branch and store that result. Should I do this in a seperate PR afterwards?
> 
> Meh - does the test take a long time to run? There's no need to add complexity just to be efficient in testing code.

I can run 250 replicates in a minute

jeromekelleher:
> Meh - does the test take a long time to run? There's no need to add complexity just to be efficient in testing code.

They do take a fair bit to run, but more than that, it helps to understand what the code is doing. If I see something done inside the loop, I expect it *needs* to be done inside the loop. I don't think it'd add complexity, it just needs a bit of refactoring.

jeromekelleher:
You can add another commit here if you like @GertjanBisschop, but as I've forced pushed to the branch, you'll need to be a bit nasty:

```
git fetch origin
git reset --hard origin/finite_sites_verification
```

then you can add a commit fixing the errors, and push as before. I think it would be better to wait for a follow up PR for any refactorings though, there's other people working on the file and we want to get the big stuff in.

jeromekelleher:
Or we can merge this now and you can fix the issues in a follow-up @GertjanBisschop - it's up to you.

GertjanBisschop:
I'll try the nasty stuff. Seems straightforward. Thanks for the commands.

GertjanBisschop:
Worked. Mistakes are fixed.

jeromekelleher:
Great, merging. Thanks @GertjanBisschop!

Have you stuff you're planning on following up on @GertjanBisschop? I can make some issues if that'll help.

GertjanBisschop:
> Great, merging. Thanks @GertjanBisschop!
> 
> Have you stuff you're planning on following up on @GertjanBisschop? I can make some issues if that'll help.

I will defnitely optimise the code: not calculating matrix exponentials unnecessarily, ...

PR #1039, opened 2020-05-24
Test the OOA model against stdpopsim

### LINKED ISSUE #997: [Docs] Move OOA example to tutorial/demography/complete example
We decided that the OOA example is good to keep around.

We should put a date/version number on the warning box, being precise about when the wrong version of the model was identified.

(I'd also like to have a blog-post or something about this to give people an idea of what they might expect if they *did* use the old model. We should link to this.)

jeromekelleher:
Make sure the OOA model 

Also moves OOA model to its new location in the tutorial.

Closes #997 

Currently draft - need to release stdpopsim 0.1.2 before we can actually test against the correct model. Parking until then.


jeromekelleher:
This is ready for merging now I think. It hasn't completely systematised the process of making code examples for the tutorial, but it's a good start. At least we're sure the OOA example is correct now, anyway. I'll open an issue to track what needs to be done to follow up.

PR #1043, opened 2020-05-26
Restore attribute default for SimpleBottleneck.

### LINKED ISSUE #1040: SimpleBottleneck breaking change between 0.7.4 and current master
Stdpopsim uses SimpleBottleneck in a couple of unit tests, despite its undocumented status. One unit test now breaks when using git master, because the population parameter is not explicitly provided. Previously SimpleBottleneck had a `population=None` default, but this has been removed in the transition to being an attr class.

0.7.4
```
class SimpleBottleneck(DemographicEvent):
    # This is an unsupported/undocumented demographic event.
    def __init__(self, time, population=None, proportion=1.0, population_id=None)
        ...
```

current
```
# This is an unsupported/undocumented demographic event.
@attr.s
class SimpleBottleneck(DemographicEvent):
    population = attr.ib()
    proportion = attr.ib(default=1.0)
    ...
```

grahamgower:
This was lost somewhere in the change to using attrs.

Fixes #1040.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -275,7 +275,7 @@ class MassMigration(DemographicEvent):
 # This is an unsupported/undocumented demographic event.
 @attr.s
 class SimpleBottleneck(DemographicEvent):
-    population = attr.ib()
+    population = attr.ib(default=None)
     proportion = attr.ib(default=1.0)
 
     def get_ll_representation(self, num_populations=None):


jeromekelleher:
Thanks @grahamgower. I'm not sure we should go through with this one though - the low-level code will error if passed a None for the population value, and for consistency we'd probably want the default value to be ``-1`` if we wanted to mean "all populations".

How is this used in stdpopsim? I think it might be better to change the test case there, as it won't work with the msprime engine, anyway.

grahamgower:
In stdpopsim, its only used for testing if the model equality infrastructure works (ie. the demographic_events_equal() method). The demographic events aren't simulated, and a population could easily be specified in the one instance where it's missing.

jeromekelleher:
Let's do that then.

PR #1045, opened 2020-05-27
start at matrix model docs

petrelharp:
I've written down some technical details - basic facts about Markov models of mutation - that will probably be useful for people interpreting what they see from their mutation model.  For reference also for #1032 .

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -556,7 +556,7 @@ then A and T alleles have a 25% lower mutation rate than do C and G alleles,
 since 25% of the time that we consider mutating them, we leave them unchanged.
 From the properties of the Poisson distribution,
 the probability that a tree of total length :math:`T`
-has no mutations at a given discrete site is :math:`exp(-rT)`,
+has no mutations at a given discrete site is :math:`\exp(-rT)`,
 if mutations are put down at a rate of :math:`r`.
 Suppose that a single tree of total length :math:`T = 1.5`
 extends over many discrete sites,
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -493,6 +494,106 @@ overlay neutral mutations on this tree sequence.
...
+with instantaneous transition rate from ``i`` to ``j`` that is equal to
+``rate`` multiplied by ``transition_matrix[i,j]``.
+The ``root distribution`` and every row in the ``transition_matrix``
+must give *probabilities*, i.e., they must be nonnegative numbers summing to 1.
+
+To interpret the implications,
+it helps to know how the underlying mutational process is implemented.
+First, "possible" mutations are placed on the tree,
+with a mean density equal to the ``rate``, per unit of time and sequence length.
+If ``discrete=False`` then this is an infinite-sites model,
```
jeromekelleher, line 517: Something that just occured to  me here - it is *possible* that someone could want a continuous position but not guaranteed infinite sites behaviour. I.e., it does happen that we get multiple hits at sites when we have continuous positions, but we rejection sample them to guarantee infinite sites. Discrete and infinite sites are not quite the opposite of each other here. I guess it's even possible that someone would want infinite sites *and* discrete (if they have a large genome and low mutation rate).

Are we making a mistake by conflating discrete and infinite sites into the same parameter? It would be easy to separate them.

petrelharp: Hm, good thought? But no, I think we've made the right decision. If someone really wants to simulate a discrete-sites model with 2^24 sites and the likelihood of multiple hits determined by the vagaries of pseudorandom number generation, then they can just set `sequence_length=2**24`.

petrelharp:
I'll leave this open for @GertjanBisschop to have an edit of, and maybe add things to? (let me know if you're done, @GertjanBisschop?)

docs/api.rst
jeromekelleher: Agreed.

GertjanBisschop:
Thanks Peter. I have nothing to add. Was/is a great help.

Issue #981, opened 2020-04-30
Fix OOA example


petrelharp:
As described here: https://github.com/popsim-consortium/stdpopsim/pull/496 the example [in the docs](https://msprime.readthedocs.io/en/latest/tutorial.html#id1) is slightly wrong - this should be fixed and the subtle reason it was wrong should be explained.

apragsdale:
I can have a go at fixing the model here as well. I also saw in the docs that there is a todo to include a digram of the model. Would something like this figure work here: https://github.com/apragsdale/demography/blob/master/examples/figures/gutenkunst_ooa.pdf

jeromekelleher:
Thanks @apragsdale, that would be very helpful. An SVG version of the plot you linked to would be really helpful in the docs.

This is a serious problem though, as a bunch of papers have used the wrong version of the model. :frowning_face:

apragsdale:
Ah, oh no. That does appear to be a more serious problem than I thought. Not sure what else to do at this point, aside from fixing it and attaching the note that is included now.. let me know if I can help with anything else with it.

jeromekelleher:
Thanks @apragsdale, I appreciate it. I'll have a think about how to best handle this.

petrelharp:
One check on how serious this is would be to compare theoretical coalescence time distributions for the two models on the last time period (where they might differ), using `coalescence_rate_trajectory()`.  This doesn't check how different linkage would be, though.  @apragsdale, perhaps you could calculate some of your linkage stats for the two models, and see how they differ?

PR #1050, opened 2020-06-01
Add DemographyDebugger warning for merged-population migration rates.

grahamgower:
This is intended to catch problems like #981.


jeromekelleher:
Good idea @grahamgower, this is very helpful. I'm not sure why Appveyor is failing, but it might go away after rebasing.

Also, a couple of tests are now emitting warnings. We should probably just change the tests so that they don't.

grahamgower:
The `test_possible_lineages_with_migration()` test seems valid, as the possible_lineages() output is accurately reporting that ancestral material could exist in the already-merged populaton. So I just suppressed the warning there.

jeromekelleher:
Great, thanks @grahamgower. Merging.

PR #1057, opened 2020-06-02
Add DemographyDebugger.as_graph().

### LINKED ISSUE #1047: [Demography] Export Demography description to "demography" format?
The [demography package](https://github.com/apragsdale/demography) is an incredibly useful debugging tool. Can we export our demography descriptions in a format that interoperates? It looks like demography uses networkx, so perhaps we could export a description in a low-level dict-of-dicts format that networkx could read in? I don't think we should have a dependency on networkx here, but exporting to a format that networkx can read would be no problem.

Is this feasible @apragsdale? This would be part of the larger refactor of the larger project to refactor the demography API (https://github.com/tskit-dev/msprime/projects/10), for which there's a few half finished PRs lying around (see #1028, #1025, #970)

grahamgower:
Closes #1047.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -1156,7 +1156,10 @@ class DemographyDebugger:
             ``growth_rate``, ``M_in`` and ``M_out``.
             ``M_in[k]`` is the migration rate from population `k` into
             the node, and ``M_out[k]`` is the migration rate out of the node
-            and into population `k`.
+            and into population `k`. ``pulse_in[(k, t)]`` is the proportion of
+            migrants into the node from population `k` at time `t`.
+            ``pulse_out[(k, t)]`` is the proportion of migrants out of the node
+            that enter population `k` at time `t`.
         """
 
         def attrs_equal(attr1, attr2, abs_tol=1e-9):
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -1119,3 +1119,144 @@ class DemographyDebugger:
         Returns the number of epochs defined by the demographic model.
         """
         return len(self.epochs)
+
+    def as_graph(self, leaves=None):
+        """
+        Build a networkx-compatible dict-of-dicts digraph of the demography.
+
+        Each node in the graph represents a single population over a time span
+        in which the population's size and migration parameters don't change.
+        Multiple nodes may exist for a given population, reflecting different
+        demographic parameters in different time periods. Nodes are represented
+        as (epoch_id, population_id) tuples, where epoch_id is an index into
+        the self.epochs array.
+        Edges in the graph are directed from ancestors to descendents. An edge
+        between nodes with the same population ID indicates one or more
+        demographic parameter changes, whereas an edge between nodes with
+        different population IDs indicates a MassMigration(proportion=1).
+
+        To convert the graph into a networkx DiGraph::
+
+            import networkx as nx
+
+            ddb = msprime.DemographyDebugger(...)
+            dict_graph, node_attrs = ddb.as_graph()
+            G = nx.DiGraph(dict_graph)
+            assert nx.is_directed_acyclic_graph(G)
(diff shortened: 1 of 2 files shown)


msprime/demography.py
```diff
@@ -1119,3 +1119,144 @@ def num_epochs(self):
...
+                        interned.add(de.dest)
+                        dest = (j, de.dest)
+                        graph[dest][source] = {}
+                        end_time = max(node_attrs[source]["end_time"], de.time)
+                        node_attrs[source].update(end_time=end_time)
+                    else:
+                        dest = get_descendent(node_attrs, (j, de.dest))
+                        assert dest is not None
+                        node_attrs[dest]["pulse_in"][(source, de.time)] = de.proportion
+                        node_attrs[source]["pulse_out"][(dest, de.time)] = de.proportion
```
grahamgower, line 1227: I *think* these are in and out wrt backwards time.

msprime/demography.py
```diff
@@ -1119,3 +1119,144 @@ def num_epochs(self):
...
+                    node = (j, k)
+                    assert node not in node_attrs
+                    node_attr = dict(
+                        start_size=pop.start_size,
+                        end_size=pop.end_size,
+                        growth_rate=pop.growth_rate,
+                        start_time=epoch.start_time,
+                        end_time=epoch.end_time,
+                        M_in=epoch.migration_matrix[k, :],
+                        M_out=epoch.migration_matrix[:, k],
```
grahamgower, line 1242: I *think* these are in and out wrt forwards time.

grahamgower:
And I reckon [this code](https://gist.github.com/grahamgower/19eef3a6cd1c4e27c850e09ca5b34b28) does the right thing for @apragsdale's `demography` library (modulo getting the pulse_in/pulse_out and M_in/M_out conventions correct). But obviously this presents a problem with how we test `as_graph()`, because we really want to test that it integrates nicely in the `demography` library.

I think that `as_graph()` presents a nice way of debugging user-defined demographic models. In particular, I wonder if the `check_graph()` test method I added should really be a method of DemographyDebugger as well, because I suspect it can catch a number of demographic misspecification issues.

msprime/demography.py
grahamgower: I was wrong, it appears this correpsonds to in/out with a backwards-time convention. This seems appropriate.

jeromekelleher:
This looks great @grahamgower, thanks! I dug into the line-by-line details, but the high-level view looks just right.

> But obviously this presents a problem with how we test as_graph(), because we really want to test that it integrates nicely in the demography library.

We can definitely test against the demography library. Is it pip installable? If so, we just add it to our dev dependencies and add a few unit tests to make sure that it accepts the input and so on.

grahamgower:
> We can definitely test against the demography library. Is it pip installable? If so, we just add it to our dev dependencies and add a few unit tests to make sure that it accepts the input and so on.

I think so, there's a `setup.py` file at least. Does that mean one can pip install from the git repository? Is it possible to pip install an open PR? I updated the PR over at https://github.com/apragsdale/demography/pull/44 but this is still going to require some back-and-forth to make sure the two projects interoperate. It would be nice be able to test that before committing changes to either repository. Any help here would be appreciated! I'll step back for a while and give you both time to review this. @apragsdale, @jeromekelleher.

jeromekelleher:
> I think so, there's a setup.py file at least. Does that mean one can pip install from the git repository?

Yes, we should be able to pip install from a git URL all right. We could probably install from your fork/branch as well, I think the pip install from git is pretty general.

jeromekelleher:
@grahamgower - should we close this in favour of the upstream work on demes, or is this still relevant?

grahamgower:
Ah, yes. I might reuse some of this code to provide an `as_demes()` method instead, but that will be a new pr.

PR #1059, opened 2020-06-06
Multiple merger common ancestor time fix

JereKoskela:
@jeromekelleher, this is what I think multiple merger coalescents without rescaling should look like. It probably isn't necessary to include a complete `get_common_ancestor_waiting_time_from_rate` function in each model separately, but I thought writing it this way was the most transparent way to see what is going on at first. Let me know if you have any comments.

@TPPSellinger, could you have a look and see if you can break this please? I have run the standard unit tests as well as the expected SFS and break point tests from verification.py, and all of those look good. I seem to recall you had a library of some more extreme cases as well, right? For some reason I couldn't run the `xi_beta_recombinations` and `xi_dirac_recombinations` tests from verification.py - python kept complaining about an invalid key even when I copied and pasted the key directly from the code. If you can get those to run, that would be good too.

@eldonb, would you mind reading through the Dirac coalescent case please? I've taken an edgucated stab at how population growth should work for that model, but to my knowledge the Cannings model leading to a mixture of Kingman and Xi-mergers with population growth has never been written down, so that is where I expect my probability of error to be largest.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -5640,6 +5640,45 @@ msp_std_common_ancestor_event(
  * Dirac coalescent
  **************************************************************/
 
+/* Given the specified rate, return the waiting time until the next common ancestor
+ * event for the specified population */
+static double
+msp_dirac_get_common_ancestor_waiting_time_from_rate(
+    msp_t *self, population_t *pop, double lambda)
+{
+    double ret = DBL_MAX;
+    double alpha = pop->growth_rate;
+    double t = self->time;
+    double u, dt, z;
+
+    if (lambda > 0.0) {
+        u = gsl_ran_exponential(self->rng, 1.0 / lambda);
+        if (alpha == 0.0) {
+            ret = pop->initial_size * pop->initial_size * u;
+        } else {
+            dt = t - pop->start_time;
+            z = 1 + alpha * pop->initial_size * pop->initial_size
+                * exp(-alpha * dt) * u;
+            /* if z is <= 0 no coancestry can occur */
+            if (z > 0) {
+                ret = log(z) / alpha;
+            }
+        }
+        if (u == 0) {
+            /* In the exceedingly rare cases where gsl_ran_exponential returns
+             * 0, we return the smallest representable value > the current time
+             * to avoid returning a tree sequence with zero length branches.
+             * Note that we can still return 0 from this function if population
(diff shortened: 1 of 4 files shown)


TPPSellinger:
I remember also having the key issue and solved it recently. I will fix it, run the recombination test  (I will send you email with the code fixed if you want (probably tomorrow)). Some scenarios might take some time to run (Actually, I wanted to ask @jeromekelleher  about this, some scenarios seem to run faster with scrm (no approximation made) than with msprime). I will let you know if everything runs or if I encounter any issue. Should I test for anything else ? 

PS: What do you mean by : without rescaling ?

JereKoskela:
> PS: What do you mean by : without rescaling ?

That everything is simulated directly in units of generations. It's still the same infinite population Xi-coalescent model, but we avoid having to first simulate the rescaled process in units of N generations, and then converting the branch lengths to generations in the end.

TPPSellinger:
Here are the recombination results for the dirac and the beta coalescent:
Beta : 1.1

n=100_alpha=1 1_growth_0_rec_check
Dirac : psi=0.9 , c = 100

n=100_psi=0 9_c=100_growt_0_rec_check

Everything seems fine to me ! Great job @JereKoskela  !

jeromekelleher:
This looks great, thanks @JereKoskela! I think the best approach would be to merge now, and create an issue to track the population growth issue. We can then either come up with some tests to check if it's working properly, or force it to be zero before  release.

Thanks for checking this out @TPPSellinger, that's very helpful.

jeromekelleher:
Looks like there's some linting issues with the C code and it might be nice to factor out the tricky u-is-exactly-zero case here. Are you happy for me to finish this one up @JereKoskela?

JereKoskela:
> This looks great, thanks @JereKoskela! I think the best approach would be to merge now, and create an issue to track the population growth issue. We can then either come up with some tests to check if it's working properly, or force it to be zero before release.

I agree. My guess is that it should be fixed by this update, but I'll have to think about how to verify that. I reran my example from #985, and at least that looks to be working properly now. The problem was the implicit Kingman-ness in the connection between `pop->initial_size` and common ancestor waiting times.

> Are you happy for me to finish this one up @JereKoskela?

Absolutely, thanks @jeromekelleher!

jeromekelleher:
OK, great. Leave it with me and I'll merge it today.

PR #1063, opened 2020-06-08
clang formatting docs

petrelharp:
It looks like our CI is running clang-format, like so:
```
  python run-clang-format.py --clang-format-executable=clang-format --exclude lib/avl.c --exclude lib/avl.h --exclude lib/dev-tools --exclude lib/subprojects -r lib
```
so we should instruct developers (eg remind myself) how to do this locally, like [in the tskit docs](https://tskit.readthedocs.io/en/latest/development.html#sec-development-c-code-style). How's this?


petrelharp:
Hah, I just noticed that it *does* give the clang-format line in the quickstart. (I swear I'd searched for "clang-format" in the docs before?) But, I think this is still a good thing to put in. I've switch the command over to Ben's more concise command.

Issue #1023, opened 2020-05-12
Amino acid mutation models


jeromekelleher:
After implementing nucleotide mutation models (#1019, #1020, #1022) we should implement some amino acid models. We probably don't need to implement all of the models from SeqGen, maybe two would be sufficient. Say, PAM and Blosum62 (arbitrarily)?

jeromekelleher:
Closed in #1065

Issue #1060, opened 2020-06-08
Intermittent test failure: test_models.py:TestMixedModels.test_models_out_of_order


grahamgower:
Sometimes `ValueError` isn't raised in:
```
    def test_models_out_of_order(self):
        with self.assertRaises(ValueError):
            msprime.simulate(
                Ne=10,
                sample_size=10,
                demographic_events=[
                    msprime.SimulationModelChange(10, "hudson"),
                    msprime.SimulationModelChange(8, "hudson"),
                ],
            )
```
I guess a fixed seed could be used to ensure the simulation lasts to time=10 (or is it time=8?). But probably it makes more sense for this class of checks to occur without actually running the simulation such that failures are 100% deterministic.

PR #1065, opened 2020-06-09
implementation of AA models

GertjanBisschop:
solves #1023


GertjanBisschop:
These are the same checks for the AA models as for the nucleotide models:

[AA.pdf](https://github.com/tskit-dev/msprime/files/4752186/AA.pdf)

`validate_model(model)` and `validate_stationary(model)` still needs to be added to `tests/test_mutations.py`. But I got an error that I didn't have before: `ValueError: 'reference_size' not specified`. Something similar seems to have been addressed a couple of days back... I will check the slack channel.

msprime/mutations.py
```diff
@@ -187,6 +209,466 @@ def __init__(
         super().__init__(alleles, root_distribution, transition_matrix)
 
 
+class BLOSUM62(MutationModel):
+    def __init__(self):
+        alleles = _AMINO_ACIDS
+        num_alleles = len(alleles)
```
jeromekelleher, line 215: I would be good to document where these values came from - can you give some notes here on where you got them? Some of the values are at 3 digits of precision, others at 12. Good to give some provenance on that also.

msprime/mutations.py
```diff
@@ -187,6 +209,466 @@ def __init__(
...
+        row_sums = transition_matrix.sum(axis=1, dtype="float64")
+        np.fill_diagonal(transition_matrix, 1.0 - row_sums)
+        super().__init__(alleles, root_distribution, transition_matrix)
+
+
+class PAM(MutationModel):
+    def __init__(self):
+        alleles = _AMINO_ACIDS
+        num_alleles = len(alleles)
+        root_distribution = [
```
jeromekelleher, line 446: Likewise here on the numbers - where do they come from?

verification.py
```diff
@@ -2107,10 +2107,12 @@ def run_xi_beta_breakpoints(self):
                 f"n=100_alpha={alpha}",
                 sample_size=100,
                 Ne=Ne,
-                r=1e-7,
```
jeromekelleher, line 2110: Looks like something funny has happened here while you were rebasing. This is cancelling out some of the changes made in #1059. Probably easiest to inspect the code and manually revert your local changes.

msprime/mutations.py
GertjanBisschop: True. These are the values that are used in both pyvolve and seqgen for their rate matrices. I still need to check whether there is a better way, meaning deriving the ratet matrix from the time-dependent probability matrices as originally published. And whether it makes sense to actually do this each time, or if the best way is to store the raw values as is done in pyvolve and seq-gen.

jeromekelleher: Copying in the values like this is totally fine - just a note on where they came from is all we need.

GertjanBisschop:
PAM:
root distribution
root_distribution

diversity (with scikit)
pi

transition probabilities
alleles

BLOSUM62:
root distribution
root_distribution

diversity
pi

transition probabilities
alleles

jeromekelleher:
Great, thanks @GertjanBisschop! There's still two accidental changes to the multiple merger coalescent verification stuff - we're good to merge once these are reverted.

GertjanBisschop:
No, I think I have now changed the corrections back to the mistake. I can't see what else is still wrong.

jeromekelleher:
Great, thanks @GertjanBisschop! Can you [squash](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) the commits down to 1 please? Ready to merge then.

GertjanBisschop:
@jeromekelleher, forgot to ping you. All commits squashed down to 1.

jeromekelleher:
Merging; thanks @GertjanBisschop!

jeromekelleher:
Dammit, looks like this is failing because of #1060. I'll fix this.

jeromekelleher:
Thanks again @GertjanBisschop! Just FYI, if you want GitHub to autoclose issues you need to use "Closes" or "Fixes". "solves" doesn't work, it seems.

https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

GertjanBisschop:
OK thanks. Good to know. Shall I add the 'closed in' comment to the other mutation model issues as well?

jeromekelleher:
Yes please!

PR #1067, opened 2020-06-09
collapse highlevel and lowlevel Simulator classes

### LINKED ISSUE #1053: Factor out high-level Simulator class
The Simulator class in [ancestry.py](https://github.com/tskit-dev/msprime/blob/6718266c48706f964364ced485d34c399a94c0c9/msprime/ancestry.py#L556) is basically a thin wrapper around the low-level Simulator object in the Python-C interface. It doesn't really do anything much add a bunch of boilerplate, and the functionality it does have could easily be pulled out into functions.

It would be nice if we could set get rid of the high level Simulator class entirely, and change the low-level Simulator so that all the methods like ``get_num_breakpoints()`` become properties (that is, so they are basically like the high-level simulator).

It would be a bit of refactoring, but it would remove a lot of boilerplate and complexity, and it would help people understand how things fit together better.

grahamgower:
closes #1053

Not done yet, but these are the changes most likely to cause merge conflicts. I wasn't sure which of the remaining code should be in simulator_factory or the Simulator subclass, so figured I'd send this in for some feedback.

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -399,41 +399,29 @@ def simulator_factory(
     if gene_conversion_track_length is None:
         gene_conversion_track_length = 1
 
-    # TODO factor out this Simulator object here and instead call a
-    # factory function to make the low-level Simulator given the
-    # state of the variables in here.
-    sim = Simulator()
-    sim.samples = samples
-    sim.model = model
-    sim.recombination_map = recombination_map
-    sim.from_ts = from_ts
-    sim.demography = demography
-    sim.model_change_events = model_change_events
-    sim.store_migrations = record_migrations
-    sim.store_full_arg = record_full_arg
-    sim.start_time = start_time
-    sim.num_labels = num_labels
-    sim.gene_conversion_rate = gene_conversion_rate
-    sim.gene_conversion_track_length = gene_conversion_track_length
-    # TODO see notes on the WrightFisherPedigree; the pedigree should
-    # be a parameter of the *model*.
-    sim.pedigree = pedigree
-    if sim.pedigree is not None:
-        sim.check_pedigree()
     if random_generator is None:
         # For the simulate code-path the rng will already be set, but
         # for convenience we allow it to be null to help with writing
         # tests.
         random_generator = _msprime.RandomGenerator(core.get_random_seed())
-    sim.random_generator = random_generator
 
-    # We always need at least n segments, so no point in making
-    # allocation any smaller than this.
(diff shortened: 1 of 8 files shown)


jeromekelleher:
YES! Thanks @grahamgower, this looks just right! Some great stuff in here.

Do you mind if we park this for a few days while a few other PRs are working their way through the pipeline? I have #1025, #970 and #1066 which will all conflict a little bit with this, so it'd be nice to flush them through first.

But this is great, I'd expect only minor changes from what you have.

grahamgower:
No probelm @jeromekelleher, just ping me once those other PRs have merged and let me know what you'd like changed here.

grahamgower:
Conflict resolution was a little tricky in some bits, so this wants careful checking that I didn't remove anything that was just introduced.

tests/test_lowlevel.py
```diff
@@ -546,45 +546,46 @@ def verify_squashed_edges(self, sorted_edges):
...
-            self.assertGreaterEqual(sim.get_num_rejected_common_ancestor_events(), 0)
-        self.assertGreater(sim.get_num_avl_node_blocks(), 0)
-        self.assertGreater(sim.get_num_segment_blocks(), 0)
-        self.assertGreater(sim.get_num_node_mapping_blocks(), 0)
-
-        edges = sim.get_edges()
+        if sim.model == "hudson":
+            self.assertEqual(sim.num_rejected_common_ancestor_events, 0)
+        elif sim.model in ["smc", "smc_prime"]:
+            self.assertGreaterEqual(sim.num_rejected_common_ancestor_events, 0)
```
jeromekelleher, line 571:
```suggestion
        model_name = sim.model["name"]
        if model_name == "hudson":
            self.assertEqual(sim.num_rejected_common_ancestor_events, 0)
        elif model_name in ["smc", "smc_prime"]:
            self.assertGreaterEqual(sim.num_rejected_common_ancestor_events, 0)
```

msprime/ancestry.py
```diff
@@ -695,195 +683,171 @@ def simulate(
 # and replace with the low-level class. If we change the low-level class
```
jeromekelleher, line 683: Can delete this comment now

tests/test_lowlevel.py
grahamgower: Ah, nice catch!

grahamgower:
squashed and ready...

jeromekelleher:
Thanks again Graham, this is fantastic.

Issue #1056, opened 2020-06-02
Decouple test definition from test running in verification.py


jeromekelleher:
#1035 made a good start on refactoring the verification.py script, which is a big horrible mess.

The next step in the process of making this comprehensible and easier to work with is to decouple the test *definitions* from the actual running of the tests.

What I'd like to see is something like the following 
```python
@attr.s
class Test:
    """
    The superclass of all tests. Each test must define a name, a group
    and a run method.
    """
    name = attr.ib(type=str)
    group = attr.ib(type=str)

    def run(self, output_dir):
        raise NotImplementedError()


@attr.s
class MsCommandLine(Test):

    command = attr.ib(type=str)

    def run(self, output_dir):
        # Rather than printing like we did before, let's use the logging infrastructure. 
        logging.info("running", self.command)
        # Port all the run stuff in here.


def register_ms_command_line_tests(runner):
    def register(name, command):
        runner.register(MsCommandLine(name=name, group="ms", command=command))

    register("size-change1", "10 10000 -t 2.0 -eN 0.1 2.0")
    register("growth-rate-change1", "10 10000 -t 2.0 -eG 0.1 5.0")
    register("growth-rate-2-pops1", "10 10000 -t 2.0 -I 2 5 5 2.5 -G 5.0")
    # etc.

@attr.s
class TestRunner:
    """
    Class responsible for registering all known tests and running
    them.
    """
    tests = attr.ib(factory=list)
    groups = attr.ib(factory=set)

    def register(self, test):
        self.tests.append(test)
        self.groups.add(test.group)

    def run(self, output_dir, names=None, group=None):
        tests = self.tests
        if names is not None:
            tests = [test for test in self.tests if test.name in names]
        elif group is not None:
            tests = [test for test in self.tests if test.group == group]
        for test in tests:
            test.run(output_dir)
```

Thus, the TestRunner does nothing except register and run the tests. All the actual test behaviour is encapsulated in the classes. The ``group`` concept is flexible, and doesn't have to be tied to a particular test type or class.

Some of the tests are most bespoke where there's very little sharing of code between different tests within a group. The simplest thing for now there may be to make a different class for each test. Hopefully we'll be able to factor out some of the common code in those later.

How does this look @saunack?

saunack:
Looks good. I got most of it. I'll get back to you in case of difficulties

saunack:
[...]

PR #1068, opened 2020-06-10
Decoupling tests from definitions

saunack:
Working on #1056 
I have started with a skeleton for now. It was getting too big for one change so I thought to do it in steps first.
The `Cmd` classes create a temporary instance of the relevant `SimulationVerifier` child class and adds test instances from there, for now. @jeromekelleher if this is a step in the right direction, I will move forward with extracting the complete `add_instance()` (and other equivalent functions) code out of the  `SimulationVerifier` child classes
Also, I forgot to add the `TestRunner` this time. I will put it in the next commit.

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -6,6 +6,7 @@ import argparse
 import ast
 import collections
 import itertools
+import logging
 import math
 import os
 import random
@@ -15,6 +16,7 @@ import tempfile
 import time
 
 import allel
+import attr
 import dendropy
 import matplotlib
 import numpy as np
@@ -226,18 +228,21 @@ def make_test_dir(test_name):
     return basedir
 
 
-class SimulationVerifier:
+@attr.s
+class Test:
     """
-    Class to compare msprime against ms to ensure that the same distributions
-    of values are output under the same parameters.
+    The superclass of all tests. Each test must define a name, a group
+    and a run method.
     """
 
-    def __init__(self, output_dir):
-        self._output_dir = output_dir
-        self._instances = {}
-        self._mspms_executable = [sys.executable, "mspms_dev.py"]
-        self._slim_executable = ["./data/slim"]
-        self._ms_executable = ["./data/ms"]
+    name = attr.ib(type=str)
+    group = attr.ib(type=str)
+    _mspms_executable = attr.ib(init=False, default=[sys.executable, "mspms_dev.py"])
+    _slim_executable = attr.ib(init=False, default=["./data/slim"])
+    _ms_executable = attr.ib(init=False, default=["./data/ms"])
+
+    def run(self, output_dir):
+        raise NotImplementedError()
 
     def _run_sample_stats(self, args):
         print("\t", " ".join(args))
@@ -275,15 +280,6 @@ class SimulationVerifier:
             pyplot.savefig(f, dpi=72)
             pyplot.close("all")
 
-    def run(self, keys=None):
-        the_keys = sorted(self._instances.keys())
-        if keys is not None:
-            the_keys = keys
-        for key in the_keys:
-            print(key)
-            runner = self._instances[key]
-            runner()
-
     def get_ms_seeds(self):
         max_seed = 2 ** 16
         seeds = [random.randint(1, max_seed) for j in range(3)]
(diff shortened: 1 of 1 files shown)


verification.py
```diff
@@ -226,6 +227,20 @@ def make_test_dir(test_name):
     return basedir
 
 
+class Test:
```
jeromekelleher, line 230: Let's start using [attrs](https://www.attrs.org/en/stable/) for these classes. There's examples of using this in, e.g., the demography.py module.

verification.py
```diff
@@ -377,205 +392,158 @@ def verify_recombination(
...
-class MsTest(SimulationVerifier):
-    def __init__(self, output_dir):
-        super().__init__(output_dir)
+class MsCmd1(Test):
+    def __init__(self, name, group, command):
+        super().__init__(name, group)
+        self.command = command
+
+    def run(self, output_dir):
+        MsTest().add_ms_instance(self.name, self.command)
```
jeromekelleher, line 401: I don't understand how this would work - we'd need to port the code from the verifier infrastructure into here so that the test can actually run. We want to avoid using the old infrastructure for adding test instances, as this won't work in parallel (that's one of the main motivations for this refactor).

verification.py
```diff
@@ -377,205 +392,158 @@ def verify_recombination(
         pyplot.close("all")
 
 
-class MsTest(SimulationVerifier):
-    def __init__(self, output_dir):
-        super().__init__(output_dir)
+class MsCmd1(Test):
```
jeromekelleher, line 395: I think it would be better to see these as Test classes, so we can call this ``MsTest``. Same for the other classes below.

verification.py
saunack: I mean to do the same thing. But I thought of doing it in steps after I saw that the changes were getting too big for one commit. I mentioned it here too
> I have started with a skeleton for now. It was getting too big for one change so I thought to do it in steps first.

jeromekelleher: That's a fair point, but we do need to keep it working though, and I don't see how what you have currently would. I don't think the diff will be that big, if the classes that you have are mostly kept as is, and just the registration etc changes a bit?

saunack: Yeah. But I was not sure if I was proceeding the correct direction, so I wanted to get your views on it first. This commit was mainly for tracking and getting my progress evaluated. I'll finally merge all the commits into one that would be a working model.
I'll hopefully get everything correct in the next commit itself.

jeromekelleher: Great, thanks.

verification.py
```diff
@@ -4399,137 +4569,101 @@ def _run_pyvolve_comparison(self, model, length=20, num_samples=10):
...
-                smc,
-                discoal,
-                ms,
-                seqgen,
-                pyvolve,
-            ]
-        )
+    if args.extended:
+        register_dtwfvscoalescent_tests(runner, add_random_dtwf=True)
+    if args.tests is not []:
```
jeromekelleher, line 4638: Comparing for identity isn't recommended. Better to use ``if len(args.tests) > 0``. Also this has broken the ``-g`` behaviour, which doesn't currently work.

verification.py
```diff
@@ -4399,137 +4569,101 @@ def _run_pyvolve_comparison(self, model, length=20, num_samples=10):
...
-    # use a simple DFS to get the names of classes and
-    classes = SimulationVerifier.__subclasses__()
-    while classes:
-        top = classes.pop()
-        tests.append(get_class_name(top))
-        for c in top.__subclasses__():
-            classes.append(c)
-    return tests
+    def run(self, output_dir, names=None, group=None):
+        if names != []:
```
jeromekelleher, line 4608: It would probably be simplest to say that ``names`` and ``group`` are mutually exclusive here, so we have something like
```python
if names is not None and groupd is not None:
     raise ValueError("Cannot specify test names and group at the same time")
if names is not None:
     tests = [test for test in self.tests if test.name in names]
elif group is not None:
    tests = [test for test in self.tests if test.group == group]
else:
    tests = self.tests

￼```

This way we're just reasoning about the default values.

jeromekelleher:
LGTM, thanks @saunack. Merging.

jeromekelleher:
Oops, looks like the rebase has gone wrong here @saunack. You've pulled in changes from another PR (see the list of files changed). Can you rebase again please?

Issue #1058, opened 2020-06-05
Declarative demography


jeromekelleher:
After a discussion with @grahamgower this morning, we decided it would be nice to have a *declarative* structure for demography, where we describe what our populations are and how they relate to each other. Here's a first pass at doing the OOA model using a [toml](https://github.com/toml-lang/toml) description:

**Warning: the parameters are WRONG - DO NOT COPY THEM!!**
```toml
description = "Gutenkunst et al three population Out-of-Africa"
generation_time = 25
time_units = "kya"

[populations]
    [populations.ancestral_human]
        size = 7300
        time = 220

    [populations.ancestral_african]
        ancestor = "ancestral_human"
        population_size = 12300
        time = 140

    [populations.ancestral_eurasian]
        ancestor = "ancestral_africans"
        size = 12300
        time = 21.2

    [populations.YRI]
        ancestor = "ancestral_africans"
        size = 12300

    [populations.CEU]
        ancestor = "ancestral_eurasian"
        size = 1000
        growth_rate = 0.0055

    [populations.CHB]
        ancestor = "ancestral_eurasian"
        size = 510
        growth_rate = 0.004


# The limitations of toml here become apparent when we try to 
# declare some migration relationships between these populations.
[migration]
    symmetric = [
        # Annoyingly, the rate must be enclosed in a list for toml
        [["ancestral_african", "ancestral_eurasian"], [25e-5]],
        [["YRI", "CEU"], [3e-5]],
        [["YRI", "CHB"], [1.9e-5]],
        [["CEU", "CHB"], [9.6e-5]],
    ]
    # Could have asymmetric migration here also
   
    # Sticking in an admixture event, just to see if we can handle mass migration.
    admixture = [
        # (source, dest), (time, fraction). We need this lameness 
        # because toml won't accept mixed array types. This isn't 
        # great.
        [["YRI", "CEU"], [1.0, 0.1]]
    ]
```

Some parts of this are quite nice, and others are pretty nasty. I think toml is too restrictive and will have another go in a minute with yaml.

This is an inherently graphical description of the populations, inspired by @apragsdale's approach  in the demography package. A "population" in this context is modelled as a unit in which the parameters don't change and no mergers with other populations occur. Each population has an ``ancestor`` attribute, which points to the population that it was derived from. There is also a ``time`` attribute, which (forwards in time) is the time at which this population goes extinct (if it hasn't split into other populations).

[...]

Issue #1054, opened 2020-06-02
Discourage population reuse in demography API?


jeromekelleher:
Defining demographic models and interpreting the results is complicated by the fact that we reuse population IDs to mean different things at different times. Should we encourage users to define population models (and give an example in OOA) that explicitly defines and names all the ancestral populations? This would make interpretation of the tree sequence a lot easier later on, and I think make the models easier to interpret. This would remove a confusing asymmetry I think.

This would involve making the explicit event types require *new* populations PopulationSplit(ancestral_population, derived_populations)  - see #1013.

The extra populations would have some perf consequences in the current code, but it would be fairly easy change to get msprime to "ignore" empty populations (which it should do anyway, for large spatial models).

jeromekelleher:
Ping @grahamgower, @apragsdale

grahamgower:
At first glance, this seems like a good idea. But it will be more onerous to specify larger demographies. I guess there'll be on the order of 2*n population IDs for `n` extant populations.

I wonder if the nominal case should be to encourage users *not* to define models manually, but give them a tool to build a demography graphically, and have a machine readable model autogenerated. In this case it would make sense for all nodes in the demography graph to be assigned different msprime population IDs (in some nice deterministic way that makes it easy to write an inverse function).

jeromekelleher:
Interesting idea @grahamgower.

apragsdale:
I think I'd be more in favor of allowing the flexibility for the ancestral population to be either a new, unused population _or_ a continuation of one of the derived populations. You can imagine cases where both make sense historically (for example, in either some range expansion model or a model with island colonization from a mainland populations for the latter case).

jeromekelleher:
I agree @apragsdale, we shouldn't be too rigid about it. But, I think PopulationSplit events should have that form ``PopulationSplit(ancestral, derived)`` where ancestral must be a new population and ``derived`` is a list that cannot contain ``ancestral``. We'll still keep the MassMigration events so it'll be fully general.

apragsdale:
Right, we just want to makes sure we don't lose generality or flexibility in defining models. So I think as long as the MassMigration events still allow for a proportion of 1, we keep that flexibility.

PR #1069, opened 2020-06-10
Make Hudson simulate work better for large sparse sets of populations

jeromekelleher:
Part of the assumptions of using populations more freely in the demographic API update (e.g. #1054 and #1058) is that populations are cheap. There's no reason we should be spinning around a O(N^2) loop when computing migration rates. I'd previously assumed that the number of populations was O(10) so it wasn't worth doing anything fancier, but I think this assumption isn't correct. Anyway, we want to be able to run large spatial models where we're likely to have only local migration, and if we have, say, 1000 demes, then the O(N^2) loop is awful.

This is a simple approach to fixing this. During Hudson simulate we maintain a set of population IDs in which there's at least one lineage, and, for each population, a list of populations where lineages can potentially migrate to (i.e., with nonzero migration rate). Then, instead of iterating over all populations, we just iterate over these subsets. The only tricky thing is maintaining the list, which I think is simple enough so long as we don't try to be too clever.

I'd imagine that for really big models you might want to do something mathematically more clever, but I'm guessing that this will take us a long way. @petrelharp, do you agree with the approach? I've outlined it here in the algoriths.py file (with a few drive-by changes - I originally started doing something much more complex and started refactoring.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,9 @@
 - Add replicate_index to simulate, allowing output of a single tree sequence
   from a set of replicates. (:user:`benjeffery` :pr:`914`).
 
+- Much better simulation performance for models with large numbers
+  of populations (:user:`jeromekelleher`, :pr:`1069`).
+
 **Breaking changes**:
 
 - The class form for specifying models (e.g., ``msprime.StandardCoalescent()``)
--- a/algorithms.py
+++ b/algorithms.py
@@ -155,10 +155,10 @@ class Population:
     """
 
     def __init__(self, id_, num_labels=1):
-        self._id = id_
-        self._start_time = 0
-        self._start_size = 1.0
-        self._growth_rate = 0
+        self.id = id_
+        self.start_time = 0
+        self.start_size = 1.0
+        self.growth_rate = 0
         # Keep a list of each label.
         # We'd like to use AVLTrees here for P but the API doesn't quite
         # do what we need. Lists are inefficient here and should not be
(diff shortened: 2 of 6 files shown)


petrelharp:
That's the sort of think I had in mind. Basically, replacing a dense migration matrix with a sparse representation of it.

jeromekelleher:
This turned out quite well. Running this extreme example (that basically does nothing except migrations):
```python
import msprime

N = 1000
model = msprime.Demography.stepping_stone_1d(N, 1, circular=False)
# Take samples from the extreme ends of the linear habitat.
samples = [
    msprime.Sample(time=0, population=0),
    msprime.Sample(time=0, population=N - 1)]

ts = msprime.simulate(
    samples=samples, demography=model, random_seed=1,
    record_provenance=False)

print(ts.tables.nodes)
```

Takes 4.5 minutes with the code in master and 1.8 seconds with this branch. That's a good note to finish the week on I think!

jeromekelleher:
I've quickly checked some stats tests, but this probably shouldn't be merged until I've run more of them and double checked there's no subtle errors after creeping in.

petrelharp:
Mulddling through a start at something like this was what I occupied myself with late on election night of 2016, so I'll take it as a Good Omen that you've got it resolved in 2020.

jeromekelleher:
> Mulddling through a start at something like this was what I occupied myself with late on election night of 2016, so I'll take it as a Good Omen that you've got it resolved in 2020.

Err, great! Maybe merging this will revert all the crappy bits of the past 4 years then! It's worth a go anyway.

jeromekelleher:
I've checked the stats tests and they all seem to be OK, so I'm going to merge. Not sure what CodeCov is complaining about, as there's no changes to the Python code.

petrelharp:
git rebase 2016

jeromekelleher:
Hell yeah.

PR #1079, opened 2020-06-19
Fix subtle numerical issues with recombination.

### LINKED ISSUE #1046: Core dumped error
I got the following error : 
~~~~
python3: lib/object_heap.c:136: object_heap_free_object: Assertion `self->top < self->size' failed.
Aborted (core dumped)
~~~~
After running the following code : 
~~~~
import matplotlib
import numpy as np
import pandas as pd
import scipy.special
import scipy.stats
from matplotlib import lines as mlines
from matplotlib import pyplot
import msprime
import msprime.cli as cli

sample_size=20
Ne=10**4
r=1*1e-8
m=1.25*1e-8
L=10 ** 8
strengh=10

population_configurations=[msprime.PopulationConfiguration(initial_size=Ne,sample_size=sample_size)]
demographic_events = [
            msprime.PopulationParametersChange(
                time=1000, growth_rate=-0.00025584278811045
            ),
            msprime.PopulationParametersChange(
                time=10000, growth_rate=0
            ),
        ]
for x in range(1,11):
    ts=msprime.simulate(population_configurations=population_configurations,recombination_rate=r,mutation_rate=m,length=L,demographic_events=demographic_events)
    name="Figure_1_Decrease_strength_"+str(strengh)+"_"+str(x)+".vcf"
    with open(name, "w") as vcf_file:
        ts.write_vcf(vcf_file, 2)
    vcf_file.close()
~~~~
The error does not always occur, it might run couple of iterations before crashing.

jeromekelleher:
Closes #1046

Tricky, tricky numerical problems. Hopefully this sorts it - I don't know much more paranoid one can be here.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1715,13 +1715,30 @@ msp_defrag_segment_chain(msp_t *self, segment_t *z)
     return 0;
 }
 
-/* TODO: is floor the correct operation here or should we round? */
-static double
-msp_dtwf_generate_breakpoint(msp_t *self, double start)
+static int
+msp_dtwf_generate_breakpoint(msp_t *self, double start, double *breakpoint)
 {
-    double k = recomb_map_sample_poisson(&self->recomb_map, self->rng, start);
-    assert(k > start);
-    return k;
+    int ret = 0;
+    double mass_to_next_recomb, x;
+    int num_breakpoint_resamplings = 0;
+    double left_limit = self->recomb_map.discrete ? start + 1 : start;
+
+    do {
+        mass_to_next_recomb = gsl_ran_exponential(self->rng, 1.0);
+        x = recomb_map_shift_by_mass(&self->recomb_map, left_limit, mass_to_next_recomb);
+        if (x > start) {
+            break;
+        }
+        num_breakpoint_resamplings++;
+        if (num_breakpoint_resamplings == 10) {
+            ret = MSP_ERR_BREAKPOINT_RESAMPLE_OVERFLOW;
+            goto out;
+        }
+    } while (true);
+
+    *breakpoint = x;
+out:
+    return ret;
 }
 
 int MSP_WARN_UNUSED
@@ -2122,7 +2139,10 @@ msp_dtwf_recombine(msp_t *self, segment_t *x, segment_t **u, segment_t **v)
     segment_t s1, s2;
     segment_t *seg_tails[] = { &s1, &s2 };
 
-    k = msp_dtwf_generate_breakpoint(self, x->left);
+    ret = msp_dtwf_generate_breakpoint(self, x->left, &k);
+    if (ret != 0) {
+        goto out;
+    }
     s1.next = NULL;
     s2.next = NULL;
     ix = (int) gsl_rng_uniform_int(self->rng, 2);
@@ -2168,7 +2188,10 @@ msp_dtwf_recombine(msp_t *self, segment_t *x, segment_t **u, segment_t **v)
             msp_subtract_segment_mass(self, x, z);
             assert(x->left < x->right);
             x = z;
-            k = msp_dtwf_generate_breakpoint(self, k);
+            ret = msp_dtwf_generate_breakpoint(self, k, &k);
+            if (ret != 0) {
+                goto out;
+            }
         } else if (x->right <= k && y != NULL && y->left >= k) {
             // Recombine in gap between segment and the next
             x->next = NULL;
@@ -2176,7 +2199,10 @@ msp_dtwf_recombine(msp_t *self, segment_t *x, segment_t **u, segment_t **v)
             while (y->left >= k) {
                 self->num_re_events++;
                 ix = (ix + 1) % 2;
-                k = msp_dtwf_generate_breakpoint(self, k);
+                ret = msp_dtwf_generate_breakpoint(self, k, &k);
+                if (ret != 0) {
+                    goto out;
+                }
             }
             seg_tails[ix]->next = y;
             if (seg_tails[ix] == &s1 || seg_tails[ix] == &s2) {
@@ -3643,18 +3669,13 @@ out:
 }
 
 /* Computes the scaling factor used to map floating point cumulative
- * recombination mass values to integers. The value is chosen so that
- * the sum of all maximum cumulative mass values is less than INT64_MAX
- * and therefore guaranteed not to overflow, whilst maintaining a very
- * high level of granularity.
+ * recombination mass values to integers.
  */
 static int MSP_WARN_UNUSED
 msp_set_recomb_map_mass_scale(msp_t *self)
 {
     size_t num_samples = self->num_samples;
     double max_mass = TSK_MAX(1, recomb_map_get_total_mass(&self->recomb_map));
-    int64_t large_int = 1LL << 61;
-    double mass_scale;
 
     if (self->from_ts != NULL) {
         /* The from_ts value can be an empty tree sequence, which will contain
@@ -3665,12 +3686,9 @@ msp_set_recomb_map_mass_scale(msp_t *self)
     /* Make sure that any ancient samples are taken into account too */
     num_samples += self->num_sampling_events;
     assert(num_samples > 0);
-
     /* We can have at most max_mass per sample */
     max_mass *= (double) num_samples;
-    /* The scale is value that maps max_mass to large_int */
-    mass_scale = ((double) large_int) / max_mass;
-    return recomb_map_set_mass_scale(&self->recomb_map, mass_scale);
+    return recomb_map_set_mass_limit(&self->recomb_map, max_mass);
 }
 
 /*
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1251,6 +1251,8 @@ msp_print_state(msp_t *self, FILE *out)
     fprintf(out, "Fenwick trees\n");
     for (k = 0; k < self->num_labels; k++) {
         fprintf(out, "=====\nLabel %d\n=====\n", k);
+        fprintf(out, "\ttotal scaled mass = %" PRId64 "\n",
+            msp_get_total_scaled_recombination_mass(self, (int) k));
         for (j = 1; j <= (uint32_t) fenwick_get_size(&self->links[k]); j++) {
             u = msp_get_segment(self, j, (label_id_t) k);
             scaled_recomb_mass = fenwick_get_value(&self->links[k], j);
@@ -2286,12 +2288,16 @@ msp_choose_uniform_breakpoint(msp_t *self, int label, double *ret_breakpoint,
         breakpoint
             = recomb_map_scaled_mass_to_position(&self->recomb_map, breakpoint_mass);
 
-        /* FIXME: can this happen now? */
-        /* Deal with various quirks that can happen with numerical
+        /* Deal with various quirks that may happen with numerical
          * imprecision from going back and forth between recombination mass
          * and physical positions. We try to make this robust by making
          * resampling the default case and only break out of the loop
-         * when the conditions we need are explicitly met. */
+         * when the conditions we need are explicitly met.
+         *
+         * It's not entirely clear whether these things can actually happen
+         * but it's best to be safe and return an actual error rather than
+         * bombing out on an assertion.
+         */
         if (x == NULL) {
             /* if there is no previous segment we cannot have breakpoint
              * <= y->left */
@@ -3685,8 +3691,19 @@ out:
 static int MSP_WARN_UNUSED
 msp_set_recomb_map_mass_scale(msp_t *self)
 {
+    int ret = 0;
     size_t num_samples = self->num_samples;
-    double max_mass = TSK_MAX(1, recomb_map_get_total_mass(&self->recomb_map));
+    double total_mass = recomb_map_get_total_mass(&self->recomb_map);
+    double mass_scale, initial_mass;
+    /* We start with a 48 bit number to represent the maximum cumulative
+     * scaled mass. For long genomes the total scaled mass does grow by
+     * quite a lot because of the effects of trapped genetic material.
+     * Empirically this seems like a safe limit, and is
+     * still a very fine level of granularity for smaller simulations
+     * so seems like a reasonable compromise. However, we may need to
+     * make this a parameter in the future if we do get overflows, so
+     * that users can at least tune the value. */
+    int64_t initial_scaled_mass = 1LL << 48;
 
     if (self->from_ts != NULL) {
         /* The from_ts value can be an empty tree sequence, which will contain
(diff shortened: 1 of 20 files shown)


jeromekelleher:
Can I get your eyes on this please @daniel-goldstein?

jeromekelleher:
Not entirely solved yet - still getting some assertions tripping. There's a lot to be said for integers...

jeromekelleher:
This turned out to be a *lot* nastier than I had hoped, and has illustrated a potentially fatal flaw in the current implementation of recombination. Since we moved to simulating in physical coordinates we've been using a simple conversion of the old integer based fenwick tree code to doubles. However, numerical imprecision means that this isn't necessarily a valid thing to do.

The fenwick tree maintains the cumulative recombination mass (rate times genome spanned) for each segment. It does this by summing the cumulative masses up the tree as segments are added and removed. With doubles, this is an inherently lossy process - we can't hope to exactly represent the input values by storing them in this cumulative tree. A little bit of precision loss is usually fine, but I think it's potentially catastrophic here. When you add two numbers of very different magnitudes, the smaller number is "lost". So, if we put lots of numbers of different magnitudes in to the fenwick tree, then the information about the smaller values will inevitably be lost. In the example I've been working with in #1046, we end up with value in the tree that differ by around 10 orders of magnitude.

Maybe this doesn't matter - if we have segments with very small recombination masses then we're very unlikely to choose these for recombination events. I worry that the fenwick tree representation is "drifting" over time, though and eventually it doesn't reflect reality. I guess the way to measure that is if the total sum in the fenwick tree is approximately equal to the total sum of the segment masses, or if it diverges over time. If they stay equal to within a reasonable relative accuracy we're probably OK.

I wonder if it might be simpler to translate the segment masses into an unsigned 64 bit integer. We could scale the mass by making the maximum cumulative sum equal to UINT64_MAX. Any losses of precision would then be predictable and we wouldn't have to worry about any losses from adding and subtracting values in the Fenwick tree.

@petrelharp, have you any thoughts here? I could use your help on this one!

(ps. the actual bug causing the assertion trip here here is solved, I think, as it had to do with returning segments mapped to zero mass (i.e., not in use) when a random mass very close to zero was generated. I hope anyway!)

molpopgen:
This is a long shot, but would something like Kahan summation help to minimize the problem?

https://en.wikipedia.org/wiki/Kahan_summation_algorithm

jeromekelleher:
> (ps. the actual bug causing the assertion trip here here is solved, I think, as it had to do with returning segments mapped to zero mass (i.e., not in use) when a random mass very close to zero was generated. I hope anyway!)

Correction: no I haven't. There's a different problem now. Goddamit I hate working with floats sometimes!

petrelharp:
Hm, ok. Let me make sure I know what's going on here. The Fenwick tree is used to pick a random segment proportional to length, for recombination, right? Does it have any other purpose? And, how are the segments ordered in the tree (like, where do newly inserted ones go)?

jeromekelleher:
> This is a long shot, but would something like Kahan summation help to minimize the problem?

It'd probably help, but it's finicky to do right and I don't think it would ultimately solve the issue (if the precision actually is a problem). We could implement the tree in long doubles, too, but that would (a) probably be slower and (b) not work on windows, as long double == double there.

Casting back to ints is looking more attractive all the time

daniel-goldstein:
Oh boy.

> I guess the way to measure that is if the total sum in the fenwick tree is approximately equal to the total sum of the segment masses, or if it diverges over time. If they stay equal to within a reasonable relative accuracy we're probably OK.

This should be checked in `verify`, have you had any luck in identifying drift with that? Or does it just take so long that turning on `verify` is hopeless?

If this is purely in the Fenwick tree then I think it shouldn't be too bad to isolate those changes. It does make me wonder if there's a simpler way to achieve the end of result of just trying to uniformly pick a segment based on the mass.

molpopgen:
> Casting back to ints is looking more attractive all the time

Makes sense.  There's no substitute for knowing exactly what's happening.

jeromekelleher:
> This should be checked in verify, have you had any luck in identifying drift with that? Or does it just take so long that turning on verify is hopeless?

Yeah - if you turn on verify it would take months to trigger the assertion! Anyway, it's hard - how do we know what's a significant difference? I don't know what the arbitrary threshold of relative difference is. If we set it pretty low, then maybe running a slightly longer simulation will exceed this threshold.

> If this is purely in the Fenwick tree then I think it shouldn't be too bad to isolate those changes. It does make me wonder if there's a simpler way to achieve the end of result of just trying to uniformly pick a segment based on the mass.

The Fenwick tree was the only way I could find that would do this in log time and do the maintenance in log time. There are absolutely gajillions of segment updates and lookups in a large simulation, so it's entirely performance critical and it'll be pretty hard to beat the Fenwick tree (which is a beautiful structure, I've been reminded!).

I've already spend longer debugging these problems that I would have spent changing the masses to integers (I think), and I've lost confidence in the code now. I want to be certain that the output of msprime is valid regardless of how large the simulation is and I don't have that certainty now. So, I'll have a go at scaling to int64s.

jeromekelleher:
> Hm, ok. Let me make sure I know what's going on here. The Fenwick tree is used to pick a random segment proportional to length, for recombination, right? Does it have any other purpose? And, how are the segments ordered in the tree (like, where do newly inserted ones go)?

The tree is used to choose segments proportional to their recombination mass (length * rate), yes. It doesn't serve any other purpose. The "tree" is pretty arbitrary, it's really just a way of keeping a running sum a list of things efficiently. The position in the tree is determined by the segment's ID, which is arbitrary (IDs are assigned sequentially from 1 as they are allocated. A segment's ID stays the same even though it is reused many times. Think of it as a position in memory rather than a program semantic ID.) Segments are "removed" from the tree by setting their weight to zero. When we randomly choose a mass value from 0 to the total mass across all segments, then Fenwick tree allows us to find this segment in log time. 

But it's not working like this with doubles I think. For small simulations it seems to be working fine, but there's weird stuff happening with these bigger simulations. The initial bug was that we were choosing a segment that had mass 0 assigned to it because numerical jitter had left it with a small positive value within the tree and our random value happened to be a very small positive value. I've fixed this issue, but hit a slightly different one later in the simulation. 

As I say, I've lost confidence in this now. I'd rather take the discreteness on up front and deal with it using 64 bit ints, than fix these worrying problems and hope that no other unforeseen numerical weirdnesses happen.

jeromekelleher:
@daniel-goldstein I've had a go at discretising the mass values by updating the recombination map object in the  algorithms.py file. This is the least invasive way of doing this I could think of. Basically, we change the RecombinationMap to think about masses as scaled masses, and it knows about the scaling factor needed to turn those into real rates. Discretisation errors should be pretty well localised then, and we are guaranteed the fenwick tree will always work properly.

Can you see any pitfalls from doing it this way?

petrelharp:
> So, I'll have a go at scaling to int64s.

Sounds good!

daniel-goldstein:
> One thing that I'm wary of is sample_poisson, which acts under the assumption that there exists some recombination mass to the right of a given starting point. Do the semantics change if suddenly these very small but extant zones of recombination vanish?

Ah, I really do come back to haunt myself, don't I.
https://github.com/tskit-dev/msprime/blob/46de3e080f94e954f3918d5bfe5ad765c1d55001/lib/msprime.c#L4144-L4149

`sample_poisson` is also not used in `algorithms.py` but I think that's because a lot of `dtwf` isn't.

jeromekelleher:
Great, thanks @daniel-goldstein - I'll push the changes through and see how it goes.

jeromekelleher:
An update here: I've pushed through the changes needed to store recombination mass values as scaled up integers, and it seems to work well so far (well, the tests are passing anyway, and the statistical checks look OK). I need to explore the limits and see what happens at the extremes, but if anyone wants to take a look, comments would be much appreciated.

The DTWF and gene conversion simulations are currently broken. Should be easy enough to fix.

jeromekelleher:
> The DTWF and gene conversion simulations are currently broken. Should be easy enough to fix.

I should really stop saying things like this :(

jeromekelleher:
OK, phew! I think this is nearly fixed now, and it'd be good to get your eyes on it please @daniel-goldstein. There's still a few tests to be patched up etc, but everything seems to work now except for some DTWF corner cases (which I think we'll need to address in a follow up).

jeromekelleher:
I *really* have to stop saying this is fixed! Turns out the original example is triggering assertions now. But, at least it's triggering them immediately and in every case rather than after a few hours in rare cases, so that's progress.

jeromekelleher:
I think this is finally ready for review. There are a few issues that need to be followed up on, but this should fix the vast majority of the problems with recombination in the current code.

The conversion to using integers to represent recombination masses didn't go as smoothly as anticipated. The idea of scaling recombination masses into a given integer range works well enough, but the problem turns out that we don't have a set maximum (except in the case of the SMC). If R is the total recombination rate, then we can have substantially more than nR total recombination mass if we have a long genome because of trapped ancestral material. Quite how much trapped ancestral material we can have was a surprise to me - the overall recombination rate can end up thousands of times higher than the original rate before the CA/RE equilibrium is reached! Hence, I've chosen an initial size of a 48 bit number, giving room for this to grow to 62 bits. I'm sure it's possible to overflow this with a sufficiently long genome, but we'll just need to experiment to see what's practical.

We could come up with better ways of doing this (maybe have a dynamic strategy, where we scale the maximum up as we overflow), but this will set the basic infrastructure in place and solve the outstanding issue for the vast majority of cases.

jeromekelleher:
@daniel-goldstein, would you mind taking a look please?

@petrelharp, have you followed the basic problem and the solution here?

petrelharp:
>  have you followed the basic problem and the solution here?

Only from a distance. This sounds reasonable, from a distance. And, that amount of trapped ancestral material is quite surprising! I suppose this is the source of a lot of the uptick in run times for big genomes.

I'll try to have a look soon. How exactly does the discretization work? I guess it's got to be doing floor(recomb_mass * 2^48) somewhere, so if you have a region of genome with less than 2^{48} recombination per bp somewhere, then you'll have adjacent nucleotides every so often that can't recombine apart? (which seems OK, but maybe good to document somewhere)

lib/msprime.c
```diff
@@ -2171,42 +2230,127 @@ msp_dtwf_recombine(msp_t *self, segment_t *x, segment_t **u, segment_t **v)
...
-        k = recomb_map_mass_to_position(&self->recomb_map, y->right_mass - (t - h));
-    } while (y->left >= k && y->prev == NULL);
+        /* Choose a scaled recombination mass uniformly from the total and find the
+         * segment y that is associated with this *cumulative* value. Note that
+         * although the scaled masses are stored as integers we need to compute
+         * the random value as a double because gsl_randint does not generate
+         * 64 bit random values. */
+        random_mass = gsl_ran_flat(self->rng, 1, (double) fenwick_get_total(tree));
+        int_random_mass = (int64_t) round(random_mass);
+        assert(int_random_mass <= fenwick_get_total(tree));
```
petrelharp, line 2281: should this be `<` not `<=`?

jeromekelleher:
> Only from a distance. This sounds reasonable, from a distance. And, that amount of trapped ancestral material is quite surprising! I suppose this is the source of a lot of the uptick in run times for big genomes.

Yes, that's a good point. Maybe there's some calculations from Wiuf and Hein that can help us out here.

> How exactly does the discretization work? I guess it's got to be doing floor(recomb_mass * 2^48) somewhere, so if you have a region of genome with less than 2^{48} recombination per bp somewhere, then you'll have adjacent nucleotides every so often that can't recombine apart? (which seems OK, but maybe good to document somewhere)

Yes, basically. I haven't thought through the exact consequences, but the thinking is that we have 2^48 units of recombination mass. I guess these units won't fall neatly on base pair boundaries, and so we will get bases which cannot recombine in regions with non-zero rate.

petrelharp:
It'd be good to get the exact details and consequences written down. I've just spent far too long trying to think through how the last version of recombination over [here](https://github.com/tskit-dev/pyslim/issues/61).

daniel-goldstein:
This looks great! Made it most of the way through the C code, and will get through the rest and give a more careful look tomorrow. I don't quite understand what is meant by trapped ancestral material and the consequences of such so @petrelharp's request for an elaboration would help me understand the pitfalls more.

petrelharp:
"Trapped ancestral material" is when there are two disjoint ancestral segments in a single ancestor - say, if node 0 inherited [0,1) from node 2, while node 1 inherited [99, 100) from node 2; then since we need to know if the two segments are inherited together or not, we need to keep track of the whole segment [0, 100), at least until they recombine apart.

jeromekelleher:
Trapped ancestral material is what makes the coalescent with recombination hard @daniel-goldstein. Suppose you have two segments initially, (0, L, a), (0, L, b). The total recombination mass now is 2rL. Then, let's say some recombination events happen, and we have 
```
(0, 1, a) 
(1, L - 1, a)
(L - 1, L, a)
(0, L, b)
```
The total recombination mass is still 2rL. But, suppose we have a common ancestor event so that we get 
```
(0, 1, a), (L - 1, a)
(1, L - 1, a)
(0, L, b)
```
Now the total recombination mass is > 2rL. In principle, the total recombination mass can become much larger than nrL, and it does seem to for large L.

jeromekelleher:
After thinking about this more, I think we'll have to do an adaptive approach. Unless we know some good bounds on the maximum amount of trapped material in the coalescent, I think we have to, as it'll always be possible to overflow the current limits with a sufficiently large simulation. I think the following approach will be easy enough to implement:

- Start with 2^62 units of recombination mass mapped to nrL and run the simulation as we do now.
- When we hit 2^63 units of recombination mass (i.e., it has doubled), we rescale so that the new maximum is mapped to 2^62. This will mean going though all the extant segments and rebuilding the Fenwick tree. This is a bit of a cost, but it's not that big a deal. We don't expect the amount of ancestral material to double many times, so overall it'll be a negible cost.

Does this sound like a reasonable approach @petrelharp? I think this is basically the level of precision that we'd get from using doubles anyway, except we're distributing the bits uniformly rather than across the vast space that's possible in floating point. I'll think through the details fully and document this with an example, but I just want to see if you can see any immediate problems first.

petrelharp:
Well, let's see: the main thing is that total recombination mass is preserved (when appropriate). You've got it set up already so that if a segment [a, c) splits into [a, b) and [b, c) then the mass of the sub-segments is equal to the mass of the original segment, even after scaling and converting to integers (I think?).

And, I *think* something simliar should be true when changing between rescaling factors. Let's say that we double the rescaling factor (or, do I mean half it?), and the integer mass of some segment changes from x to y. If x != 2y then the difference, (x - 2y), should go somewhere - for instance, at least if you double the rescaling factor, the total mass in the tree should half?  The main thing we want to be true is that the total mass of a chromosome shouldn't change, no matter how many blocks it's broken up into.

daniel-goldstein:
@jeromekelleher Had a couple look-throughs and looks really solid. The explanation helped and seems like the conversions back and forth are well-handled. I'm mainly still not confident about how this plays with breakpoint sampling for DTWF, but like you said better to leave that to a smaller follow-up.

jeromekelleher:
Abandoned in favour of #1108. Changing to integers just introduced too many problems because there's no fixed upper bound on total recombination mass in practise.

petrelharp:
Argh, what a pain.

jeromekelleher:
Yeah, this has been "fun" all right: simulating recombination is actually pretty hard as it happens! Happily #1108 seems to be working. :crossed_fingers:

PR #1080, opened 2020-06-24
Better debugging tools for simulations

jeromekelleher:
Cause I need them for #1079.


jeromekelleher:
Going to merge this so I can use it. New logging stuff should be uncontroversial enough, and it's marked as an "advanced feature" so I think it's fine.

PR #1086, opened 2020-07-03
Refactored verification.py.

jeromekelleher:
Building on the heroic work from @saunack to tidy up the verification.py script, I've made a significant update. It's now organised in a way a bit like unittest, where we organise the tests into classes, and each test is a method that starts with "test", and is intended to be executed independently of all others.

There's still some rough edges, but it's a big improvement I think. There's a few tests in there that I'm not clear about, but I think I'll follow these up in issues rather than in review.

I'll update this with some documentation for developers soon.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -214,7 +214,7 @@ Meson is best installed via ``pip``:
 
     $ python3 -m pip install meson --user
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Gets through all 141 tests in 15 minutes on my machine now, woohoo! This is going to make my life so much easier.

PR #1097, opened 2020-07-09
Update verification.py

TPPSellinger:
fixes the XivsHudson test

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -1635,12 +1635,10 @@ class RecombinationMutationTest(Test):
         )
 
 
-# FIXME These tests are miles off at the moment. There are *way* more
-# nodes edges and trees in the Dirac model than Hudson.
 class XiVsHudsonTest(Test):
     """
-    Test that Xi coalescents are equivalent to the Hudson model in the
-    appropriate regimes.
+    Test that Xi dirac coalescent is equivalent to the Hudson model in the
+    appropriate regime.
     """
 
     def _run(self, xi_model, **kwargs):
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Looks good to me @TPPSellinger. The checks are complaining about a minor formatting issue. See the [documentation](https://tskit.readthedocs.io/en/latest/development.html#pre-commit-checks) for more details on this (there's info on how to install pre-commit too).

verification.py
```diff
@@ -1651,6 +1649,7 @@ def _run(self, xi_model, **kwargs):
             model_str = "hudson"
             if model != "hudson":
                 model_str = "Xi"
+                simulate_args["Ne"] = 2*(math.sqrt(int(simulate_args["Ne"])))
```
jeromekelleher, line 1652: A comment here would be helpful I think. "We need to rescale Ne by blah because blah blah".

TPPSellinger: Okay. I hope the comment is clear.

jeromekelleher: Perfect. Just need to get the formatting fixed now and we're good. All you really need to do is run "black" on the code, like
```
python3 -m pip install black
python3 -m black verification.py
git commit -a --amend
```
and that should do it.

TPPSellinger: Thanks ! Sorry  for all the trouble, as  I thought is was a small commit I went through the GitHub UI.

TPPSellinger:
Everything is fine (or some squashing is expected) ?

jeromekelleher:
It would be nice to squash them all right @TPPSellinger  - 6 commits is a bit more than we want for this change!

PR #1098, opened 2020-07-09
Update verification.py

TPPSellinger:
fixes the XivsHudson tets. I hope this does the trick.


TPPSellinger:
Let me know if this does the trick  @jeromekelleher . Thanks again for all your help !

PR #1100, opened 2020-07-10
Demography interchange with JSON

jeromekelleher:
It would be useful to define a non-programmatic interface to allow external programs describe an msprime demography to condition an msprime simulation on. For example, you might simulate a complex demography with a grid of populations with population sizes and migration rates changing over time using [Quetzal](https://becheler.github.io/pages/quetzal.html), and then run a simulation through this randomly generated demography.

This is somewhat similar to what we're talking about in #1058 , but more focused on the problem of  interchanging this information between programs rather than allowing humans describe a complicated demography easily.

Here's an initial proposal for this using JSON, and what it looks like:
```python
N = 5
model = msprime.Demography.stepping_stone_1d(N, migration_rate=0.1, circular=False)
for j in range(20):
    model.events.append(
        msprime.PopulationParametersChange(time=j + 1, population=j, initial_size=10))

s = model.asjson()
with open("demography_example.json", "w") as f:
    print(s, file=f)

other_model = msprime.Demography.fromjson(s)
assert other_model == model
```
And the json:
```json
{
  "format": "msprime-demography-json_1.0",
  "populations": [
    {
      "initial_size": 1,
      "growth_rate": 0,
      "name": null,
      "description": null,
      "temporary_hack_for_metadata": null
    },
    {
      "initial_size": 1,
      "growth_rate": 0,
      "name": null,
      "description": null,
      "temporary_hack_for_metadata": null
    },
    {
      "initial_size": 1,
      "growth_rate": 0,
      "name": null,
      "description": null,
      "temporary_hack_for_metadata": null
    },
    {
      "initial_size": 1,
      "growth_rate": 0,
      "name": null,
      "description": null,
      "temporary_hack_for_metadata": null
    },
    {
      "initial_size": 1,
      "growth_rate": 0,
      "name": null,
      "description": null,
      "temporary_hack_for_metadata": null
    }
  ],
  "migration_matrix": [
    [
      0,
      0.1,
      0,
      0,
      0
    ],
    [
      0.1,
      0,
      0.1,
      0,
      0
    ],
    [
      0,
      0.1,
      0,
      0.1,
      0
    ],
    [
      0,
      0,
      0.1,
      0,
      0.1
    ],
    [
      0,
      0,
      0,
      0.1,
      0
    ]
  ],
 "events": [
    {
      "type": "population_parameters_change",
      "time": 1,
      "population": 0,
      "initial_size": 10
    },
    {
      "type": "population_parameters_change",
      "time": 2,
      "population": 1,
      "initial_size": 10
    },
    {
      "type": "population_parameters_change",
      "time": 3,
      "population": 2,
      "initial_size": 10
    },
    {
      "type": "population_parameters_change",
      "time": 4,
      "population": 3,
      "initial_size": 10
    },
    {
      "type": "population_parameters_change",
      "time": 5,
      "population": 4,
      "initial_size": 10
    }
  ]
}
```
I've pretty-printed this using jq, so it's quite a long-form. There's also a bunch of optional stuff in the Population object I haven't bothered removing.

It's not the most efficient possible way of encoding this, but it is quite simple and would probably serve as a reasonably good interchange for all but the largest models (JSON parsing is surprisingly efficient).

Thoughts?

cc @Becheler

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -189,6 +189,30 @@ class Demography:
     def asdict(self):
         return attr.asdict(self)
 
+    def asjson(self):
+        """
+        Returns a JSON encoding of this demography.
+        """
+        M = np.array(self.migration_matrix)
+        data = {
+            # TODO do this properly with mimetype string or something.
+            "format": "msprime-demography-json_1.0",
+            "populations": [pop.asdict() for pop in self.populations],
+            "migration_matrix": M.tolist(),
+            "events": [event.asjson() for event in self.events],
+        }
+        return json.dumps(data)
+
+    @staticmethod
+    def fromjson(json_demography):
+        data = json.loads(json_demography)
+        populations = [Population(**pop) for pop in data["populations"]]
+        migration_matrix = np.array(data["migration_matrix"])
+        events = list(map(parse_json_event, data["events"]))
+        return Demography(
+            populations=populations, migration_matrix=migration_matrix, events=events
+        )
+
     def debug(self):
         return DemographyDebugger(self)
 
@@ -498,6 +522,19 @@ class PopulationConfiguration:
         )
 
 
+def parse_json_event(json_event):
+    type_map = {
+        "population_parameters_change": PopulationParametersChange,
+        "migration_rate_change": MigrationRateChange,
+        "mass_migration": MassMigration,
+        # TODO add the other events
+    }
+    # Don't modify the input
+    json_event = dict(json_event)
+    type_str = json_event.pop("type")
+    return type_map[type_str](**json_event)
+
+
 @attr.s
 class DemographicEvent:
     """
@@ -505,6 +542,9 @@ class DemographicEvent:
     """
 
     time = attr.ib()
+
+    def asjson(self):
+        return self.get_ll_representation()
 
     def asdict(self):
         return {


Becheler:
Wow great, thank you, it will help me a lot !

I don't understand several aspects of the form, but I guess it's my lack of documentation/msprime culture. 
For example, if N is the number of demes, I would expect and N*N migration matrix for a generation. I know the example was a first proposal, I know there is somewhere in the code an allocation to a pop_num*pop_num for initializing a migration matrix, but it's just to be sure we are talking about the same thing ? 

Also, I'm not sure about the exact meaning of the information stored in the migration matrix (what you call migration rate). If e(x,y,t) is the emigration rate from deme x to y at time t, does the sum of e(x,y,t)*N(x,t) over y sum to N(x,t) ? 

On a second note, I get that we are talking about non-programming interface, but i had just a random thought on ... programming interface (ofc) that I am throwing here ;) Feel free to move it/ delete it if not relevant :)

Demographic informations are accessed in the Quetzal coalescence module using function objects abstractions, with a syntax close from:
N(x,t) to access the population size at time x at time t
M(x,y,t) to access the effective number of migrants from deme x to deme y at time t
In C++, function object abstractions are lambda functions or any class implementing the function call operator. In C I think we only have function pointer to represent this concept.

So, I don't know too much the internal of msprime (and maybe i'm wrong about the migration_matrix use) , but I guess that the following steps might be an easy step forward towards a lower level library:
- locate the 2 (?) points at which the coalescence process access these demographic informations to compute (i) the probability of coalescence and (ii) the backward transition probabilities
- replace the bracket operator call of the migration_matrix to a pointer of function, and similar thing with the population size
- that would decouple the demographic data representation from the coalescence process code, providing a reasonably simple/easy to maintain programming interface (at the cost of dereferencing the pointer though)
- in the msprime client code, this function can simply be the bracket operator of the migration_matrix
- with quetzal client code it can be a member function of the migration_history class.

Just throwing that here :)
Thanks again for opening this conversation :)

jeromekelleher:
Thanks for the feedback @Becheler! Regarding the migration rates, I think the best resource for explaining this is our tutorial [book chapter](https://link.springer.com/protocol/10.1007/978-1-0716-0199-0_9). We've also tried to be precise about the model [here](https://msprime.readthedocs.io/en/latest/api.html#simulation-model).

For the migration matrix, yes, we have an NxN matrix which is in the JSON above (it does look a bit crappy because of the formatting). We declare the initial state of this matrix at the start of the simulation (present time), and then modify it via [MigrationRateChange](https://msprime.readthedocs.io/en/latest/api.html#msprime.MigrationRateChange) events. The underlying assumption is that the matrix doesn't change too much, over time. If it is totally different every generation, then this model won't work well.

Re internal C API changes and integration - I'm afraid that's something that just out of scope for the project. Honestly, it's hard enough to keep things still developing and backward compatible with just having to keep track of the Python API. Version 1.0 would be the last version of msprime if we had to keep the C API stable as well. I've tried to make this a community platform that lots of people can work on and extend, while still keeping it efficient and easy to install. Some compromises have to be made, I'm afraid.

jeromekelleher:
I'm going to close this as it's basically made redundant by the upstream [demes](https://github.com/popsim-consortium/demes-python) project.

Becheler:
Ok! I've been caught by other stuff, but I'll sure keep an eye on the demes project for interfacing quetzal in a nearish future.

PR #1102, opened 2020-07-24
Fix codecov issues with new versions of meson

jeromekelleher:


jeromekelleher:
@benjeffery, can you approve this please so it can be automerged? ta.

PR #1104, opened 2020-07-24
wrote discoal command line to msprime simulation, other assorted bits…

andrewkern:
… on verification.py

getting all the bits of the verification for sweeps using discoal in place...

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -683,22 +683,17 @@ class DiscoalTest(Test):
         for i in range(3, len(tokens)):
             # pop size change case
             if tokens[i] == "-en":
-                logging.error(
-                    "sweeps with population size changes remain\
-                              unimplemented"
+                raise ValueError(
+                    "sweeps with population size changes remain unimplemented"
                 )
-                return
             # migration rate case
             if (tokens[i] == "-m") or (tokens[i] == "-p"):
-                logging.error(
-                    "sweeps with multiple populations remain \
-                              unimplemented"
+                raise ValueError(
+                    "sweeps with multiple populations remain unimplemented"
                 )
-                return
             # split or admixture case
             if (tokens[i] == "-ea") or (tokens[i] == "-ed"):
-                logging.error("sweeps with splits or admixture not supported")
-                return
+                raise ValueError("sweeps with splits or admixture not supported")
             # sweep params
             if tokens[i] == "-x":
                 sweep_site = float(tokens[i + 1])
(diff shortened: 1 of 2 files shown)


andrewkern:
i know @jeromekelleher will be out of earshot for a bit and that's fine-- i'm getting this up here for @gbinux to see and will continue to work on the discoal verification through this PR over the next week.

andrewkern:
lol added analytic comparison of sojourn times and i already see a factor of two error 😭

gbinux:
Is this the synptom of "a factor of two error"?
D
pi
ss

andrewkern:
these must be for sweeps? then yes, probably. i will have to find this bug.

the neutral comparisons between msprime and discoal look 100% ok

andrewkern:
btw @gbinux the factor of two error that i'm talking about is from `test_sojourn_time` in the new `SweepAnalytical` test class

gbinux:
Thanks @andrewkern. I just ran `python3 verification.py -c DiscoalSweeps` with the new version of verification.py in this PR. The above picutures were from the `test_sweep_ex1` folder. I guess we are refering to different things then.

 I tried to grep `SweepAnalytical`  in the root folder but did not find any .py files containing `SweepAnalytical`.  If the new `SweepAnalytical class` related to what I am supposed to work with, may I have peek at it?

andrewkern:
sorry just saw your comment. the class is here https://github.com/andrewkern/msprime/blob/66972a7d213a0625d5dc3cdb0f370726dc924755/verification.py#L825

gbinux:
Never mind. I saw your last commit when you submitted it to the PR number. :).

verification.py
```diff
@@ -669,8 +671,88 @@ def _run_mutation_discoal_stats(self, args):
...
+        nreps = int(tokens[1])
+        seq_length = int(tokens[2])
+        # parse discoal command line for params
+        # init ones we definitely need for comparison
+        theta = rho = alpha = sweep_site = sweep_mod_time = None
+        refsize = 1e6
+        for i in range(3, len(tokens)):
+            # pop size change case
+            if tokens[i] == "-en":
+                logging.error(
```
jeromekelleher, line 690: I think these should be exceptions - just ``raise ValueError(...)`` instead of calling logging.error and returning.

andrewkern:
given that #1109 was merged, should we merge this PR and then deal with the scaling?

verification.py
gbinux: > I think these should be exceptions - just `raise ValueError(...)` instead of calling logging.error and returning.

I think this might be change request before merging the PR. @andrewkern

andrewkern:
ah missed that. looks like i need to rebase this PR too

jeromekelleher:
Yes, looks good @andrewkern - you just need to rebase and change the logging.errors() to ValueErrors.

jeromekelleher:
@andrewkern - any chance you could finish this up please? The changes are minor and it'll just get more out of date the longer it stays here.

jeromekelleher:
I just pushed a few minor updates here to address the issues so we can merge @andrewkern.

jeromekelleher (approved):

PR #1108, opened 2020-08-04
Alt fix recombination bug

jeromekelleher:
Here's an alternative method to #1079 I've come up with to try and solve the nasty numerical problems with recombination in #1046. The mapping-to-integers method I tried in #1079 was just coming up with more and more problems, so I thought I'd have another go at working in floats.

I *think* the only real problem we have is that the Fenwick tree is drifting from the true values, which we can solve by periodically rebuilding it. Let's see what happens when we try it on some big sims though.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -1951,7 +1951,7 @@ MatrixMutationModel_init(MatrixMutationModel *self, PyObject *args, PyObject *kw
             PyErr_SetString(PyExc_TypeError, "alleles must be unicode strings");
             goto out;
         }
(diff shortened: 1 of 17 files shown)


jeromekelleher:
Having been wrong about this so many times before, I really hesitate to say this, but I think this does fix the numerical issues with recombination. I've run a few hundred replicates of the original simulation that provoked the issue in  #1046 and they seem to be running OK.

I've added in some verification tests where we compare with the released version of msprime, and that all looks good too (although we could definitely do more with this!).

I think I'll merge this tomorrow morning if the test simulations are still running, as it's at least better than what we have currently.

PR #1109, opened 2020-08-06
Add MsmsSweeps test class in verification.py

gbinux:
An initial version of `MsmsSweeps` test class in verification.py.  
I will update this class when the genic sweep model or related APIs are updated.

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -761,14 +761,14 @@ class DiscoalSweeps(DiscoalTest):
 
 class MsmsSweeps(Test):
     """
-    Compare Msms with msprime/discoal for selective sweeps.
+    Compare msms with msprime/discoal for selective sweeps.
 
     NOTE:
     1. As msms and msprime use different numbers of seeds, the specified
     seed(s) in the testing cmd string will be ignored.
     2. Msms allows user to specify selection starting time/frequency (-SI),
     or, alternatively, specify selectiy selection ending time/frequency (-SF);
-    msprime is able simulate selection similar to the '-SF' option in msms
+    msprime is able to simulate selection similar to the '-SF' option in msms
     3. Msms allows user to specify different selection coefficient for AA and Aa,
     but in msprime only selection coefficient for AA can be specified, and
     use h=0.5 to calculate that for Aa.
@@ -776,7 +776,7 @@ class MsmsSweeps(Test):
 
     def _msms_str_to_parameters(self, msms_cmd):
         """
-        parse msms cmdline arguments into a dictionary.  This method is called
+        Parse msms cmdline arguments into a dictionary.  This method is called
         by `_run_msp_sample_stats`
 
         msms cmdline pattern:
@@ -792,6 +792,7 @@ class MsmsSweeps(Test):
         num_sweeps = 0
         sAA = saA = sel_pos = -1.0
         saA = -0.5
+        refsize = 1
         rand_seed = (random.randint(1, 2 ** 16),)
 
         # parse arguments
@@ -836,7 +837,7 @@ class MsmsSweeps(Test):
             "num_sweeps": num_sweeps,
             "end_time_lst": end_time_lst,
             "end_frequency_lst": end_frequency_lst,
-            "Ne": refsize,
+            "refsize": refsize,
             "alpha": sAA,
             "theta": theta,
             "rho": rho,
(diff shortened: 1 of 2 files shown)


verification.py
```diff
@@ -758,6 +759,271 @@ def test_sweep_ex1(self):
...
+
+    def _run_msms_vs_discoal(self, cmd):
+        df_discoal = self._run_discoal_sample_stats(cmd)
+        df_msms = self._run_msms_sample_stats(cmd)
+        self._plot_stats("discoal_msms", df_discoal, df_msms, "discoal", "msms")
+
+    def test_Neutral_msms_vs_msp(self):
+        self._run_msms_vs_msp("100 300 -t 200 -r 200 500000 -N 10000")
+
+    def test_Selective_msms_vs_msp(self):
```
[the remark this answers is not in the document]
jeromekelleher: I've created an issue to track this: #1111, can you mention this in the comment please?

verification.py
```diff
@@ -758,6 +759,271 @@ def test_sweep_ex1(self):
...
+            Ne_temp = params["Ne"]
+
+        ts = msprime.simulate(
+            sample_size=params["nsam"],
+            Ne=Ne_temp,
+            length=params["num_sites"],
+            mutation_rate=mutation_rate,
+            recombination_rate=recombination_rate,
+            model=model,
+            random_seed=random.randint(1, 2 ** 16),
```
jeromekelleher, line 889: There's no need to provide seeds here, msprime will generate good seeds on its own.

verification.py
```diff
@@ -758,6 +759,271 @@ def test_sweep_ex1(self):
...
+            "alpha": sAA,
+            "theta": theta,
+            "rho": rho,
+            "num_sites": num_sites,
+            "sel_pos": sel_pos,
+            "rand_seed": rand_seed,
+        }
+
+    def _msms_parms_to_run_msp(self, params):
+        """ run smulation for a single sample and return a tree sequence.
```
jeromekelleher, line 849: This isn't quite formatted in the usual way, can you change to
```
"""
Runs a single simulation and returns a tree sequence.
"""

File: verification.py [hunk 4]
```diff
@@ -758,6 +759,271 @@ def test_sweep_ex1(self):
...
+            this method is called by `_run_msp_sample_stats` in a loop to
+            generate nrep samples.
+        """
+        Ne_temp = None
```
jeromekelleher, line 853: I don't see what Ne_temp is for. I think it would be simpler to just set Ne=0.25 throughout, unless we're explicitly setting population sizes from the msms command line. It's confusing to have ``params["Ne"] and also this Ne parameter. is there a way we can simplify this?

verification.py
```diff
@@ -758,6 +759,271 @@ def test_sweep_ex1(self):
...
+            # the scaling is mimicing verficatin.py - DiscoalTest class -
+            # _discoal_str_to_simulation methods.
+            recombination_rate = params["rho"] / (params["num_sites"] - 1)
+            mutation_rate = params["theta"] / params["num_sites"]
+            Ne_temp = 0.25
+
+        else:
+            model = "hudson"
+            # To circumvent huson/genicselection scaling inconsistency
+            recombination_rate = params["rho"] / params["num_sites"] / 4 / params["Ne"]
```
jeromekelleher, line 878: It'd be good to factor the scaling factor out here rather than repeating the same calculation twice. Also, I find it difficult to read things like ``1 / n / x`` isn't that said more plainly as x / n?

verification.py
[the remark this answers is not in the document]
gbinux: @jeromekelleher I am glad you caught this. This happens randomly but pretty frequently. In my hand, it could happen in any of the tests involved with the msprime sweep model. I can try to use smaller numbers in the cmd line to see if this can go away for now.

verification.py
gbinux: I know this is confusing. I was also confused too. For Hudson model, I need to use Ne= refsize, but for sweep model, I need to use Ne=0.25. I will re-organize this part to make it easier to see the difference.

gbinux:
@jeromekelleher I have updated the code accordingly. To avoid test crashes, I tried to use smaller numbers in cmd parameters. It indeed reduces the probabilities of crashes but the error still happens. So I just hide all of the selective sweep tests until the bug is fixed.

jeromekelleher:
Fantastic, thanks @gbinux! Merging.

PR #1110, opened 2020-08-06
Initial ploidy variable

JereKoskela:
Ping @jeromekelleher and @andrewkern.

This is a possible starting point for a ploidy variable. I've put it in as a simulator (`msp_t`) parameter under the assumption that nearly all (or actually all?) conceivable simulations will be exclusively haploid or exclusively diploid. The alternative would be to give each `population_t` its own ploidy variable, which could facilitate simulating populations with different ploidies.

There is also already a ploidy variable in the `pedigree_t` struct. I'm completely unfamiliar with pedigrees and don't know if the variables can clash somewhere, except that current tests don't pick it up if they do.

I've made ploidy default to 2 to match the current implementation. That way everyone's scripts will keep working as they do at present.

There is no checking yet that the variable takes sensible values. We also need to decide what range of permissible values we want. In principle, there is no reason we can't have, say, hexaploid organisms by setting ploidy = 6, at least in the Hudson model, but at least it should be be a positive integer. Do we want to restrict it further to, say, just 1 or 2?

Test coverage and a docstring still need to come, but I thought we should settle on the basic implementation first. Once we have that pinned down, I'll also update the multiple merger coalescents to allow monoecious, haploid reproduction. As usual, variable ploidy is a bit more involved for those models, with both ploidy and the number of parents being important.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -230,6 +230,13 @@ msp_set_store_full_arg(msp_t *self, bool store_full_arg)
 }
 
 int
+msp_set_ploidy(msp_t *self, uint32_t ploidy)
+{
+    self->ploidy = ploidy;
+    return 0;
+}
+
+int
 msp_set_dimensions(msp_t *self, size_t num_populations, size_t num_labels)
 {
     int ret = 0;
@@ -555,6 +562,8 @@ msp_alloc(msp_t *self, size_t num_samples, sample_t *samples, recomb_map_t *reco
     self->state = MSP_STATE_NEW;
     /* Set up pedigree */
     self->pedigree = NULL;
+    /* Set default to diploid */
+    self->ploidy = 2;
 out:
     return ret;
 }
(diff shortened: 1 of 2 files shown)


jeromekelleher:
This is excellent, thanks @JereKoskela! I'm slightly wary of the high-level Python changes and want to go through the consequences carefully. One thing that we could do is just include the stuff in the C library and fix ``self->ploidy = 2``, then push the other stuff off to a follow up PR? That way we could get this merged right away, since there's no interface changes.

JereKoskela:
> This is excellent, thanks @JereKoskela! I'm slightly wary of the high-level Python changes and want to go through the consequences carefully. One thing that we could do is just include the stuff in the C library and fix `self->ploidy = 2`, then push the other stuff off to a follow up PR? That way we could get this merged right away, since there's no interface changes.

Works for me @jeromekelleher. Do you want me to leave this PR open for reference and start a different one, or shall I overwrite this to just cover the changes in C?

jeromekelleher:
Probably easiest to overwrite this one I'd imagine - sorry, I know it's annoying to undo work. I'll follow up with a proposal for the high-level stuff, because there's some tricky things I want to do with allocating individuals etc, and I'm not entirely sure about how to do it.

JereKoskela:
> Probably easiest to overwrite this one I'd imagine - sorry, I know it's annoying to undo work. I'll follow up with a proposal for the high-level stuff, because there's some tricky things I want to do with allocating individuals etc, and I'm not entirely sure about how to do it.

No problem - I expected individuals and pedigrees might cause problems. I've reduced the update to just the C code. I've left in the `ms_set_ploidy` method for future infrastructure, but it doesn't yet get called by anything: `self->ploidy` is hard-coded to 2 in `msp_alloc'.

jeromekelleher:
Merged, thanks @JereKoskela!

PR #1113, opened 2020-08-06
Support ploidy in simulate and add individuals

### LINKED ISSUE #557: Add ploidy argument to simulate and allocate individuals

jeromekelleher:
Here's a proposal for adding the ploidy argument to ``simulate`` and also inserting individuals to the table collection. @JereKoskela just cleared up the low-level calculations in #1110 which means that we should scale time correctly now for a given number of k-ploid individuals. I haven't tested this statistically, which I should do as part of this PR.

This PR also adds in k-ploid individuals into the table collection - which is a seriously welcome addition. There's only two controversial things here I think:

1. The ``sample_size`` argument still refers to the number of monoploid nodes (gametes, I guess?) **not** the number of k-ploid individuals (unless ploidy==1). This will be confusing, but I don't see how we can change this without breaking all sorts of code. I guess we could consider adding another ``sample_size`` keyword argument, but that seems clumsy. The only other alternative I can think of is that we have ``ploidy`` default to None, in which we keep the default diploid timescale but *don't* allocate any individuals. Then, if ploidy is explicitly set, the ``sample_size`` argument refers to the number of k-ploid individuals. It's not obvious to me which is more confusing, but not creating individuals by default feels like the wrong thing to do.
2. Currently if we ask for a number of samples that doesn't evenly fit into k-ploid individuals, we raise a warning. This seems like a good idea under the current formulation - but it's going to result in a lot of warnings which will confuse people.

Thoughts? 

This is based on #1108, so contains a bunch of other stuff right now (I'll hopefully merge #1108 tomorrow).

See #557 for more discussion (but some of it is out of date now).

Closes #557.

### DIFF
--- a/_msprimemodule.c
+++ b/_msprimemodule.c
@@ -3380,7 +3380,7 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
         "demographic_events", "model", "avl_node_block_size", "segment_block_size",
         "node_mapping_block_size", "store_migrations", "start_time",
         "store_full_arg", "num_labels", "gene_conversion_rate",
-        "gene_conversion_track_length", NULL};
+        "gene_conversion_track_length", "ploidy", NULL};
     PyObject *py_samples = NULL;
     PyObject *migration_matrix = NULL;
     PyObject *population_configuration = NULL;
@@ -3403,11 +3403,12 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
     double start_time = -1;
     double gene_conversion_rate = 0;
     double gene_conversion_track_length = 1.0;
+    int ploidy = 2;
 
     self->sim = NULL;
     self->random_generator = NULL;
     self->recombination_map = NULL;
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!O!O!O!|O!OOO!O!nnnidindd", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!O!O!O!|O!OOO!O!nnnidinddi", kwlist,
             &PyList_Type, &py_samples,
             &RecombinationMapType, &recombination_map,
             &RandomGeneratorType, &random_generator,
@@ -3421,7 +3422,7 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
             &avl_node_block_size, &segment_block_size,
             &node_mapping_block_size, &store_migrations, &start_time,
             &store_full_arg, &num_labels,
-            &gene_conversion_rate, &gene_conversion_track_length)) {
+            &gene_conversion_rate, &gene_conversion_track_length, &ploidy)) {
         goto out;
     }
     self->random_generator = random_generator;
@@ -3498,6 +3499,11 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
             gene_conversion_track_length);
     if (sim_ret != 0) {
         handle_input_error("set_gene_conversion_rate", sim_ret);
+        goto out;
+    }
+    sim_ret = msp_set_ploidy(self->sim, ploidy);
+    if (sim_ret != 0) {
+        handle_input_error("set_ploidy", sim_ret);
         goto out;
     }
 
@@ -3760,6 +3766,18 @@ Simulator_get_time(Simulator  *self, void *closure)
         goto out;
     }
     ret = Py_BuildValue("d", msp_get_time(self->sim));
+out:
+    return ret;
+}
+
+static PyObject *
+Simulator_get_ploidy(Simulator  *self, void *closure)
+{
+    PyObject *ret = NULL;
+    if (Simulator_check_sim(self) != 0) {
+        goto out;
+    }
+    ret = Py_BuildValue("i", self->sim->ploidy);
 out:
     return ret;
 }
@@ -4543,6 +4561,8 @@ static PyGetSetDef Simulator_getsetters[] = {
             "The tables"},
     {"time", (getter) Simulator_get_time, NULL,
             "The current simulation time" },
+    {"ploidy", (getter) Simulator_get_ploidy, NULL,
+            "The ploidy value used in scaling coalescent time." },
     {NULL}  /* Sentinel */
 };
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -247,10 +247,17 @@ msp_set_store_full_arg(msp_t *self, bool store_full_arg)
 }
 
 int
-msp_set_ploidy(msp_t *self, uint32_t ploidy)
+msp_set_ploidy(msp_t *self, int ploidy)
 {
-    self->ploidy = ploidy;
-    return 0;
+    int ret = 0;
+
+    if (ploidy < 1) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+    self->ploidy = (uint32_t) ploidy;
+out:
+    return ret;
 }
 
 int
@@ -3400,6 +3407,7 @@ msp_reset_from_samples(msp_t *self)
         }
     }
     /* Set up the sample */
+    tsk_ind = TSK_NULL;
     for (u = 0; u < (node_id_t) self->num_samples; u++) {
         if (self->samples[u].time <= self->start_time) {
             ret = msp_insert_sample(self, u, self->samples[u].population_id);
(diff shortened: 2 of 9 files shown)


petrelharp:
> I guess we could consider adding another sample_size keyword argument, but that seems clumsy

Gee, I think adding an `individual_sample_size` argument would be a nice way to do it?

> Currently if we ask for a number of samples that doesn't evenly fit into k-ploid individuals, we raise a warning.

I agree with this.

jeromekelleher:
Pinging @tskit-dev/all here for more opinions as this is a big decision. We can finally support k-ploid individuals, but it's not obvious what the best approach is. Here's three possibilities:

- ``msprime.simulate(10, ploidy=2)`` runs the diploid coalescent and allocates **5** individuals. If ploidy is 1 we get 10 individuals, and if ploidy is 3 we get 4 (3 and 1/3 triploids, effectively) and we raise a warning. Ploidy defaults to 2. This is what's currently implemented.
- ``msprime.simulate(num_sample_individuals=5, ploidy=2)`` creates 5 individuals and 10 nodes. It's an error to use the ``num_sample_individuals`` argument at the same time as the original ``sample_size`` arg. I guess we'd still have to have the sample behaviour as above when using the ``sample_size`` arg and ``ploidy``.
- An alternative to the above (and probably the cleanest approach) but potentially most confusing is to have the current ``sample_size`` argument's behaviour depend on the value of ``ploidy``. If ``ploidy`` is None (the default) we use the diploid time scale (to keep compatibility) and we create num_samples nodes with no individuals (the current behaviour). If ``ploidy`` is specified, then the ``num_samples`` argument becomes the number of sampled *individuals*, so we get ploidy * num_samples sample nodes in the returned tree sequence.

I think I prefer the last approach, except for the fact that we'll then return tree sequences without individuals by default, and it'll surely lead to errors and bugs in client code if they have to remember to specify ``ploidy`` just so their VCFs will work properly. So, on balance I'm leaning towards the first option, as having yet another way to specify samples seems problematic. I'm very much open to any other ideas though!

bhaller:
I have a question, which I'll preface by saying that I've never 100% understood how polyploidy really works.  My question is: if you simulate tetraploids, are you simulating autotetraploids or allotetreploids, or does it somehow work out to being the same thing for simulation purposes?

hyanwong:
Should we think about how the parameters interact with population specifications? For instance, if the sample size is the number of genomes (and not number of individuals), and we use the same terminology within populations, might we accidentally create an individual that belongs to multiple populations?

Having the sample number specify the number of individuals, rather than the total number of genomes, seems less likely to lead to this sort of problem.

petrelharp:
The `ploidy` option here is "how many chromosomes-that-might-recombine-with-each-other does each individual carry around". Some tetraploids have four of each chromosome, and any of them might pair with any others. Some (more likely the allotetraploids?) have two pairs, but we still call them tetraploids because the chromosomes are homologous; for simulation purposes this latter situation would count as diploid.

hyanwong:
Could we default to `ploidy=None` but issue a warning that this is deprecated and in future versions `ploidy` will become a required parameter?

bhaller:
> The `ploidy` option here is "how many chromosomes-that-might-recombine-with-each-other does each individual carry around". Some tetraploids have four of each chromosome, and any of them might pair with any others. Some (more likely the allotetraploids?) have two pairs, but we still call them tetraploids because the chromosomes are homologous; for simulation purposes this latter situation would count as diploid.

I see.  This would probably be a good thing to be explicit about in the doc; although maybe people who work with k-ploidy stuff would expect it anyway.

gbinux:
> * `msprime.simulate(10, ploidy=2)` runs the diploid coalescent and allocates **5** individuals. If ploidy is 1 we get 10 individuals, and if ploidy is 3 we get 4 (3 and 1/3 triploids, effectively) and we raise a warning. Ploidy defaults to 2. This is what's currently implemented.

I like this method.  I am thinking that our simulator simulates haplotypes (`sample_size` alias: `num_haps` ). Haplotypes can be then used to generate `k`-poid individuals in different ways. 

By default, we use `k` different haplotypes to make a single *heterozygous* `k`-ploid individual (k haps to 1 ind).  But what if we want to simulate *homozygous* diploid? In that case, we use one haplotype to combine with itself to make one diploid (1 hap to 1 ind). This can change how many individuals we can generate from a given number of haplotypes simulated. So, I would suggest having an additional argument (like `is_het` default to True) which together with `ploidy=2` could better control specify the way individuals are generated.

petrelharp:
Instead of "homozygous", do you mean a "selfed" individual, or a member of an "inbred line", @gbinux? When would you want to have `is_het=False`?

gbinux:
> Instead of "homozygous", do you mean a "selfed" individual, or a member of an "inbred line", @gbinux? When would you want to have `is_het=False`?

@petrelharp Thank you for the question. I was thinking of malaria parasite *Plasmodium falciparum*.  

"Humans are the intermediate hosts in which asexual reproduction occurs, and female anopheline mosquitos are the definitive hosts harbouring the sexual reproduction stage." (From Wikipedia). 

In human samples, the parasites are haploids but sometimes need to be treated as *fake* diploids for some analyses. For example, to generate diploid VCF files from simulated haplotypes,  I would like to have each individual homozygous at each locus (this is what I meant "homozygous diploid", sorry if I used it incorrectly)

hyanwong:
> In human samples, the parasites are haploids but sometimes need to be treated as _fake_ diploids for some analyses. For example, to generate diploid VCF files from simulated haplotypes...

But why would you want to generate fake diploid VCFs in the first place? This sounds like a hack to get certain tools to work. As such, would it not be better to have a bespoke VCF outputting function that produces fake diploids?

jeromekelleher:
> Should we think about how the parameters interact with population specifications? 

Yes, good question @hyanwong. With the new Demography object, sampling is decoupled from the Population specification so you can't specify sample size with a Population (the old style PopulationConfiguration is still supported, though). The new way of doing things is to do something like
```
demography = msprime.Demography.stepping_stone_1d(10, migration_rate=1)
# There will ultimately be a bunch of different ways of generating a Demography object, this is just 
# one example.
samples = demography.sample(2, 2) # 2 samples each from the first two demes
ts = msprime.simulate(samples=samples, demography=demography)
```
So, the Sample objects passed in as arguments to the ``samples`` parameter will have to either be considered haploid genomes or k-ploid individuals. That's ultimately the choice here I guess.

hyanwong:
> So, the Sample objects (basically a `(time, population)` tuple) passed in as arguments to the `samples` parameter will have to either be considered haploid genomes or k-ploid individuals. That's ultimately the choice here I guess.

So won't the sample objects need a ploidy too?

I can also imagine cases where we have mixed ploidies - e.g. where there is alternation between haploid & diploid each generation. But while it is something that needs to be possible in a tree sequence, perhaps it'll never really be an issue for msprime?

grahamgower:
Presumably one could get mixed ploidy by recapitating with a different ploidy. Unless that would be explicitly disallowed?

petrelharp:
Summarizing, there's two issues here:
1. the time scaling: since Ne (reasonably) refers to the number of individuals, but coalescence rates are inversely proportional to the number of chromosomes, we need the ploidy to convert "population size" to coalescence rates.
2. the output: how many chromosomes (ie sample nodes) to include per individual?

I think the proposal is to have the `ploidy` argument specify both of these at the same time. More fine-grained control of which nodes exactly go into which individuals could be provided as a separate method (`ts.set_individuals()`?) that would let you get e.g. some haploid and some diploid samples.

Note that the coalescent does *not* explicitly deal with individuals, separate sexes, haploid/diploid life stages, or anything like that except for the two points above. We should avoid giving the impression that it does.

hyanwong:
> Note that the coalescent does _not_ explicitly deal with individuals, separate sexes, haploid/diploid life stages, or anything like that except for the two points above. We should avoid giving the impression that it does.

That is a very good point, Peter.

jeromekelleher:
I'm going to close this - let's leave simulate as it is, and allocate individuals using the new ``sim_ancestry`` function (#1119).

PR #1114, opened 2020-08-07
Tskit 0.3.0 updates

jeromekelleher:
Minimal updates to use the tskit C API 0.99.4. The idea is that this will bump the version of tskit required to 0.3, and be ready to merge once tskit 0.3 has been released in order to minimise disruption for developers.

Msprime 0.7.4 works perfectly well with tskit 0.3.0b1 though, which is great.

One minor bug did show up while doing this, which we should sort out:  https://github.com/tskit-dev/tskit/issues/753


jeromekelleher:
I tried to make the test on Appveyor and Travis work by using the binaries from pip. However, this ended up causing problems with HDF5 as we're mixing between h5py from pip and conda. It's not worth figuring out how to fix this, since it seems like basically everything is working from the Circle CI tests.

msprime/cli.py
```diff
@@ -293,7 +293,15 @@ def print_trees(self, tree_sequence, output):
         breakpoints = list(self._simulator.breakpoints) + [self._num_loci]
         if self._num_loci == 1:
             tree = next(tree_sequence.trees())
-            newick = tree.newick(precision=self._precision)
+            # newick = tree.newick(precision=self._precision)
```
jeromekelleher, line 296: This has been fixed upstream, so we can revert once 0.3.0 is released.

PR #1115, opened 2020-08-07
Add discrete_coordinates argument to simulate.

### LINKED ISSUE #880: API for 'finite_sites'?
With #862 merged it's now possible to simulate recombination in discrete or continuous coordinates by specifying the ``discrete`` parameter to RecombinationMap. This means we can have a "finite sites" or "infinite sites" model of recombination (although strictly, we would have to check and enforce the infinite sites recombination thing). It's probably a good idea to have a high-level API for this in ``simulate`` so we don't force users to specify a RecombinationMap if they want to simulate in discrete coordinates.

One idea is to have a new ``finite_sites=True`` argument which would apply to both recombination **and** mutation. We haven't implemented the mutation bit yet, but we could just ValueError for now if ``finite_sites`` is true and ``mutation_rate > 0``. It seems to me that users would probably want finite sites semantics in both cases, and that this would be a user friendly option in the long run.

pinging @tskit-dev/all for opinions on this one.

jeromekelleher:
WIP.

One notable change this includes is  to make all arguments after the first to simulate keyword only. I doubt this will break anyone's code, and it's a good thing to get formalised for 1.0. Any objections?

Closes #880

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -281,7 +281,7 @@ def simulator_factory(
     random_generator=None,
     random_seed=None,
     length=None,
-    discrete_coordinates=None,
+    discrete_genome=None,
     recombination_rate=None,
     recombination_map=None,
     population_configurations=None,
@@ -367,13 +367,13 @@ def simulator_factory(
             raise ValueError("Cannot provide non-positive sequence length")
         if the_rate < 0:
             raise ValueError("Cannot provide negative recombination rate")
-        if discrete_coordinates and length != math.floor(length):
+        if discrete_genome and length != math.floor(length):
             raise ValueError(
-                "Cannot specify non-integer sequence length when discrete_coordinates "
+                "Cannot specify non-integer sequence length when discrete_genome "
                 "is True"
             )
         recombination_map = RecombinationMap.uniform_map(
-            the_length, the_rate, discrete=discrete_coordinates
+            the_length, the_rate, discrete=discrete_genome
         )
     else:
         if not isinstance(recombination_map, RecombinationMap):
@@ -443,7 +443,7 @@ def simulate(
     *,
     Ne=1,
     length=None,
-    discrete_coordinates=None,
+    discrete_genome=None,
     recombination_rate=None,
     recombination_map=None,
     mutation_rate=None,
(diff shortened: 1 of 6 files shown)


jeromekelleher:
I think this is ready for review - be great to get some eyes on this one. Although, I think there's an argument for changing the parameter name to ``discrete_genome`` instead of ``discrete_coordinates``.

We can think about maybe changing the default value of this to True after we merge (#1116).

msprime/ancestry.py
```diff
@@ -364,8 +367,14 @@ def simulator_factory(
             raise ValueError("Cannot provide non-positive sequence length")
         if the_rate < 0:
             raise ValueError("Cannot provide negative recombination rate")
+        if discrete_coordinates and length != math.floor(length):
+            raise ValueError(
+                "Cannot specify continuous sequence length when discrete_coordinates "
```
petrelharp, line 372: maybe instead of "continuous" say "non-integer"

msprime/ancestry.py
```diff
@@ -477,6 +488,17 @@ def simulate(
     :param float length: The length of the simulated region in bases.
         This parameter cannot be used along with ``recombination_map``.
         Defaults to 1 if not specified.
+    :param bool discrete_coordinates: If True, use discrete coordinates
+        in simulations such that recombination breakpoints and mutational
+        sites can only occur at integer positions along the genome.
+        Multiple mutations can occur at the same site.
+        If False (the default) the recombination breakpoints and mutational
+        sites are continuous floating point values. All sites in the
```
petrelharp, line 496: "and mutational sites" -> "and coordinates of mutational sites"

msprime/ancestry.py
```diff
@@ -477,6 +488,17 @@ def simulate(
...
+        returned tree sequence will have exactly one mutation. Please see
+        the :func:`.mutate` function for a more powerful approach to
+        simulating mutations on a tree sequence. It is an error to specify
+        the ``discrete_coordinates`` parameter at the same time as the
+        ``recombination_map`` argument.
```
petrelharp, line 501: ... since this is a property of the recombination_map? Since this also controls mutations maybe it'd be better to allow it to be set but to check if it agrees with the recombination map?

petrelharp: oh, I see your note below, now.

petrelharp:
Looks good to me; minor doc suggestions above.

msprime/ancestry.py
jeromekelleher: done.

msprime/ancestry.py
jeromekelleher: done.

jeromekelleher:
Thanks for the comments @petrelharp . The exact relationship between the recomb-map and discreteness will depend on what the default values are, so didn't see any point in worrying about it until we've decided what the default is for certain (#1116).

Issue #1116, opened 2020-08-10
Change to discrete coordinates by default?


jeromekelleher:
Currently the discrete parameter defaults to False, which is done for compatibility reasons. However, this is probably the wrong choice for most of the models that people might specify, and will probably lead to bugs in people's code. There's really no point in specifying a JukesCantor model without specifying  ``discrete=True`` at the same time.

Perhaps this would be confusing/hard to document though? Perhaps we should default ``discrete=True``, and accept the (probably very minor) breakage that ensues? I don't imagine many people are using ``msprime.mutate`` yet and this is probably our only chance to git it sensible defaults.

jeromekelleher:
Of course, we could argue that ``msprime.simulate`` should default to ``discrete_coordinates=True`` also (#1115). Let's see, what would break? Not a whole lot, actually, except simulations where L=1 and recombination rate != 0. This is an obvious case here discrete is wrong. A more subtle case is when L is small and recombination rate is large - the difference between the discrete and continuous cases will be very large then. I guess what we could do is see if the value of L * recombination_rate and if this is greater than some threshold **and** discrete_coordinates is left to it's default value we emit a warning. If the user explicitly sets discrete_coordinates=False, then we shut up and do what we're told, regardless.

jeromekelleher:
The intuition here is that for a simulation like
```
ts = msprime.simulate(10, Ne=10**4, length=10**8, recombination_rate=1e-8)
```
there's basically no difference between discrete and continuous coordinates, and since most people want discrete coordinates we should use them by default.

On the other hand, if some has a simulation like
```
ts = msprime.simulate(10, length=10, recombination_rate=100)
```
then using discrete coordinates would certainly give different answers. In this case, we would emit a warning like
```
Warning! Since 1.0, msprime uses discrete coordinates by default and the simulation that you are running will have quite different results when discrete and continuous coordinates. If this is an existing simulation developed using older versions of msprime, you should set ``discrete_coordinates=False`` to ensure that that the same type of simulation is run. If you really want to run this simulation with discrete coordinates, set ``discrete_coordinates=True`` to make this warning go away.
```

petrelharp:
Ooooo, I think this is just the right think to do (default to discrete, emit the warning). For when to warn, maybe: if length is less than 1000 and recomb rate is positive?

This means we should go through the docs and check/modify all the places that `simulate()` is used.

jeromekelleher:
[...]

PR #1117, opened 2020-08-10
Simulate discrete genomes by default

jeromekelleher:
See #1116

This is an initial proposal to simulate genomes on discrete coordinates by default. This is a major change from the pre-1.0 semantics, but hopefully it won't break too much. Basically we warn if we think the simulation might be affected significantly. Currently this is just L<1000 and mutation/recombination rate > 0. But, I guess this should really be some function of L * rate > some threshold that makes multiple hits on a single base reasonably likely (we could do the maths on this for the basic coalescent model, but maybe this would be badly wrong in some cases and so worse than a simple rule of thumb).

Basically seems to work. Some stuff is broken, but the warning mostly works. 

Some thoughts:

1.  If we're making discrete the default, then perhaps we should make the parameter ``continuous_genome``? It seems more natural to specify the option this way. We'd also have to update the ``RecombinationMap`` class.
2. It's a bit weird being able to specify discrete/continuous in both simulate and the recombination map. Really, we'd only want to specify ``discrete_genome`` when the recombination_map argument is not provided, since it should be canonical. 

I haven't touched ``mutate`` yet, but presumably we'd do something similar there.

### DIFF
--- a/docs/examples/segregating_sites.py
+++ b/docs/examples/segregating_sites.py
@@ -6,7 +6,10 @@ import msprime
 def segregating_sites(n, theta, num_replicates):
     S = np.zeros(num_replicates)
     replicates = msprime.simulate(
-        sample_size=n, mutation_rate=theta / 4, num_replicates=num_replicates
+        sample_size=n,
+        mutation_rate=theta / 4,
+        num_replicates=num_replicates,
+        discrete_genome=False,
     )
     for j, tree_sequence in enumerate(replicates):
         S[j] = tree_sequence.get_num_mutations()
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1463,6 +1463,10 @@ API for replication: by providing the ``num_replicates`` argument to the
 in a straightforward manner. Here is an example where we compare the
 analytical results for the number of segregating sites with simulations:
 
+.. warning:: TODO add a helpful warning that older versions of this
+   example will give incorrect answers with msprime 1.0 and link to the
+   docs section where we talk about the discrete genome stuff.
+
 .. literalinclude:: examples/segregating_sites.py
 
 Running this code, we get:
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -40,8 +40,6 @@ from . import provenance
 
 logger = logging.getLogger(__name__)
 
-SHORT_GENOME_THRESHOLD = 1000
-
 
 def model_factory(model):
     """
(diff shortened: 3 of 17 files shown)


jeromekelleher:
I've gone through the updates here and fixed all the tests and warnings. Overall, I think this is a good change which will give the right defaults for 1.0 onwards, enable the old behaviour where desired and hopefully not result in much silent code breakage. The warnings work well and allowed me to find all the stuff in the test suite (which was a good dogfooding exercise). Some remaining issues:

- It's very annoying that we use ``discrete`` for RecombinationMap and mutate, but ``discrete_genome`` for simulate. We should pick one and use it throughout. I guess ``discrete_genome`` is less ambiguous, but it's a bit less concise. (We had this discussion before in #880. I'm open to suggestions!)
- Should we allow a RecombinationMap with non-integer coordinates to have discrete=True? It's hard to imagine a real case where you'd want to specify recombination rates varying between sites but then constrain the sites to be at integer locations. I toyed with disallowing them for a bit, but it caused a fair bit of breakage in the test suite. :shrug:

petrelharp:
> maybe we should be more worried about strict compatibility. Maybe we should have a vote or something.

I like this because it's pushing people towards using the methods which are closer to correct, and also are easier for us to maintain in the long run (down with floating point numbers! =).  But I'm also worried about the compatibility. More input would be good.

jeromekelleher:
It'll be a totally biased sample, but here's an attempt to get some input: https://twitter.com/jeromekelleher/status/1293254293508755456

castedo:
It might be worth framing this as, not a 2-way trade-off, but a 3-way trade-off between:

- keep backwards compatibility
- use more sensible (new) defaults
- use the full English word 'simulate' as the function name

The value of sticking to the English word 'simulate' seems rather esthetic and subjective. If that value is small, then another option is:

1. introduce a new function name, say 'sim' (or perhaps 'run' or 'runsim' or 'simular' :))
2. give the new desirable default behaviour to 'sim' 
3. deprecate the function 'simulate' while preserving its default of continuous coordinates

The warnings generated by calling 'simulate' can be increasingly annoying over time. Then eventually 'sim' can become an alias to the new function name. By that point it will be breaking backwards compatibility with a function that has been screaming deprecation warnings for a long time.

FWIW, I like short function names and don't find it problematic that they are not full English words.  But that's just one subjective esthetic style opinion of many. Probably because I try to write code in 80 column terminal windows in vim. :)

jeromekelleher:
It's a good idea @castedo - creating a new top-level function *would* solve a lot of significant problems (see also our dilemma with #1113 ).

jeromekelleher:
I'm going to close this - let's go with the new top-level function instead in #1119.

PR #1128, opened 2020-08-18
Ploidy for multiple merger models

JereKoskela:
This is a proposed infrastructure for k-ploid individuals in multiple merger models. It's a bit more complicated than in the Hudson case because the family structure also plays a role. I've assumed ploidy = 1 means that each individual only has one parent, while ploidy > 1 means two-parent families. I'm sure that is the best thing to do for ploidy = 1 or 2, but am less sure about higher ploidies. Comments are very welcome.

The statistical tests in verification.py are all hard-coded to the diploid case for now. I'll update them once we have a set way to enter arbitrary ploidies as simulation arguments. As far as I could see, that has not been implemented yet, right?

Ping @jeromekelleher, @eldonb, @TPPSellinger

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -6028,10 +6028,26 @@ msp_dirac_get_common_ancestor_waiting_time_from_rate(
     if (lambda > 0.0) {
         u = gsl_ran_exponential(self->rng, 1.0 / lambda);
         if (alpha == 0.0) {
-            ret = pop->initial_size * pop->initial_size * u;
+            if (self->ploidy == 1) {
+                ret = pop->initial_size * pop->initial_size * u;
+            } else {
+                /* For ploidy > 1 we assume N/2 two-parent families, so that the rate
+                 * with which 2 lineages belong to a common family is (2/N)^2 */
+                ret = pop->initial_size * pop->initial_size * u / 4.0;
+            }
         } else {
             dt = t - pop->start_time;
-            z = 1 + alpha * pop->initial_size * pop->initial_size * exp(-alpha * dt) * u;
+            if (self->ploidy == 1) {
+                z = 1
+                    + alpha * pop->initial_size * pop->initial_size * exp(-alpha * dt)
+                          * u;
+            } else {
+                /* For ploidy > 1 we assume N/2 two-parent families, so that the rate
+                 * with which 2 lineages belong to a common family is (2/N)^2 */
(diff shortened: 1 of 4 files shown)


TPPSellinger:
Yes, I think you are right. I don't think it has been implemented. However, I don't think all verification.py test should be modified. I think we can keep those checking recombination and mutation follow the same process (except if you assume ploidy=1 cannot have recombinations or other ploidy generally affects recombinations, then I can modify the test for you). But we might need to update the SFS tests  (I'm not so familiar with the effect of ploidy on chromosome structure and evolution). If you have papers in mind I could read them and then build new SFS tests.

jeromekelleher:
> The statistical tests in verification.py are all hard-coded to the diploid case for now. I'll update them once we have a set way to enter arbitrary ploidies as simulation arguments. As far as I could see, that has not been implemented yet, right?

No, this hasn't been implemented in Python yet @JereKoskela. I think we'll end up putting the ``ploidy`` argument into the new ``sim_ancestry`` function and leave it out of the current ``simulate`` function entirely. I can whip up a bare-bones version of that when you need it. I think it'd be best to get the basic additions that you've made here merged first with C test coverage, and we can follow up with the updates needed to run the verification tests later.

I'll leave it to you, @TPPSellinger and @eldon to decide whether this is the appropriate model or not - seems reasonable to me.

TPPSellinger:
Okay, we can do it this way. I will check Bjarki's code and paper. As soon as you think the implementation is robust, you can ping me and I will build  and do the tests. I will also think about the k-ploidy test.

JereKoskela:
> No, this hasn't been implemented in Python yet @JereKoskela. I think we'll end up putting the `ploidy` argument into the new `sim_ancestry` function and leave it out of the current `simulate` function entirely. I can whip up a bare-bones version of that when you need it. I think it'd be best to get the basic additions that you've made here merged first with C test coverage, and we can follow up with the updates needed to run the verification tests later.

I updated the multiple merger C tests to cover both the ploidy = 1 and ploidy > 1 branches. I'm happy to get this merged, and @TPPSellinger and I can return to the statistical tests once you've had a chance to finish the `sim_ancestry` interface @jeromekelleher. No rush on our account.

jeromekelleher (approved):
Great, thanks @JereKoskela! I'll open an issue to track documenting this.

PR #1129, opened 2020-08-18
Add wheel building in github action

### LINKED ISSUE #1004: Binary wheels
Infrastructure for building binary wheels automatically is in the pipeline (https://github.com/tskit-dev/kastore/pull/118). I'm making this issue to track the progress here for msprime.

The extra complexity here is whether we can pull in GSL as part of the wheel (there's no real point in having the wheel if we can't). I'm hoping that we can use the version of GSL that's available in the manylinux Docker image (i.e., we ``yum install gsl-devel`` in there) and it'll all just work. It doesn't matter what version of GSL it is, anything should work unless it's truly prehistoric.

Assuming we can, how much of GSL do we bring in (i.e., is the software doing something clever and just pulling out the bits of libgsl that we're using, or is it just copying in libgsl and libgslcblas), and how big does this make our wheel?

benjeffery:
Closes #1004


benjeffery:
I've spent enough time on this without figuring out how to statically link on OSX and WINDOWS. On OSX we might be able to do something like https://dropline.net/2015/10/static-linking-on-mac-os-x/ 
I think we just do manylinux wheels for now. I've left the other OS code commented out for future versions of ourselves to maybe have a go.

benjeffery:
BTW linux wheels with static GSL are ~6MB

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,276 @@
...
+      - master
+      - test
+    tags:
+      - '*'
+  release:
+    types: [published]
+  pull_request:
+
+jobs:
+#  windows:
```
jeromekelleher, line 15: I think we should probably delete these blocks for Windows and OSX, or at least put them at the bottom of the file. It'll be confusing for people to come in and see huge chunks of commented out code.

I wouldn't be in any hurry revisiting OSX or Windows binaries, tbh, so probably by the time anyone comes back to this these commented blocks will just be misleading anyway, and (whoever they are) would be better off starting fresh.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,276 @@
...
+          mkdir dist
+          cp */*.{whl,gz} dist/.
+      - name: Publish distribution to Test PyPI
+        if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && !contains(github.event.ref, 'C_')
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.test_pypi_password }}
+          repository_url: https://test.pypi.org/legacy/
+      - name: Publish distribution to PRODUCTION PyPI
+        if: github.event_name == 'release' && !startsWith(github.event.release.tag_name, 'C_')
```
jeromekelleher, line 273: We can get rid of the ``startswith C_`` bit here - there'll never be C releases of msprime. There's  couple more instances of this elsewhere I think.

setup.py
```diff
@@ -109,6 +109,15 @@ def finalize_options(self):
 
 libraries = ["gsl", "gslcblas"]
 defines = []
+extra_objects = []
+
+if os.environ.get("STATIC_GSL", False):
```
jeromekelleher, line 114: This is a bit of a hack - it's not very nice that we're hard-coding paths here. What about having an env variable STATIC_GSL_PREFIX which, if defined, gives the prefix to where libgsl.a and libgslcblas.a can be found? We can os.path.join that to the files.

.github/workflows/docker/buildwheel.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+DOCKER_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$DOCKER_DIR/shared.env"
+
+set -e -x
+
+ARCH=`uname -p`
```
jeromekelleher, line 7: It'd be simpler/more reliable to install GSL using yum (this is centos, isn't it?), I would have thought. We don't really care about the GSL version, so maybe just use that?

.github/workflows/docker/buildwheel.sh
```diff
@@ -0,0 +1,32 @@
...
+cd ..
+
+for V in "${PYTHON_VERSIONS[@]}"; do
+    PYBIN=/opt/python/$V/bin
+    rm -rf build/       # Avoid lib build by narrow Python is used by wide python
+    # Instead of letting setup.py install a newer numpy we install it here
+    # using the oldest supported version for ABI compatibility
+    $PYBIN/pip install oldest-supported-numpy
+    cat /usr/include/gsl/gsl_version.h
+    STATIC_GSL=1 $PYBIN/python setup.py build_ext --inplace
```
jeromekelleher, line 24: Actually, do we need to do the static build thing? Doesn't ``auditwheel repair`` do some magic where it finds the external libs and copies them into the wheel?

jeromekelleher (commented):
Looks good, thanks @benjeffery. I'd imagine there was a lot of force-pushes to get to this point!

I have a few comments above - mainly, do we actually need to do static link for GSL? If not, maybe this would make OSX feasible too?

Let's not do Windows - it's just a waste of time. Windows users use conda, end-of. We'd just confuse them by giving them options, and then not be able to help them when they hit issues because none of us know anything about windows.

benjeffery:
If we make a wheel and don't statically link then you get a error like `missing library libgsl.so` at import time, with a similar message for OSX. I'll have one shot at OSX static linking with the technique in that URL.

jeromekelleher:
> If we make a wheel and don't statically link then you get a error like missing library libgsl.so at import time, with a similar message for OSX.

I'm confused about what ``auditwheel repair`` actually *does* then...

jeromekelleher:
This is weird, when I run auditwheel on some locally built wheels it's not seeing GSL:
```
$ auditwheel show dist/msprime-0.7.5.dev342+g49e2192d-cp38-cp38-linux_x86_64.whl 

msprime-0.7.5.dev342+g49e2192d-cp38-cp38-linux_x86_64.whl is
consistent with the following platform tag: "linux_x86_64".

The wheel references external versioned symbols in these system-
provided shared libraries: libpthread.so.0 with versions
{'GLIBC_2.2.5'}, libc.so.6 with versions {'GLIBC_2.4', 'GLIBC_2.14',
'GLIBC_2.3.4', 'GLIBC_2.2.5', 'GLIBC_2.7'}, libm.so.6 with versions
{'GLIBC_2.29', 'GLIBC_2.2.5'}

This constrains the platform tag to "linux_x86_64". In order to
achieve a more compatible tag, you would need to recompile a new wheel
from source on a system with earlier versions of these libraries, such
as a recent manylinux image.
```
but if I do an ``lddtree`` we get the right libraries:
```
 auditwheel lddtree _msprime.cpython-38-x86_64-linux-gnu.so                                                                                                                [21/440]
INFO:auditwheel.main_lddtree:{                                                                                                                                                                
    "interp": null,                                                                                                                                                                           
    "path": "_msprime.cpython-38-x86_64-linux-gnu.so",                                                                                                                                        
    "realpath": "_msprime.cpython-38-x86_64-linux-gnu.so",                                                                                                                                    
    "needed": [                                                                                                                                                                               
        "libgsl.so.23",                                                                                                                                                                       
        "libpthread.so.0",                                                                                                                                                                    
        "libc.so.6"                                                                                                                                                                           
    ],                                         
    "rpath": [],                                                                               
    "runpath": [],                                                                             
    "libs": {                                  
        "libgsl.so.23": {             
            "realpath": "/usr/lib/x86_64-linux-gnu/libgsl.so.23.1.0",
            "path": "/usr/lib/x86_64-linux-gnu/libgsl.so.23",
            "needed": [          
                "libgslcblas.so.0",                                                            
                "libm.so.6",                                                                   
                "libc.so.6"
            ]                                  
        },                                     
        "libgslcblas.so.0": {                                                                  
            "realpath": "/usr/lib/x86_64-linux-gnu/libgslcblas.so.0.0.0",
            "path": "/usr/lib/x86_64-linux-gnu/libgslcblas.so.0",
            "needed": [     
                "libm.so.6",          
                "libc.so.6"                    
            ]                                  
        },                                     
        "libm.so.6": {                         
            "realpath": "/lib/x86_64-linux-gnu/libm-2.31.so",
            "path": "/lib/x86_64-linux-gnu/libm.so.6",   
            "needed": [                                                                        
                "libc.so.6",
                "ld-linux-x86-64.so.2"                                                         
            ]                                                                                  
        },                                     
        "libc.so.6": {                                                                         
            "realpath": "/lib/x86_64-linux-gnu/libc-2.31.so",
            "path": "/lib/x86_64-linux-gnu/libc.so.6",              
            "needed": [                                                                        
                "ld-linux-x86-64.so.2"
            ]                                  
        },                                                                                     
        "ld-linux-x86-64.so.2": {                                                              
            "realpath": "/lib/x86_64-linux-gnu/ld-2.31.so",           
            "path": "/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2",
            "needed": []
        },                                     
        "libpthread.so.0": {
            "realpath": "/lib/x86_64-linux-gnu/libpthread-2.31.so",
            "path": "/lib/x86_64-linux-gnu/libpthread.so.0",
            "needed": [
                "libc.so.6",
                "ld-linux-x86-64.so.2"
            ]                                  
        }                                      
```
So, it's like auditwheel somehow isn't seeing the _msprime extension module at all.

jeromekelleher:
Ah, no, I'm wrong, when we do ``auditwheel -v show`` get (with a bunch of other stuff):
```
msprime-0.7.5.dev342+g49e2192d-cp38-cp38-linux_x86_64.whl is
consistent with the following platform tag: "linux_x86_64".

The wheel references external versioned symbols in these system-
provided shared libraries: libpthread.so.0 with versions
{'GLIBC_2.2.5'}, libc.so.6 with versions {'GLIBC_2.7', 'GLIBC_2.3.4',
'GLIBC_2.14', 'GLIBC_2.4', 'GLIBC_2.2.5'}, libm.so.6 with versions
{'GLIBC_2.2.5', 'GLIBC_2.29'}

This constrains the platform tag to "linux_x86_64". In order to
achieve a more compatible tag, you would need to recompile a new wheel
from source on a system with earlier versions of these libraries, such
as a recent manylinux image.

The following external shared libraries are required by the wheel:
{
    "libc.so.6": "/lib/x86_64-linux-gnu/libc-2.31.so",
    "libgsl.so.23": "/usr/lib/x86_64-linux-gnu/libgsl.so.23.1.0",
    "libgslcblas.so.0": "/usr/lib/x86_64-linux-gnu/libgslcblas.so.0.0.0",
    "libm.so.6": "/lib/x86_64-linux-gnu/libm-2.31.so",
    "libpthread.so.0": "/lib/x86_64-linux-gnu/libpthread-2.31.so"
}

In order to achieve the tag platform tag "manylinux2014_x86_64" the
following shared library dependencies will need to be eliminated:

libgsl.so.23, libgslcblas.so.0

In order to achieve the tag platform tag "manylinux2010_x86_64" the
following shared library dependencies will need to be eliminated:

libgsl.so.23, libgslcblas.so.0

In order to achieve the tag platform tag "manylinux1_x86_64" the
following shared library dependencies will need to be eliminated:

libgsl.so.23, libgslcblas.so.0
```

jeromekelleher:
Can you try this with the dynamically linked libs using the centos version of GSL, and run ``auditwheel -v show`` and ``auditwheel -v repair`` please @benjeffery? The whole point of auditwheel is that we *don't* have to do funny things with static linking, and I'd like to get this working properly. Hopefully some debug output from the manylinux container will help track things down.

benjeffery:
> centos version of GSL
Do you mean the one installed by `yum install libgsl`? We can't use that one as the version installed in the `manylinux` Centos6 container is too old (1.13) and not compatible with msprime - hence my manual install.

jeromekelleher:
> Do you mean the one installed by yum install libgsl? We can't use that one as the version installed in the manylinux Centos6 container is too old (1.13) and not compatible with msprime - hence my manual install.

Whoa, that is old! OK, manual install is good then - but we should document this sort of stuff with some comments.

jeromekelleher:
It *looks* like audiwheel is doing what it's suppose to [here](https://github.com/tskit-dev/msprime/pull/1129/checks?check_run_id=1002091100#step:6:6009). I wonder if there's some cached version of _msprime lying around somewhere that's confusing things. Can we download these built wheels and try them out locally?

jeromekelleher:
The versions looks weird too (``msprime-0.1.dev1+g030bfe3-cp38-cp38-manylinux2010_x86_64.whl``) - do you know what's going on there?

benjeffery:
> The versions looks weird too (`msprime-0.1.dev1+g030bfe3-cp38-cp38-manylinux2010_x86_64.whl`) - do you know what's going on there?

Yep, in `setup.py` we have  `use_scm_version={"write_to": "msprime/_version.py"},` which needs to be changed to the tskit way of doing things I think.

jeromekelleher:
> Yep, in setup.py we have use_scm_version={"write_to": "msprime/_version.py"}, which needs to be changed to the tskit way of doing things I think.

Why, is setuptools_scm not compatible with wheel building?

jeromekelleher:
I'm not sure running the full test suite is actually helping here, because we're pulling in all these extra dependencies and stuff to do it. It's making things much more complicated than they need to be. Really all we want to do to test these wheels is
```
msp simulate 10 tmp.trees
tskit info tmp.trees
```
If both those commands succeed, that's all we want to know.

benjeffery:
If we created a tag then I think that the version number would match the tag, as the version is being set automatically. Currently we don't so its creating a version based on last tag plus a commit hash. It's not incompatible. We could change tskit to use this method if we wanted.

jeromekelleher:
> If we created a tag then I think that the version number would match the tag, as the version is being set automatically. Currently we don't so its creating a version based on last tag plus a commit hash. It's not incompatible. We could change tskit to use this method if we wanted.

Ah, OK, that's just setuptools_scm doing it's thing. The last tag is 0.7.4, though, so it's surprising it's not picking that up.

We can't use setuptools_scm on tskit, as it doesn't like the multi-directory setup. I spent ages messing with it, and couldn't get it to work properly.

benjeffery:
Great, it seems there was some issue with using the same instance to build and test - now I've separated it into separate steps. This has the added benefit that the wheels are uploaded (to github, not PyPI) even if the tests fail. I'll try OSX now that's sorted.

jeromekelleher:
> I'll try OSX now that's sorted.

Great - hopefully installing GSL from brew and auditwheel will do the trick there. It would be good to have OSX wheels all right.

benjeffery:
auditwheel is linux only, trying https://pypi.org/project/delocate/

benjeffery:
To use `delocate` we would need to move the C extension module into it's own directory. See "EXTERNAL LIBRARY IS NOT BUNDLED INTO WHEEL" here: https://pypi.org/project/delocate/

benjeffery:
@jeromekelleher I think this finally all works! I had to move the location of the c extension module to get `delocate` to work. Final wheels can be seen at https://github.com/tskit-dev/msprime/actions/runs/215076800.

benjeffery:
Ok, I've squashed and removed the PR trigger, so this only runs on push to master and release. Should be good to merge.

tests/test_dict_encoding.py
```diff
@@ -11,8 +11,8 @@
 import numpy as np
 import tskit
 
-import _msprime as c_module
```
jeromekelleher, line 14: Much smaller change is to have ``import msprime._mspriem as c_module``

benjeffery: Yes, but why a different convention in this module? Is it for readability?

jeromekelleher:
Hooray! Congrats @benjeffery, this is a big power-up.

I'm slightly nervous about moving ``_msprimemodule.c`` though - this is a big change and will have unexpected consequences. How about we do this first in a separate PR so we can see the changes in isolation?

benjeffery:
Sure, I'll separate it out.

tests/test_dict_encoding.py
jeromekelleher: It's because we've copied it directly from the tskit source code, and the idea was to be able to update it over time by copying the new code in and not having to replace loads of ``_tskit`` with ``_msprime``'s. Also we can diff the two versions of the file more easily then to see if they're in sync.

benjeffery: Ah, of course, makes sense.

benjeffery:
Gah, gihub has done that "force push closes PR" thing again. Fixing.

Issue #1105, opened 2020-07-25
Something odd with migration to pop 0 in DTWF


nspope:
Hi, thank you for developing this wonderful tool.

When simulating from the DTWF model with island-with-migration population structure (where migration rate varies; e.g. with distance) and very low effective population sizes per deme, I notice something like an artefact in simulated tree topology with regards to tip order.

For example, (A) I generate a random migration matrix by simulating points in the unit square and mapping distance to migration rate with an exponential function; (B) permute rows and columns of the migration matrix; (C) set deme (diploid) Ne to 1; (D) simulate independent genealogies of a single haploid per deme, with deme Ne set to 1; (E) calculate the average pairwise TMRCA of a given tip, averaged over all other tips and loci; (F) repeat B through E ... then this is what the distribution of the average pairwise TMRCA looks like, across permutations:

![tmrca_density](https://user-images.githubusercontent.com/5767783/88448124-0bf98080-cdef-11ea-8b1e-6a99eb9f7126.png)

I am using msprime 0.7.4 and can replicate with the latest version of msprime/master. This does not seem like intended behavior, because the first tip generally has a greater TMRCA (regardless of the spatial location of the first tip, which is really the only thing in the model that makes tips non-exchangeable)? Here is code to reproduce:

```
import msprime
import numpy
import scipy.spatial.distance as dist
import pandas as pd
import matplotlib.pyplot as plt

# expectation of TMRCA under DTWF island model
def island_model_TMRCA (migr_mat, reps = 100, seed = 1):
    migr_mat   = numpy.array(migr_mat, dtype=float)
    num_popul  = migr_mat.shape[0]
    pop_confg  = []
    for i in range(num_popul):
      pop_confg += [msprime.PopulationConfiguration(
        sample_size=1, initial_size=1.)]
    sim = msprime.simulate(
            population_configurations = pop_confg,
            migration_matrix = migr_mat,
            mutation_rate = 0.,
            Ne = 1.,
            num_replicates = reps,
            random_seed = seed,
            model = "dtwf")
    tmrca = numpy.zeros(migr_mat.shape)
    for locus in sim:
       for deme1 in range(num_popul):
         for deme2 in range(num_popul):
           tmrca[deme1,deme2] += locus.at(0.).tmrca(deme1, deme2) 
    return tmrca / reps

# random "spatial" migration matrix
def simulate_migration_matrix(seed, dim):
  numpy.random.seed(seed)
[...]

Issue #988, opened 2020-05-04
[Docs] Review migration rates documentation


jeromekelleher:
A persistent source of confusion in msprime has been what the migration in rates in msprime actually mean.

#978 updated the simulation model section with some clarifications from @petrelharp, and harmonised the usage of source/dest in the MigrationRateChange and MassMigration events. We should review this, and try to clarify as much as possible - perhaps adding a section in the tutorial to really emphasise the point with some simple examples, which we link to the in terse "simulation model" section.

jeromekelleher:
Note also the discrepancy between migration rates in the DTWF and Hudson models, #1105

nspope:
I am looking at [docs/api.rst](https://github.com/tskit-dev/msprime/blob/master/docs/api.rst#simulation-model) in relation to #1105 :

> Population structure is modelled by specifying a fixed number of subpopulations `d`, and a `d \times d` matrix `M` of per generation migration rates. Each element of the matrix `M_{j,k}` defines the fraction of population `j` that consists of migrants from population `k` in each generation. Note that migration rates are specified in a way that makes sense for the coalescence process, so that the `(j, k)^{th}` entry of the migration matrix, `M_{j, k}`, gives the rate at which an ancestral lineage moves from population `j` to population `k` as one follows it back through time.

The description of `M_{j,k}` as a "fraction" is confusing (to me) for continuous time. The way I think of it is that `M_{j,k}` is the expected number of migrations from `k` to `j`, per lineage per generation, forward in time. For discrete time `\sum_{k \neq j} M_{j,k}` has to be less than or equal to 1, because there is at most one migration event per lineage per generation. For continuous time, "generations" are just units and not a fundamental way to organize events: so migration per lineage can happen more "quickly" than 1/generation, and the rates aren't constrained to the unit simplex.

Does this seem like a reasonable change (from 'fraction' to 'expected migrations per individual per generation')? I am after a definition that applies equally well to DTWF and the coalescent.

petrelharp:
Well, let's see, thinking through things: Suppose there's a population of size `N` and an inmigration rate of `mN`, i.e., at rate `m` new individuals show up from elsewhere and replace a randomly chosen individual. The expected number of replacements per unit time is `mN`, and since the rate at which migrants arrive and replace a *given* individual is `mN/N = m`, the probability a given individual is replaced is `1 - exp(-m) = m + O(m^2)`.

[...]

Issue #1122, opened 2020-08-12
Split up C unit tests


jeromekelleher:
The C unit tests are all currently in one file. Split this up into (say) three files, ``test_likelihood``, ``test_ancestry``, and ``test_mutations``. We can copy some of the code from tskit's [test_main.c](https://github.com/tskit-dev/tskit/blob/master/c/tests/testlib.c) to avoid repeating all the CUnit boilerplate in each file.

This is a good introduction to the C code (see also #1211 and #1084)

castedo:
msprime/docs/development.rst needs updating to reflect the recent new improvement by @nspope (splitting up the C unit tests). The dev docs still reference running `./build/tests` and `./build/tests test_fenwick` so those lines are now out of date. This is in the section C Library > Unit Tests.
Perhaps we still want a simple script `./build/tests` that runs all the tests?

jeromekelleher:
Good spot, thanks @castedo.

> Perhaps we still want a simple script ./build/tests that runs all the tests?

Running ``ninja -C build test`` does this, so we'd be better off documenting this useful trick.

castedo:
or `find ./build/test_* -type f -executable | bash` ;-)

PR #1130, opened 2020-08-18
Fix DTWF migration bug in #1105

nspope:
For #1105:
* Because multiple entries of the migration matrix can change in a single generation, the least intrusive place to check the row sum constraint is in `msprime.c::msp_run_dtwf`. It would be nicer to check for validity immediately after parsing demographic events, but I don't see how to do this without doing a 'dry run' through all the migration rate changes/population mergers. However, I could implement this if it's preferred to checking on the fly.
* Definition of migration rates are updated throughout docs/tutorial as per comments in #988
* Fixed invalid migration matrix (invalid for DTWF) in `tests.c::test_demographic_events`
* Added unit test `tests.c::test_dtwf_migration_matrix_not_stochastic`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -67,25 +67,22 @@ However, recombination (with or without recombination maps) and a constant gene 
 rate along the genome can be combined in ``msprime``.
 
 Population structure is modelled by specifying a fixed number of subpopulations
-:math:`d`, and a :math:`d \times d` matrix :math:`M` of per generation migration rates.
-Each element of the matrix :math:`M_{j,k}` defines
-the fraction of population :math:`j` that consists of migrants from
-population :math:`k` in each generation.
-Note that migration rates are specified in a way that makes sense for the
-coalescence process, so that the :math:`(j, k)^{th}` entry of the migration matrix,
-:math:`M_{j, k}`, gives the rate at which an ancestral lineage moves from
-population :math:`j` to population :math:`k` as one follows it back through time.
-In terms of the demographic process of the populations, if a total of
-:math:`n` migrants move from population :math:`k` to
-population :math:`j` per generation, and population :math:`j` has a total of
-:math:`N_{j}` individuals, then the migration matrix has :math:`M_{j,k} = n/N_j`.
-This differs from the migration matrix one usually uses in population
-demography: if :math:`m_{k,j}` is the proportion
-of individuals (in the usual sense; not lineages) in population :math:`k`
-that move to population :math:`j` per generation,
-then translating this proportion of population :math:`k`
-to a proportion of population :math:`j`, we have
-:math:`M_{j,k} = m_{k,j} \times N_k / N_j`.
+:math:`d`, and a :math:`d \times d` matrix :math:`M` of per-generation
+migration rates. The :math:`(j,k)^{th}` entry of :math:`M` is the expected number
+of migrants moving from population :math:`k` to population :math:`j` per
+generation, divided by the size of population :math:`j`. In terms of the
+coalescent process, :math:`M_{j,k}` gives the rate at which an ancestral
+lineage moves from population :math:`j` to population :math:`k`, as one follows
+it back through time. In continuous-time models, when :math:`M_{j,k}` is close
+to zero, this rate is approximately equivalent to the fraction of population :math:`j`
+that is replaced each generation by migrants from population :math:`k`. In
+discrete-time models, the equivalence is exact and each row of :math:`M` has
+the constraint :math:`\sum_{k \neq j} M_{j,k} \leq 1`. This differs from the
+migration matrix one usually uses in population demography: if :math:`m_{k,j}`
+is the proportion of individuals (in the usual sense; not lineages) in
+population :math:`k` that move to population :math:`j` per generation, then
+translating this proportion of population :math:`k` to a proportion of
+population :math:`j`, we have :math:`M_{j,k} = m_{k,j} \times N_k / N_j`.
 
 Each subpopulation has an initial absolute population size :math:`s`
 and a per generation exponential growth rate :math:`\alpha`. The size of a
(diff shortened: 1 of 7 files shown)


jeromekelleher (changes requested):
This is great, thanks @nspope! We're not quite catching all the C code changes in the test cases (see the diff [here](https://codecov.io/gh/tskit-dev/msprime/compare/9a4561e6c67f9bfd027a40c19384456b16197343...d7a4b0983181a4abcea7ea6952f0a2c5980a4070/diff) so it would be great if you could add a quick test to make sure we cover that last case.

docs/api.rst
```diff
@@ -67,25 +67,22 @@ However, recombination (with or without recombination maps) and a constant gene
...
+it back through time. In continuous-time models, when :math:`M_{j,k}` is close
+to zero, this rate is approximately equivalent to the fraction of population :math:`j`
+that is replaced each generation by migrants from population :math:`k`. In
+discrete-time models, the equivalence is exact and each row of :math:`M` has
+the constraint :math:`\sum_{k \neq j} M_{j,k} \leq 1`. This differs from the
+migration matrix one usually uses in population demography: if :math:`m_{k,j}`
+is the proportion of individuals (in the usual sense; not lineages) in
+population :math:`k` that move to population :math:`j` per generation, then
+translating this proportion of population :math:`k` to a proportion of
+population :math:`j`, we have :math:`M_{j,k} = m_{k,j} \times N_k / N_j`.
```
petrelharp, line 85: nice, this sounds good

docs/tutorial.rst
```diff
@@ -1577,6 +1589,12 @@ generations in the past, there were only two lineages left. The continuous
 time standard coalescent model was then used to simulate the ancient past of
 these two lineages.
 
+Note that migration rates between populations in the discrete-time
+Wright-Fisher model obey a sum constraint (see the :ref:`API
+<sec_api_simulation_model>`), whereas those in continuous-time models
+do not. Therefore migration rates may need to be modified simultaneously
+with transitions between simulation models.
+
```
petrelharp, line 1597: not sure if we need to say this caveat again, as usually people will not have migration rates that are so high?

lib/msprime.c
```diff
@@ -4462,6 +4469,13 @@ msp_run_dtwf(msp_t *self, double max_time, unsigned long max_events)
             }
             assert(mig_tmp[j] == 0);
 
+            // Must check that migration matrix is right stochastic in the main
```
petrelharp, line 4472: perhaps say "must have rows that sum to <= 1" instead of "right stochastic", to be less jargoney

lib/msprime.c
```diff
@@ -431,6 +433,11 @@ msp_set_migration_matrix(msp_t *self, size_t size, double *migration_matrix)
                     goto out;
                 }
             }
+            sum += migration_matrix[j * N + k];
+        }
+        if (self->model.type == MSP_MODEL_DTWF && sum > 1) {
+            ret = MSP_ERR_DTWF_MIGRATION_MATRIX_NOT_STOCHASTIC;
+            goto out;
```
petrelharp, line 440: Hm - won't this be caught by the other check below, anyways? If so, is there a reason to keep it in both places? It seems like the code would be tidier with just one check...

jeromekelleher: I agree it'd be better if we just had one check, but maybe there's some loophole we're not seeing?

petrelharp: Don't know. I guess it doesn't hurt to have it in here, and arguably makes the constraint more obvious.

jeromekelleher: If we can't figure out a way of testing it, though, we should definitely remove it - no point in having unreachable code cluttering the place up.

nspope: I agree, this really isn't needed if checking on the fly, I will go ahead and amend.

Actually I think the check that's not covered by unit tests is the one in `msp_run_dtwf` ... this one is explicitly tested. I'll add that.

Thanks!

nspope:
OK, now have a single check that is covered in unit tests, and incorporated suggestions on docs by @petrelharp.

Question: having an "aborted" run with an invalid migration matrix, and then resetting by
https://github.com/tskit-dev/msprime/blob/e2a9116d4abe677530c267a4f9a5da51fafeee23/lib/tests/tests.c#L988-L992
is fine, correct? The original suite of tests in `tests.c::test_demographic_events` are passing, so am assuming so ...

lib/tests/tests.c
```diff
@@ -973,8 +974,23 @@ test_demographic_events(void)
...
+            ret = msp_initialise(&msp);
+            CU_ASSERT_EQUAL(ret, 0);
+            ret = msp_run(&msp, DBL_MAX, ULONG_MAX);
+            CU_ASSERT_EQUAL(ret, MSP_ERR_DTWF_MIGRATION_MATRIX_NOT_STOCHASTIC);
+            /* Reset to valid state */
+            ret = msp_set_migration_matrix(&msp, 4, migration_matrix);
+            CU_ASSERT_EQUAL(ret, 0);
+            ret = msp_reset(&msp);
+            CU_ASSERT_EQUAL(ret, 0);
+        }
```
petrelharp, line 993: Gee, I would have pulled this out into a separate test, but testing everything all at once seems to be how we're rolling in this file already, so I'll let Jerome weigh in on this if he wants. LGTM otherwise.

nspope:
Sure, happy to refactor if that's preferred.

lib/tests/tests.c
jeromekelleher: Yes, I agree with Peter - this test is already badly factored and too complicated, so let's pull this new check out into it's own test,  ``test_dtwf_migration_rates``. I would use  ``test_dtwf_zero_pop_size`` as inspiration here (i.e., copy most of it). It's annoying there's so much copying and duplication in this file, but hopefully we can split it up a bit soon (#1122) and that'll make things easier to refactor.

nspope:
ugh, this is wrapped up after rebase, but now appveyor fails (not sure why, comparison shows that it is able to automerge)

nspope:
Re-triggered checks and this seems good to go.

jeromekelleher (approved):

jeromekelleher:
Looks great, thanks again @nspope!

Issue #983, opened 2020-04-30
no simulation happens if all samples are ancestral


petrelharp:
For instance:
```
ts = msprime.simulate(samples=[msprime.Sample(0, 1.0), msprime.Sample(0, 1.0)])
print(ts.tables.nodes)
# id	flags	population	individual	time	metadata
# 0	1	0	-1	1.00000000000000	
# 1	1	0	-1	1.00000000000000	
```
... so, there's no trees! Nothing has happened!  Adding a sample at time 0 gets trees.

It's a bit wierd, I know, but it'd be easy to at least check if this is going to happen and throw an error if so.

jeromekelleher:
Should we try to detect it and handle it properly or just error?

petrelharp:
Depends how easy handling it properly is; I haven't looked into it.

jeromekelleher:
I don't think it'd be too bad, probably just a case of detecting the condition in C and winding the clock forward until the first sampling event. Would have to make sure that demographic events etc, are properly handled too of course.

PR #1131, opened 2020-08-19
Allow initial_size=0.

### LINKED ISSUE #1125: population sizes can't be zero, so positive sizes are used where they don't make sense
Consider a three population model, `(0, (1, 2))`, where `0` is the ancestor of `1` and `2`. Furthermore, we have that `0` goes extinct at some time `T`, and `1` and `2` pop into existance at this time.

The typical way of describing this model is as follows.
```python
import msprime

N = 1000
T = 500
population_configurations = [
    msprime.PopulationConfiguration(initial_size=N),
    msprime.PopulationConfiguration(initial_size=N),
    msprime.PopulationConfiguration(initial_size=N),
]
demographic_events = [
    msprime.MassMigration(T, source=1, dest=0),
    msprime.MassMigration(T, source=2, dest=0),
]

ddb = msprime.DemographyDebugger(
    population_configurations=population_configurations,
    demographic_events=demographic_events,
)
ddb.print_history()
```
Which results in the following output:
```
Model =  hudson(reference_size=1)
=============================
Epoch: 0 -- 500.0 generations
=============================
     start     end      growth_rate |     0        1        2
   -------- --------       -------- | -------- -------- --------
0 |  1e+03    1e+03               0 |     0        0        0
1 |  1e+03    1e+03               0 |     0        0        0
2 |  1e+03    1e+03               0 |     0        0        0

Events @ generation 500.0
   - Mass migration: Lineages moved with probability 1.0 backwards in time with source 1 & dest 0
                     (equivalent to migration from 0 to 1 forwards in time)
   - Mass migration: Lineages moved with probability 1.0 backwards in time with source 2 & dest 0
                     (equivalent to migration from 0 to 2 forwards in time)
===============================
Epoch: 500.0 -- inf generations
===============================
     start     end      growth_rate |     0        1        2
   -------- --------       -------- | -------- -------- --------
0 |  1e+03    1e+03               0 |     0        0        0
1 |  1e+03    1e+03               0 |     0        0        0
2 |  1e+03    1e+03               0 |     0        0        0
```

But I would argue that the population sizes should be zero outside the epochs where the populations exist. This would be a more accurate representation of the demograpy (but neither `PopulationConfiguration` nor `PopulationParametersChange` permit `initial_size=0`):
```
N = 1000
T = 500
population_configurations = [
    msprime.PopulationConfiguration(initial_size=0),
    msprime.PopulationConfiguration(initial_size=N),
    msprime.PopulationConfiguration(initial_size=N),
]
demographic_events = [
    msprime.PopulationParametersChange(T, initial_size=N, population_id=0),
    msprime.MassMigration(T, source=1, dest=0),
    msprime.MassMigration(T, source=2, dest=0),
    msprime.PopulationParametersChange(T, initial_size=0, population_id=1),
    msprime.PopulationParametersChange(T, initial_size=0, population_id=2),
]
```

A possible simpler alternative would be to have `PopulationConfiguration` and `PopulationParametersChange` accept an `enable` parameter that defaults to `True`?

grahamgower:
Closes #1125.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -404,7 +404,7 @@ msp_set_population_configuration(
         ret = MSP_ERR_POPULATION_OUT_OF_BOUNDS;
         goto out;
     }
-    if (initial_size <= 0) {
+    if (initial_size < 0) {
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
     }
@@ -3308,6 +3308,11 @@ msp_insert_sample(msp_t *self, node_id_t sample, population_id_t population)
     double seq_len = self->sequence_length;
     segment_t *u;
 
+    if (self->populations[population].initial_size == 0) {
+        ret = MSP_ERR_BAD_SAMPLES;
+        goto out;
+    }
+
     u = msp_alloc_segment(self, 0, seq_len, 0,
         rate_map_position_to_mass(&self->recomb_map, seq_len), sample, population, 0,
         NULL, NULL);
@@ -5405,7 +5410,7 @@ msp_add_population_parameters_change(
         ret = MSP_ERR_POPULATION_OUT_OF_BOUNDS;
         goto out;
     }
-    if (initial_size <= 0) {
+    if (initial_size < 0) {
         assert(!gsl_isnan(initial_size));
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
--- a/lib/tests/test_ancestry.c
+++ b/lib/tests/test_ancestry.c
@@ -2011,10 +2011,8 @@ test_demographic_events(void)
         CU_ASSERT_EQUAL(ret, 0);
         CU_ASSERT_EQUAL_FATAL(msp_set_recombination_rate(&msp, 1), 0);
 
-        /* Zero or negative population sizes are not allowed */
+        /* Negative population sizes are not allowed */
         ret = msp_set_population_configuration(&msp, 0, -1, 0);
-        CU_ASSERT_EQUAL(ret, MSP_ERR_BAD_PARAM_VALUE);
-        ret = msp_set_population_configuration(&msp, 0, 0, 0);
         CU_ASSERT_EQUAL(ret, MSP_ERR_BAD_PARAM_VALUE);
 
         ret = msp_set_num_populations(&msp, 2);
@@ -2034,6 +2032,8 @@ test_demographic_events(void)
             CU_ASSERT_EQUAL(ret, 0);
         }
 
+        ret = msp_set_population_configuration(&msp, 0, 0, 0);
+        CU_ASSERT_EQUAL(ret, 0);
         ret = msp_set_population_configuration(&msp, 0, 100, 0);
         CU_ASSERT_EQUAL(ret, 0);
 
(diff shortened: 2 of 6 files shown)


grahamgower:
It seems that something more is needed. Sampling from a population with N=0 should fail, but doesn't in all cases. And sampling from an ancient population that had initial_size=0 in the population configuration, but population size increased at an older time, should be reasonable. In the latter case each sample is (erroneously) a singleton root!

jeromekelleher:
We can check for population size = 0 when we add samples [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L3298). It should be a simple matter to check if ``population->initial_size == 0``, and that will catch all cases in which we try to sample from zero-size pops, as it's checking the initial_size at the appopriate time.

We'll want to get some simple C tests in for this as well @grahamgower - just to make sure we're covering these new code paths.

grahamgower:
> We can check for population size = 0 when we add samples [here](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.c#L3298). It should be a simple matter to check if `population->initial_size == 0`, and that will catch all cases in which we try to sample from zero-size pops, as it's checking the initial_size at the appopriate time.

Is there an easy way to find out the size of a population at a particular time (without running the simulation or the demography debugger)? Because for a model like in the tests I added here, the *time* of sampling dictates whether sampling should be permitted.

> We'll want to get some simple C tests in for this as well @grahamgower - just to make sure we're covering these new code paths.

Yep, can do.

Do you have any insight about why what's going on in this case where we attempt to sample from the ancestral population at an appropritate time point?
```python
import msprime

N = 1000
T = 500
population_configurations = [
    msprime.PopulationConfiguration(initial_size=0),
    msprime.PopulationConfiguration(initial_size=N),
    msprime.PopulationConfiguration(initial_size=N),
]
demographic_events = [
    msprime.MassMigration(T, source=1, dest=0),
    msprime.MassMigration(T, source=2, dest=0),
    msprime.PopulationParametersChange(T, initial_size=N, population_id=0),
    msprime.PopulationParametersChange(T, initial_size=0, population_id=1),
    msprime.PopulationParametersChange(T, initial_size=0, population_id=2),
]

ts = msprime.simulate(
    population_configurations=population_configurations,
    demographic_events=demographic_events,
    samples=[msprime.Sample(0, 2 * T)] * 10,
)

print(ts.first().draw(format="ascii"))
print("num_roots =", ts.first().num_roots)
```
Which with the current pr outputs the following:
```
0  1  2  3  4  5  6  7  8  9

num_roots = 10
```

jeromekelleher:
> Is there an easy way to find out the size of a population at a particular time (without running the simulation or the demography debugger)? Because for a model like in the tests I added here, the time of sampling dictates whether sampling should be permitted.

No, we'll just have to catch that at run-time. That's fine.

jeromekelleher:
> Do you have any insight about why what's going on in this case where we attempt to sample from the ancestral population at an appropritate time point?

Ah yes, that's #983

jeromekelleher:
Is this ready for final review @grahamgower?

grahamgower:
> Is this ready for final review @grahamgower?

Yep, sorry. I've squashed/rebased now.

jeromekelleher (approved):
Looks great, thanks @grahamgower!

PR #1133, opened 2020-08-19
Move c module to /msprime

benjeffery:
Needed for #1129, separated out for discussion.

The main breakage will be if users are doing `import _msprime` but as that is considered internal that might be acceptable.


tests/test_dict_encoding.py
```diff
@@ -11,8 +11,8 @@
 import numpy as np
 import tskit
 
-import _msprime as c_module
```
jeromekelleher, line 14: Let's change this to ``import msprime._msprime as c_module``. This'll make the diffs to the tskit version smaller and make our lives easier.

jeromekelleher (changes requested):
I thought we might have some stuff in the documentation that would be misleading, but it looks like we're all OK. This is quite an improvement, actually, keeping all the msprime stuff withing the package directory. Excellent!

One minor change above.

tests/test_dict_encoding.py
benjeffery: Done!

benjeffery:
@jeromekelleher changes made, should be good to go.

jeromekelleher (approved):

PR #1134, opened 2020-08-19
Initial outline of sim_ancestry.

jeromekelleher:
First pass of the sim_ancestry function, basically just enough so that we can run interesting models with ploidy (see #1128).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
       - run:
           name: Run gcov manually, as the one used in codecov doesn't work here.
           command: |
-            gcov -pb -o ./build/temp.linux*/ _msprimemodule.c
+            gcov -pb -o ./build/temp.linux*/msprime msprime/_msprimemodule.c
             cd build-gcc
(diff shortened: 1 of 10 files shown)


jeromekelleher:
I think this is ready to go - this is the simplest and least controversial subset of parameters I could add in. This should be enough to unblock @JereKoskela in #1137

jeromekelleher:
I think I'm going to merge this so we have the initial draft for sim_ancestry in place and a way to run k-ploid sims. The main differences between this and ``msprime.simulate`` are:

- samples refer to k-ploid **individuals**. So, ``msprime.sim_ancestry(10, ploidy=2)`` results in a tree sequence with 20 samples. Ploidy defaults to 1.
- The discrete_genome parameter defaults to True
- The equivalent to the ``length`` argument to msprime.simulate is ``sequence_length``.
- The ``recombination_rate`` argument can either be a single value *or* a RateMap instance. If you specify a ``recombination_rate`` scalar, you must specify a sequence_length.

This seems like a useful and uncontroversial set of changes, so let's try them out for a bit. Over to you @JereKoskela and @TPPSellinger!

PR #1138, opened 2020-08-20
Generalise to many Fenwick trees and rate maps

jeromekelleher:
The code currently assumes that we have only one Fenwick tree, which is very restrictive. This update puts all the calculations for computing the masses associated with segments in once place, so that we can optionally have a Fenwick tree (or probably 2) also for working with gene conversion.

WIP.


fbaumdicker:
Great to see some progress here @jeromekelleher. I have been thinking about this issue for a while now. In short: Yes, I think we can do better than this!

However, it is a little bit more complicated and we will have to make a few design decisions. So let me try to explain the current GC_in + GC_left approach and how we could change it:

The current GC_in part is indeed very similar to recombination and I don‘t think we need to change this part. However, it is worth mentioning that currently we already have rejection sampling within the GC_in part. If both breakpoints fall between the same two segments, nothing will happen but this case occurs within the GC_in part. This is mainly relevant if two small parts merge that have a huge distance between them. But such parts are frequently broken apart again by recombination. So unless recombination is zero this is not a big issue.

The GC_left part has currently no rejection sampling at all. It conditions on two events:
A) The gene conversion track actually reaches the left end of the individual
B) The gene conversion track does not span the whole individual (= chain of segments)

If I understood you correctly you are suggesting to remove both A and B such that we simulate all GC events that start left of the right end of the last segment. However, I think it would suffice to remove only condition B). That way we would no longer need the information of the last segment within the first segment but would avoid the ton of events left of the first segment.

So far, I was in favor of using a Fenwick tree that replaces the current costly computation of cleft that needs to iterate over all segments. A Fenwick tree could avoid this in the following way:

Assign a weight to each segment, which is either zero if it is not the first segment in a chain. Or otherwise, the weight is some function of the first position in the first segment and the last position in the last segment. That way we could update cleft using a Fenwick tree without iterating over everything again and again. However, at each gc, rec, coal event, we would have to check for the involved individuals and change the weights of the new and old first segments in dependency of the new and old last segments. We would need to do this really carefully.

As an alternative, it might be much easier and less error-prone to remove the condition B) on the last segment instead. (But still keep the condition A that each event covers at least the first position). For long segments, removing B) really doesn‘t make a big difference, as almost all events will end before the right end. And for short segments, there won‘t be that many noneffective events as A) still significantly limits the number of events. The benefit would be that segment weights do not depend on the following segments.

For example, if we would remove condition B) in the current implementation the computation of cleft would be really easy (for constant gc initiation rates gamma). It would be the current number of individuals * gamma / tl, where tl is the expected track length. (But we would need to change some distributions in the implementation of GC_left)

If the gc initiation rate is variable (i.e. defined by a custom map or proportional to recombination) we would need to compute for each position m along the genome the rate at which gene conversion tracks arrive at this position given by something like

w_m = \sum_k^m  \gamma_{m-k} (1-1/tracklength)^k ,

where gamma_j is the local gene conversion initiation rate at position j. We need to keep in mind though that this will result in fewer gc events at the left end of the genome compared to the right end unless we add some „virtual“ gc initiation weight to the left of the first position.
We would need to compute w_m only once at the beginning. During the simulation, we need to keep track of the sum over all weights w_m at the first position of the first segment of all individuals. This could be achieved with a Fenwick tree.

Does this make sense to you? Any thoughts?

jeromekelleher:
Thanks @fbaumdicker, this is super helpful. I considered the approach in which we have a fenwick tree which gives a weight to the first segment which is a function of the distance between the first and last segments, but I think this would be (a) tricky and (b) probably inefficient for RE events because we'd need to iterate to both the start and end of the segment chain to compute the new cleft values for both individuals. A lot of things are built around the assumption that you can consider segments mostly in isolation, and this breaks that assumption. (Maybe this won't actually be significant, though, and I'm overthinking it.)

I think it would probably be best to think about improving the cleft performance in a follow-up PR though, as most of the infrastructure here is needed in whatever way we do it. Do you agree that my current approach should produce the correct distribution of GC events (even if a bit inefficiently)?

fbaumdicker:
Yes, as I outlined above I agree we should avoid the dependency of the left and right positions in the individuals.

I also had a look at the changes you made to the gc processes and I believe the suggested "simplified" approach should produce the correct gc distribution. However, we should also check whether the gc verification tests are still passed. One of them is designed to detect differences/errors of the gc_left and gc_in part at the left and right end of segments.

However, I fear the current approach will not only be _a bit inefficiently_ but significantly slower than other approaches. Especially for longer sequences. In the conditioned version an individual with only one position will produce no gc events at all, while in the current approach it will create a lot of them. In addition, those short individuals are created at each gc event.
But we should be able to deal with this later if you want to finish the multiple Fenwick trees part first and separately.
Just keep in mind that we will need another (slightly different working) Fenwick tree for this.

jeromekelleher:
OK, thanks @fbaumdicker - I'll try and get this finished up as soon as possible so, and we can circle back on the left-break process then.

jeromekelleher:
Note to self: looks like the label tracking for sweeps is probably broken (see algorithms.py update). Need to update the C tests so that this is found and fix.

jeromekelleher:
I'm going to merge this as it's turned into a major diff. Some really significant updates here though:

1) Fixed a major perf regression on large simulations that was caused by previous attempts to handle numerical drift in the fenwick trees.
2) Generalised the Fenwick tree code so that we can have more than one of them and they can be turned on and off as we like.
3) The GC code is much nicer now - but broken, unfortunately. I'll open an issue to track this.
4) Good progress on setting up the fundamental benchmarks for ASV.

In terms of perf, we're doing pretty well on this branch but we may find some improvements by working with the benchmarks.

PR #1140, opened 2020-08-21
Fix docs build

jeromekelleher:
Addresses omission from changing the location of the docs build in #1133


benjeffery:
Ah, of course. I think changing the CI to build docs before the extension might be good? Would have caught my omission.

jeromekelleher:
> I think changing the CI to build docs before the extension might be good?

Yes, excellent idea! I've pushed a commit doing just that.

I'm not at all surprised this has happened btw - we were never going to catch all the little quirks around the low-level module in one pass. If this is the only breakage we get then I think we've done a great job.

PR #1146, opened 2020-08-25
Add initial benchmarks

benjeffery:
Add initial benchmarks using airspeed velocity. As well as being used for local dev, these can be run when commits are pushed to master to give a running report like https://pv.github.io/numpy-bench/ although this is not setup yet. I've added some basic `simulate`` benchmarks, certainly there are more that could be written.


jeromekelleher (approved):
Wohoo!

PR #1147, opened 2020-08-25
Update verification.py

TPPSellinger:
Add SFS checks for different ploidy levels. Ping @JereKoskela @jeromekelleher .


TPPSellinger:
I tried to 
~~~~~
python3 -m black verification.py
~~~~~
but it didn't seem to do the trick. I don't really understand why it's complaining.

grahamgower:
It looks like you added new functions, but there are already functions with those names. Black doesn't find this, but flake8 does.

TPPSellinger:
My bad,  I will fix this. Thanks for the help @grahamgower !

Issue #1111, opened 2020-08-06
Assertion bug in sweeps


jeromekelleher:
The test ``test_Selective_msms_vs_msp`` added in #1109 results in an assertion being thrown:
```
jk@tempest$ python3 verification.py test_Selective_msms_vs_msp
  0%|                                                                   | 0/1 [00:00<?, ?it/s]2020-08-06 16:30:14,257 [154387] INFO     root: running 1 tests using 1 processes
2020-08-06 16:30:14,257 [154387] INFO     root: Running TestInstance(test_class='MsmsSweeps', method_name='test_Selective_msms_vs_msp')
python3: lib/msprime.c:5845: msp_std_common_ancestor_event: Assertion `y_node != NULL' failed.
Aborted (core dumped)
```

andrewkern:
this is curious. any insights here @jeromekelleher?

jeromekelleher:
Nope, no idea. I didn't spend any time on it, though. How about we get your new tests merged and I'll take a look then?

jeromekelleher:
This also happens with ``sweep_ex1`` in the DiscoalTests sometimes.

jeromekelleher:
Clsoed in #1149

PR #1149, opened 2020-08-26
add a test to test_sweeps.c to reproduce the bug in issue 1111

gbinux:
In the test_sweep,c, I set all the input parameters of msp_t  to match those in verficiation.py script captured by running gdb —args /usr/bin/python3 verification.py -c MsmsSweeps (break at msp_run_sweep function and print *self). Then I put the c code in a loop to mimic the num_repeats. Still, I was not able to reproduce the bug. Suggestions? @jeromekelleher 

I tried to simplify this from a list of 3 models `[None, (t_start, temp_model), (None, None)] ` to a single model object `temp_model` (not in this commit) to better match the test function in test_sweeps.c. And `python3 verification.py -c MsmsSweeps` can reproduce the bug but `test_sweeps` can not.

```
        if params["num_sweeps"] > 0:
            model = [None]
            for i in range(params["num_sweeps"]):
                temp_model = msprime.SweepGenicSelection(
                    position=params["sel_pos"],
                    end_frequency=params["end_frequency_lst"][i],
                    start_frequency=0.5 / params["refsize"],  # should be small
                    alpha=params["alpha"],
                    dt=1.0 / (40 * params["refsize"]),  # should be small enough?
                )
                t_start = params["end_time_lst"][i]
                model.append((t_start, temp_model))
            model.append((None, None))
```

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4736,6 +4736,8 @@ msp_run_sweep(msp_t *self)
         goto out;
     }
 
+    curr_step = 1;
+
     while (msp_get_num_ancestors(self) > 0 && curr_step < num_steps) {
         events++;
         /* Set pop sizes & rec_rates */
@@ -4746,7 +4748,6 @@ msp_run_sweep(msp_t *self)
             rec_rates[j] = recomb_mass;
         }
 
-        curr_step++;
         event_prob = 1.0;
         event_rand = gsl_rng_uniform(self->rng);
         sweep_over = false;
@@ -4754,24 +4755,34 @@ msp_run_sweep(msp_t *self)
             sweep_dt = time[curr_step] - time[curr_step - 1];
             /* using pop sizes grabbed from get_population_size */
             pop_size = get_population_size(&self->populations[0], time[curr_step]);
-            p_coal_B = ((sweep_pop_sizes[1] * (sweep_pop_sizes[1] - 1)))
-                       / allele_frequency[curr_step] * sweep_dt / pop_size;
-            p_coal_b = ((sweep_pop_sizes[0] * (sweep_pop_sizes[0] - 1)))
-                       / (1.0 - allele_frequency[curr_step]) * sweep_dt / pop_size;
+
+            p_coal_B = 0;
+            if (avl_count(&self->populations[0].ancestors[1]) > 1) {
+                p_coal_B = ((sweep_pop_sizes[1] * (sweep_pop_sizes[1] - 1)))
+                           / allele_frequency[curr_step] * sweep_dt / pop_size;
+            }
+            p_coal_b = 0;
+            if (avl_count(&self->populations[0].ancestors[0]) > 1) {
+                p_coal_b = ((sweep_pop_sizes[0] * (sweep_pop_sizes[0] - 1)))
+                           / (1.0 - allele_frequency[curr_step]) * sweep_dt / pop_size;
+            }
             p_rec_b = rec_rates[0] * sweep_dt;
             p_rec_B = rec_rates[1] * sweep_dt;
             sweep_pop_tot_rate = p_coal_b + p_coal_B + p_rec_b + p_rec_B;
             /* doing this to build in generality if we want >1 pop */
+
             total_rate = sweep_pop_tot_rate;
             event_prob *= 1.0 - total_rate;
             curr_step++;
+
             sweep_over = total_rate == 0;
         }
         if (sweep_over) {
             break;
         }
-        /* passed check, choose event */
+
         tmp_rand = gsl_rng_uniform(self->rng);
+
         e_sum = p_coal_b;
         self->time = time[curr_step - 1];
         if (tmp_rand < e_sum / sweep_pop_tot_rate) {
--- a/lib/tests/test_sweeps.c
+++ b/lib/tests/test_sweeps.c
@@ -307,6 +307,78 @@ test_sweep_genic_selection_time_change(void)
     tsk_table_collection_free(&tables);
 }
 
+static void
+sweep_genic_selection_mimic_msms_single_run(unsigned long int seed)
+{
+
+    /* Try to mimic the msms parameters used in verification.py
+           "100 300 -t 200 -r 200 500000"
+           " -SF 0 0.9 -Sp 0.5 -SaA 5000 -SAA 10000 -N 10000"
+     */
+    int ret;
+    uint32_t n = 100;
+    double num_loci = 500001;
+    double position = 0.5;
+    double alpha = 10000;
+    double recom_rate = 0.0004;
+    double start_frequency = 0.5 / 10000;
+    double end_frequency = 0.9;
+    double dt = 1.0 / 400000;
+
+    msp_t msp;
+    sample_t *samples = malloc(n * sizeof(sample_t));
+    gsl_rng *rng = gsl_rng_alloc(gsl_rng_default);
+    tsk_table_collection_t tables;
+
+    // Test over differnt seeds
+    gsl_rng_set(rng, seed);
+
+    CU_ASSERT_FATAL(samples != NULL);
+    CU_ASSERT_FATAL(rng != NULL);
+    memset(samples, 0, n * sizeof(sample_t));
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = num_loci;
+    ret = msp_alloc(&msp, n, samples, &tables, rng);
+    CU_ASSERT_EQUAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(msp_set_recombination_rate(&msp, recom_rate), 0);
+    ret = msp_set_dimensions(&msp, 1, 2);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    // To mimic the verfication.py call
+    msp_set_discrete_genome(&msp, 0);
+    msp_set_gene_conversion_rate(&msp, 0, 1);
+    msp_set_avl_node_block_size(&msp, 65536);
+    msp_set_node_mapping_block_size(&msp, 65536);
+    msp_set_segment_block_size(&msp, 65536);
+
+    ret = msp_set_simulation_model_sweep_genic_selection(
+        &msp, position, start_frequency, end_frequency, alpha, dt);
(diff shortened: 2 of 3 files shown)


lib/tests/test_sweeps.c
```diff
@@ -307,6 +307,77 @@ test_sweep_genic_selection_time_change(void)
...
+    double start_frequency = 0.5 / 10000;
+    double end_frequency = 0.9;
+    double dt = 1.0 / 400000;
+
+    msp_t msp;
+    sample_t *samples = malloc(n * sizeof(sample_t));
+    gsl_rng *rng = gsl_rng_alloc(gsl_rng_default);
+    tsk_table_collection_t tables;
+
+    gsl_rng_set(rng, 1000);
```
jeromekelleher, line 333: Ah, so you're using the same random seed here each time. Try making ``seed`` a parameter and passing in your index variable i in the function below.

jeromekelleher:
Looks great, thanks @gbinux! I'd recommend trying a bunch of seeds now and see if you can reproduce. Also worth running things through valgrind, just to see if there's any memory issues going on.

lib/tests/test_sweeps.c
gbinux: I tried this method which always leads to crashes.
```
Suite: msprime
  Test: test_genic_selection_trajectory ...passed
  Test: test_sweep_genic_selection_bad_parameters ...passed
  Test: test_sweep_genic_selection_events ...passed
  Test: test_sweep_genic_selection_single_locus ...passed
  Test: test_sweep_genic_selection_recomb ...passed
  Test: test_sweep_genic_selection_time_change ...passed
  Test: test_sweep_genic_selection_reproduce_bug_1111 ...seed: 1seed: 2seed: 3seed: 4seed: 5seed: 6seed: 7seed: 8seed: 9seed: 10seed: 11seed: 12seed: 13seed: 14seed: 15seed: 16seed: 17seed: 18seed: 19seed: 20seed: 21seed: 22seed: 23seed: 24seed: 25seed: 26seed: 27seed: 28seed: 29seed: 30seed: 31seed: 32seed: 33seed: 34seed: 35bug!gsl: ../gsl/gsl_rng.h:199: ERROR: invalid n, either 0 or exceeds maximum value of generator
Default GSL error handler invoked.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff79f2537 in __GI_abort () at abort.c:79
#2  0x00007ffff7c2311d in gsl_error () from /lib/x86_64-linux-gnu/libgsl.so.25
#3  0x00007ffff7cf340d in gsl_rng_uniform_int () from /lib/x86_64-linux-gnu/libgsl.so.25
#4  0x0000555555568a56 in msp_std_common_ancestor_event (self=0x7fffffffe1d0, population_id=0, label=1) at ../msprime.c:6006
#5  0x0000555555566179 in msp_run_sweep (self=0x7fffffffe1d0) at ../msprime.c:4782
#6  0x00005555555664e8 in msp_run (self=0x7fffffffe1d0, max_time=1.7976931348623157e+308, max_events=4294967295) at ../msprime.c:4879
#7  0x000055555555891f in sweep_genic_selection_reproduce_bug_1111_single_run (seed=35) at ../tests/test_sweeps.c:363
#8  0x00005555555589c1 in test_sweep_genic_selection_reproduce_bug_1111 () at ../tests/test_sweeps.c:379
#9  0x00007ffff7faba97 in ?? () from /lib/x86_64-linux-gnu/libcunit.so.1
#10 0x00007ffff7fabd39 in ?? () from /lib/x86_64-linux-gnu/libcunit.so.1
#11 0x00007ffff7fac137 in CU_run_all_tests () from /lib/x86_64-linux-gnu/libcunit.so.1
#12 0x0000555555586e0e in test_main (tests=0x7fffffffe7a0, argc=1, argv=0x7fffffffe918) at ../tests/testlib.c:97
#13 0x0000555555558aaf in main (argc=1, argv=0x7fffffffe918) at ../tests/test_sweeps.c:400
```

There seems to be a bug in `msp_std_common_ancestor_event` function. Some seeds can lead to the  `n-1` equals  `0` and then gsl error.

`j = (uint32_t) gsl_rng_uniform_int(self->rng, (n - 1)); `

gbinux:
Since some seeds could lead to the error in ` msp_std_common_ancestor_event ` due to `gsl_rng_uniform_int` parameter `n - 1` less than 1

```
       j = (uint32_t) gsl_rng_uniform_int(self->rng, (n - 1));
```

I tried this:

```
       ancestors = &self->populations[population_id].ancestors[label];
       /* Choose x and y */
       n = avl_count(ancestors);
       ////////////////////////////////////////////
       if(n <= 1) {
               ret = 0;
               goto out;
       }
       //////////////////////////////////////////////
       j = (uint32_t) gsl_rng_uniform_int(self->rng, n);


out:
       return ret;
```
This returns 0 when n <=1. It looks like both the gsl error and the bug in issue#1111 are addressed! After this fix, I have unhidden all test methods in MsmsSweeps test class (which were hidden to avoid crashes in the verfication.py tests), they run well without errors. I also checked the Q-Q plots for Hudson model (from the `test_neutral_msms_vs_msp` method) and they looked pretty consistent. This makes me think that return 0 when n <= 1 does not affect the role of `msp_std_common_ancestor_event ` in simulation. I hope this is true.

jeromekelleher:
Thanks @gbinux, this is great. I wonder if there's a slightly simpler explanation though - are we calling ``msp_common_ancestor_event`` from the sweep code where there's < 2 lineages present? If this was a general problem it would certainly have happened before now, so it must be something that's happening in the context of a sweep.

gbinux:
Thanks for the input @jeromekelleher . I will look into the sweep code and see if I can locate the bug.

gbinux:
I think this can fix the bug in #1111 . When there are < 2 lineages of either `B` label or `b` label, the coalescent rate is set to 0 and ignore the corresponding coalescent events.
```
            /* using pop sizes grabbed from get_population_size */
            pop_size = get_population_size(&self->populations[0], time[curr_step]);
            ////////////////////////////////////////////////////////////
            if (avl_count(&self->populations[0].ancestors[1]) > 1)
                p_coal_B = ((sweep_pop_sizes[1] * (sweep_pop_sizes[1] - 1)))
                       / allele_frequency[curr_step] * sweep_dt / pop_size;
            else
                p_coal_B = 0;
            if (avl_count(&self->populations[0].ancestors[0]) > 1)
                p_coal_b = ((sweep_pop_sizes[0] * (sweep_pop_sizes[0] - 1)))
                       / (1.0 - allele_frequency[curr_step]) * sweep_dt / pop_size;
            else
                p_coal_b = 0;
             ////////////////////////////////////////////////////////////
            p_rec_b = rec_rates[0] * sweep_dt;
            p_rec_B = rec_rates[1] * sweep_dt;

```
Does this break the assumptions behind the code? @andrewkern If not, I will push the commit

lib/msprime.c
```diff
@@ -4746,32 +4754,45 @@ msp_run_sweep(msp_t *self)
...
         while (event_prob > event_rand && curr_step < num_steps && !sweep_over) {
             sweep_dt = time[curr_step] - time[curr_step - 1];
             /* using pop sizes grabbed from get_population_size */
             pop_size = get_population_size(&self->populations[0], time[curr_step]);
-            p_coal_B = ((sweep_pop_sizes[1] * (sweep_pop_sizes[1] - 1)))
-                       / allele_frequency[curr_step] * sweep_dt / pop_size;
-            p_coal_b = ((sweep_pop_sizes[0] * (sweep_pop_sizes[0] - 1)))
-                       / (1.0 - allele_frequency[curr_step]) * sweep_dt / pop_size;
+
+            // to avoid # lineages < 2
```
jeromekelleher, line 4765: I would structure this slightly differently to save a bit of vertical space @gbinux 

```c
p_coal_B = 0;
if (avl_count(&self->populations[0].ancestors[1]) > 1) {
     p_coal_B = ((sweep_pop_sizes[1] * (sweep_pop_sizes[1] - 1)))
￼                           / allele_frequency[curr_step] * sweep_dt / pop_size;
}
p_coal_b = 0;
...
```
I think the intention is very clear then and we don't need the comments.

Note as part of the project style we always use braces around indented blocks.

lib/msprime.c
```diff
@@ -4736,6 +4736,14 @@ msp_run_sweep(msp_t *self)
         goto out;
     }
 
+    // let curr_step = 1 before the loops and then add 1 to it at the end of
```
jeromekelleher, line 4739: This seems correct to me @gbinux, and this is a good explanation. However, I think the comment will be confusing to people who read the code in the future, who have no idea what the old way was. Comments should describe the code *as it is*.

jeromekelleher: Maybe we don't need the comment at all?

lib/msprime.c
```diff
@@ -4746,32 +4754,45 @@ msp_run_sweep(msp_t *self)
...
             /* doing this to build in generality if we want >1 pop */
+
             total_rate = sweep_pop_tot_rate;
             event_prob *= 1.0 - total_rate;
             curr_step++;
-            sweep_over = total_rate == 0;
+
+            // Not using total_rate == 0; otherwise crashes happen due to precision
+            // issue.
+            sweep_over = total_rate <= DBL_EPSILON;
```
jeromekelleher, line 4788: Does this really happen? I thought we'd end up with total_rate = 0 because we multiply by zero at some point. How would we end up with very tiny total_rate values?

jeromekelleher (commented):
Thanks for tracking this down @gbinux, just a few stylistic issues.

lib/msprime.c
gbinux: You are right. This must have happened when I was messing around to locate the bug. It seems changing it back to 
`sweep_over = total_rate == 0;` does not cause any issue in multiple tests.  I will change this line back to the previous version.

jeromekelleher (approved):
LGTM ,thanks @gbinux. Can you [squash](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) the commits please?

PR #1150, opened 2020-08-26
C structures refactoring : Rename/elim some tsk_id_t typedefs; rename data members

### LINKED ISSUE #1121: C structures refactoring
There's a number of long-standing refactoring issues noted at the top of [msprime.h](https://github.com/tskit-dev/msprime/blob/master/lib/msprime.h#L66).

- [ ] Get rid of the population_id_t, label_id_t typedefs and use tsk_id_t everywhere instead.
- [ ] Change segment_t.population_id to segment_t.population
- [ ] Change node_mapping_t.left to node_mapping_t.position.
- [ ] Change sweep_t.locus to sweep_t.position

This would be a good issue for someone looking to get familiar with the C code.

### LINKED ISSUE #1145: requirements/readthedocs.txt missing sphinx_rtd_theme
In building a minimal container image that can make the msprime docs, I call
```
pip3 install -r requirements/readthedocs.txt
```
but results in:
```
(ubuntu-18.04-pip)castedo@fire:~/shr/src/msprime/docs$ make
python3 -m sphinx -b html -d _build/doctrees  -W . _build/html
Running Sphinx v3.2.1
loading pickled environment... failed
failed: build environment version not current

Theme error:
sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.(pip install sphinx_rtd_theme)
Makefile:52: recipe for target 'html' failed
make: *** [html] Error 2
```
I'm in the process of committing the full Dockerfile to `tskit-dev/containers` that shows all the details.

Adding `sphinx_rtd_theme` to `requirements/readthedocs.txt` fixes the problem.

Feel free to assign to me to fix.

castedo:
Closes #1121

  - rename/elim typedefs:
    - population_id_t
    - label_id_t
    - node_id_t
    - mutation_id_t
    - site_id_t
  - rename some struct data members:
    - tsk_id_t *.population_id -> *.population
    - node_mapping_t.left -> node_mapping_t.position.
    - sweep_t.locus -> sweep_t.position

Also:
Closes #1145

### DIFF
--- a/lib/msprime.h
+++ b/lib/msprime.h
@@ -63,36 +63,30 @@
 #define MSP_PED_STATE_CLIMBING 1
 #define MSP_PED_STATE_CLIMB_COMPLETE 2
 
-/* FIXME: Using these typedefs to keep the diff size small on the initial
- * tskit transition. Can remove later. */
 typedef tsk_id_t population_id_t;
-typedef tsk_id_t node_id_t;
-typedef tsk_id_t mutation_id_t;
-typedef tsk_id_t site_id_t;
 typedef tsk_id_t label_id_t;
 
 typedef struct segment_t_t {
-    /* TODO change to population */
-    population_id_t population_id;
+    population_id_t population;
     label_id_t label;
     /* During simulation we use genetic coordinates */
     double left;
     double right;
     double left_mass;
     double right_mass;
-    node_id_t value;
+    tsk_id_t value;
     size_t id;
     struct segment_t_t *prev;
     struct segment_t_t *next;
 } segment_t;
 
 typedef struct {
-    double left; /* TODO CHANGE THIS - not a good name! */
+    double position;
     uint32_t value;
 } node_mapping_t;
 
 typedef struct {
-    population_id_t population_id;
+    population_id_t population;
     double time;
 } sample_t;
 
@@ -129,8 +123,8 @@ typedef struct {
 
 typedef struct {
     double time;
-    node_id_t sample;
-    population_id_t population_id;
+    tsk_id_t sample;
+    population_id_t population;
 } sampling_event_t;
 
 /* Simulation models */
@@ -157,8 +151,7 @@ typedef struct {
 } genic_selection_trajectory_t;
 
 typedef struct _sweep_t {
-    /* TODO change the name of this to position */
-    double locus;
+    double position;
     union {
         /* Future trajectory simulation models would go here */
         genic_selection_trajectory_t genic_selection_trajectory;
@@ -262,7 +255,7 @@ typedef struct _msp_t {
 
 /* Demographic events */
 typedef struct {
-    population_id_t population_id;
+    population_id_t population;
     double initial_size;
     double growth_rate;
 } population_parameters_change_t;
@@ -279,12 +272,12 @@ typedef struct {
 } mass_migration_t;
 
 typedef struct {
-    population_id_t population_id;
+    population_id_t population;
     double proportion;
 } simple_bottleneck_t;
 
 typedef struct {
-    population_id_t population_id;
+    population_id_t population;
     double strength;
 } instantaneous_bottleneck_t;
 
--- a/lib/dev-tools/dev-cli.c
+++ b/lib/dev-tools/dev-cli.c
@@ -118,7 +118,7 @@ read_samples(config_t *config, size_t *num_samples, sample_t **samples)
         if (t == NULL) {
             fatal_error("population not specified");
         }
-        ret_samples[j].population_id = (population_id_t) config_setting_get_int(t);
+        ret_samples[j].population = (population_id_t) config_setting_get_int(t);
         t = config_setting_get_member(s, "time");
         if (t == NULL) {
             fatal_error("population not specified");
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -145,7 +145,7 @@ cmp_node_mapping(const void *a, const void *b)
 {
     const node_mapping_t *ia = (const node_mapping_t *) a;
     const node_mapping_t *ib = (const node_mapping_t *) b;
-    return (ia->left > ib->left) - (ia->left < ib->left);
+    return (ia->position > ib->position) - (ia->position < ib->position);
 }
 
 static int
@@ -491,7 +491,7 @@ out:
 
 static segment_t *MSP_WARN_UNUSED
 msp_alloc_segment(msp_t *self, double left, double right, double left_mass,
-    double right_mass, node_id_t value, population_id_t population, label_id_t label,
+    double right_mass, tsk_id_t value, population_id_t population, label_id_t label,
     segment_t *prev, segment_t *next)
 {
     segment_t *seg = NULL;
@@ -517,7 +517,7 @@ msp_alloc_segment(msp_t *self, double left, double right, double left_mass,
     seg->left_mass = left_mass;
     seg->right_mass = right_mass;
     seg->value = value;
-    seg->population_id = population;
+    seg->population = population;
     seg->label = label;
 out:
     return seg;
@@ -568,7 +568,7 @@ msp_alloc(msp_t *self, size_t num_samples, sample_t *samples,
             goto out;
         }
         for (j = 0; j < num_samples; j++) {
-            self->samples[j].population_id = samples[j].population_id;
+            self->samples[j].population = samples[j].population;
             self->samples[j].time = samples[j].time;
             if (self->samples[j].time < 0) {
                 ret = MSP_ERR_BAD_PARAM_VALUE;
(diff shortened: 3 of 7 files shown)


castedo:
Also closes #1145

lib/msprime.h
```diff
@@ -63,36 +63,27 @@
...
-typedef tsk_id_t node_id_t;
-typedef tsk_id_t mutation_id_t;
-typedef tsk_id_t site_id_t;
-typedef tsk_id_t label_id_t;
-
 typedef struct segment_t_t {
-    /* TODO change to population */
-    population_id_t population_id;
-    label_id_t label;
+    tsk_id_t population;
```
jeromekelleher, line 67: I'm a bit hesitant about changing ``population_id_t`` and ``label_id_t``, actually, because of this struct. Once the dust settles on #1138, we might be able to make this struct a bit smaller by making population and label 16 bit (which is plenty big enough). This struct is important as we do have gazillions of them in large simulations. But, given that most things are 8 byte aligned these days I don't know if that would make any difference anyway.

jeromekelleher (commented):
Looks great, thanks @castedo! If you did this with a sed script, would you mind holding off on it until after #1138 lands? There's a bunch of changes in there too, and it would be a bit easier to not deal with the merge mess.

Also, see my comment above about label and population, it might be better to back out of changing those because we could potentially save some significant space at some point by typedefing them to int16.

castedo:
> If you did this with a sed script, would you mind holding off on it until after #1138 lands? There's a bunch of changes in there too, and it would be a bit easier to not deal with the merge mess.

The sed script with a forced push to master was more a joke. ;-) But I did use some sed lines for the first pass. Yes, let's proceed avoiding the messy merge. Better you have a clear diff with #1138 and it'll be easier for me to merge my branch after #1138 merges with master. I'm going to practice some advanced git moves and follow and work off the branch you mention in #1138.

lib/msprime.h
castedo: In general, makes sense to me to keep typedefs, like population_id_t and label_id_t. So I'll update to KEEP all the tsk_id_t typedefs that I initially removed in this PR.

jeromekelleher: Getting rid of ``node_id_t``,  ``mutation_id_t`` and  ``site_id_t`` is good, they're definitely pointless.

jeromekelleher:
Sounds good, thanks @castedo. I'll try and get #1138 ship-shape ASAP

lib/msprime.h
castedo: Good to know, I'll keep those of those three pointlessly precise typedefs out.

castedo:
@jeromekelleher 
> But, given that most things are 8 byte aligned these days I don't know if that would make any difference anyway.
With the caveat I'm not very familiar with HPC, I'd predict most users are running x86-64 and data is aligned the same way we're building in linux containers and personal computers. I'm going to see if can tell.

jeromekelleher:
> I'd predict most users are running x86-64

This is definitely true.

castedo:
I'm not sure if you are wondering whether all data members will get 8 byte
aligned or just 64 bit data members will. In gdb on x86-64 I'm looking at
segment_t and I see `population`(_id) and `label` both getting placed together and
fitting within 8 bytes. However I see 32 bit `tsk_id_t value;` sitting all alone
with a 32 bit gap to the next data member `size_t id;`.

If `population_id_t` and `label_id_t` were both 16 bits and then `tsk_id_t value;`
was moved to be next to them, then I bet all data members of segment_t will be snug
and tight, and the whole struct will fit in 64 bytes rather than the current 72 bytes.

Do you use valgrind's cachegrind tool to evaluate memory cache hits?

Do you know of any easier to see how structs are laid out in memory rather than gdb printing memory addresses of data members?

jeromekelleher:
Actually there's also #1114 that needs to go in to since tskit has just been released @castedo, so it'd be better to hold off on this until the dust has settled on this big PRs early next week. Apologies for the churn here.

jeromekelleher:
It's just the overall size of the struct I'm thinking about @castedo - because of alignment requirements, things tend to get padded out to 8-byte multiples. Cache-hits and stuff is another level, really.

castedo:
My hunch is that on x86-64 only 64 bit data members will get aligned to 
addresses of 8 byte multiples. As long as that 'tsk_id_t value;' in 
segment_t is next to population(_id) and label if they become 16 bit, 
then 8 bytes will get shaved off segment_t.

castedo:
Closes #1145

jeromekelleher:
Looks like this is merging cleanly now @castedo, so let's get it in. Can you fix the clang-format errors please?

castedo:
It looks like github is using clang-format version 6.0 whereas Ubuntu 20.04 defaults to clang-format 10.0.
@jeromekelleher should I update the
["official" :) msprime dev container](https://github.com/tskit-dev/containers/pull/3/files#diff-22d371c5bd5a6a3b8ee9b7e772a19c7a
)
to install specifically clang-format version 6.0? or should the github version be 10?

jeromekelleher (approved):
Great stuff, thanks @castedo!

jeromekelleher:
> It looks like github is using clang-format version 6.0 whereas Ubuntu 20.04 defaults to clang-format 10.0.
@jeromekelleher should I update the "official" :) msprime dev container to install specifically clang-format version 6.0? or should the github version be 10?

This is very irritating all right. I ended up installing clang-format version 6 on my Mint machine rather that fight with it. I don't know what the right approach is - I guess we should wait until the majority of people are using version 10 before upgrading our CI to it. @benjeffery, what do you think?

benjeffery:
> This is very irritating all right. I ended up installing clang-format version 6 on my Mint machine rather that fight with it. I don't know what the right approach is - I guess we should wait until the majority of people are using version 10 before upgrading our CI to it. @benjeffery, what do you think?

We picked v6 as it was the version in the Ubuntu LTS at the time. I'm reluctant to change this on everyone until either another version is ubiquitous, or that the resultant formatted code is stable. For example if we know that v8, 9 and 10 give the same output then specifying minimum v8 makes sense - I'd need to check if this is the case though. I suspect it is not.

Debian stable is now on [v7](https://packages.debian.org/stable/clang-format), ubuntu [v10](https://packages.ubuntu.com/focal/clang-format). I think centOS is v9 and fedora v10, (looking at https://pkgs.org/search/?q=clang-format). So it seems things have moved on. Let me check which versions give identical output and I'll suggest a move if it looks wise.

jeromekelleher:
Sounds good, thanks @benjeffery. It's surprising/a bit crap that clang-format doesn't doesn't have an option to say "use v6 formatting", that would solve all our problems.

jeromekelleher:
Yep, looks like there's [no good way to do this](https://stackoverflow.com/questions/59395507/how-to-maintain-a-clang-format-file-for-different-clang-format-versions).

molpopgen:
Ubuntu 20.04 has 7,8,9, and 10 available, via `apt install clang-format-X`.  Then, `update-alternatives` to have `clang-format` refer to, say, `clang-format-7`, and you may be in business.

castedo:
> Ubuntu 20.04 has 7,8,9, and 10 available, via `apt install clang-format-X`.

and also `clang-format-6.0` for ubuntu 20.04.

molpopgen:
> and also `clang-format-6.0` for ubuntu 20.04.

hah--I missed that b/c of the change in package number pattern.

castedo:
@jeromekelleher in this particular case, the annoying diff is merely white space. The lint-c test could ignore white space changes. It is C after all, not Python. That run-clang-format.py in the list-c test is ~400 lines of Python which doesn't seem to do much more than these 3 single line shell functions:
```
files () { find lib -name '*.[hc]' -not -path 'lib/avl.*' -not -path 'lib/dev-tools/*' -not -path 'lib/subprojects/*'; }
clang-diff () { clang-format $1 | diff --brief $1 - ; }
run_it() { DIFF=0; for F in $(files); do clang-diff $F; let "DIFF |= $?"; done; return $DIFF; }
run_it && echo PASS || echo FAIL
```
If one inserts `--ignore-space-change` in the `diff` call then white space changes will pass.

jeromekelleher:
It's a good idea @castedo, but it'll mean that we get noise in the diffs as those whitespace changes go back and forth each time someone edits a file (say I'm on version 6 and you're on version 7). I think we just have to standardise on clang-format 6 and be done with it.

castedo:
Good point @jeromekelleher that continued committing of whatever `clang-forman -i` does would lead to diff noise.
I've updated the Ubuntu 20 LTS container image I develop in:
https://github.com/tskit-dev/containers/pull/3/commits/778842776c22de64c2cf9f196885505eace17b7c
to follow the approach suggested by @molpopgen  and make clang-format be version 6.0.
For those that are in a pinch with a newer clang version, using this bash function might be useful:
```
chill_format() { (clang-format $1 || cat $1) | diff --ignore-space-change $1 - | patch $1; }
```
Running `chill_format` instead of `clang-format` will drop changes in amount of whitespace by clang-format.

PR #1155, opened 2020-09-03
use gsl_ran_discrete to calc random integer

### LINKED ISSUE #1084: worry about floating point and probabilities in mutation generation
For instance, in [`mutation_matrix_choose_root_state`](https://github.com/tskit-dev/msprime/blob/master/lib/mutgen.c#L232):
```
    while (u > params.root_distribution[j]) {
        u -= params.root_distribution[j];
        j++;
        assert(j < params.num_alleles);
    }
```
that assert might be hit if the root distribution doesn't sum to exactly 1. Since we do error checking that the probabilities make sense, why don't we instead change this to
```
    while (u > params.root_distribution[j] && j + 1 < params.num_alleles) {
        u -= params.root_distribution[j];
        j++;
    }
```
ie, taking the last entry? (if I did that right)

castedo:
Closes #1084

I have yet to convert mutation_matrix_t::transition_matrix over to an array of GSL gsl_ran_discrete_t.

But here's a draft sneak peak of first stage: double *root_distribution getting replaced with a GSL struct.

### DIFF
--- a/lib/msprime.h
+++ b/lib/msprime.h
@@ -24,6 +24,7 @@
 #include <stdbool.h>
 
 #include <gsl/gsl_rng.h>
+#include <gsl/gsl_randist.h>
 #include <tskit.h>
 
 #include "util.h"
@@ -327,7 +328,7 @@ typedef struct {
     size_t num_alleles;
     char **alleles;
     tsk_size_t *allele_length;
-    double *root_distribution;
+    gsl_ran_discrete_t *root_distribution;
     double *transition_matrix;
 } mutation_matrix_t;
 
--- a/lib/msprime.h
+++ b/lib/msprime.h
@@ -32,7 +32,6 @@
 #include "object_heap.h"
 
 #define USE_GSL_DISCRETE_DISTRIBUTION
-#define KEEP_NON_GSL_DISTRIBUTIONS
 
 #ifdef USE_GSL_DISCRETE_DISTRIBUTION
 #include <gsl/gsl_randist.h>
@@ -336,11 +335,11 @@ typedef struct {
     tsk_size_t *allele_length;
 #ifdef USE_GSL_DISCRETE_DISTRIBUTION
     gsl_ran_discrete_t *gsl_root_distribution;
-#endif
-#ifdef KEEP_NON_GSL_DISTRIBUTIONS
+    gsl_ran_discrete_t **gsl_transition_distributions;
+#else
     double *root_distribution;
-#endif
     double *transition_matrix;
+#endif
 } mutation_matrix_t;
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


petrelharp:
Ah good! This will be nice.

jeromekelleher:
Looks very nice @castedo, please ping when it's ready for review.

castedo:
I'm not blocked, but y'all might have some insight on what's going wrong here.

Using the GSL distribution function is causing
`test_BLOSUM62 (tests.test_mutations.TestMutationStatistics)`
and
`test_PAM (tests.test_mutations.TestMutationStatistics)`
to fail.
But when the code is flipped to stick to the old code and not GSL, then all tests pass fine.

I have not yet studied those two unit tests to understand why they might be failing when the GSL function is used.

I've refactored the code (temporarily) so that merely commenting out
`#define USE_GSL_DISCRETE_DISTRIBUTION`
will compile essentially the same code but will opt not to use the GSL function
(and then all tests pass fine).

For now I'm continuing to rework transition_matrix.

petrelharp:
Well, [those](https://github.com/tskit-dev/msprime/blob/b8f5fb60b0ca0885b7f1c2d7ac7a9ae3b95fc543/tests/test_mutations.py#L195) are just checking if [root_distribution and transition probabilities have the right properties](https://github.com/tskit-dev/msprime/blob/b8f5fb60b0ca0885b7f1c2d7ac7a9ae3b95fc543/tests/test_mutations.py#L91), so it looks like there's something wrong with how you're passing that information up to python?

castedo:
> Well, [those](https://github.com/tskit-dev/msprime/blob/b8f5fb60b0ca0885b7f1c2d7ac7a9ae3b95fc543/tests/test_mutations.py#L195) are just checking if [root_distribution and transition probabilities have the right properties](https://github.com/tskit-dev/msprime/blob/b8f5fb60b0ca0885b7f1c2d7ac7a9ae3b95fc543/tests/test_mutations.py#L91), so it looks like there's something wrong with how you're passing that information up to python?

@petrelharp I believe I have figured out why `test_BLOSUM62` and `test_PAM` were failing. The discrete distributions when converted into GSL a lookup table end up loosing some precision. GSL does not preserve the probabilitie values within a fidelity of around 1e-14 with the root and transition probabilities used in those two tests. When the tests failed, I had root distribution converting into GSL lookup tables but not the transition matrix. I believe it's specifically the test making sure the root distribution is stationary given the transition matrix. So when one was GSL and the other not, then the 1e-14 noise from GSL lookup table I bet was making np.allclose(..) fail. Looks like the default for np.allclose() is 1e-5 relative tolerance. I bet a small probability was tweaked by more than 1e-14 by GSL and this was causing a greater than 1e-5 relative change in the test.

petrelharp:
Hah, good detective work, @castedo. It sounds pretty annoying.

castedo:
Closing per #1084 discussion.

PR #1156, opened 2020-09-04
fix potential rounding bug in calc random integer

### LINKED ISSUE #1084: worry about floating point and probabilities in mutation generation

castedo:
Closes #1084

Also closes #1084 like PR  #1155, but this PR does not use the GSL function gsl_ran_discrete like PR #1155. I'm going to add more details for discussion in #1084 as to the pros and cons of these two PR alternatives.

### DIFF
--- a/lib/mutgen.c
+++ b/lib/mutgen.c
@@ -146,41 +146,18 @@ mutation_matrix_print_state(mutation_model_t *self, FILE *out)
     fprintf(out, "\n");
 }
 
-static int MSP_WARN_UNUSED
-mutation_matrix_check_validity(mutation_matrix_t *self)
+static bool MSP_WARN_UNUSED
+valid_probabilities(size_t num_alleles, double *distribution)
 {
-    int ret = 0;
-    size_t j, k;
-    double prob, min_prob;
-    double *mutation_row;
-
+    double prob = 0;
+    double min_prob = 0;
+    size_t j;
     /* Check probabilities sum to one */
-    prob = 0;
-    min_prob = 0;
-    for (j = 0; j < self->num_alleles; j++) {
-        prob += self->root_distribution[j];
-        min_prob = GSL_MIN(min_prob, self->root_distribution[j]);
+    for (j = 0; j < num_alleles; j++) {
+        prob += distribution[j];
+        min_prob = GSL_MIN(min_prob, distribution[j]);
     }
-    if (!doubles_almost_equal(prob, 1.0, 1e-12) || min_prob < 0.0) {
-        ret = MSP_ERR_BAD_ROOT_PROBABILITIES;
-        goto out;
-    }
-    for (j = 0; j < self->num_alleles; j++) {
-        prob = 0;
-        min_prob = 0;
-        mutation_row = self->transition_matrix + j * self->num_alleles;
-        for (k = 0; k < self->num_alleles; k++) {
-            prob += mutation_row[k];
-            min_prob = GSL_MIN(min_prob, mutation_row[k]);
-        }
-        if (!doubles_almost_equal(prob, 1.0, 1e-12) || min_prob < 0.0) {
-            ret = MSP_ERR_BAD_TRANSITION_MATRIX;
-            goto out;
-        }
-    }
-
-out:
-    return ret;
+    return doubles_almost_equal(prob, 1.0, 1e-12) && min_prob >= 0.0;
 }
 
 static int MSP_WARN_UNUSED
@@ -227,13 +204,8 @@ mutation_matrix_choose_root_state(mutation_model_t *self, gsl_rng *rng, site_t *
     int ret = 0;
     mutation_matrix_t params = self->params.mutation_matrix;
     double u = gsl_ran_flat(rng, 0.0, 1.0);
-    tsk_size_t j = 0;
-
-    while (u > params.root_distribution[j]) {
-        u -= params.root_distribution[j];
-        j++;
-        assert(j < params.num_alleles);
-    }
+    size_t j = probability_list_select(u, params.num_alleles, params.root_distribution);
+    assert(j < params.num_alleles);
     site->ancestral_state = params.alleles[j];
     site->ancestral_state_length = params.allele_length[j];
     return ret;
@@ -261,12 +233,7 @@ mutation_matrix_transition(mutation_model_t *self, gsl_rng *rng,
         goto out;
     }
     probs = params.transition_matrix + (tsk_size_t) pi * params.num_alleles;
-    j = 0;
-    while (u > probs[j]) {
-        u -= probs[j];
-        j++;
-        assert(j < (tsk_id_t) params.num_alleles);
-    }
+    j = (tsk_id_t) probability_list_select(u, params.num_alleles, probs);
     ret = 1;
     if (j != pi) {
         /* Only return 0 in the case where we perform an actual transition */
@@ -543,11 +510,23 @@ matrix_mutation_model_alloc(mutation_model_t *self, size_t num_alleles, char **a
 {
     int ret = 0;
     mutation_matrix_t *params = &self->params.mutation_matrix;
+    size_t i;
 
     memset(self, 0, sizeof(*self));
     if (num_alleles < 2) {
         ret = MSP_ERR_INSUFFICIENT_ALLELES;
         goto out;
+    }
+    if (!valid_probabilities(num_alleles, root_distribution)) {
+        ret = MSP_ERR_BAD_ROOT_PROBABILITIES;
+        goto out;
+    }
+    for (i = 0; i < num_alleles; ++i) {
+        double *probs = transition_matrix + i * num_alleles;
+        if (!valid_probabilities(num_alleles, probs)) {
+            ret = MSP_ERR_BAD_TRANSITION_MATRIX;
+            goto out;
+        }
     }
     params->num_alleles = num_alleles;
     params->alleles = calloc(num_alleles, sizeof(*params->alleles));
@@ -572,11 +551,6 @@ matrix_mutation_model_alloc(mutation_model_t *self, size_t num_alleles, char **a
     self->transition = &mutation_matrix_transition;
     self->print_state = &mutation_matrix_print_state;
     self->free = &mutation_matrix_free;
-
-    ret = mutation_matrix_check_validity(params);
-    if (ret != 0) {
-        goto out;
-    }
 out:
     return ret;
 }
--- a/lib/tests/test_core.c
+++ b/lib/tests/test_core.c
@@ -48,12 +48,82 @@ test_strerror_tskit(void)
     }
 }
 
+static void
+test_probability_list_select(void)
+{
+    double short_one = 1.0 - 0.5 * DBL_EPSILON;
+    CU_TEST(short_one == nexttoward(1.0, 0));
+    double short_half = 0.5 - 0.25 * DBL_EPSILON;
+    CU_TEST(short_half == nexttoward(0.5, 0));
+    double long_half = 0.5 + 0.5 * DBL_EPSILON;
+    CU_TEST(long_half == nexttoward(0.5, 1));
+    {
+        double probs[2] = { 0.5, 0.5 };
+        CU_ASSERT_EQUAL(0, probability_list_select(0.0, 2, probs))
+        CU_ASSERT_EQUAL(0, probability_list_select(short_half, 2, probs))
+        CU_ASSERT_EQUAL(1, probability_list_select(0.5, 2, probs))
+        CU_ASSERT_EQUAL(1, probability_list_select(1.0, 2, probs))
+        CU_ASSERT_EQUAL(1, probability_list_select(1.1, 2, probs))
+    }
+    {
+        double probs[2] = { 1.0, 0.0 };
+        CU_ASSERT_EQUAL(0, probability_list_select(short_one, 2, probs))
+        /* Note: gsl_ran_flat does not return 1.0 per documentation */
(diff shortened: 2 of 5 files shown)


lib/mutgen.c
```diff
@@ -146,41 +146,17 @@ mutation_matrix_print_state(mutation_model_t *self, FILE *out)
     fprintf(out, "\n");
 }
 
-static int MSP_WARN_UNUSED
-mutation_matrix_check_validity(mutation_matrix_t *self)
+static bool MSP_WARN_UNUSED
+ok_probability_masses(size_t num_alleles, double *distribution)
```
petrelharp, line 150: to better resemble other functions we have, how about `validate_probabilities`? or `check_probabilities`?

lib/tests/test_core.c
```diff
@@ -48,12 +48,82 @@ test_strerror_tskit(void)
     }
 }
 
+static void
```
petrelharp, line 51: gee, nice tests here

petrelharp:
Very nice solution. Minor comments, but looks good to me otherwise.

lib/tests/test_core.c
castedo: Ah yes, those cases at the end I put in since I thought you might be interested in seeing the case you brought up. You can see how in your case with DBL_EPSILON / 8.0 in the middle a random pick is possible, but then DBL_EPSILON / 16.0 it becomes "very small" in the sense that it is effectively zero and will make that pick impossible.

lib/mutgen.c
castedo: `ok_probability_masses` does seem to pretty much be another `check_validity` type function. And this `ok_` function is returning `bool` rather than returning zero for pass and non-zero for a specific error. So I'm thinking I should change the name and return type to be more consistent.

castedo:
@jeromekelleher I'm assuming you want me to do a squash/rebase when everybody is done with comments and there are no more tweaks.

lib/util.c
```diff
@@ -312,3 +312,39 @@ doubles_almost_equal(double a, double b, double eps)
...
+ *     (the sum of all the lengths)
+ *   - zero if `x` is negative
+ *   - `num_interval` if `x` is NaN
+ * Pre-conditions:
+ *   - `lengths` must point to `num_interval` non-negative non-NaN doubles
+ * Post-condition:
+ *   - returned value is less than or equal to `num_intervals` (and non-negative)
+ *
+ */
+size_t
```
jeromekelleher, line 334: We're not actually using this in the test suite or externally, so the usual convention would be to make it static. Then we could inline it and avoid needing to put C code in the header for the same call overhead as we have now.

lib/util.c
```diff
@@ -312,3 +312,39 @@ doubles_almost_equal(double a, double b, double eps)
...
+positive_interval_select(double x, size_t num_intervals, double const *lengths)
+{
+    size_t i = 0;
+    double sum = 0.0;
+    while (i < num_intervals) {
+        assert(lengths[i] >= 0.0);
+        sum += lengths[i];
+        if (x < sum)
```
jeromekelleher, line 342: We always use braces:
```c
if (x < sum) {
    break;
}
```

lib/util.c
```diff
@@ -312,3 +312,39 @@ doubles_almost_equal(double a, double b, double eps)
...
+            break;
+        ++i;
```
jeromekelleher, line 344: I'm being totally pedantic here, but we use ``i++`` pretty consistently elsewhere. I don't think there's any perf difference these days between ``i++`` and ``++i`` and (as you can probably tell!) consistency is nice. It doesn't make any difference though, as we never use the value returned.

Please do ignore this one if you think I'm going overboard with the consistency-is-super-important thing though!

jeromekelleher (commented):
Looks great, thanks @castedo! I've a few structural suggestions and some very pedantic style comments which you should feel entirely free to argue with if you disagree!

lib/util.c
castedo: OK. I'm trying to recall some of the gotchas when omitting braces. I think merge issues is one of the dangers. Does a particular bug-inducing danger in omitting braces come to mind?

lib/util.c
castedo: I see now that I'm thinking like I'm speaking C++. My thoughts on ++i are not applicable in C.

lib/mutgen.c
```diff
@@ -147,40 +147,17 @@ mutation_matrix_print_state(mutation_model_t *self, FILE *out)
...
-            ret = MSP_ERR_BAD_TRANSITION_MATRIX;
-            goto out;
-        }
-    }
-
-out:
-    return ret;
+    bool valid = (doubles_almost_equal(prob, 1.0, 1e-12) && min_prob >= 0.0);
+    return (valid ? 0 : MSP_ERR_BAD_PARAM_VALUE);
 }
```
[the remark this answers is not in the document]
castedo: Haha, that's what I had before and @petrelharp suggested making the naming consistent with other functions. :) The way I've reconciled this in my mind is there are two kinds of functions: (A) error checking functions vs (B) just general purpose functions evaluating whether some condition is true or false. This function is kind of in a grey area between the two. In this recent commit I went for (A) but I think simpler is better so i think (B) is better choice.

I do think that _IF_ the return convention for this function is bool like a (B) and not error code like an (A) then the function name should suggest it is like an (B) function and sounds different than an (A) function (in other words intentionally inconsistent with naming of (A) functions because the return type is different in an extremely important way). For this reason I recommend prefixing such function names with `ok_....` or `is_....` etc... This makes is more clear to the reader of a function call that indeed it is the kind of function that returns a boolean rather than an int return code.

Is there any convention already in the codebase about function names that suggestion bool funtion return type?

If not, I can suggest `is_...` because the C++ library uses that a lot I believe, C library too probably.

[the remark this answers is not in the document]
castedo: I've tended to do `ok_...` in my code, I'm not sure why. Maybe because `ok_` isn't jarring with plurals, unlike `is_` which is grating with plurals: e.g. `is_valid_proabilities`.
Regardless, with all these functions returning int values of `0` meaning "good/valid" I think it's good idea to make sure a call to a function returning a bool in which `zero`/`false` means "bad/invalid" is different.

lib/tests/test_core.c
```diff
@@ -48,12 +48,82 @@ test_strerror_tskit(void)
...
+test_probability_list_select(void)
+{
+    double short_one = 1.0 - 0.5 * DBL_EPSILON;
```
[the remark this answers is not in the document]
castedo: Kind of related, I think, but I think I should make the double const so that it's more clear this an alias for a value.

What is the thinking behind not mixing declarations and code? Just avoiding expressing too many ideas on a single line or something else?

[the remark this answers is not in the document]
grahamgower: Aside from being an organising principal, older C standards (and thus compilers) don't allow mixed code/declarations. A compiler has to generate low-level code that moves the stack pointer for each variable declared on the stack. The old way assumed all variables were declared together in one block, so the compiler could generate a single instruction to move the stack pointer once, instead of separate instructions for each variable. In addition, at the start of a function, the compiler is already shifting the stack pointer because the stack is a commonly used for passing function arguments. The newer standards (and compilers) just move the stack pointer again when there are mixed code/declarations (or maybe some compilers reorder instructions to avoid this?).

lib/util.c
jeromekelleher: Yep - forgetting about them is absolutely the main issue here. Especially for people used to Python where indentation is meaningful, stuff like this is absolutely lethal:
```c
if (x < sum) 
    x++;
    y++;
y++;
```

jeromekelleher: One could argue that this isn't a danger with a ``break``, but why bother? It's much simpler to pick one rule and to stick with it rather than wasting brain cycles on choosing the appropriate thing based on the situation.

lib/mutgen.c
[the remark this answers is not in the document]
jeromekelleher: I don't think there's a convention for ``is_`` or ``ok_``. We could use ``are_valid_probablities``, if you like. It just seemed to read well to me in the intended usage ``if (!valid_probablities)`` reads literally "if not valid probablilities" to me. Since it's an internal static function that's only used in one place I'm not too bothered either way.

lib/tests/test_core.c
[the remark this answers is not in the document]
jeromekelleher: The thinking originally was to keep older compilers happy. Since we've fixed on C99, this probably isn't an issue now. 

One of the key reasons for keeping all the declarations at the top of a function is making sure that this idiom is always guaranteed to work:
```
double *some_pointer = NULL;

// later
some_pointer = malloc(...);
if (some_pointer == NULL) {
    ret = no memory error
    goto out;
}
// still later
out:
     tsk_safe_free(some_pointer);
     return ret;
```

making sure that all the pointers are initialised at the top of the function makes sure that we're always guaranteed that there's a valid pointer when we take the error path. This is something we really use heavily in the Python-C layer in particular.

For other variables it's a good sign your function is getting too complicated if you've got loads of variables declared at the top.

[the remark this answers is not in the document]
castedo: In older C I believe this kind of thing would not be uncommon, many standard headers files are full of similar:
```
#define SHORT_ONE (1.0 - 0.5 * DBL_EPSILON)
```
in contrast to
```
double const short_one = 1.0 - 0.5 * DBL_EPSILON;
```
Is the first line not mixing declaration vs code? whereas the 2nd one is?

Does the first one make @jeromekelleher grumble but not the other? If so I'm curious why.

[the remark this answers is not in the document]
jeromekelleher: I think it's fine here @castedo - this is testing code which gets pretty ugly anyway. I would suggest different approaches for production code.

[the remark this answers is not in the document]
jeromekelleher: We're targetting a much narrower set of compilers here too, since it's only developers that'll be running the C tests. We don't have to worry about things working on Windows for example.

lib/util.c
castedo: Good point on the Python programmers!

I'm going to waste some brain cycles trying to find a merge scenario where my code results in a bug. :) I have this hunch it's something one could call "merge unfriendly".

Not that I really care, I'm more just curious, like a field linguist of the "languages" "spoken" by programmers. :) But some, like those implementing clang, think the simple rule to just follow is NOT put braces with single simple statements. I suspect this might be part of the reason clang-format does not have a feature for forcing braces after if statements:
https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

jeromekelleher: Yeah, I think it all comes down to what you were brought up on essentially. I was taught to always use braces and it seemed like a good idea so I did it. Others disagree - no coding convention is perfect.

benjeffery: I've been bitten by the situation where you come along some months later to add a line to the loop, forgetting to add the braces, donuts always been my preference to always use braces.

lib/util.c
```diff
@@ -312,3 +312,39 @@ doubles_almost_equal(double a, double b, double eps)
...
+ *   - zero if `x` is negative
+ *   - `num_interval` if `x` is NaN
+ * Pre-conditions:
+ *   - `lengths` must point to `num_interval` non-negative non-NaN doubles
+ * Post-condition:
+ *   - returned value is less than or equal to `num_intervals` (and non-negative)
+ *
+ */
+size_t
+positive_interval_select(double x, size_t num_intervals, double const *lengths)
```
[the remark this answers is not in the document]
castedo: I never new about `restrict`. Why doesn't C++ have that!!!

You're asking about 'const' in a param at all or why the ordering 'double const' vs 'const double'?

castedo:
OK, all suggestions worked into the latest commit. When do I rebase/squash? This is the end of a very very long chain of commits if an #ifdef party of gsl code in the middle.

jeromekelleher:
Looks great, thanks @castedo! It'd be good to squash to one commit if you don't mind, please.

I can't find the comment above now, but my question was around ``double const *x`` vs ``const double *x`` - I've never gotten the hang of that.

jeromekelleher (approved):
Ready to merge after squash.

castedo:
Squish squish squoosh.

castedo:
> Looks great, thanks @castedo! It'd be good to squash to one commit if you don't mind, please.
> 
> I can't find the comment above now, but my question was around `double const *x` vs `const double *x` - I've never gotten the hang of that.

This is a very interesting language development. I have a bunch of interesting example to share. But I'm going to put them in a dev doc idea I'm playing around with and will show you. Idea is a list of "TICL"s ... "Tskit-dev Interpretations of Code Language" :) ... i'll send it soon if it looks interesting. I think it might be more fun than coding rules and conventions, which sounds and feel too conservative and fascist. :)

jeromekelleher (approved):
Thanks @castedo !

PR #1158, opened 2020-09-08
Deprecate tskit module attributes

### LINKED ISSUE #991: Remove tskit compatability layer?
To avoid breaking people's code while we split msprime from tskit, we have a compatibility shim in ``__init__.py`` which imports tskit classes into the msprime namespace. So, for people who used things like ``msprime.TableCollection`` their code would still work.

For 1.0, we need to decide whether we keep supporting this or not. The pros for removing are, we'll need to do it at *some* point, and if we don't do it at 1.0, then when can we do it? The break from tskit was at version 0.7.

The cons are that we will break some code that's out there. I don't think there's a lot of code out there that uses ``msprime.TableCollection`` etc though, so I'd be surprised if it broke much.

#990 removed all dependence on old classes like msprime.TableCollection, etc from the test suite.

benjeffery:
Fixes #991

### DIFF
--- a/tests/test_deprecation.py
+++ b/tests/test_deprecation.py
@@ -1,0 +1,71 @@
+#
+# Copyright (C) 2020 University of Oxford
+#
+# This file is part of msprime.
+#
+# msprime is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# msprime is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


tests/test_deprecation.py
```diff
@@ -0,0 +1,72 @@
...
+            "pack_strings",
+            "pack_bytes",
+            "unpack_bytes",
+            "unpack_strings",
+            "validate_provenance",
+            "NODE_IS_SAMPLE",
+            "FORWARD",
+            "REVERSE",
+        ]:
+            print(name)
```
jeromekelleher, line 70: print

jeromekelleher (approved):
Yep, LGTM. I guess it's *slightly* more general because it allows people to import new stuff from tskit as well, but since it's throwing a warning that seems harmless.

petrelharp (commented):
Looks good!

tests/test_deprecation.py
benjeffery: Dang it. Fixed. If anyone knows how to set a breakpoint that only fires on failing tests that would be helpful!

benjeffery:
Fixed up and PR number added to changelog.

Issue #1161, opened 2020-09-08
GC is not being simulated correctly for long tracklengths


jeromekelleher:
Since #1138, the GC code has been updated to remove the ``gc_left`` event type. The idea was to replace this (for now) with a single class of gc event, where we compensated for gc events which start to the left of the first segment by letting the first segment span back to the start of the sequence. However, this doesn't seem to work:

![breakpoints](https://user-images.githubusercontent.com/2664569/92504822-12a35200-f1fb-11ea-9e68-d020a38122de.png)

This is a pretty extreme example with a very long track length, where we're clearly biasing towards the end of the sequence for breakpoints.

We'll probably fix this by putting in a fenwick tree to track the gc_left process like discussed in #1138 and elsewhere.

GC is working fine for short track lengths, though.

jeromekelleher:
See also the gc_correlation_between_trees test:

![prob_first](https://user-images.githubusercontent.com/2664569/92507284-a1659e00-f1fe-11ea-9022-d4b0fcb5d7f3.png)

fbaumdicker:
Ok, this looks indeed like the gcleft part is misbehaving. I will try to look into this issue.
@jeromekelleher: Are the otherfigures for the middle and last positions in the gc_correlation_between_trees test closer to the prediction?

jeromekelleher:
Thanks @fbaumdicker - I'm actually working on it at the moment. Hold off a little bit and I'll update with a proposal.

fbaumdicker:
Great! Looking forward to see your proposal.

jeromekelleher:
I've reinstated the old way of doing things in #1168, if you want to take a look @fbaumdicker  - I'll probably tidy that up a little bit and merge before trying to make it more efficient in a follow up.

fbaumdicker:
I think there might be an easy way to fix it in this new version without gc_left here. 
I didn't look at the code yet but did you add a weight of mean_gc_rate * track_length to the leftmost position in the gc_map? The distribution looks like this is missing and might cause the error.

jeromekelleher:
> I didn't look at the code yet but did you add a weight of mean_gc_rate * track_length to the leftmost position in the gc_map? The distribution looks like this is missing and might cause the error.

No, I just had a standard weight of ``gc_rate`` according to the map. That was probably the issue. There's still the problem of generating too many events then, though, isn't there?

fbaumdicker:
> There's still the problem of generating too many events then, though, isn't there?
[...]

PR #1168, opened 2020-09-09
Fix GC left process

jeromekelleher:
Reverts to the old way of doing things so that we're at least simulating the correct distribution.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -3,15 +3,20 @@ Python version of the simulation algorithm.
 """
 import argparse
 import heapq
+import logging
 import math
 import random
 import sys
 
 import bintrees
+import daiquiri
 import numpy as np
 import tskit
 
 import msprime
+
+
+logger = daiquiri.getLogger()
 
 
 class FenwickTree:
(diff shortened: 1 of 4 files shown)


jeromekelleher:
TODO: figure out what bits are redundant now and remove them.

fbaumdicker:
This looks fine. If the tests still pass this should work correctly. However, it might be possible to quickly fix the new gc_left free simulation method as outlined here https://github.com/tskit-dev/msprime/issues/1161#issuecomment-689764259

jeromekelleher:
> This looks fine. If the tests still pass this should work correctly. However, it might be possible to quickly fix the new gc_left free simulation method as outlined here #1161 (comment)

Thanks @fbaumdicker - I'll try out your idea to see if it works (easy enough to get rid of gc_left again if we don't want it). I'll merge this first though, in the interest of keeping the simulations correct in master.

jeromekelleher:
So, I won't remove the redundant bits of the gc_event yet, so we can experiment with the single-process version.

PR #1177, opened 2020-09-10
Simulate GC left as part of GC within process.

jeromekelleher:
This isn't as straightforward as we might hope @fbaumdicker  - it's easy enough to add in the extra GC mass for the first segment, but it does break the assumption that we can convert a mass to a position value when we're choosing segments. I think we could probably work around this (as I mention in the comments) but I wonder if it's worth it. We know that using a single GC process like this will result in a "many" GC events that do nothing, as the left and right breakpoints both end up being left of the left endpoint of the individual.

The question is, how much will this matter? If we do a large simulation, then won't we have lots of individuals consisting of little segments, and then the "GC null" events will end up dominating the simulation? Is there any back-of-the-envelope calculation we could do to guess the basic magnitudes of the number of events for, say, a human chromosome?

Maybe it's better to instead  focus energy on improving the GC left process? (Which, I have some fairly solid ideas for).

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -528,6 +528,7 @@ class RateMap:
         return self.total_mass / self.sequence_length
 
     def mass_between(self, left, right):
+        assert left <= right
         left_mass = self.position_to_mass(left)
         right_mass = self.position_to_mass(right)
         return right_mass - left_mass
@@ -967,11 +968,14 @@ class Simulator:
         return total_rate
 
     def get_total_gc_left_rate(self, label):
-        gc_left_total = self.get_total_gc_left(label)
+        # gc_left_total = self.get_total_gc_left(label)
         # NOTE: this is an approximation. Should we be multiplying by the
         # local GC rate when iterating over the segments?
-        mean_gc_rate = self.gc_map.mean_rate
-        return mean_gc_rate * self.track_length * gc_left_total
+        # mean_gc_rate = self.gc_map.mean_rate
+        # return mean_gc_rate * self.track_length * gc_left_total
+        # FIXME returning 0 for now to turn this event off so we can
+        # experiment with the single process.
+        return 0
 
     def get_individual_length(self, head):
         tail = head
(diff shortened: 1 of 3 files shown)


fbaumdicker:
> Is there any back-of-the-envelope calculation we could do to guess the basic magnitudes of the number of events for, say, a human chromosome?

It can be really a ton of events, especially for high gene conversion rates. In this case, all positions would recombine/convert frequently and thus null gc_events would occur at rate 0.5 * genome_length^2 * gc_rate. This might be a lot for human genomes.

fbaumdicker:
I tried using
`self.gc_map = RateMap([0, 1, self.m], [gc_rate * gc_length, gc_rate, 0])`
instead of 
`self.gc_map = RateMap([0, self.m], [gc_rate, 0])`
but wasn't able to get this working in the C version yet.

fbaumdicker:
I have, however, a few ideas to make the gc_left part independent of the right end of segments, which would be fairly easy to implement. So if you prefer this let us improve the original gc_left process.

jeromekelleher:
> I have, however, a few ideas to make the gc_left part independent of the right end of segments, which would be fairly easy to implement. So if you prefer this let us improve the original gc_left process.

Oooh, that's interesting! If we didn't need to know about the right end of the segments I'm sure we could do gc_left efficiently. Can you send me an email with the ideas, or maybe we should have a chat about it?

fbaumdicker:
I figured the idea is most easily explained using an implementation where the dependency has been removed.
I did that in https://github.com/fbaumdicker/msprime/commit/c8be1a4b3300f5adff87a4509eacdc510e4dfd61
This passes the verification tests for constant gc rates. 
But note that there are still a few things missing before I would like to create a pull request for that. But we should be able to add those in the next week(s).

The plan on what to do next would be:

1. remove the dependency on the right end --> done
2. add a flag for constant gc_maps and add fast methods to compute cleft rate in this case.
All we need is the current number of individuals/ancestors  (ideally only those with length > 1 if we want to speed this further up)
3. a) for variable gc rates compute cleft value for all positions along the genome and save it globally (this needs to be done only once in the beginning but requires the genome to be discrete)
4. b) for variable gc_rates compute the cleft value of the leftmost position (iteratively) (again only once)
4. change the current count of ancestors in get_total_gc_left and  find_cleft_individual to a sum of the values in the array from 3.+4. over all first positions of the current ancestors
5. solve 4. using a Fenwick tree instead of iterating over all individuals

Maybe we really should have a short chat about it, if you want (Wed, Thur, Fri would be possible for me).

jeromekelleher:
> I figured the idea is most easily explained using an implementation where the dependency has been removed. I did that in fbaumdicker@c8be1a4

OMG this is amazing @fbaumdicker, it's so simple! If I'm reading this right, all we're doing in gc_left is choosing an individual uniformly (which we can do much more efficiently - see the hudson CA event), and then generating a tracklength?

If so, I say we make a PR with this much immediately and worry about having non-uniform GC rates as a follow-up. This is a game-changer!

fbaumdicker:
Yes, that is all we are doing now. I was hoping you would love this way of simulating it. Thanks for the cheers. Though, the cost is that I removed the condition that gc events do not span a whole individual. Thus for long track lengths and short individuals we will create lots of non-effective gc events that do nothing. Similar to the already existing non-effective "gc within" events when both bp are between two consecutive segments. Also, note that the gene conversion code is now no longer working exactly the same way as Hudson's ms.
I think, however, that the time saved not iterating over all individuals is worth the extra events.

So I will do the following:

Implement 2. from the list above, making the gc_left part ultra-effective just as in the hudson CA events.
Create a pull request with those changes.
Create an issue that addresses the non-uniform gc rate maps with a Fenwick tree.

jeromekelleher:
Fantasic, thanks @fbaumdicker. I'm super excited about getting this in and trying out big sims with GC!

Issue #1173, opened 2020-09-09
Statistical problems with sweeps


jeromekelleher:
It looks like there are some statistical problems in the sweeps, where we are not matching up to either analytical expectations or with msms/discoal.

@gbinux, would you mind summarising the status here please (and maybe paste in a few of the verification figures to help folks see what the problems are)?

gbinux:
@jeromekelleher Still working on this but here are some updates:

1. It seems that when comparing `msp` and `discoal` sweeps we need to always explicitly specify `refsize` for `discoal` (the `-N` option). Otherwise the the selected allele frequency trajectory will not be correctly calculated. Internally, `discoal` uses 1000,000 as refsize by default.

2. I have compared the selected allele frequency trajectories generated by `msprime` and `discoal` side by side using the same parameters. If `refsize` is properly set, the trajectories will be comparable in terms of num of steps and delta t between steps.  So the we are good with trajectory generating function.

3. Similarly, I compared probability calculations for `p_coal_B`, `p_coal_b`, `p_rec_B`, and `p_rec_b`.  There seems to be certain probability inconsistency. My impression is that p_coal_B and p_caol_b calculated in msprime are almost always 8 times larger that that from `discoal`; `p_rec_B` and `p_rec_b` is two times larger. I simply divided `p_coal_B/b` by 8 and `p_rec_B/b` by 2 in `msprime.c` but the stats issue still persists. I think more time is needed to dig into this and these probability calculations are good starting places to work on.

jeromekelleher:
Excellent, thanks @gbinux. Would you mind pasting in a few of the plots output by verification.py illustrating the statistical problems here please? I'm hoping to get a few more eyes on this.

gbinux:
In the discoal original paper, a few parameters mentioned in Supplemental figure 1. With this set parameters, msms and discoal were found by the authors to generate consistent results and stats. The same set of parameters are used here to compare any two of msms, discoal and msprime for a simple selective sweeps simulation. 

The command arguments in msms format is 

`100 300 -t 40 -r 40 50000 -SF 0 0.99995 -Sp 0.5 -SaA 1000 -SAA 2000 -N 10000`

# discoal vs msprime

![image](https://user-images.githubusercontent.com/13037898/92949090-7c547380-f428-11ea-99f4-70d8c215d94d.png)

# msms vs discoal
[...]

PR #1178, opened 2020-09-11
update MsmsSweeps tests in verification.py

gbinux:
Fixed some bugs in MsmsSweeps test class and added a few tests.

The tests used to generate the plots in #1173 are:

1. `test_selective_discoal_vs_msp_use_discoal_paper_param`
2. `test_selective_msms_vs_discoal_use_discoal_paper_param`
3. `test_selective_msms_vs_msp_use_discoal_paper_param`

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -36,6 +36,9 @@ msHOT:
 discoal:
 	wget https://github.com/kern-lab/discoal/archive/v0.1.5.tar.gz
 	tar -xf v0.1.5.tar.gz
+	# To allow larger num_sites, a macro in the header file is updated
+	sed -i -e 's/^#define MAXSITES.*$$/#define MAXSITES 500000/' \
+		discoal-0.1.5/discoal.h 
 	cd discoal-0.1.5/ && make
 	cp discoal-0.1.5/discoal ./
 
--- a/verification.py
+++ b/verification.py
@@ -835,19 +835,17 @@ class MsmsSweeps(Test):
     Compare msms with msprime/discoal for selective sweeps.
 
     NOTE:
-    1. As msms and msprime use different numbers of seeds, the specified
-    seed(s) in the testing cmd string will be ignored.
-    2. Msms allows user to specify selection starting time/frequency (-SI),
-    or, alternatively, specify selectiy selection ending time/frequency (-SF);
-    msprime is able to simulate selection similar to the '-SF' option in msms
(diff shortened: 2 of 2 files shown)


data/Makefile
```diff
@@ -36,6 +36,10 @@ msHOT:
 discoal:
 	wget https://github.com/kern-lab/discoal/archive/v0.1.5.tar.gz
 	tar -xf v0.1.5.tar.gz
+	# To allow larger num_sites, a macro in the header file is updated
+	sed -e 's/^#define MAXSITES.*$$/#define MAXSITES 500000/' \
```
jeromekelleher, line 40: Could use ``sed -i`` here to do it in-place?

verification.py
```diff
@@ -845,6 +845,14 @@ class MsmsSweeps(Test):
     use h=0.5 to calculate that for Aa.
     """
 
+    def _set_cmp_to_discoal(self, cmp_to_discoal):
```
jeromekelleher, line 848: I'm not crazy about this pattern, modifying the state of the Test objects isn't a great idea for various reasons. Is there some reason we can't pass ``cmp_to_discoal`` to the calling function?

jeromekelleher (commented):
Looks good, thanks @gbinux, minor comments.

gbinux:
`test_selective_..._use_discoal_paper_param` tests match with parameters used in the `discoal` paper, but use large num_sites and are quite slow. Please let me know if I should hide them away or keep them.

jeromekelleher:
> test_selective_..._use_discoal_paper_param tests match with parameters used in the discoal paper, but use large num_sites and are quite slow. Please let me know if I should hide them away or keep them.

How slow? If it takes more than a minute, maybe we could reduce the number of sites a bit (and maybe put in a comment saying "set sites to X" to reproduce results from paper?)

gbinux:
> How slow? If it takes more than a minute, maybe we could reduce the number of sites a bit (and maybe put in a comment saying "set sites to X" to reproduce results from paper?)

longer than 1 minute. I will update them as you recommended above @jeromekelleher

jeromekelleher:
(ps. this PR is still marked as a draft so I'm assuming you don't want it merged yet @gbinux)

gbinux:
Commits were squashed into a single one (added tests into the MsmsSweeps test class and fixed a few bugs/typos.) @jeromekelleher

jeromekelleher (approved):
Great, thanks @gbinux !

Issue #977, opened 2020-04-24
Update mutation code once mutations have `time`


petrelharp:
As discussed in #946, this would make adding mutations to different time chunks of a tree sequence really possible, and would be a good idea more generally.

jeromekelleher:
Needs to get added to tskit first, https://github.com/tskit-dev/tskit/issues/513

petrelharp:
whoops, sorry, wrong repository. =)  But this will remind us to update the mutation code.

jeromekelleher:
Waiting on #1114

jeromekelleher:
We now have a time column, anyone want to fill it in? Should be a straightforward update.

mufernando:
I'll give it a go.

brianzhang01:
Hello -- chiming in with a particular question, and I hope this is an appropriate thread for it. I haven't followed all the discussions super closely but just caught up now.

I've been playing with the mutation time feature, which is awesome, and right now I've taken a tree sequence, sampled a lot of mutations, and plotted the distribution over times. Since `msprime` gave me unknown times, I used the following code:
```
temp_ts = msprime.mutate(simulation, rate=args.mu*10, random_seed=seed)
temp_tables = temp_ts.dump_tables()
temp_tables.compute_mutation_times()
temp_ts = temp_tables.tree_sequence()
mutation_times = np.array([m.time for m in temp_ts.mutations()])
histogram = np.histogram(mutation_times, interval_bins, density=False)[0]
```

However, I was surprised that `compute_mutation_times()` currently gives the midpoint time from the node below and node above, rather than a uniform sampling over the height of the branch. Am I correct in guessing that uniform sampling, because it involves randomness, is instead a feature for `msprime`? This is really exciting, thanks @mufernando and everyone else for working on it.

benjeffery:
@brianzhang01 thanks for reminding me of this. We've discussed this issue in the original mutation time work: https://github.com/tskit-dev/tskit/issues/513#issuecomment-639539212 The end result was to add as a future feature sampling uniformly, and getting round the need for random-ness by passing in a random array. However it seems there are other uses for randomness (e.g. https://github.com/tskit-dev/tskit/pull/815#issuecomment-682989666) so I'm starting to think we bite the bullet and solve the MIT-compatible source of randomness. 

Long story short - this is planned and I should have made an issue! (thought I had but can't find it!)

benjeffery:
[...]

PR #1179, opened 2020-09-11
adding mutation time

mufernando:
solves #977

TODO:

- [x] adapt C code to report back mutation times
- [x] modify mutation struct to deal with unknown times in kept mutations
- [x] add the option ``kept_mutations_before_end_time``, so that people can bypass our checks for consistency of child mutations coming before parents.
- [x] C tests for the behavior of mutation times and ``kept_mutations_before_end_time``. 
- [x] Py-C and Python infra for the new ``kept_mutations_before_end_time``.
- [x] Python tests for mutation time
    - [x] Test times are being produced
    - [x] Test allow ancestral option
    - [x] Statistical check for mutation times for not-discrete or parent-independent models (w/ allow_ancestral).
        - [x] Check exactly which mutation models are parent-independent and flag them.
    - [x] the probability that there is no more mutations on a branch after a given mutation is exp(-rate * remaining time on branch)
- [x] add mutation time to tsutil.py insert_branch_mutations
- [x] remove unknown times in tests (except to test for error)


lib/tests/test_mutations.c
```diff
@@ -78,13 +78,13 @@ insert_single_tree(tsk_table_collection_t *tables, int alphabet)
         ret = tsk_site_table_add_row(&tables->sites, 0.1, "0", 1, NULL, 0);
         CU_ASSERT_FATAL(ret >= 0);
         ret = tsk_mutation_table_add_row(
-            &tables->mutations, 0, 0, -1, TSK_UNKNOWN_TIME, "1", 1, NULL, 0);
+            &tables->mutations, 0, 0, -1, 0.0, "1", 1, NULL, 0);
```
petrelharp, line 81: we should still have unknown times in the tests, btw.

mufernando: why?

petrelharp: We will still need to add mutations to tables with unknown mutation times.

petrelharp: (I think we decided we could only do that with not-discrete mutation positions, though.)

petrelharp:
This brings up some things we'd previously postponed regarding `kept` mutations.  Notes:

- we [do not allow mixed known/unknown times](https://github.com/tskit-dev/tskit/pull/726) at a site.
- `mutate` will [throw an error](https://github.com/tskit-dev/msprime/blob/master/lib/mutgen.c#L1205) if a kept mutation is below a new mutation.
- currently in the mutation code, mutations with unknown time get assigned the time of their node, i.e., the latest they could possibly have occurred.  (this is just *internally*, on output they are still unknown)

The idea is that under most models of mutation, it doesn't really make sense to generate older mutations after younger ones. So, if you want to keep existing mutations, then one of these things ought to be true:

1. you're using infinite sites mutation (`discrete=False`), or
2. you're using a `start_time` which is after any already-existing mutations.

The handling of unknown times noted above is the right one given all this.

So, we should decide whether to sanity check input: if `keep and discrete and (max(mutations.time) < start_time)` then throw an error? (replacing mutations.time with the time of the mutations node if the time is unknown)  Probably, but we *could* just allow things to proceed, only getting the error if one of the existing mutations ends up below a new mutation.

Like I said, we should probably throw an error, but here's a situation where not throwing an error would be convenient: suppose you've got a single mutation already on your tree sequence somewhere (maybe it was a selected mutation that you simulated a sweep of) and want to throw down neutral mutations. Then it would make sense to just `mutate( )` and if you get unlucky enough to hit the site that one mutation is on, try again. The better way to do it, I suppose, would be to (a) save a copy of the site and mutation tables, (b) add mutations without keeping existing ones, then (c) add the existing site/mutation back on, dealing with coindidences somehow. I'm going to have to do something like this to add neutral mutations onto a SLiM simulation.

lib/tests/test_mutations.c
petrelharp: given the below, I guess we only need to make sure that unknown-ness is preserved at pre-existing mutations that aren't at newly mutated sites

mufernando:
ping @jeromekelleher

mufernando:
> * currently in the mutation code, mutations with unknown time get assigned the time of their node, i.e., the latest they could possibly have occurred.  (this is just _internally_, on output they are still unknown)

we will also have to decide what to do with the kept mutations which had unknown times. given now we will be outputting the times to the mutation tables, we have to know whether the kept mutation had unknown time before mutate. @petrelharp suggested we add a flag to the mutation struct to deal with this.

petrelharp:
I ran across another related TODO in the code: https://github.com/tskit-dev/msprime/blob/master/lib/mutgen.c#L307 - what we'd like to have happen here is to put `1 - time` into `origin_generation` (check that 1).

petrelharp:
> we will also have to decide what to do with the kept mutations which had unknown times.

I think that definitely we want to keep the times of input mutations - so, if they were unknown on input, they should remain unknown. The question is how to do this.

jeromekelleher:
Hmm, this is more complicated than I expected. How about we put in the strictest check we can think of that rules out awkwardness in dealing with kept mutations for now, and implement times for the simple cases first? We can then follow up with an issue/PR dealing with the awkward cases separately (making it easier to see which code is for dealing with the awkward cases).

mufernando:
Ok, so I'll throw an error when you try to `mutate()` with either `discrete=True` or `start_time` is less than any of the existing mutations time. For now, I will replace any existing kept mutations that had unknown time with the time of their node.

petrelharp:
I think you mean: when `discrete=True` *and* `start_time` is *greater* than any existing mutations' time.

mufernando:
yes! I meant not throw an error when discrete=False or start_time < kept mutations times.

petrelharp:
Minor correction: not throw an error when discrete=False or start_time <= all kept mutations times.

tests/test_mutations.py
```diff
@@ -1570,8 +1572,11 @@ def cmp_mutation(a, b):
     # in order already). Kept mutations are given an id that is
     # their order in the initial tables, and new mutations have id -1.
     out = a.id * (not a.new) - b.id * (not b.new)
-    if out == 0:
+    either_unknown = tskit.util.is_unknown_time(a.time) or tskit.util.is_unknown_time(b.time)
```
petrelharp, line 1575: Maybe your'e in the middle of this, but:  because of the `unknown_time` flag, and sticking in the node.time, aren't these times never unknown?

mufernando: yes! this is a relic I need to get rid of

petrelharp:
Note: we should also put `time` information into the `slim_time` of the metadata in the SlimMutationModel. (After rounding, and multiplying by -1, or something?) But - we should deal with this separately, after this goes through.

lib/tests/test_mutations.c
mufernando: done

lib/mutgen.c
```diff
@@ -42,6 +42,7 @@ cmp_mutationp(const void *a, const void *b)
      * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81631 */
     const mutation_t *ia = *(const mutation_t **) (uintptr_t) a;
     const mutation_t *ib = *(const mutation_t **) (uintptr_t) b;
+    // if either mutation is new, then sort by time (or node time if unknown).
```
mufernando, line 45: remove this

lib/mutgen.c
```diff
@@ -271,7 +272,8 @@ mutation_matrix_free(mutation_model_t *self)
  *  int32_t mutation_type_id_; // 4 bytes (int32_t): the id of the mutation type the
  *                             // mutation belongs to
  *  float selection_coeff_;    // 4 bytes (float): the selection coefficient
- *  int32_t subpop_index_; // 4 bytes (int32_t): the id of the subpopulation in which the
+ *  int32_t subpop_index_; // 4 bytes (int32_t): the id of the subpopulation in which
+ * the
```
mufernando, line 276: undo this

lib/mutgen.c
```diff
@@ -980,14 +983,16 @@ mutgen_add_existing_mutation(mutgen_t *self, site_t *site, tsk_id_t id, tsk_id_t
 }
 
 static int MSP_WARN_UNUSED
-mutgen_initialise_sites(mutgen_t *self)
+mutgen_initialise_sites(mutgen_t *self, bool discrete_sites, bool allow_ancestral)
```
petrelharp, line 986: instead just an argument `check_ancestral`

lib/mutgen.c
```diff
@@ -1201,7 +1219,8 @@ mutgen_choose_alleles(mutgen_t *self, tsk_id_t *parent, mutation_t **bottom_muta
         } else {
             parent_mut = bottom_mutation[u];
             mut->parent = parent_mut;
-            if (mut->time > parent_mut->time || (parent_mut->new && !mut->new)) {
+            if (!allow_ancestral
+                && (mut->time > parent_mut->time || (parent_mut->new && !mut->new))) {
```
petrelharp, line 1223: move `mut->time > parent_mut->time` to an assert. Or delete it since we'll find out about the error later - we won't have a valid tree sequence.

petrelharp: oh delete this whole thing since we can't trigger it any more

mufernando: it is being triggered by one of the tests, check which.

lib/tests/test_mutations.c
```diff
@@ -186,9 +186,12 @@ test_mutgen_errors(void)
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     ret = mutgen_set_time_interval(&mutgen, 0.5, 10.0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
+    /*TODO: Change the check in initialise_sites to not be triggered with
```
mufernando, line 189: remove todo

lib/tests/test_mutations.c
```diff
@@ -541,6 +546,78 @@ test_single_tree_mutgen_keep_sites_many_mutations(void)
...
+        &tables.mutations, 1, 0, -1, TSK_UNKNOWN_TIME, "1", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret >= 1);
+
+    gsl_rng_set(rng, 2);
+    ret = mutgen_alloc(&mutgen, rng, &tables, &mut_model, 1);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = mutgen_set_rate(&mutgen, 1);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    /* should error out with discrete times and keeping mutations above
```
petrelharp, line 576: *discrete sites

lib/tests/test_mutations.c
```diff
@@ -541,6 +546,78 @@ test_single_tree_mutgen_keep_sites_many_mutations(void)
...
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = mutgen_set_rate(&mutgen, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_site_table_add_row(&tables.sites, 0.0, "0", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret >= 0);
+    ret = tsk_mutation_table_add_row(
+        &tables.mutations, 0, 0, -1, TSK_UNKNOWN_TIME, "1", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret >= 0);
+    ret = mutgen_generate(
+        &mutgen, MSP_DISCRETE_SITES | MSP_KEEP_SITES | MSP_ALLOW_ANCESTRAL_MUTATIONS);
```
petrelharp, line 612: check some mutations got added

lib/util.c
```diff
@@ -199,7 +199,7 @@ msp_strerror_internal(int err)
         case MSP_ERR_MUTATION_GENERATION_OUT_OF_ORDER:
             ret = "Tree sequence contains mutations that would descend from "
                   "existing mutations: finite site mutations must be generated on "
-                  "older time periods first.";
+                  "older time periods first";
```
mufernando, line 202: fix period

msprime/mutations.py
```diff
@@ -763,11 +767,18 @@ def mutate(
     0/1 states. By default mutations from the infinite sites model with a binary
     alphabet are generated.
 
-    By default, sites and mutations in the parameter tree sequence are
+    By default, sites and mutations in the input tree sequence are
     discarded. If the ``keep`` parameter is true, however, *additional*
     mutations are simulated. Under the infinite sites mutation model, all new
     mutations generated will occur at distinct positions from each other and
-    from any existing mutations (by rejection sampling).
+    from any existing mutations (by rejection sampling). Furthermore, by
```
petrelharp, line 774: modify to explain start_time condition instead

msprime/mutations.py
```diff
@@ -795,7 +806,10 @@ def mutate(
     :param bool discrete: Whether to generate mutations at only integer positions
         along the genome.  Default is False, which produces infinite-sites
         mutations at floating-point positions.
-    :return: The :class:`tskit.TreeSequence` object  resulting from overlaying
+    :param bool allow_ancestral: Whether to allow mutations to be added
+        ancestrally to existing (kept) mutations. This flag has no effect
```
petrelharp, line 810: modify this to "whether to allow start time before existing mutations" or somethign

petrelharp: change to "allow_kept_mutations_after_start_time" or "kept_after_start_time_ok" "kept_after_start_time"...?

tests/test_mutations.py
```diff
@@ -441,7 +440,7 @@ def mutate(self, ts, model, rate=1.0, keep=False, discrete=False, seed=42):
         lib_tables.provenances.clear()
         py_tables = py_ts.dump_tables()
         py_tables.provenances.clear()
-        self.assertEqual(lib_tables, py_tables)
+        self.assertEqual(lib_tables.mutations, py_tables.mutations)
```
petrelharp, line 443: change back

petrelharp: add allow-ancestral to this method?

tests/test_mutations.py
```diff
@@ -919,6 +918,17 @@ def test_keep_mutation_parent_zero_rate(self):
...
+        ts = msprime.simulate(12, recombination_rate=3, random_seed=3)
+        ts_mut = msprime.mutate(ts, rate=1, random_seed=1, discrete=True)
+        self.assertGreater(ts_mut.num_sites, 0)
+        with self.assertRaises(_msprime.LibraryError):
+            msprime.mutate(ts_mut, rate=1, random_seed=1, keep=True, discrete=True)
+        ts_2mut = msprime.mutate(
+            ts_mut, rate=1, random_seed=3, discrete=True, allow_ancestral=True
+        )
+        self.assertGreater(ts_2mut.num_mutations, ts_mut.num_mutations)
+
```
petrelharp, line 931: test the error is thrown with another few start_time / mutations combinations: start_time = Inf,  mutation time is unknown and node is isolated, mutation time is before start time but node time is after (should not error)

tests/test_mutations.py
```diff
@@ -1077,14 +1089,49 @@ def verify_mutation_rates_ts(self, ts, model, rate, discrete):
...
+    def verify_mutation_times_ts(self, ts):
+        start = np.full(ts.num_mutations, -1, dtype=np.float32)
+        end = np.full(ts.num_mutations, -1, dtype=np.float32)
+        time = np.full(ts.num_mutations, -1, dtype=np.float32)
+        for mut in ts.mutations():
+            time[mut.id] = mut.time
+            start[mut.id], end[mut.id] = self.find_start_end_times(
+                ts, mut.node, mut.site
+            )
```
petrelharp, line 1113: could do something like:
```
>>> for t in ts.trees():
...   for mut in t.mutations():
...     end_time = t.time(mut.node)
...     start_time = t.parent(mut.node).time
```
but this looks correct

tests/test_mutations.py
```diff
@@ -1077,14 +1089,49 @@ def verify_mutation_rates_ts(self, ts, model, rate, discrete):
...
+        assert np.all(start >= 0)
+        assert np.all(end >= 0)
+        assert np.all(time >= 0)
+        greater_end = end > start
```
petrelharp, line 1117: remove

tests/test_mutations.py
```diff
@@ -1077,14 +1089,49 @@ def verify_mutation_rates_ts(self, ts, model, rate, discrete):
...
+        centered_time = time[greater_end] - start[greater_end]
+        time_inter = end[greater_end] - start[greater_end]
+        rng = random.Random(3937)
+        generated_times = np.array([rng.uniform(0, high) for high in time_inter])
+        if centered_time.shape[0] > 0:
```
petrelharp, line 1122: instead do a sign test for how oftern generated times > time_inter

tests/test_mutations.py
```diff
@@ -1077,14 +1089,49 @@ def verify_mutation_rates_ts(self, ts, model, rate, discrete):
...
+            res = stats.ks_2samp(centered_time, generated_times)
+            print(res)
+            self.assertGreater(res[1], 0.01)
+
     def test_binary_model(self):
         model = msprime.BinaryMutations()
         self.verify_model(model)
         self.verify_mutation_rates(model)
```
petrelharp, line 1130: state_independent

petrelharp: also arbitrary model below

mufernando:
When the column `time` was added to the MutationTable we decided that we would not allow mixed unknown and known times at a site. See [docs](https://tskit.readthedocs.io/en/latest/data-model.html#sec-mutation-requirements) and [PR](https://github.com/tskit-dev/tskit/pull/726#issuecomment-663157912).

The goal of this PR is to make use of the time column in msprime simulations. `msprime.mutate()` allows existing mutations to be kept in the tree sequence, which can in theory hav unknown times. How should we deal with mixed unknown and known times in msprime?

- Do not allow unknown times in any of the kept mutations have unknown times?
- Error out if any of the new generated mutations fall at a site with any existing unknown times?

The first option is much more clear the user, because the error will not be stochastic. However, this could be too restrictive and break a lot of people's codes.

ping @jeromekelleher @benjeffery @hyanwong

benjeffery:
Really appreciate the clear statement of the problem!

I wonder if a third option might be good where msprime refuses to even try to add times to a TS with unknown times?

petrelharp:
> I wonder if a third option might be good where msprime refuses to even try to add times to a TS with unknown times?

I think that's the first option - throw an error if the user tries to add mutations to a ts with any unknown mutation times?

Also, btw: @mufernando did the hard work of retaining existing unknown times (this would make sense if, for instance, all new mtuations happened below the branch the mutations were on), before we remembered about our requirement that there be no mixing of known/unknown times at a site.

jeromekelleher:
So is the issue that there will be tree sequences out there from older versions of SLiM that will not have times associated with mutations, and we want to be able to accommodate those? I wonder if this is something that could/should be handled at the pyslim layer, i.e., pyslim reads in the data, checks for older versions and if so fills in sensible times for the mutations (which I presume are held in the metadata?)

petrelharp:
I don't think there's an issue with SLiM - the new SLiM will assign mutation times, and it's pretty much ready to go, so we're all good there. If someone does really want to use a previously-generated tree sequence (from msprime or SLiM!) and keep mutations, then they can go ahead and use [compute_mutation_times](https://tskit.readthedocs.io/en/latest/python-api.html?#tskit.TableCollection.compute_mutation_times) to add in times.

The question is about the API: should we throw an error early, because we know that the requested procedure *could* generate illegal mixed-type site? or do we wait to see if the procedure actually *does* generate an illegal site before throwing the error? I'd say to definitely throw the error early, except that for instance if someone has a tree sequence with unknown-time mutations at floating-point positions, and wants to add mutations at integer positions, then this would in practice be fine but we'd still throw an error. But, I can't imagine this actually being a good part of a workflow, so meh.

tests/test_mutations.py
mufernando: I added more time tests, but am waiting for a decision on what to do with unknown time mutations.

tests/test_mutations.py
```diff
@@ -1096,14 +1126,52 @@ def verify_mutation_rates_ts(self, ts, model, rate, discrete):
...
+            for mut in t.mutations():
+                mut_time[mut.id] = mut.time
+                end_time[mut.id] = t.time(t.parent(mut.node))
+                start_time[mut.id] = t.time(mut.node)
+                if mut.parent != tskit.NULL:
+                    end_time[mut.id] = min(
+                        end_time[mut.id], ts.mutation(mut.parent).time
+                    )
+                    start_time[mut.parent] = max(
+                        start_time[mut.parent], t.time(mut.node)
```
petrelharp, line 1143: found it!! This line should be
```
                        start_time[mut.parent], mut.time                 

```

jeromekelleher:
I say error early and just raise an error if there's any unknown times at all. We can always relax the restriction later if it turns out to be something someone actually needs to do.

mufernando:
I think I'm done. One thing we might want to change is the flag `kept_mutations_before_end_time`. Is the name of the flag clear enough? We could do `allow_kept_mutations_before_end_time` but that seems too long...

petrelharp:
Yay, excellent!!!

We do need a different name for that flag, though. How about `allow_kept_as_derived`? Or even just `check_kept_times`, as in the C code - you need to read the docs to see what is being checked, but that's OK.

jeromekelleher:
Yeah, this is looking great, thanks @mufernando!

The name of the flag is a bit clumsy, and doesn't really communicate what's happening to me. I've been wondering if "keeping" mutations is the right way to think about this, anyway. What if we changed the sense of this so that ``mutate`` expects to get an empty site/mutation table by default, and throws an error unless this flag is set. Then, we can talk about "existing" sites or "input" sites/mutations rather than "kept" (which I keep seeing as "klept", making it feel somewhat sinister!). I don't know if this helps much, but "keeping" is something that's been low-level bothering me for a bit.

petrelharp:
I like that, @jeromekelleher - that'd avoid the current problem where it's easy to accidentally overwrite existing mutations when you meant to do `keep=True`. We could even just change `keep` to `add`, since then you're *adding* mutations? And `allow_existing_derived?

jeromekelleher:
Sounds good to me @petrelharp. How about we merge this much, and file an issue to track the renaming?

jeromekelleher:
This one went in with 40 commits @petrelharp - it would have been nice to squash it down to something more meaningful first. Future-us will need deal with git bisect landing on non-passing commits. 

Also, it's better to use the "AUTOMERGE-REQUESTED" label rather than pressing the "merge" button manually - that way we know the branch is up to date and that the tests will definitely pass when the branch is merged into main.

petrelharp:
Damn it, sorry, @mufernando just pointed that out also. =( =( =(

mufernando:
Can we just revert this merge and do it again?

jeromekelleher:
No worries! @benjeffery - is it worth doing a bit of retrospective git surgery here?

jeromekelleher:
> Can we just revert this merge and do it again?

No, the revert button is pointless, unfortunately. It just makes another commit which undoes the previous one.

mufernando:
related to this, part of why this history is so messy is because we don't have much guidance on how to use clang on the C side of msprime. If I just mindlessly use `clang-format` on some C files it changes a bunch of stuff. idk if it's worth it to open an issue for this though?

benjeffery:
Ok, I'm going to jump in and fix this, it's only been 1 hour and we'll let everyone know on slack in case they have pulled.

jeromekelleher:
> related to this, part of why this history is so messy is because we don't have much guidance on how to use clang on the C side of msprime. If I just mindlessly use clang-format on some C files it changes a bunch of stuff. idk if it's worth it to open an issue for this though?

Yeah, this isnt' great @mufernando - would you mind opening an issue with the problems you hit?

benjeffery:
Right, I've undone this PR. Please open a new one! :+1:

Issue #1024, opened 2020-05-12
Documentation for mutation models


jeromekelleher:
This is a meta-issue which can be closed when we have 'good' documentation for general mutation models. This means

- [ ] Thorough API documentation for all implemented models
- [ ] A quick lookup table listing the models implemented with links to the model documentation
- [ ] Some sort of tutorial

See also #950

jeromekelleher:
I think we can close this now.

PR #1176, opened 2020-09-10
Refactor mutation model API to add string references

benjeffery:
Fixes #1070.
Fixes #1189

PR #1180, opened 2020-09-12
Mutation model docs

petrelharp:
Here I've added some documentation for
- some of the mutation matrix models
- the infinite alleles model
- the slim mutation model
so is a step towards #1024.


petrelharp:
Whoops - initially I had some commits off a different PR in here. Took those out.

msprime/mutations.py
```diff
@@ -131,9 +153,42 @@ def __init__(self):
...
+    :math:``1/(mu * P[i,j])`` generations. In this model, :math:``\\kappa`` is
+    the transition:transversion ratio and :math:``\\pi`` is the vector of equilbrium
+    frequencies, so that we define :math:``P`` by first defining
+    :math:``p[i,j] = \\pi[j]`` if :math:``i \to j`` is a transversion, and
+    ``p[i,j] = \\kappa \\pi[j]`` otherwise, and then letting
+    :math:``P[i,j] = p[i,j] / M`` where :math:``M = 1 + (\\gamma - 1) \\max(\\pi)``
+    is the largest row sum of :math:``p``. This means that if you mutate with the
+    HKY model with ``mutation_rate`` equal to :math:``\\mu``, then the mutation
+    rate from "A" to "G", for instance, is
+    :math:``\\mu \\kappa \\pi[G] / (1 + (\\gamma - 1) \\max(\\pi))``.
```
petrelharp, line 171: This lengthy explanation of how to get the actual mutation rates from the parameters is kinda unfortunate. You must have come across this, @GertjanBisschop - what do you think? We could change the mutation generation code to avoid this, by not requiring the the transition probability matrix have row sums equal to one (and doing the renormalization at the C level), or adding an extra "rate multiplier" property to mutation models to make this `M` go away? Or, probably, we should just say it's a minor annoyance and leave well enough alone?

petrelharp: A few more thoughts: I was looking at [the HKY model on Wikipedia](https://en.wikipedia.org/wiki/Models_of_DNA_evolution#HKY85_model_(Hasegawa,_Kishino_and_Yano_1985)), and thinking that their rate matrix is much nicer: e.g. Q[A, C] = pi[A] instead of mu * pi[A] / (1 + (gamma-1) * max(pi)). But, Wikipedia is *not* giving the rate matrix, just giving it up to an overall constant.  So, Wikpedia's parameterization is Q[A,C] = mu * pi[A], which is simpler than ours, but we just have a somewhat different interpretation of mu - their mu is our mu / M. So, I think maybe our parameterization is fine?

petrelharp: I moved this to docs/api.rst, so that I could get it to compile, and rewrote it, btw.

petrelharp: I'm trying to figure out if our parameterization *does* differ from seq-gen's? I see that kappa differs by a factor of 2, which I understand, but how about the overall mutation rate? Is this something you got to the bottom of, @GertjanBisschop? Gah, I wish seq-gen documented what it's doing.

docs/api.rst
```diff
@@ -449,8 +454,182 @@ with instantaneous transition rate from ``i`` to ``j`` that is equal to
...
+root distribution ``[1/4, 1/4, 1/4, 1/4]``, and transition matrix
+``[[0, 1/3, 1/3, 1/3], [1/3, 0, 1/3, 1/3], [1/3, 1/3, 0, 1/3], [1/3, 1/3, 1/3, 0]]``.
+It has no free parameters.
+
+Citation: *Jukes TH, Cantor CR (1969). Evolution of Protein Molecules. pp. 21-132.*
+
+.. autoclass:: msprime.HKY
+
+TODO: move to msprime/mutations.py
+
```
petrelharp, line 487: Ok, the following is correct now, I believe. Note that the discussion below is somewhat complicated by the scaling factor `M`, required to make `P` a valid transition matrix. It looks like seq-gen avoids this issue by not documenting what the overall mutation rate actually is. Do you know, @GertjanBisschop - what's the overall mutation rate for seq-gen under HKY?  Are we testing for equality of total number of mutations or just relative frequencies of different transition types?  And, how does this description seem to you - clear? correct?

We *could* make the discussion below so that the mutation rates are just `mu * q[i,j]`, if we added a "rate scaling factor" to the mtuation model object, and put M in there, transparent to the user. But I think this is probably not worth it - what do you think?

petrelharp:
Ok, I think this is good for this PR? There's bigger picture rearranging that needs to happen, pending the API changes, but these can go in there somehow.

petrelharp:
FYI, to review the docs, check out this branch and then `make` them locally:
```
git fetch origin pull/1180/head:mutation_model_docs
git checkout mutation_model_docs
cd docs
make
```
and then look at the html files in `docs/_build/html`

docs/api.rst
```diff
@@ -449,8 +454,345 @@ with instantaneous transition rate from ``i`` to ``j`` that is equal to
...
+:math:`M = \max_i \left( \sum_{j \neq i} q[i,j] \right)`. Then, this implementation as
+a :class:`.MatrixMutationModel` has the transition matrix
+:math:`P[i,j] = q[i,j] / M`, and :math:`P[i,i]` chosen so rows sum to one.
+
+Citation: *Felsenstein J, Churchill GA (January 1996). "A Hidden Markov
+Model approach to variation among sites in rate of evolution". Molecular
+Biology and Evolution. 13 (1): 93–104.*
+
+Note that this implementation has the root distribution as a separate
+parameter, although it defaults to the equlibrium distribution. If you
```
jeromekelleher, line 593: typo, equlibrium

docs/api.rst
```diff
@@ -449,8 +454,345 @@ with instantaneous transition rate from ``i`` to ``j`` that is equal to
...
+Protein Sequence Structures, Vol. 5, Suppl. 3, National Biomedical Research
+Foundation, Washington DC, pp. 345-352.*
+
+
+++++++++++++++++++++++++++++++++++++
+Defining your own finite-sites model
+++++++++++++++++++++++++++++++++++++
+
+If you want to define your own :class:`.MatrixMutationModel`, you have a good
+deal of freedom. For instance, here's a "decomposition/growth/disturbance"
```
jeromekelleher, line 740: Haha, love it!

docs/api.rst
```diff
@@ -536,8 +878,126 @@ which can be computed as follows:
...
+++++++++++++++++++++++++++++++++
+Infinite Alleles Mutation Models
+++++++++++++++++++++++++++++++++
+
+You can also use a model of *infinite alleles* mutation: where each new mutation produces a unique,
+never-before-seen allele. The underlying mutation model just assigns the derived state
+to be a new integer every time a new mutation appears.
+By default these integers start at zero, but a different starting point can be chosen,
+with the ``start_allele`` argument.
+It does this separately for each site, so that the ancestral state of every site will be the ``start_allele``,
```
jeromekelleher, line 895: We'll need to update this for #1182

jeromekelleher (commented):
Looks great, thanks @petrelharp!

docs/api.rst
GertjanBisschop: This is great work. Will review this and answer all your questions tonight. Whichever way we decide on the scaling factor `M`, we will have one up on SeqGen given that this is all now very well documented.

docs/api.rst
petrelharp: following your lead here... =)

docs/api.rst
petrelharp: done in 6f071b6

docs/api.rst
```diff
@@ -449,8 +454,345 @@ with instantaneous transition rate from ``i`` to ``j`` that is equal to
...
+and follow the "Dayhoff DCMut" model as described in *Kosiol, C., and Goldman,
+N. 2005. Different versions of the Dayhoff rate matrix. Molecular Biology and
+Evolution 22:193-199.* The original citation is: *Dayhoff, M.O., Schwartz,
+R.M., Orcutt, B.C. (1978). A model of evolutionary change in proteins. Atlas of
+Protein Sequence Structures, Vol. 5, Suppl. 3, National Biomedical Research
+Foundation, Washington DC, pp. 345-352.*
+
+
+++++++++++++++++++++++++++++++++++++
+Defining your own finite-sites model
```
GertjanBisschop, line 736: Looking at the different subsections of 'simulating mutations', I was wondering whether this section could be called 'Defining your own Matrix Mutation model', just to make the sequence of paragraphs more logical. Can these really only be finite-sites (not a comment, just a question)?

GertjanBisschop (commented):
Still really love the *Parameterization of Matrix Mutation Models*. This is super informative!

docs/api.rst
petrelharp: > But I would argue against adding the rescaling factor to the mutation model object. I

Oh, good. =)

> If I understand things correctly, you would never allow the user, defining their own MatrixMutationModel to specify any M value, right?

The way I'd imagine it would be that the MatrixMutationModel would have a `rate_scaling` argument ("M"), that gets multiplied by the mutation rate when it is used with mutate( ). So, yes, the user would be able to set this, but only if they are doing their own model. Then, if we defined HKY as a MatrixMutationModel(... rate_scaling=M), then we'd be able to delete all references to M from the documentation here.

docs/api.rst
petrelharp: Oh, whoops - I meant to write "finite alleles" model. But yes, "matrix mutation" model would be better.

petrelharp:
Should I wait on more comments here, @GertjanBisschop? Otherwise, I'll wait on #1176 and then tidy this up.

GertjanBisschop:
No more comments. Sorry @petrelharp, should have made that clearer.

docs/api.rst
benjeffery: This is amazing.

petrelharp:
Ok, I've merged together the docs from #1176. I think this is reasonable, but suggestions of other ways to organize it are welcome.

jeromekelleher (approved):
This is a massive step forward, thanks @petrelharp! LGTM.

I think this is ready to merge, @benjeffery?

docs/api.rst
```diff
@@ -454,24 +467,193 @@ in these models see :ref:`sec_api_mutation_matrix_models_details`.
...
+equal to :math:`Q_{ij}`, where
+
+:math:`\mathbf{Q} =
+\frac{\mu}{M} \times
+\begin{bmatrix}
+\cdot        &  \pi_C         & \left(1 + \frac{\kappa-1}{\pi_A + \pi_G}\right) \pi_G & \pi_T           \\
+\pi_A        &  \cdot         & \pi_G        & \left(1 + \frac{\kappa-1}{\pi_A + \pi_G}\right) \pi_T    \\
+\left(1 + \frac{\kappa-1}{\pi_A + \pi_G}\right) \pi_A &  \pi_C         & \cdot        & \pi_T           \\
+\pi_A        &  \left(1 + \frac{\kappa-1}{\pi_A + \pi_G}\right) \pi_C  & \pi_G        & \cdot
+\end{bmatrix}`
```
benjeffery, line 535: The equations here are not the same as in the code, I _think_ the code has this:
```suggestion
\begin{bmatrix}
\cdot        &  \pi_C         & \left(1 + \frac{\kappa-1}{\pi_A + \pi_G}\right) \pi_G & \pi_T           \\
\pi_A        &  \cdot         & \pi_G        & \left(1 + \frac{\kappa-1}{\pi_C + \pi_T}\right) \pi_T    \\
\left(1 + \frac{\kappa-1}{\pi_A + \pi_G}\right) \pi_A &  \pi_C         & \cdot        & \pi_T           \\
\pi_A        &  \left(1 + \frac{\kappa-1}{\pi_C + \pi_T}\right) \pi_C  & \pi_G        & \cdot
\end{bmatrix}`
```

msprime/mutations.py
```diff
@@ -308,8 +333,24 @@ def __init__(self, kappa, equilibrium_frequencies=None, root_distribution=None):
...
+    parameterization of a time-reversible mutation process (Tavaré et al.
+    1986). It allows specification of per-nucleotide equilibrium frequencies
+    and equilibrium transition rates.
+
+    See the documentation for more detail on the precise
+    parameterization.
+
+    :param array-like[float] relative_rates: Controls the relative rates of
+        mutation between each pair of nucleotides, in order "A<->C", "A<->T",
+        "C<->G", "C<->T", and "G<->T".
```
benjeffery, line 346: This part is missing a set of nucleotides:
```suggestion
    :param array-like[float] relative_rates: Controls the relative rates of
        mutation between each pair of nucleotides, in order "A<->C", "A<->G", "A<->T",
        "C<->G", "C<->T", and "G<->T".
```

docs/api.rst
```diff
@@ -454,24 +467,193 @@ in these models see :ref:`sec_api_mutation_matrix_models_details`.
...
+Evolution 22:193-199.* The original citation is: *Dayhoff, M.O., Schwartz,
+R.M., Orcutt, B.C. (1978). A model of evolutionary change in proteins. Atlas of
+Protein Sequence Structures, Vol. 5, Suppl. 3, National Biomedical Research
+Foundation, Washington DC, pp. 345-352.*
 
-.. autoclass:: msprime.InfiniteAllelesMutationModel()
 
-.. autoclass:: msprime.SLiMMutationModel()
 
 .. _sec_api_mutation_matrix_models_details:
```
benjeffery, line 658: I think this is my bad, but this sections title should be "Matrix Mutation Models Details" not  "Mutation Matrix Models Details"

msprime/mutations.py
```diff
@@ -164,13 +170,17 @@ class InfiniteAllelesMutationModel(
...
+    allele": each time the model is asked to produce a new allele (either for an
+    ancestral or derived state), the "next allele" is provided, and then
+    incremented.
+
+    :param int start_allele: The nonnegative integer to start assigning alleles from.
+       (default: 0)
+    :ivar int next_allele: The *next* allele to be assigned. This increments by
+        one each time an allele is assigned to a new mutation, and resets to
+        ``start_allele`` each time a new ancestral state is assigned.
+    :param int block_size: TODO
```
benjeffery, line 183: I don't see any reference to `block_size` in the code for this class, unless I missed something inherited?

msprime/mutations.py
```diff
@@ -213,25 +234,26 @@ def __init__(self):
...
     The Hasegawa, Kishino and Yano mutation model (Hasegawa et al. 1985). Based on the
-    standard ACGT nucleotides as alleleic states, this model takes into account
+    standard ACGT nucleotides as alleleic states, this model allows different rates for
     transitions and transversions, and sets an equilibrium frequency for each nucleotide.
     In addition a custom ancestral frequency (``root_distribution``) can be specified.
     With ``kappa=1.0`` and the default values of the other arguments this model is equal
     to :class:`JC69MutationModel`. This model is similar to :class:`F84MutationModel`
     but with a differing parametrisation for ``kappa``.
 
+    See the documentation for more detail on the precise parameterization.
```
benjeffery, line 244: There's quite a few places where it says to "see the documentation" Would be nice to say the section name (or URL?) as I'm not sure I'd know where to look otherwise.

benjeffery (commented):
This is looking great @petrelharp. (I'm learning a lot about mutation models!)
Just a few comments.

msprime/mutations.py
petrelharp: Right - I wasn't sure how to handle this, since on RTD we'd want to say "See below"; while in `help(msprime.HKYMutationModel)` we want to say "see https://msprime.readthedocs.io/en/latest/api.html#msprime.HKYMutationModel"... which on RTD would link to the very same place that the person is reading.

Alternatively, we could just put all of the details, that I've now got after the short description on RTD, into the docstring itself? How about that - or is that too much information?

docs/api.rst
petrelharp: whoops! good catch.

petrelharp:
Ah, one more thing: the `block_size` argument to the SLiM model is still undocumented. And, sadly, I still don't know quite what it does. It's the block size for the tsk_blkalloc... but, what happens if it is wrong? If it is too small, will the user get an error? When would the user want to adjust this number?

jeromekelleher:
The reasoning for keeping ``block_size`` is given [here](https://github.com/tskit-dev/msprime/blob/master/lib/mutgen.c#L570). Basically, we have to have it because in principle the amount of memory each slim mutation requires is unbounded, and we can't alloc more than the ``block_size`` in one go using ``tsk_blkalloc``. So, the user shouldn't every need to worry about this unless their model is pretty daft and has 8K worth of mutations stacked on top of each other.

It's too much of a tedious faff to do the memory allocations via direct mallocs, so I think we're stuck with it.

petrelharp:
I read that, but still wasn't sure what happens if it's too small. But, come on peter, it's easy to check... I see: a `_msprime.LibraryError: Out of memory.`. I'll say "You should not need to change this unless you get a "out of memory" error due to a very large number of stacked mutations."

jeromekelleher:
Looks like this is ready to go, other than the discussion about how to refer to the documentation? I don't really have much of an opinion on that, so happy to merge whenever.

msprime/mutations.py
benjeffery: I think the second option, I don't have any issue with large docstrings, if they are all useful info.

petrelharp:
I moved the docs, and fixed up a bit in the F84 documentation that I think was only correct for the case of equal base frequencies. I think this is good to go!

jeromekelleher:
Great! Merging.

PR #1182, opened 2020-09-13
don't restart alleles at every site for infinite alleles

### LINKED ISSUE #1181: InfiniteAllelesMutationModel may duplicate alleles when adding to previously-mutated sequence
The way the InfiniteAllelesMutationModel works is by (a) whenever it's asked to make a root state, it [sets the "next" allele to the "start" allele](https://github.com/tskit-dev/msprime/blob/master/lib/mutgen.c#L474),a nd then (b) whenever it's asked to make a derived state, it [returns the 'next' allele and increments it](https://github.com/tskit-dev/msprime/blob/master/lib/mutgen.c#L461). The result, when mutating an unmutated tree sequence, is that at each site the root state is `start_allele` and a site with `n` mutations will have alleles `start_allele`, `start_allele + 1`, ... , `start_allele + n`.  However, if we use it to mutate an already-mutated tree sequence, then if it mutates an already-mutated site, the 'next' allele doesn't get reset, and so  the counter stays at wherever it was for the last site.

This is a problem because it could produce more than one identical allele at the same site. For instance:
```
t = tskit.TableCollection(sequence_length=1)
t.nodes.add_row(time=0)
t.nodes.add_row(time=10)
t.sites.add_row(ancestral_state='0', position=0)
t.mutations.add_row(derived_state='1', node=1, site=0)
t.edges.add_row(parent=1, child=0, left=0, right=1)

model = msprime.InfiniteAllelesMutationModel()
mts = msprime.mutate(t.tree_sequence(), rate=1, model=model, start_time=2, random_seed=1, keep=True, discrete=True)
print(mts.tables.mutations)
```
produces
```
id	site	node	time	derived_state	parent	metadata
0	0	1	nan	1	-1	
1	0	0	nan	0	0	
2	0	0	nan	1	1	
3	0	0	nan	2	2	
4	0	0	nan	3	3	
```

Maybe this is ok, and we just say that this model only makes sense when not keeping existing mutations. But I think a more natural option is to do like the SlimMutationModel does, having a *global* (not site-specific) counter. Then we could do like
```
model = msprime.InfiniteAllelesMutationModel()
mts = msprime.mutate(ts, rate=1, model=model, discrete=True, start_time=1)
next_model = msprime.InfiniteAllelesMutationModel(start_allele=model.next_allele)
mmts = msprime.mutate(mts, rate=1, model=model, discrete=True, end_time=1)
```
and we'd be guaranteed to keep getting new alleles. It's kinda nice to have alleles 0...n at a site, but oh well? The important thing here is *infinite alleles*.  Any objections to making this switch?

Note that this proposal would still require user input to ensure the alleles were actually unique. We could, say, assign alleles to be random floats in [0,1] and thus almost always unique, but that would I think only be natural to probabilists.

petrelharp:
Closes #1181.

The only thing I'm not sure about is the test for mutation overflow - yes, it overflows? Do we need that test? Perhaps we should be detcting overflow and throwing an error?

### DIFF
--- a/lib/mutgen.c
+++ b/lib/mutgen.c
@@ -425,6 +425,7 @@ infinite_alleles_print_state(mutation_model_t *self, FILE *out)
     infinite_alleles_t params = self->params.infinite_alleles;
     fprintf(out, "infinite alleles:\n");
     fprintf(out, "\tstart_allele:%" PRIu64 "\n", params.start_allele);
+    fprintf(out, "\tnext_allele:%" PRIu64 "\n", params.next_allele);
 }
 
 /* The maximum number of digits for an unsigned 64 bit integer is 20
@@ -469,9 +470,6 @@ static int
 infinite_alleles_choose_root_state(
     mutation_model_t *self, gsl_rng *MSP_UNUSED(rng), site_t *site)
 {
-    infinite_alleles_t *params = &self->params.infinite_alleles;
-
-    params->next_allele = params->start_allele;
     return infinite_alleles_make_allele(
         self, &site->ancestral_state, &site->ancestral_state_length);
 }
@@ -612,6 +610,7 @@ infinite_alleles_mutation_model_alloc(
         goto out;
     }
     params->start_allele = start_allele;
+    params->next_allele = start_allele;
 out:
     return ret;
 }
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -2595,13 +2595,12 @@ class TestInfiniteAllelesMutationModel(unittest.TestCase):
         for start_allele in [0, 10, 2 ** 64 - 1]:
             iamm = _msprime.InfiniteAllelesMutationModel(start_allele)
             self.assertEqual(iamm.start_allele, start_allele)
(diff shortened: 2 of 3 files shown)


jeromekelleher:
Sure, this seems logical. I didn't really have a strong feeling about whether we should have per-site of globally unique alleles. Globally unique alleles it is.

I'm not sure why you're nuking ``start_allele`` though - I think it's confusing to mixup the current state and the initial state and maybe you'll sometimes want to know how many alleles were generated?

jeromekelleher:
> The only thing I'm not sure about is the test for mutation overflow - yes, it overflows? Do we need that test? Perhaps we should be detcting overflow and throwing an error?

I was just checking that overflow happens as expected. Overflow behaviour isn't defined on int64, so it's a good way to make sure that the types are correctly set all the way down the stack.

petrelharp:
I took out start allele because it was easier. ;) But, it's back in now, as well as the overflow test. Note also that I've fixed the docs over at #1180. This should be good to go, assuming it passes!

jeromekelleher (approved):
Looks good,thanks @petrelharp. Just needs a quick squash.

PR #1183, opened 2020-09-14
Refactor to put simulation information into tables

jeromekelleher:
This is a pretty significant refactoring to simplify putting information into the tables as much as we can. At the low-level this means that we don't have separate code paths for the samples and from_ts case any more, but instead it's all handled in using the from_ts code (which is nice and elegant). If this turns out to make the simple case slower than it needs to be we can put in some special cases, say when the input tables has zero edges.

I've done a fair bit of test refactoring to make the tests less crufty.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -221,7 +221,7 @@ class Population:
 
     def get_ind_range(self, t):
         """ Returns ind labels at time t """
-        first_ind = np.sum([self.get_size(t_prev) for t_prev in range(0, t)])
+        first_ind = np.sum([self.get_size(t_prev) for t_prev in range(0, int(t))])
         last_ind = first_ind + self.get_size(t)
 
         return range(int(first_ind), int(last_ind) + 1)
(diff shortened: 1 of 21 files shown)


jeromekelleher:
Notes:

- Need to add support for negative times
- This will involves closing #983 - but basically involves solving the startup problems we have here. So maybe best to just sort it out.

jeromekelleher:
I don't think anyone really wants to review this (it's a big diff), but this is a heads-up that it contains some fairly significant updates. Every simulation is now being treated like an old "from_ts" simulation, as we're initialising the simulation from the input tables. This really clears up a lot of cruft and opens the door for cool new stuff. It does loosen things up a bit by allowing for negative start times and letting demographic events happen before ``start_time`` etc - these were mainly needed to make what we were already doing with the ``from_ts`` code path, so I'm hopeful this won't really break anything.

jeromekelleher:
Dammit, benchmarks say no. Significant regressions on a few cases, will have to look into these before merging.

jeromekelleher:
Actually no, benchmarks are fine! I was just comparing against older stuff before we'd made the big generalisations. So great, let's merge!

PR #1185, opened 2020-09-16
Move physical_to_genetic into RateMap

hyanwong:
I think this is a generally useful thing for any RateMap. Not sure if the existing tests are sufficient, but they should cover it, I think?

### DIFF
--- a/msprime/intervals.py
+++ b/msprime/intervals.py
@@ -33,39 +33,111 @@ class RateMap:
     """
     A class mapping a numeric value to a set of adjacent intervals along the
     genome.
+
+    :param list position: A list of :math:`n` positions, starting at 0, and ending in the
+        sequence length over which the RateMap will apply.
+    :param list rate: A list of :math:`n-1` positive rates that apply between each
+        position.
+    :param list cumulative_rate: A alternative method to specify rates between each
+        position: a list of :math:`n` increasing values starting at 0 providing the
+        cumulative rate.
+    :param float start_position: When returning a mean_rate over the genome, ignore the
+        part of the map from zero up to ``start_position``, as long as the rate is 0 in
+        that region. If ``None`` (default), the ``start_position`` is 0.
+    :param float end_position: When returning a mean_rate over the genome, ignore the
+        part of the map from ``end_position`` onwards, as long as the rate is 0 in that
+        region. If ``None`` (default), the ``end_position`` is the end of the map.
     """
 
-    def __init__(self, position, rate, map_start=0):
-        # We take copies to make sure there are no unintended consequences
-        # later when a user modifies the arrays in this map.
-        self.position = np.array(position, dtype=float, copy=True)
-        self.rate = np.array(rate, dtype=float, copy=True)
-        self.map_start = map_start
-        size = len(self.rate)
-        if size < 1:
+    def __init__(
+        self,
+        position,
+        rate=None,
+        cumulative_rate=None,
+        *,
+        start_position=None,
+        end_position=None,
+    ):
+        # Taking copies ensures there are no unintended consequences later if a user
+        # modifies the input arrays.
+
+        # Save positions into a read-only array
+        self._position = np.array(position, dtype=float, copy=True)
+        size = len(self.position)
+        if size < 2:
             raise ValueError("Must have at least two positions")
-        if self.position.shape[0] != size + 1:
-            raise ValueError(
-                "Rate array must have one less entry than the position array."
-            )
         if self.position[0] != 0:
             raise ValueError("First position must be zero")
         if np.any(self.position[:1] >= self.position[1:]):
             raise ValueError("Position values must be in increasing order")
-        if np.any(self.rate < 0):
-            raise ValueError("Rates must be non-negative")
+        self._position.flags.writeable = False
 
-        self.cumulative = np.insert(np.cumsum(np.diff(self.position) * self.rate), 0, 0)
+        # Save rates / cumulative rates into 2 read-only arrays
+        if rate is None and cumulative_rate is None:
+            raise TypeError("One of 'rate' or 'cumulative_rate' must be specified")
+        if rate is not None and cumulative_rate is not None:
+            raise TypeError("Cannot specify both 'rate' and 'cumulative_rate'")
+
+        if rate is not None:
+            self._rate = np.array(rate, dtype=float, copy=True)
+            if self.rate.shape[0] != size - 1:
+                raise ValueError(
+                    "Rate array must have one less entry than the position array"
+                )
+            if np.any(self.rate < 0):
+                raise ValueError("Rates must be non-negative")
+            self._cumulative = np.insert(
+                np.cumsum(np.diff(self.position) * self.rate), 0, 0
+            )
+        else:
+            self._cumulative = np.array(cumulative_rate, dtype=float, copy=True)
+            if self.cumulative.shape[0] != size:
+                raise ValueError(
+                    "Cumulative rate array must be the same size as the position array"
+                )
+            if self.cumulative[0] != 0:
+                raise ValueError("Cumulative rate array must start with 0")
+            self._rate = np.diff(self.cumulative) / np.diff(self.position)
+            if np.any(self.rate < 0):
+                raise ValueError("Cumulative rates must be strictly increasing")
+
+        if start_position is None:
+            start_position = 0
+        elif self.cumulative_at(start_position) > 0:
+            raise ValueError("Rates before the start_position must all be zero")
+        self.start_position = start_position
+
+        if end_position is None:
+            end_position = self.sequence_length
+        elif self.cumulative_at(end_position) < self.total_mass:
+            raise ValueError("Rates after the end_position must all be zero")
+        self.end_position = end_position
+
+        # Make all of the internal arrays read-only, so they can't get out of sync
+        self._rate.flags.writeable = False
+        self._cumulative.flags.writeable = False
+
+    @property
+    def position(self):
+        return self._position
+
+    @property
+    def rate(self):
+        return self._rate
+
+    @property
+    def cumulative(self):
+        return self._cumulative
 
     def __len__(self):
-        return len(self.rate)
+        return len(self._rate)
 
     @staticmethod
     def uniform(sequence_length, rate):
         return RateMap([0, sequence_length], [rate])
 
     def asdict(self):
-        return {"position": self.position, "rate": self.rate}
+        return {"position": self._position, "rate": self._rate}
 
     @property
     def sequence_length(self):
@@ -73,6 +145,11 @@ class RateMap:
 
     @property
     def total_mass(self):
+        """
+        The cumulative total over the entire map
+        """
+        # Since we insist that the rate up to start_position and past end_position is 0
+        # we can just return the cumulative total.
         return self.cumulative[-1]
 
     @property
@@ -82,13 +159,10 @@ class RateMap:
     @property
     def mean_rate(self):
         """
-        Return the weighted mean across all windows of the entire map.
+        The mean rate over this map, from map.start_position to map.end_position.
         """
-        window_sizes = self.position[1:] - self.position[:-1]
-        weights = window_sizes / self.sequence_length
-        if self.map_start != 0:
-            weights[0] = 0
-        return np.average(self.rate, weights=weights)
+        span = self.end_position - self.start_position
+        return self.total_mass / span
 
     def at(self, x):
         """
@@ -114,28 +188,41 @@ class RateMap:
         :return: An array of cumulative rates, the same length as ``x``
         :rtype: numpy.ndarray(dtype=np.float64)
         """
+        if np.any(np.array(x) < 0) or np.any(np.array(x) > self.sequence_length):
+            raise ValueError(
+                f"Cannot have physical positions < 0 or > {self.sequence_length}"
+            )
         return np.interp(x, self.position, self.cumulative)
 
-    def slice(self, start=None, end=None, trim=False):  # noqa: A003
+    def slice(self, start=None, end=None, *, trim=False):  # noqa:A003
         """
-        Returns a subset of this rate map between the specified end
-        points. If start is None, it defaults to 0. If end is None, it defaults
-        to the end of the map. If trim is True, remove the flanking
-        zero rate regions such that the sequence length of the
-        new rate map is end - start.
+        Returns a subset of this rate map between the specified end points.
+
+        :param float start: The start of the region to keep. If ``None``, it defaults to
+            0.
+        :param float end: The end of the region to keep. If end is ``None``, it defaults
+            to the end of the map.
+        :param bool trim: If True, remove the flanking regions such that the
+            sequence length of the new rate map is ``end`` - ``start``. If ``False``
+            (default), do not change the coordinate system, but instead replace each
+            flanking region with 0, and set the ``start_position`` and ``end_position``
+            of the returned map to ``start`` and ``end``.
+
+        :return: A new RateMap instance
+        :rtype: RateMap
         """
         if start is None:
             i = 0
             start = 0
         if end is None:
-            end = self.position[-1]
+            end = self.sequence_length
             j = len(self.position)
 
         if (
             start < 0
             or end < 0
-            or start > self.position[-1]
-            or end > self.position[-1]
+            or start > self.sequence_length
+            or end > self.sequence_length
             or start > end
         ):
             raise IndexError(f"Invalid subset: start={start}, end={end}")
@@ -153,28 +240,25 @@ class RateMap:
         rate = self.rate[i : j - 1].copy()
         position[0] = start
         position[-1] = end
-        map_start = 0
 
         if trim:
-            position -= start
-        else:
-            # Prepend or extend zero-rate region at start of map.
-            if position[0] != 0:
-                map_start = position[0]  # TODO: is this what we want here?
-                if rate[0] == 0:
-                    position[0] = 0
-                else:
-                    position = np.insert(position, 0, 0)
-                    rate = np.insert(rate, 0, 0)
-            # Append or extend zero-rate region at end of map.
-            if position[-1] != self.position[-1]:
-                if rate[-1] == 0:
-                    position[-1] = self.position[-1]
-                else:
-                    position = np.append(position, self.position[-1])
-                    rate = np.append(rate, 0)
+            # Return trimmed map with changed coords
+            return self.__class__(position - start, rate)
 
-        return self.__class__(position, rate, map_start=map_start)
+        # Need to check regions before & after sliced region are filled out:
+        if start != 0:
+            if rate[0] == 0:
+                position[0] = 0  # Extend
+            else:
+                rate = np.insert(rate, 0, 0)  # Prepend
+                position = np.insert(position, 0, 0)
+        if end != self.position[-1]:
+            if rate[-1] == 0:
+                position[-1] = self.sequence_length  # Extend
+            else:
+                rate = np.append(rate, 0)  # Append
+                position = np.append(position, self.position[-1])
+        return self.__class__(position, rate, start_position=start, end_position=end)
 
     def __getitem__(self, key):
         """
@@ -224,9 +308,11 @@ class RecombinationMap:
         the largest possible value, allowing the maximum resolution
         in the recombination process. However, for a finite sites
         model this can be set to smaller values.
+    :param str name: A name to use for this map, and stored in the ``.name``
+        attribute. This will only be used if the map is written out to a file
     """
 
-    def __init__(self, positions, rates, num_loci=None, map_start=0):
+    def __init__(self, positions, rates, num_loci=None, name=None, **kwargs):
         if num_loci is not None:
             if num_loci == positions[-1]:
                 warnings.warn("num_loci is no longer supported and should not be used.")
@@ -237,8 +323,8 @@ class RecombinationMap:
                     "scale positions to span the desired number of loci "
                     "and set discrete=True"
                 )
-        self.map = RateMap(positions, rates[:-1])
-        self.map_start = map_start
+        self.map = RateMap(positions, rates[:-1], **kwargs)
+        self.name = name
 
     @classmethod
     def uniform_map(cls, length, rate, num_loci=None):
@@ -255,16 +341,33 @@ class RecombinationMap:
         return cls([0, length], [rate, 0], num_loci=num_loci)
 
     @classmethod
-    def read_hapmap(cls, filename):
+    def read_hapmap(cls, filename, sequence_length=None):
         """
-        Parses the specified file in HapMap format.
+        Parses the specified file in HapMap format. These files must be
+        white-space-delimited, and contain a single header line (which is
+        ignored) and a set of rows that pertain to a single chromosome.
+        Each line after the header is expect to contains a name (e.g. chr1),
+        and the starting position and recombination rate for the segment from
+        that position (inclusive) to the starting position on the next line
+        (exclusive). Starting positions of each segment are given in units of
+        bases, and recombination rates in centimorgans/Megabase. A warning is
+        given if the name in the first column differs between rows. Otherwise
+        only the second (Position) and third (Rate) columns are read: any
+        further columns are ignored. If the first starting position is not equal
+        to zero, then a zero-recombination region is inserted at the start of
+        the chromosome.
 
         .. warning::
             This method is deprecated, use the module-level
-            :func:`read_hapmap` function instead.
+            :func:`read_hapmap` function instead, which reads the genetic position
+            in centiMorgans from the file, rather than using the rate column.
 
         :param str filename: The name of the file to be parsed. This may be
-            in plain text or gzipped plain text.
+            in plain text or gzipped plain text. It may also be a file object.
+        :param str sequence_length: The total length of the map. If ``None``, then
+            assume it is the final physical position in the file. Otherwise pad the
+            map with a zero rate region between the final position in the file and
+            the value specified here.
         :return: A RecombinationMap object.
         """
         warnings.warn(
@@ -272,9 +375,77 @@ class RecombinationMap:
             "Use msprime.read_hapmap() instead.",
             FutureWarning,
         )
-        rate_map = read_hapmap(filename)
-        rate = np.append(rate_map.rate, 0)
-        return cls(rate_map.position, rate, map_start=rate_map.map_start)
+        rownames, position, rate = np.loadtxt(
+            filename,
+            skiprows=1,
+            dtype=[("chrom", object), ("position", int), ("rate", float)],
+            usecols=(0, 1, 2),
+            unpack=True,
+        )
+        if len(rownames) == 0:
+            raise ValueError("Empty hapmap file")
+
+        if len(np.unique(rownames)) > 1:
+            warnings.warn(
+                f"Different chromosome names found in the first column of {filename}. "
+                "Check that the file contains data for a single chromosome."
+            )
+        name = rownames[0]
+        # Rate is expressed in centimorgans per megabase: convert to per-base rates
+        rate *= 1e-8
+
+        start = position[0]
+        end = position[-1]
+        if start != 0:
+            position = np.insert(position, 0, 0)
+            rate = np.insert(rate, 0, 0)
+        if rate[-1] != 0:
+            raise ValueError(
+                "The last rate provided in the recombination map must be zero"
+            )
+        if sequence_length is not None:
+            if sequence_length < end:
+                raise ValueError(
+                    f"Cannot specify a sequence_length < {end} (the last position)"
+                )
+            elif sequence_length > end:
+                position = np.append(position, sequence_length)
+                rate = np.append(rate, 0)
+        return cls(position, rate, name=name, start_position=start, end_position=end)
+
+    def write_hapmap(self, filename):
+        """
+        Write out the recombination map in a format with genetic distances in cM,
+        so that it can be read in by the non-deprecated read-hapmap function.
+        Omit the rate up to the start_position and after the end_position.
+        """
+        headers = ["Chromosome", "Position(bp)", "Rate(cM/Mb)", "Map(cM)"]
+        name = "Unknown" if self.name is None else self.name
+        with open(filename, "wt") as file:
+            file.write("\t".join(headers) + "\n")
+            for pos, rate, cumul in zip(
+                self.position, self.get_rates(), self.cumulative
+            ):
+                if self.map.start_position <= pos <= self.map.end_position:
+                    file.write(
+                        "{}\t{:.12g}\t{:g}\t{:g}\n".format(
+                            name, pos, rate * 1e8, cumul * 100
+                        )
+                    )
+
+    @property
+    def position(self):
+        # For backwards compatibility
+        return self.map.position
+
+    @property
+    def rate(self):
+        # For backwards compatibility
+        return self.map.rate
+
+    @property
+    def cumulative(self):
+        return self.map.cumulative
 
     @property
     def mean_recombination_rate(self):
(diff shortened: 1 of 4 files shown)


hyanwong:
NB - we might want to change the name to "cumulative" or something like that, which would be more meaningful if this was a rate map for mutation rates, for instance.

jeromekelleher:
This seems like a useful feature, but we shouldn't call it ``physical_to_genetic``. I don't have a better suggestion right now, though. (Given that the functionality is so straightforward with numpy, I think we should probably just implement it in tsinfer, if we need it there, rather than adding a dependency on msprime just for this.)

We would want some sort of tests here besides relying on what we're getting through the legacy

hyanwong:
> This seems like a useful feature, but we shouldn't call it `physical_to_genetic`. 

Agree. `cumulative_values` perhaps?

>(Given that the functionality is so straightforward with numpy, I think we should probably just implement it in tsinfer, if we need it there, rather than adding a dependency on msprime just for this.)

Also agree. I was going to implement it in msprime, then copy the entire `intervals.py` file into tsinfer, and leave it there with a note to delete once (if) we move this into tskit.

> We would want some sort of tests here besides relying on what we're getting through the legacy

Fair enough. I guess we might delete the legacy stuff at some point anyway.

hyanwong:
Actually, I wonder if we want a simple `interpolate` method? Then we could do `RateMap.interpolate('cumulative')`, and we could also have  `RateMap.interpolate('flat')` (or 'linear', or whatever) for getting other intermediate values? What do you think @jeromekelleher ?

petrelharp:
What would `interpolate` do, @hyanwong? I think of interpolation as guessing at values between observed points, but I think a RateMap is a piecewise-constant function where we know the values at every point. It does seem useful to be able to ask "what's the value of the function at location x" (ie what's the rate here) and "what's the integral up until x" (eg the genetic map position) - are these what you're talking about?

hyanwong:
> What would `interpolate` do, @hyanwong? I think of interpolation as guessing at values between observed points, but I think a RateMap is a piecewise-constant function where we know the values at every point. 

I'm not sure that most real-world recombination maps, such as provided by read_hapmap (or indeed real mutation rate maps) are based on knowing the value at every base pair are they? Instead, I thought they were usually based around a set of point estimates at "changeable" places along the genome.

> It does seem useful to be able to ask "what's the value of the function at location x" (ie what's the rate here) and "what's the integral up until x" (eg the genetic map position) - are these what you're talking about?

That is exactly what I'm talking about. But if the map is simply a set of point estimates, then "what is the rate here" can be found by e.g. a stepwise line fit (the norm), or e.g. smoothing at the transition points somehow, e.g. using a cubic spline. There is some literature on this, ISTR. But I may be spouting rubbish.

petrelharp:
> I'm not sure that most real-world recombination maps, such as provided by read_hapmap (or indeed real mutation rate maps) are based on knowing the value at every base pair are they? Instead, I thought they were usually based around a set of point estimates at "changeable" places along the genome.

It's true! But, the way we're using the RateMap object is not as a collection of coarse estimates, but as a complete specification of the rate everywhere.

> That is exactly what I'm talking about. But if the map is simply a set of point estimates, then "what is the rate here" can be found by e.g. a stepwise line fit (the norm), or e.g. smoothing at the transition points somehow, e.g. using a cubic spline. There is some literature on this, ISTR. But I may be spouting rubbish.

True! As above, though, I don't think a RateMap is a set of point estimates. Even if it is, how to do this right is a major question, and depends on how the map was inferred, probably. For instance, we don't get point estimates of the rate at particular positions, we get point estimates of the difference in cumulative map position between particular physical positions. And, it's not clear that you want to smooth, anyhow, as we know that the true map is actually pretty hotspott-ey (well, in some species, anyhow!).

hyanwong:
> True! As above, though, I don't think a RateMap is a set of point estimates. Even if it is, how to do this right is a major question, and depends on how the map was inferred, probably. 

Yes, I'm sure this is right, which is why the default setting would simply be to treat it as piecewise constant (I think of this as interpolation_method="stepwise"), and we might not even have any other methods. But perhaps we should allow the possibility. I don't mind if it's not called `interpolation` though. Any other suggestions for "value at x, where x is not in the list of positions used to define the map"?

petrelharp:
`.value()` and `.cumulative_value()`?

hyanwong:
Alternatively (and I have no preference here) `value()` and `value(cumulative=True)`.

hyanwong:
(NB: I I quite like the succinctness of `.at()`, which also matches `ts.at()`

petrelharp:
I do too, but since there's these two things (the rate and cumulative rate), I worry that just plain `at` will get confused. How about `rate_at` and `cumulative_rate_at`? I think they should be different methods, rather than a `cumulative=True` flag, since they have different units, and the basic sort of thing that a method returns should not change.

hyanwong:
I think `cumulative` makes it clear, though. E.g. `at` and `cumulative_at` is pretty clear to me: since the class is a `*Rate*Map` anyway, I'm not sure we need `rate_at`?

petrelharp:
Sure, either way.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/1185/)

hyanwong:
This is now reworked to use `at` & `cumulative_at`, following the discussion above with @petrelharp, and to use the `pos_cm` value of the HapMap file, allowing us to read in maps that don't start at 0cM (Fixes #1255, and I changed the doc example) and remove the `map_start` parameter, as discussed with @jeromekelleher in #1255  - I've also allowed the column positions to be specified, as there are some other maps out there (e.g. from Beagle, I think) that omit the rate column. And I added a note there about 1-based versus 0-based indexing: 
```
         Positions in a RateMap are zero-based, meaning that rate values in the returned
         RateMap object are defined from position 0 onwards. Thus in the example above,
         a rate of 0.002664/48232 cM/bp (i.e. 0.0552 cM/Mb) is assumed to apply from
         position 0 (inclusive) to 48232 (exclusive). This may cause the first rate to
         differ minutely from a rate calculated on the basis of the first position being
         one.
```
It all seems to work.

There are some other changes in there too. As also [discussed](https://github.com/tskit-dev/msprime/issues/1255#issuecomment-716717287) in #1255 I've made the RateMap read-only. And (more controversially, needing discussion), I've reformulated `RaterMap.slice` because I think it's unclear what should happen to sliced-off regions when you `slice(trim=False)`. I don't think setting those regions to a rate of 0 is necessarily the obvious thing, so there's a parameter `replacement` which can be used to specify this, and a note which say why you might want to set replacement to 0. I'm not sure what we *should* set it to be default, however? At the moment it uses the mean of the replaced region, but we could use NaN, or perhaps force a replacement value to be specified if trim is False?

jeromekelleher:
@grahamgower, any chance you could give a review here please? I'm guessing we'll be using this infrastructure downstream in stdpopsim, so this is your chance to make sure it's done properly!

tests/test_recombination_map.py
```diff
@@ -396,34 +440,76 @@ def test_mean_recombination_rate(self):
...
-                    1   9477943    0.0864382908650907      0.971188757364862
-                    1   9696341    4.76495005895746        0.990066707213216
-                    1   9752154    0.0864316558730679      1.25601286485381
-                    1   9881751    0.0                     1.26721414815999"""
+                        prev_rate = 0
+                    weights.append(pos - prev_pos)
+                    rates.append(prev_rate)
+                    prev_pos = pos
+                    prev_rate = rate
+            return np.average(rates, weights=weights) * 1e-8  # convert to morgans / bp
```
grahamgower, line 507: This weighted average produces incorrect values, because it fails to account for zero-rate regions at the start of files. I believe the existing code in this test was producing correct values.

grahamgower (commented):
I like the docs changes, and the `at()`/`cumulative_at()` functions seems genuinely useful and make things generally tidier.

Unfortunately, this introduces a regression of #844. See also
https://github.com/popsim-consortium/stdpopsim/issues/630

I'm also not convinced we need to support `slice(..., trim=False)` at all. The use case with `replacement=0` is maybe justified if you're going to ignore simulated regions outside the sliced interval (but then, why not just simulate a shorter region, and alter coordinates in downstream analyses as required?). But what are the circumstances where other replacement values make sense? E.g. why would one want to simulate with a uniform rate everywhere except a specific interval?

grahamgower:
> why would one want to simulate with a uniform rate everywhere except a specific interval?

Oh. Telomeres! So the use case is: get the start/end coordinates of the genetic map, and call slice(..., trim=False, replacement=chromosome_mean_rate).

grahamgower:
Thinking about this some more... Maybe what we want is the option to artificially extend the genetic length of a map when it's read with `read_hapmap()`? This would be useful under the assumptions that (a) a genetic map file starts and ends inside the bounds of the chromosome, and (b) this is because crossovers there are hard to observe, rather than because they don't occur. Please correct me if assumption (b) is wrong, I'm really not familiar with the details here.

So, given the hapmap genetic map for chr22, which starts 16mb into the chromosome (and ends a few kb short of the end), one might load it like this: `msprime.RecombinationMap.read_hapmap("genetic_map_GRCh37_chr22.txt", length=51304566)`. Then `read_hapmap()` would fill zero-rate regions at the start and end of the map (if they exist) with the mean value (using the `slice()` function with the changes proposed in this PR). This would mean that the `mean_recombination_rate` property would accurately reflect the mean rate of the genetic map. It would also work with the other changes proposed in this PR (e.g. removal of the `map_start` attribute).

hyanwong:
> > why would one want to simulate with a uniform rate everywhere except a specific interval?
> 
> Oh. Telomeres! So the use case is: get the start/end coordinates of the genetic map, and call slice(..., trim=False, replacement=chromosome_mean_rate).

True about telomeres - although I wasn't explicitly thinking about that. It was more that I didn't like the "invisibly assume 0" behaviour that we previously had. I wonder if we should explicitly require a replacement value if `replace=True`? Unfortunately that doesn't allow the current behaviour (which allows a different mean rate for each side), but perhaps that's OK.

hyanwong:
> Thinking about this some more... Maybe what we want is the option to artificially extend the genetic length of a map when it's read with `read_hapmap()`? This would be useful under the assumptions that (a) a genetic map file starts and ends inside the bounds of the chromosome, 

I don't think this is quite true. With the PR, we are no longer reading in the "rate" value from the map, but the `pos_cM` value. If this doesn't start at 0 (as in the example I give in the docs, which is from an existing human map), we can assume a constant, non-zero rate at the start of the chromosome. What we do at the end of the chromosome is more tricky. I guess if the last rate is not zero, we could use the rate value to fill in to the end of the passed-in sequence_length.

> and (b) this is because crossovers there are hard to observe, rather than because they don't occur. Please correct me if assumption (b) is wrong, I'm really not familiar with the details here.

I think you are right that a rate of 0 is usually put here because they are hard to measure.

> Then read_hapmap() would fill zero-rate regions at the start and end of the map (if they exist) 

I think we don't want to fill out the region at the start with the mean value by default, because if the first row is at (say) 16mb with a genetic position at 0.0cM, this really does indicate that the initial region has a rate of 0. And similarly, if the initial genetic position is (say) 0.08 cM, then we should trust the user to mean that the rate from 0-16mb is 8/16e6 = 5e-07 M/bp.

We could, however, have an optional parameter to "change" the start of the map. The issue I have with this is that we would no longer recover the same cM coordinates as were in the original file, which seems a bit odd to me.

tests/test_recombination_map.py
hyanwong: Hmm. The start of the files may not be zero-rate. It depends if the first genetic position is 0cM. I'll look into the regression and see what the issue is.

grahamgower:
What if we have `start` and `end` parameters, which a user may optionally set to extend towards the start and end respectively. So, `start=0` would extend towards the start (using map-wide average only if the first row has 0 in the `Map(cM)` column), and `end=END_COORD` would extend to END_COORD using the final row's `Rate(cM/Mb)` value, or alternately by the map-wide average if that is zero. 

> The issue I have with this is that we would no longer recover the same cM coordinates as were in the original file, which seems a bit odd to me.

Yeah, that is a slight concern. I'm not sure how big a problem it is in practice though. Probably this is more likely to do the right thing than simulating a big chunk of chromsome without recombinations (e.g. see https://github.com/popsim-consortium/stdpopsim/issues/663)

petrelharp:
To be sure we're doing the right thing, I'd still like to know why we're getting map files with a nonzero start position. This is only observed in the wild as produced by liftOver, right? And, is it a bug there? If so, we should not just work around it. Maybe @jradrion knows? (see #1255, Jeff?)

grahamgower:
> to use the `pos_cm` value of the HapMap file, allowing us to read in maps that don't start at 0cM

It's probably worth double checking the hapmap files have consistency between the cM and rate columns. I've just now looked at some other examples of maps in stdpopsim, and see that at the HomSap/DeCodeSexAveraged_GRCh36 maps have garbage in the cM column, and the AraTha/SalomeAveraged_TAIR7 don't have this column at all!

jradrion:
The original human HapMap files from GRCh37 have nonzero start positions. This is why the liftOver maps also have nonzero start positions. It was my understanding (perhaps incorrectly) that the Map(cM) column of these files is never used. My liftOver files for hg38 don't even include this column, and IIRC, this column is also absent from the DroMel map (along with the above mentioned maps).

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1185/)

hyanwong:
@grahamgower - a slight hitch before finalising this PR. In the new code we are reading the genetic distance (the pos (cM) column) rather than the rate, which is what Gil and I both thought was sensible. This also naturally deals with some of those weird files where the genetic distance for the first entry in the map file (say at physical position X) doesn't start at 0 cM, but at some positive number instead (say Y). In this case it's pretty clear to me that the start of the chromosome, from 0..X, has a non-zero rate of Y/X, and this should be included in the reported mean rate over the whole chromosome.

BUT, if the first entry in the map file *is* at 0cM, should we treat it as a special case and exclude the start of the chromosome from the mean rate calculation. In fact, some to think of it, if we extend the Recombination Map to the end of the chromosome with a 0 rate region, so that the sequence length of the map matches up with the sequence length of the chromosome, should we exclude this region too?

Perhaps there should be a parameter to `mean_rate()`, "exclude_flanking_zeros" - if so should this default to "True"?

grahamgower:
I reckon a mean rate should be a mean over the map's extent. In the case of the first entry in the file being 0cM, at position > 1, then the map's extent doesn't cover the start of the chromosome, and thus the mean rate calculation should ignore it. An extreme example is the HapmapII genetic map for chr22 which starts at position~=16mb. If you want to assume that the `RateMap`'s coordinates correspond with the extent of the map in the file, then we need to have some process by which existing map files can be fixed for this assumption. E.g. using the new behaviour with `trim=False, replacement=None`.

hyanwong:
> I reckon a mean rate should be a mean over the map's extent. In the case of the first entry in the file being 0cM, at position > 1, then the map's extent doesn't cover the start of the chromosome, and thus the mean rate calculation should ignore it. An extreme example is the HapmapII genetic map for chr22 which starts at position~=16mb. If you want to assume that the `RateMap`'s coordinates correspond with the extent of the map in the file, then we need to have some process by which existing map files can be fixed for this assumption. E.g. using the new behaviour with `trim=False, replacement=None`.

But if the first entry in the file is at (say) 2 cM, then the map *does* cover the start of the chromosome, right?

hyanwong:
Also, what happens if we extend the map (in python, not in the map file) to make the sequence lengths the same between the map and the chromosome (which seems desirable, and we will end up doing). Then we would claim that the "extent" of the map didn't cover the last section either, right?

In both case we have to store some state in the map which says "the original, read-in data doesn't include the start and/or end" - we had something like that before, but it seems a horrible hack. Wouldn't it be better to say "by default we trim the first and last regions if they have rate 0"? Then we don't need to store state. The only difference would be if we have a map that explicitly starts at position 0 with rate 0, in which case we would end up excluding that region from the mean rate calculation, even though it was in the file. But I think we would probably want to do that anyway!

hyanwong:
Oh, I should also say that we are going to (presumably) implement this in the new RateMap class, which isn't just for recombination rates, but for any rates along the genome (e.g. mutation rates, whatever). I don't know if that changes the situation?

grahamgower:
> But if the first entry in the file is at (say) 2 cM, then the map _does_ cover the start of the chromosome, right?

Sounds right.

> Also, @grahamgower, what happens if we extend the map (in python, not in the map file) to make the sequence lengths the same between the map and the chromosome (which seems desirable, and we will end up doing)? Then we would claim that the "extent" of the map didn't cover the last section either, right?

That sounds right too.

> In both case we have to store some state in the map which says "the original, read-in data doesn't include the start and/or end" - we had something like that before, but it seems a horrible hack. Wouldn't it be better to say "by default we trim the first and last regions if they have rate 0"? Then we don't need to store state.

It seems that biology is messy, and bioinformatics doubly so. Can't we keep the current behaviour for `RecombinationMap.read_hapmap()` (the hack), and deprecate the method? And change the assumed behaviour for `RateMap.read_hapmap()` to match your nice clean implementation. Then we can load the old files, and we have a way to transition that messy old data to nice clean new data that is allowed to use the non-deprecated API. I haven't looked through your diff again, but I thought this is what we talked about previously?

hyanwong:
Phew - I finally got round to fixing all this @grahamgower . No hurry, but could you take a peek at the interface when you have time. I implemented `start_position` *and* `end_position`, so that we can specify a sequence_length when reading a hapmap file, and the start and end will be padded with zero-recombination regions which will not be reported back in the mean_rate.

I removed the ability to specify a value for the start and end regions in `slice(... trim=True)`, setting these flanks to 0, and simply used the new `start_position` and `end_position` functionality to mark the slice boundaries.

There's also a `write_hapmap` function that writes the old (`msprime.RecombinationMap`) maps to a text file to include the genetic position in centiMorgans, and which can then be read back in again by the new `msprime.read_hapmap()` function. I needed to get the first column of the hapmap file (the chromosome name) so I could write that back out again, so I also now check that the names in the 1st column are the same for all rows - if that does break any stdpopsim maps, they were probably faulty anyway. There's some tests for the conversion functionality in test_recombination_map.py::TestReadHapmap::test_read_old_hapmap`. Hopefully that will allow the old stdpopsim maps to be converted to the new format. It won't work for maps that have a starting cM value >0, but you don't have any of those in stdpopsim anyway.

The old `msprime.RecombinationMap` class has some methods/attributes that don't have equivalent names in the RateMap class, for instance `.mean_recombination_rate` and `get_total_recombination_rate()` are not provided, but are accessed through `.mean_rate` and `.total_mass`. I don't know if we want to derive a specific `RecombinationRateMap` class from the plain `RateMap` so we can add these aliases on top and maintain like-for-like functionality?

Finally - I guess I should test this in the current stdpopsim repo. Is it just a question of installing my PR of msprime locally and running the tests? It doesn't look like you are using msprime as a submodule in the stdpopsim repo, or anything like that.

msprime/intervals.py
```diff
@@ -229,26 +342,104 @@ def uniform_map(cls, length, rate, num_loci=None):
...
         .. warning::
             This method is deprecated, use the module-level
-            :func:`read_hapmap` function instead.
+            :func:`read_hapmap` function instead, which uses the genetic position
+            in centiMorgans rather than the rate column.
 
         :param str filename: The name of the file to be parsed. This may be
-            in plain text or gzipped plain text.
+            in plain text or gzipped plain text. It may also be a file object.
+        :param float sequence_length: The .
```
grahamgower, line 366: Looks like the docstring here needs finishing.

msprime/intervals.py
```diff
@@ -229,26 +342,104 @@ def uniform_map(cls, length, rate, num_loci=None):
...
         )
-        rate_map = read_hapmap(filename)
-        rate = np.append(rate_map.rate, 0)
-        return cls(rate_map.position, rate, map_start=rate_map.map_start)
+        hapmap = np.loadtxt(filename, skiprows=1, usecols=(1, 2))
+        try:
+            filename.seek(0)  # If a file object, we need to read from start again
+        except AttributeError:
+            pass
+        rownames = np.genfromtxt(filename, skip_header=True, usecols=0, dtype="str")
```
grahamgower, line 379: It should be possible to avoid reading the file twice. How about something like the following?
```python
rownames, position, rate = np.loadtxt(
    filename,
    skiprows=1,
    dtype=[("chrom", object), ("position", int), ("rate", float)],
    usecols=(0,1,2),
    unpack=True
)
```

msprime/intervals.py
```diff
@@ -229,26 +342,104 @@ def uniform_map(cls, length, rate, num_loci=None):
...
+        if map_start != 0:
+            position = np.insert(position, 0, 0)
+            rate = np.insert(rate, 0, 0)
+        if rate[-1] != 0:
+            raise ValueError(
+                "The last rate provided in the recombination map must be zero"
+            )
+        if sequence_length is not None:
+            position = np.append(position, sequence_length)
+            rate = np.append(rate, 0)
```
grahamgower, line 401: I think this has three cases that need to be handled differently. 1) `if sequnce_length < position[-1]`, raise an error, 2) `elif sequence_length == position[-1]`, do nothing. 3) `else:` append as you have here.

grahamgower (approved):
Very nice @hyanwong, it looks great!

> Finally - I guess I should test this in the current stdpopsim repo. Is it just a question of installing my PR of msprime locally and running the tests? It doesn't look like you are using msprime as a submodule in the stdpopsim repo, or anything like that.

Um, yes, for the most part. There might be issues with some stdpopsim tests relating to recapitation and the SLiM engine. You can ignore those tests, or clone the stdpopsim `msprime-1.0` branch where these should be fixed.

msprime/intervals.py
```diff
@@ -198,9 +309,11 @@ class RecombinationMap:
         the largest possible value, allowing the maximum resolution
         in the recombination process. However, for a finite sites
         model this can be set to smaller values.
+    :param int name: A name to use for this map, and stored in the ``.name``
```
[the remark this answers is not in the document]
hyanwong: Yep - copy/paster typo. Thanks.

msprime/intervals.py
hyanwong: Ah, I didn't realise you could give a dtype to np.loadtxt. Much better. Thanks.

msprime/intervals.py
hyanwong: Good point. I haven't done the unit tests for sequence_length yet.

msprime/intervals.py
```diff
@@ -82,34 +159,70 @@ def size(self):
     @property
     def mean_rate(self):
         """
-        Return the weighted mean across all windows of the entire map.
+        The mean rate over this map, from map.start_position to map.end_position.
+        """
+        span = self.end_position - self.start_position
+        return self.total_mass / span
+
+    def at(self, x):
```
jeromekelleher, line 167: I think it's clearer to call these ``get_rate`` and ``get_cumulative_mass``.

msprime/intervals.py
```diff
@@ -82,34 +159,70 @@ def size(self):
     @property
     def mean_rate(self):
         """
-        Return the weighted mean across all windows of the entire map.
+        The mean rate over this map, from map.start_position to map.end_position.
```
jeromekelleher, line 162: Worth noting here the semantics of what start and end mean, or maybe just giving a pointer back to the contructor docs.

jeromekelleher: We should be a bit more precise here, too. It's the mean rate weighted by genome span, right? Someone might think this was just the unweighted mean of the different values.

msprime/intervals.py
```diff
@@ -198,9 +308,11 @@ class RecombinationMap:
         the largest possible value, allowing the maximum resolution
         in the recombination process. However, for a finite sites
         model this can be set to smaller values.
+    :param str name: A name to use for this map, and stored in the ``.name``
```
jeromekelleher, line 311: What's this for? This is a legacy interface, we don't want to add any more functionality to it.

msprime/intervals.py
```diff
@@ -198,9 +308,11 @@ class RecombinationMap:
...
+        attribute. This will only be used if the map is written out to a file
     """
 
-    def __init__(self, positions, rates, num_loci=None, map_start=0):
+    def __init__(self, positions, rates, num_loci=None, name=None, **kwargs):
```
jeromekelleher, line 315: Again, no extra features here, this is legacy code which we just want to preserve for compatability.

msprime/intervals.py
```diff
@@ -307,54 +523,102 @@ def get_length(self):
...
+        Positions in a RateMap are zero-based, meaning that rate values in the returned
+        RateMap object are defined from position 0 onwards. Thus in the example above,
+        a rate of 0.002664/48232 cM/bp (i.e. 0.0552 cM/Mb) is assumed to apply from
+        position 0 (inclusive) to 48232 (exclusive). If the map positions use a
+        one-based, system, with the first position being 1 rather than 0, this may
+        result in an miniscule difference in the first rate value calculated.
 
     :param str filename: The name of the file to be parsed. This may be
         in plain text or gzipped plain text.
+    :param str sequence_length: The total length of the map. If ``None``, then assume
```
jeromekelleher, line 568: float sequence_length

msprime/intervals.py
```diff
@@ -307,54 +523,102 @@ def get_length(self):
...
+        it is the final physical position in the file. Otherwise pad the map with a zero
+        rate region between the final position in the file and the value specified here.
+    :param str has_header: If True, assume the file has a header row and ignore the
+        first line of the file.
+    :param str bp_column: The zero-based index of the column in the file specifying
```
jeromekelleher, line 573: int not str - likewise for the next

msprime/intervals.py
```diff
@@ -229,26 +341,111 @@ def uniform_map(cls, length, rate, num_loci=None):
...
+            if sequence_length < end:
+                raise ValueError(
+                    f"Cannot specify a sequence_length < {end} (the last position)"
+                )
+            elif sequence_length > end:
+                position = np.append(position, sequence_length)
+                rate = np.append(rate, 0)
+        return cls(position, rate, name=name, start_position=start, end_position=end)
+
+    def write_hapmap(self, filename):
```
jeromekelleher, line 416: This is handy, but lets put it into a module function, like ``read_hapmap``. The file argument should be a fileobj rather than a filename too.

msprime/intervals.py
```diff
@@ -229,26 +341,111 @@ def uniform_map(cls, length, rate, num_loci=None):
...
         """
         warnings.warn(
             "RecombinationMap.read_hapmap() is deprecated. "
             "Use msprime.read_hapmap() instead.",
             FutureWarning,
         )
-        rate_map = read_hapmap(filename)
-        rate = np.append(rate_map.rate, 0)
-        return cls(rate_map.position, rate, map_start=rate_map.map_start)
+        rownames, position, rate = np.loadtxt(
```
jeromekelleher, line 378: It's pretty confusing having two different methods doing the same thing and having slightly different functionality, and one being deprecated. I think we'll have to support the hapmap-rate-column version indefinitely, so we can't just lump it into the deprecated RecombinationMap interface.

I think we'll need to have to think of a way to make the ``read_hapmap`` function work in both cases.

jeromekelleher (commented):
There's lots of good stuff in here, thanks @hyanwong, but there's too much for one PR - I've spent an hour reviewing already, and would need to spend at least another hour to figure it all out. I'm not sure we're taking the right approach with the hapmap functionality, and should maybe try to read the header to find the column indexes instead.

Can you take out the changes to the deprecated RecombinationMap, and all of the hapmap stuff please? This should just leave the API changes to RateMap, which will be easier to review and discuss in isolation.

hyanwong:
> There's lots of good stuff in here, thanks @hyanwong, 

Thanks for taking the time to look over it.

> but there's too much for one PR - I've spent an hour reviewing already, and would need to spend at least another hour to figure it all out. 

Yeah, I can see that. Breaking it down would indeed be helpful.

> I'm not sure we're taking the right approach with the hapmap functionality, and should maybe try to read the header to find the column indexes instead.

Actually, there's no standard header used for these files, so I think we're just stuck with specifying the column indexes, sorry.

> Can you take out the changes to the deprecated RecombinationMap, and all of the hapmap stuff please? This should just leave the API changes to RateMap, which will be easier to review and discuss in isolation.

Right, that make sense. I'll see what I can untangle.

hyanwong:
Oh, and another comment / question for @grahamgower - should `len(rate_map)` return the number of rates, or the number of positions (I'm guessing the first). And should we include rates that are before `start_position` and after `end_position` (i.e. those which were auto-inserted to make the map to the correct length)? I suspect not, but that would be a change.

grahamgower:
> should `len(rate_map)` return the number of rates, or the number of positions (I'm guessing the first)

I think there are even more ways to guess at what this length would mean. E.g. one might guess that it's the sequence_length. So, in light of the many ways one could incorrectly guess at what this does, it might be better to not provide a `__len__()` method at all, and force the user to directly grab the property they're interested in.

hyanwong:
> > should `len(rate_map)` return the number of rates, or the number of positions (I'm guessing the first)
> 
> it might be better to not provide a `__len__()` method at all, and force the user to directly grab the property they're interested in.

Just tried that. Unfortunately in places like `msprime/ancestry.py` we have e.g. `assert len(gene_conversion_map) == 1`. That does have a "fix" above it, so we could possibly change that and deprecate the `__len__` method, though?

grahamgower:
> we could possibly change that and deprecate the `__len__` method, though?

It's not yet in any released version, so I'd be in favour of yanking it out.

jeromekelleher:
> It's not yet in any released version, so I'd be in favour of yanking it out.

Sure, we can do that. But let's do it as a follow-up issue and not make this PR any bigger.

hyanwong:
I'm going to close this PR and open the broken-down changes into other PRs, as there's a lot of now not-quite-relevant discussion above.

msprime/intervals.py
```diff
@@ -33,46 +33,123 @@ class RateMap:
...
-        self.position = np.array(position, dtype=float, copy=True)
-        self.rate = np.array(rate, dtype=float, copy=True)
-        self.map_start = map_start
-        size = len(self.rate)
-        if size < 1:
+    def __init__(
+        self,
+        position,
+        rate=None,
+        cumulative_rate=None,
```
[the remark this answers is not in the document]
hyanwong: I'm replying to these comments here, even though the issue is closed, pending a new one @jeromekelleher - is that OK?

If we do have just one way of specifying the rate, I would vote for using the cumulative rate, not the actual rate. The cumulative rate should be the canonical value that is being set - it is what is usually measured, is more reliable, and doesn't suffer from rounding errors when there are lots of values. That's why I added it in. I see rate as (almost always) a *derived* value, not the original, fed-in data.

msprime/intervals.py
hyanwong: We discussed this with Peter, and came up with "at" and "cumulative_at", ISTR?

msprime/intervals.py
hyanwong: We need it to be able to convert old recombination maps (in stdpopsim) to the new style, which can then be output using the (again, new) write_hapmap function, and read-in by the new code. I discussed this with @grahamgower - perhaps you have a different suggestion for how to produce new maps from the old ones, that are (sometimes) lacking genetic position?

msprime/intervals.py
hyanwong: This is only for "old" maps, to convert them to the new style, so I didn't think it was worth implementing functionality to write out the RateMaps. I guess I could generalise it though.

msprime/intervals.py
hyanwong: bool. Sorry.

PR #1190, opened 2020-09-18
Full functionality for sim_ancestry

### LINKED ISSUE #1171: Fill out functionality for sim_ancestry and document
The current version of sim_ancestry is missing quite a few features. We should support everything in ``simulate``, except for

- mutation_rate 
- population_configurations
- migration_matrix
- demographic_events (these are replaced by the demography argument)
- pedigree (this isn't properly implemented yet anyway)
- mutation_generator

We should document it all as well. Most of it can be copied directly from msprime.simulate. We'll probably want to refactor the ``simulator_factory`` and other code a bit so we don't duplicate too much, but also without making divergent code paths within simulator_factor to deal with the two different front ends.

jeromekelleher:
Closes #1171

This is a complete-ish proposal for the new ``sim_ancestry`` function - see below for summary. I've requested reviews from a whole bunch of people. Don't worry if you don't have time to go through everything (or at all), I just wanted to keep people in the loop and given them an early chance to complain if they hate the new interface!


jeromekelleher:
This is ready for review now, I think. It needs documentation, obviously, but I'm hoping the code is fairly clear about what all the various corner cases etc are and how we're trying to accomodate them. I'll file some follow-up issues to track the stuff that's come up, and for documenting the API.

The main differences from simulate are:

- Samples refer to k-ploid *individuals* and we fill out the individual table
- Renamed a few parameters so they are clearer, ``length`` -> ``sequence_length``, ``Ne``->``population_size``
- Also renamed ``from_ts`` to ``initial_state``, as this is much more descriptive of what it actually *is* now. I'm imagining documenting this as an advanced way of initialising your simulations, as well as recapitating.
- Much more use of type polymorphism on the input values, so we now just have ``recombination_rate`` which can be either a single value or a RateMap instance.
- ``discrete_genome`` is True by default.

I tried to refactor things a bit so that we're not duplicating too much code between ``sim_ancestry`` and ``simulate``.

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+    # Doing this with numpy is slighty obscure, but it's *much* faster
+    # than simple loops. For 10^7 samples with ploidy=2 we go from
+    # several minutes to a few seconds.
+    ind_flags = np.zeros(n, dtype=np.uint32)
+    tables.individuals.set_columns(flags=ind_flags)
+    N = n * ploidy
+    node_individual = np.zeros(N, dtype=np.int32)
+    a = np.arange(n, dtype=node_individual.dtype)
+    for j in range(ploidy):
+        node_individual[j::ploidy] = a
```
petrelharp, line 799: shorter (and arguably easier to read?) is
```
node_individual = np.repeat(range(n), ploidy)
```

jeromekelleher: Aha, yes, much better, thanks.

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+
+    # TODO we many need to have different ploidy values for the different models.
+    # For example, dtwf is an inherently diploid process and anything else doesn't
+    # make sense. However, having different defaults for ploidy will lead to
+    # confusing problems when we have mixed models, so perhaps we should raise
+    # an error from models that require ploidy=2 and force them to explicitly
+    # set it.
+    # TODO Might also be worth considering setting ploidy=2 by default. This
+    # *is* likely to be the most common application, and those doing haploid
+    # stuff can set ploidy=1, if they like.
```
petrelharp, line 967: I support a default of ploidy=2.

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+        random_generator = _parse_random_seed(random_seed)
+    elif random_seed is not None:
+        raise ValueError("Cannot specify both random seed and random generator")
+
+    return Simulator(
+        tables=initial_state,
+        random_generator=random_generator,
+        recombination_map=recombination_map,
+        gene_conversion_map=gene_conversion_map,
+        gene_conversion_track_length=gene_conversion_track_length,
```
petrelharp, line 1032: I think it's "tract" not "track": [891 hits](https://scholar.google.com/scholar?q="gene+conversion+tract") versus [22](https://scholar.google.com/scholar?q="gene+conversion+track").

msprime/demography.py
```diff
@@ -225,6 +236,17 @@ def from_old_style(
             demography.events = demographic_events
         return demography
 
+    @staticmethod
+    def simple_model(population_size=1):
+        """
+        Returns a simple single-population model.
+        """
+        check_population_size(population_size)
+        # TODO what should the population name be here??
```
petrelharp, line 245: `pop0` seems good?

petrelharp (approved):
Looks great! I've looked over everything, and it seems to do what we want, except the details about populations that aren't solid yet. I'm a bit worried that so many existing tests are switched to using `_parse_sim_ancestry` instead of `sim_ancestry`, though?

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+    of the tables.
+    """
+    if isinstance(samples, collections.abc.Iterable):
+        # We check if the samples can be interpreted as a list first because
+        # we may want to extend the semantics here later to include specifying
+        # a set of individuals from a pedigree defined in the initial state.
+        # The most natural way to do this would be a numpy array of individual
+        # IDs. By checking the iterable condition first, we make sure that we're
+        # not interpreting input numpy arrays of length 1 as a numeric argument.
+        # Interpret as a list of Sample objects.
```
petrelharp, line 821: maybe the following should be done with numpy, also, e.g.
```
pops = np.array([s.population for s in samples])
node_pops = np.repeat(pops, ploidy)
node_inds = np.repeat(range(n), ploidy)
... set_columns...
```

msprime/ancestry.py
JereKoskela: I agree. It also makes the defaults consistent with ms, which might be helpful for users.

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
                 )
+        population_size = 1 if population_size is None else float(population_size)
+        demography = demog.Demography.simple_model(population_size)
+    elif isinstance(demography, demog.Demography):
+        if population_size is not None:
+            raise ValueError("Cannot specify demography and population size")
+    else:
+        raise TypeError("demography argument must be an instance of msprime.Demography")
+    # TODO should we validate that the population sizes are positive integers
+    # for DTWF? It's not actually required, but could help catch some errors?
```
JereKoskela, line 998: Is there a reason to not do this for an explicitly finite population model?

JereKoskela:
Most of these changes are to parts of code I've never looked at, but the `sim_ancestry` interface seems very clear, and definitely covers everything we might want for multiple merger models.

JereKoskela (approved):

msprime/ancestry.py
jeromekelleher: Opened issue #1202 to track (or tract?)

msprime/ancestry.py
jeromekelleher: Mainly because it gets caught at runtime during the simulation - we'll get a "population size has gone to zero" (or similar) error. It's slightly murky, because DTWF rounds the population sizes because of exponential growth, so we do actually handle non-integer sizes within the simulation. It seems a bit odd to say you must have integer sizes on the way in, then, and could end up being awkward. :shrug: I don't know though.

jeromekelleher:
>  I'm a bit worried that so many existing tests are switched to using _parse_sim_ancestry instead of sim_ancestry, though?

These are all new tests @petrelharp - all the old tests are untouched. Since the actually simulations are already pretty well tested, I focused on validating the argument parsing side of things.

msprime/ancestry.py
JereKoskela: Good point about needing to round anyway. In that case, I think it feels cleaner to not bother with a check here either.

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+
+    # Simple defaults.
+    start_time = 0 if start_time is None else float(start_time)
+    end_time = sys.float_info.max if end_time is None else float(end_time)
+    # Note: casting to bool here means that we support Truthiness, *except* for
+    # None, which is interpreted as "give me the default". This may lead to
+    # confusing problems for (misguided) users of truthiness.
+    discrete_genome = True if discrete_genome is None else bool(discrete_genome)
+    record_migrations = False if record_migrations is None else bool(record_migrations)
+    record_full_arg = False if record_full_arg is None else bool(record_full_arg)
```
grahamgower, line 899: Could use a coercion function like this?
```python
def coerce_tristate(b, meaning_of_None):
    if b is None:
        return meaning_of_None
    if not isinstance(b, bool):
        raise TypeError("expected True, False, or None")
    return b

# truth table
assert(True == coerce_tristate(None, True))
assert(True == coerce_tristate(True, True))
assert(False == coerce_tristate(False, True))
assert(False == coerce_tristate(None, False))
assert(True == coerce_tristate(True, False))
assert(False == coerce_tristate(False, False))
```

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+            raise ValueError("Either samples or initial state must be specified")
+        initial_state = tskit.TableCollection(sequence_length)
+        demography.insert_populations(initial_state)
+        _parse_samples(samples, ploidy, initial_state)
+    else:
+        if samples is not None:
+            raise ValueError("Cannot specify both samples and initial_state")
+        if sequence_length != initial_state.sequence_length:
+            raise ValueError(
+                "The initial_state sequence length must agree with other parameters"
```
grahamgower, line 1012: It might help the user to see a list of other parameters that this must agree with. Although I suppose such a list might get longer in the future (e.g if a mutation map is added), so this risks becoming outdated and thus confusing if the problem parameter isn't in the list.

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+            )
+        if len(initial_state.populations) == 0:
+            raise ValueError(
+                "initial_state tables must defined at least one population."
```
grahamgower, line 1016: s/defined/define/

grahamgower (approved):
LGTM!

msprime/ancestry.py
jeromekelleher: As in, enforce lack of truthiness, so people have to cast to bool explicitly themselves if they want it? Good idea...

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+        if is_dtwf:
+            # A default size of 1 isn't so smart for DTWF and almost certainly
+            # an error.
+            if population_size is None:
                 raise ValueError(
-                    "Must specify a sequence_length for simulations of a "
-                    "genome with recombination or gene conversion without "
-                    "rate map"
+                    "When using the DTWF model, the population size must be set "
+                    "explicity, either using the population_size or demography "
```
gtsambos, line 987: (minor typo) - explicitly

msprime/ancestry.py
```diff
@@ -721,113 +762,375 @@ def simulate(
...
+    demography.validate()
 
-    sim = simulator_factory(
-        sample_size=sample_size,
-        samples=samples,
-        ploidy=ploidy,
-        length=sequence_length,
+    if initial_state is None:
+        if samples is None:
+            raise ValueError("Either samples or initial state must be specified")
```
gtsambos, line 1003: By initial state you mean a tree sequence inputted with the optional argument `initial_state` here right? I'd make this more specifically clear by using the exact argument name.

msprime/ancestry.py
```diff
@@ -851,6 +1154,9 @@ def __init__(
         ll_tables = _msprime.LightweightTableCollection(tables.sequence_length)
         ll_tables.fromdict(tables.asdict())
 
+        assert len(gene_conversion_map) == 1
```
gtsambos, line 1157: Is this assertion here because you want the user to specify a uniform gene conversion rate for now? It might be nice to raise a verbose error for the user about this

gtsambos (approved):
Nice! Only very minor comments from me, though I'm not very familiar with all of this code base so some of it I just scanned. Is there a reason why you changed the various `_parse` methods on Ancestry objects etc. to have the underscore in front? To me that makes them seem like purely internal methods that you aren't expecting people to actively use, but you have invoked them directly in some of the tests -- will users ever need to do this too?

jeromekelleher:
> Is there a reason why you changed the various _parse methods on Ancestry objects etc. to have the underscore in front? To me that makes them seem like purely internal methods that you aren't expecting people to actively use, but you have invoked them directly in some of the tests -- will users ever need to do this too?

Excellent question @gtsambos! Putting the underscores in front of the function names is an attempt to clean up the namespace and to make it clear that these *are* internal functions that are not intended for end-users. It's fine for us to use them internally (i.e., for testing), but it's giving a (hopefully) unequivocal signal to end users that they should *not* be using these functions.

msprime/ancestry.py
jeromekelleher: Done. Thanks for the input.

msprime/ancestry.py
jeromekelleher: Good call. Opened #1211 to track.

msprime/ancestry.py
jeromekelleher: Done, ta.

msprime/ancestry.py
jeromekelleher: Done.

msprime/ancestry.py
jeromekelleher: opened #1212 to track.

jeromekelleher:
Thanks very much for the comments all, this was super helpful. I've changed the default ploidy to 2,  implemented other suggestions, and punted a few off to new issues. Let's merge!

PR #1191, opened 2020-09-19
Move rate_map_t declarations into new rate_map.h

castedo:
There is currently ratemap.c and test/test_rate_map.c. Looking at file sizes, seems to make sense to move the rate_map_t declarations into its own rate_map.h. For the new perf improving code I'm writing I'm thinking of adding them into rate_map.h/c since they are targeted at speeding up rate_map_t execution (at least for now).

Let me know what you think. Feel free to merge this PR into master or wait until a less trivial PR. Either way, I'll keep the branch name around for cleaner diffs.

### DIFF
--- a/lib/rate_map.h
+++ b/lib/rate_map.h
@@ -1,0 +1,48 @@
+/*
+** Copyright (C) 2020 University of Oxford
+**
+** This file is part of msprime.
+**
+** msprime is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** msprime is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


lib/rate_map.h
```diff
@@ -0,0 +1,48 @@
+/*
+** Copyright (C) 2015 University of Oxford
```
jeromekelleher, line 2: s/2015/2020

jeromekelleher (approved):
Sure, totally reasonable thing to do and it'll make it a bit easier to test in isolation. Just need to change the copyright year in the header.

lib/rate_map.h
castedo: replaced and rebased/squashed

jeromekelleher (approved):

PR #1193, opened 2020-09-21
Add asv update script and version benchmarks

### LINKED ISSUE #1165: Document running airspeed velocity benchmarks and set up site
#1146 and #1138 added some initial benchmarks using [airspeed velocity](https://asv.readthedocs.io/en/stable/). For the record, here's what the current set look like 

image

We have some basic benchmarks for Hudson and DTWF simulations. Getting these up and running and working for the problems I was working on was pretty frustrating and there's a lot of gotchas involved in running it. Hopefully these will settle down a bit now, though. One of the reasons it's so awkward is that ASV is built to accurately time short-lived functions, but we're mainly interested in things that take 10s of seconds (at least) to run.

Ideally, we'd like to run these regularly and keep a site updated with the results. The simplest way to do this is probably for us in Oxford to run the benchmarks on one of our servers on a cron job, and publish the results to github pages. We can look into having a dedicated machine at some point, but this will get us most of the way without extra faff of trying to find appropriate hardware.

@benjeffery, any thoughts here?

benjeffery:
Fixes #1165 
Site is live at https://tskit-dev.github.io/msprime-asv and being updated by a cronjob from an old laptop `poplar` in my loft.
It will slowly cover the old commits over a few days.
Specs:
 - Core i5-3317U @ 2.60GHz
 -  4GB RAM
 - Python 3.8.

We somehow ended up with two benchmarking sections in the docs - I've fixed that too.

### DIFF
--- a/benchmarks/update_asv_site.sh
+++ b/benchmarks/update_asv_site.sh
@@ -1,0 +1,19 @@
+source env/bin/activate
+git pull upstream main
+sudo cpufreq-set -c 3 -g performance
+#asv does recent commits first, so by letting it run for 55min, and the cron to 1hr 
+#it will always keep up with new commits, but also process the backlog
+timeout 3300s asv run -j 4 --show-stderr --cpu-affinity 3 --skip-existing ALL
+sudo cpufreq-set -c 3 -g powersave
+asv publish
+cd .asv
+rm -rf docs
+mv html docs #Github doesn't let you choose arbitary subfolders to serve so we have to use docs
+rm -rf .git
+git init
+git remote add origin git@github.com:tskit-dev/msprime-asv.git
+git checkout -b main
+git add *
+git commit -m "Automated asv upload"
+git push -f origin main
+cd ..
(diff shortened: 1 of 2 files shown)


benchmarks/benchmarks.py
```diff
@@ -47,6 +47,9 @@ class LargeSimulationBenchmark:
     rounds = 1
     min_run_count = 1
     timeout = 120
+    # As we may wish to change code but not have ASV reset benchmarks we set a
+    # fixed version.
+    version = "1"
```
jeromekelleher, line 52: I wasn't sure about this - does this mean we have a single version for all the benchmarks in this class? So, if we change one do we then invalidate all the benchmarks? If we do tweak the code, isn't it most likely to be to change parameters? I think we'd be better off using ASV's approach until we're sure that it doesn't work and resort to manual versioning of each individual benchmark as a last resort.

jeromekelleher (commented):
LGTM, thanks. See comment about versioning.

benchmarks/benchmarks.py
benjeffery: I added this in before I realised it was no bother to regenerate past benchmarks, I've removed it now.

benjeffery:
@jeromekelleher should be good to go!

jeromekelleher (approved):

PR #1196, opened 2020-09-21
Speed up gc left

fbaumdicker:
We now avoid parsing individuals and segments to get the rate of gene conversion events that start left of the first segment.
This should be faster compared to the previous implementation.
I added a flag gc_rate_constant to indicate whether we can use the most simple approach to compute the rate from left (counting individuals) or whether we need to parse the individuals. Currently, if gc_rate_constant is false the parsing will lead to the same result if the gc_map has actually a constant rate. However, it does not yet give the correct gc rate from the left if the gc rate varies.
This will be addressed in another pull request.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -670,7 +670,7 @@ class Simulator:
         for j in range(N):
             assert N == len(migration_matrix[j])
             assert migration_matrix[j][j] == 0
-        assert gene_conversion_length > 1
+        assert gene_conversion_length >= 1
 
         self.tables = tables
         self.model = model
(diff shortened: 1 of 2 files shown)


algorithms.py
```diff
@@ -1622,26 +1629,27 @@ def wiuf_gene_conversion_left_event(self, label):
...
-            lnpc = math.log(1.0 - 1.0 / self.track_length)
-        # generate tracklength
-        u = random.random()
-        tl = math.floor(1.0 + math.log(1.0 - u * (1.0 - pc ** distance)) / lnpc)
+        # generate track_length
+        tl = np.random.geometric(1 / self.track_length)
 
         bp = y.left + tl
-        while y.right <= bp:
+        while (y is not None and y.right <= bp):
```
jeromekelleher, line 1641: Don't need the brackets here.

algorithms.py
```diff
@@ -968,36 +969,42 @@ def get_total_gc_rate(self, label):
...
-            for ind in pop.iter_label(label):
-                dist = self.get_individual_length(ind)
-                gc_left_total += 1 - x ** dist
-                if gc_left_total >= cleft_value:
-                    return ind
+        if self.gc_rate_constant:
+            mean_gc_rate = self.gc_map.mean_rate
+            for pop in self.P:
+                num_ancestors = pop.get_num_ancestors()
+                if gc_left_total + num_ancestors * mean_gc_rate * self.track_length >= cleft_value:
```
jeromekelleher, line 996: There's a lot of multipling and dividing by constants here. Wouldn't it be simpler to have
```python
individual_index = math.floor(gc_left_value / (mean_gc_rate * self.track_length))
for pop in self.P:
    num_ancestors = pop.get_num_ancestors()
    if individual_index < num_ancestors:
        return pop._ancestors[label][individual_index]
    individual_index -= num_ancestors
```

(I haven't tested this!)

jeromekelleher (commented):
Looks great, thanks @fbaumdicker! I wouldn't bother with having two code paths though. Currently there's no way to specify a gc rate map, so it's not really a problem.

jeromekelleher:
How's this looking @fbaumdicker? I'd like to get this simplest version of this we can incorporated ASAP - the code is moving pretty quickly at the moment, so any diffs get left behind quickly!

fbaumdicker:
Almost done. Should I remove the preliminary parts for non_constant gc_rates or keep and cover them?

algorithms.py
```diff
@@ -942,36 +943,48 @@ def get_total_gc_rate(self, label):
...
+        return gc_left_total
 
     def get_total_gc_left(self, label):
         gc_left_total = 0
-        x = (self.track_length - 1) / self.track_length
-        for pop in self.P:
-            for ind in pop.iter_label(label):
-                dist = self.get_individual_length(ind)
-                gc_left_total += 1 - x ** dist
+        if self.gc_rate_constant:
```
jeromekelleher, line 950: Let's just get rid of this branch (and the gc_rate_constant flag) for now - you can keep a branch with what you have, but let's get some working code in as soon as possible.

lib/msprime.c
```diff
@@ -326,11 +326,19 @@ msp_set_gene_conversion_map(msp_t *self, size_t size, double *position, double *
     return ret;
 }
 
+int
+msp_set_gene_conversion_constant(msp_t *self, bool is_constant)
```
jeromekelleher, line 330: I don't think there's much point in adding this infrastructure in - we're just going to delete it as soon as we have a good solution anyway.

jeromekelleher (commented):
Thanks @fbaumdicker - as I say, let's get the simplest working version in here as soon as we can. I don't think there's much point in trying to cover the non-constant case here, since it's not actually doing anything different to the constant case at the moment, and it's just dead code. I'm really keen to get the constant case in ASAP so we can benchmark and see if it's fast enough.

fbaumdicker:
This should work now. Not sure how to increase the code coverage further. But I am super curious how fast this version is.

jeromekelleher:
Perfect! Just needs a quick rebase and squash and we're ready to go @fbaumdicker - I'm also super curious to see how it performs.

fbaumdicker:
Should be rebased and squashed now. I was struggling a little bit with the switch from master to main as the default branch. If something is missing or messed up let me know. But I am afk for two days so any further change would have to wait until Friday.

jeromekelleher (approved):

jeromekelleher:
Looks great, thanks @fbaumdicker! Merging.

jeromekelleher:
@mergifyio rebase

jeromekelleher:
Looks like mergify is confused by the base branch changes. going to merge these manually. This one first.

Issue #1195, opened 2020-09-21
Non-standard binary search semantics


castedo:
In the msprime code there is [a TODO to call `tsk_search_sorted`](https://github.com/tskit-dev/msprime/blob/1ee8d287b15857ac8638c369b99d06bba77e235a/lib/rate_map.c#L168) instead of `msp_binary_interval_search`. In [the comments for `tsk_search_sorted`](https://github.com/tskit-dev/tskit/blob/bc50b5a92e56b578af12caea43d13a82c2008b59/c/tskit/core.c#L622) it reads 'Mirrors the semantics of numpy's searchsorted ...'.

There are two issues highlighted here. One issue is that the behavior of `msp_binary_interval_search` can not be replicated by `tsk_search_sorted` when there are repeated values.  In a sense, the loops of each are fundamentally different. One loop is searching for a strict non-equal upper bound and the other loop is searching for an upper bound that may be equal.

The other problem is that `tsk_search_sorted()` does not have the semantics of `numpy.searchsorted`. It is closest to `numpy.searchsorted(..., side='right')` which one could call "find_first_strict_upper_bound". But what `tsk_search_sorted()` actually does would be called "find_first_strict_upper_bound_or_last_match".

I'm thinking a good way to clean this up is to have functions in msprime and/or tskit that really mirror the semantics of numpy.searchsorted and equivalent standard C++ <algorithm> library functions [std::lower_bound()](https://en.cppreference.com/w/cpp/algorithm/lower_bound) and [std::upper_bound()](https://en.cppreference.com/w/cpp/algorithm/upper_bound). (just two semantics, both reflected in both libraries).

The semantics of these standard functions make a lot of sense to me and I think reflect the wisdom of decades of usage. The existing function can be derived from these two standard binary search function semantics. At some point, some of the hacky tweaks and adjustments in the code can be eliminated using these standard semantics.

One piece of feedback some people might want to provide is on naming of these two functions with standard semantics. If it weren't for the names in the C++ library, I would think good names are:

`find_first_upper_bound` and `find_first_strict_upper_bound`

but I suspect this a little confusing with the C++ library function names. I actually think the C++ library function names are more confusing than the names I suggest. :) But the C++ library function have already been named so that's tough reality.

Suggestions on naming? Other issues? Concerns?

jeromekelleher:
[...]

PR #1197, opened 2020-09-22
fast search lookup

### LINKED ISSUE #1159: Add LRU cache rate_map_position_to_mass
The [rate_map_position_to_mass](https://github.com/tskit-dev/msprime/blob/master/lib/rate_map.c#L149) function gets called a vast number of times during large simulations, and the binary search lookups that it incurs are a significant fraction of the time taken on the [time_human_chr22](https://github.com/tskit-dev/msprime/blob/master/benchmarks/benchmarks.py#L98) benchmark. I think we can significantly improve on this by having a small (say, 3-5 value) LRU cache for this lookup.

I guess the way to do this is to have a small ``lru_cache`` class which we put in its own file and write a few unit tests for. The API would look something like:
```c
int lru_cache_alloc(lru_cache_t *self, size_t size);
int lru_cach_free(lru_cache_t *self);
bool lru_cache_get(lru_cache_t *self, double key, double *value); // returns true for a hit
void lru_cache_put(lru_cache_t *self, double key, double value); // put at the front of the queue and evict LRU value
```

Maybe we'd want ``lru_cache_put`` to return an int just in case we need to return an error, I'm not sure. The only error condition I can think of is putting in a duplicate key, though, and this seems fairly harmless.

@castedo - this seems right up your street, any interest?

castedo:
- to speed up search for index of position

Closes #1159

@jeromekelleher For the Hudson.time_human_chr22 benchmark I'm seeing approx a 25% speed up. I was hoping for a far more spectacular speed boost. I'm a bit surprised the speed up isn't greater. This first implementation hasn't gotten detailed unit tests, clean up and deeper perf analysis, streamlined, etc... I have a hunch this implementation can be sped up a lot more. I just have to follow through more on this first implementation.

Let me know if you see any issues or have ideas.

I need to do some code clean up and streamlining and investigations so this is still a draft. This also relates to #1195.


benjeffery:
> so this is still a draft.

Github has "draft" PRs - there is a drop down when you make the PR, or you can convert to draft with the little button under the reviewers list. Then a "ready for review" button appears that you click when you want it to be non-draft again.

castedo:
Thx for catching that. Don't know why I forgot to do that.

lib/rate_map.c
```diff
@@ -203,3 +210,95 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
     double result_mass = rate_map_position_to_mass(self, pos) + mass;
     return rate_map_mass_to_position(self, result_mass);
 }
+
+#ifdef _Static_assert
+_Static_assert(FLT_RADIX == 2, "Base 2 floating point types required");
```
jeromekelleher, line 215: Looks like _Static_assert is only in C11, which we'll probably never use. On the one hand this is reasonably safe assumption so maybe we don't need it, but I guess we could have 
```
#if FLT_RADIX != 2
    abort()
#endif 
```
or similar?

lib/rate_map.c
```diff
@@ -203,3 +210,95 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+        self->lookups[i].num = (size_t) num;
+    }
+out:
+    return ret;
+}
+
+static inline long
+ceil_to_long(double x)
+{
+    return lrint(ceil(x)); // TODO: test use of fast non-portable equivalent
```
jeromekelleher, line 253: isn't this the same as ``return (long) ceil(x)``? What's the ``lrint`` for?

lib/rate_map.c
```diff
@@ -203,3 +210,95 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    const int power = ilogb((double) n_values) + 1;
+
+    if (*values != 0.0 || !isfinite(max) || max < 0) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+    self->power_shift = power - (ilogb(max) + 1);
+    const long max_index = ceil_to_long(ldexp(max, self->power_shift));
+    self->num_lookups = 1 + (size_t) max_index;
+    self->lookups = malloc(self->num_lookups * sizeof(*(self->lookups)));
```
jeromekelleher, line 271: Need:
```
if (self->lookups == NULL) {
    ret = MSP_ERR_NO_MEMORY;
    goto out;
}
```

lib/rate_map.c
```diff
@@ -203,3 +210,95 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    ret = fast_search_lookup_init_lookups(self, values, n_values);
+out:
+    if (ret != 0) {
```
jeromekelleher, line 274: We don't need to do this conditional free here. The API contract is that once ``x_alloc(&obj)`` has been called the client *must* call `x_free(&obj)``. It is always safe to call x_free after x_init has been called (and hence, x_init must take care to zero out any pointers at the start.

lib/rate_map.c
```diff
@@ -203,3 +210,95 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+}
+
+int
+fast_search_lookup_alloc(
+    fast_search_lookup_t *self, const double *values, size_t n_values)
+{
+    int ret = 0;
+    const double max = values[n_values - 1];
+    const int power = ilogb((double) n_values) + 1;
+
```
jeromekelleher, line 263: We need ``memset(self, 0, sizeof(*self))`` here. See the comment below for why.

lib/util.c
```diff
@@ -334,6 +335,52 @@ msp_binary_interval_search(double query, const double *values, size_t n_values)
...
+const double *
+find_first_upper_bound(const double *values, size_t n_values, double query)
+{
+    if (query > values[n_values - 1]) {
+        return values + n_values;
+    }
+    return values + find_first_upper_bound_or_back(values, n_values, query);
+}
+
+#ifndef __OPTIMIZE__
```
jeromekelleher, line 351: I don't think we need to bother with conditional compile paths here. I'd rather compile in the old code as part of the test suite (i.e., copy the implementation in the test C file) and do the comparison as part of unit tests.

jeromekelleher (commented):
This looks ingenious @castedo, and a 25% improvement is absolutely excellent here!

My main input is that I have no idea how this works - would you mind adding some commentary on the fast lookup as to how the floating point wizardry works?

lib/rate_map.c
castedo: I got some interesting details on this at https://stackoverflow.com/questions/3385515/static-assert-in-c

_Static_assert has been in gcc since version 4.6 and it sounds like it shows up in clang and even VS. It just wasn't standardized until C11. I'm estimating that few if any users are going to compile with a compiler that doesn't have it. And if they don't then this static assert won't happen. But they probably are using IEEE base 2 floats anyway.

There are many ways to re-implement a static assert.

lib/rate_map.c
castedo: `lrint` is a misguided attempt at getting something more optimized than a simple cast. My line here is compiling to 14 machine instructions with one call.

Doing `(long) ceil(x)`, not only is much easier to read, but compiles to 2 machine instructions with one call.

It's pretty cool to see all these super fancy new x64_64 instructions. Like one that does the cast in one instruction. I learned x64 long ago and 64-bit is new to me.

lib/util.c
castedo: This code is going to need to change with some of my next changes. After that it should look less confusing. Also after I add some explanation as to how this overall lookup technique works, this test will make more sense I think. It's not really testing old code vs new code. In its current state it does kind of look like that.
I'll leave this un-resolved to remind me to circle back to you about it.

castedo:
@jeromekelleher thx for pointing out the proper API behavior. That makes a lot of sense.

I only just now realized that I can look at review comments in the file view and not just this conversation view. Now comments make a lot more sense. :)

So I will continue with this fast lookup table approach.

This is what I'm thinking as next steps:
1) make a decision as to whether the searches should be for strict upper bounds or just upper bounds
2) massage the existing **pre-**fast-lookup-table code on a separate branch/PR to match 1) and close #1195
3) add awesome unit tests given 1) and make sure lookup table is as expected
4) add more comments/explanation which can reference 3) as examples

Let me know if you want explanation sooner than 4).

jeromekelleher:
Sounds great, thanks @castedo!

lib/rate_map.c
benjeffery: @jeromekelleher @castedo For tskit we do static asserts with `#error` as in https://github.com/tskit-dev/tskit/blob/master/c/tskit/core.h#L74

jeromekelleher: Yes, thinking about it again - given it's such an obscure and unlikely thing we're guarding against, it seems a bit perverse to depend on something else that we have to test existence for, just to check it. I.e., *if* we're on a platform where FLT_RADIX != 2, then I think there's a very good chance that ``_Static_assert`` doesn't exist either, therefore rendering the exercise pointless.

castedo: thx @benjeffery, that is a far better. It should have occurred to me that nothing needs to get compiled, it's entirely a check that the preprocessor can handle with just `#error`.

lib/rate_map.c
```diff
@@ -235,66 +235,124 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
 #error "Base 2 floating point types required"
 #endif
 
+static bool
+valid_sorted_nonempty_array(const double *array, size_t size)
+{
+    size_t idx;
+
+    if ((ptrdiff_t) size < 1 || isnan(array[0])) {
```
jeromekelleher, line 242: Clever, I haven't seen this way of detecting a negative size before. But what if we have an array of 2**63 elements?!! :wink:

lib/rate_map.c
```diff
@@ -235,66 +235,124 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    }
+    return true;
+}
+
+static bool
+fast_search_lookup_valid(fast_search_lookup_t *self)
+{
+    const double *start, *stop;
+
+    if (!(self->query_multiplier >= 0.0)) { // NaN not valid
```
jeromekelleher, line 259: Just so I'm clear, you're guarding against ``query_multiplier`` being NaN here by negating >=, right?

lib/rate_map.c
```diff
@@ -235,66 +235,124 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
 fast_search_lookup_alloc(
-    fast_search_lookup_t *self, const double *values, size_t n_values)
+    fast_search_lookup_t *self, const double *elements, size_t n_elements)
 {
     int ret = 0;
-    const double max = values[n_values - 1];
-    const int power = ilogb((double) n_values) + 1;
+    double max_element, upper_bound;
+
+    const size_t max_input_size = 1L << (DBL_MANT_DIG - 1); // 4096 terabytes
```
jeromekelleher, line 330: Is the literal ``1L`` safe here? I've had issues before with 1L being interpreted as a 4 byte integer. Safer use ``1LL``?

lib/rate_map.c
```diff
@@ -235,66 +235,124 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    if (elements[0] != 0.0 || isnan(elements[0])) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+    max_element = elements[n_elements - 1];
+    if (max_element < 0.0 || !isfinite(max_element)) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+    upper_bound = fmax(max_element, DBL_MIN); // want a positive upper bound
```
jeromekelleher: Ah, didn't realise we have ``fmax`` in C99. :+1:

jeromekelleher (commented):
Shaping up nicely @castedo - I'm learning all sorts of new stuff here!

lib/rate_map.c
castedo: that will only happen with future computers with 128 bit size_t. :)

lib/rate_map.c
castedo: Yep.

lib/rate_map.c
castedo: Good point. Making sure it's not compiled as 32-bits with 1LL does seem safer.

castedo:
@jeromekelleher, here's some quick status on next steps. I've now done steps 1-3 mentioned earlier. I had a hard time on 1). This was to make a decision whether the performance enhanced search should be for an upper bound or a strict upper bound. Now that I've implemented 2-3 and done perf testing and compared alternatives I'm pretty convinced that searching for a strict upper bound is the way to go. Rough summary of the reasons is slight perf boost and makes this new perf enhanced code simpler and easier to understand I think.

With this in mind, I'm about to create a separate PR which is ONLY introducing a new function for strict upper bound search and using it in the key search getting optimized here. This PR #1197, or perhaps yet another new PR, will build on top of this pivot to strict upper abound. Stay tuned. The new PR is going to be called "pr/strict_upper_bound".

lib/rate_map.c
castedo: Good eye @jeromekelleher thx! This fixed a unit test failure on appveyor running on Windows. After saw it fail on (I suspect) Windows more than once I was thinking "oh boy, this is going to be a pain in the A to figure out." So you saved me a lot of time. 64-bit Windows keeps long as 32-bit.

castedo:
In my latest commit https://github.com/tskit-dev/msprime/pull/1197/commits/56c69be53234d4b57658b45f37d0d448dc348ca3 I've attempted completing the previously mentioned step 4), which is add an explanation on how it works. But after writing it, I'm not sure it's very helpful explanation/documentation/commenting. I'm not very experienced commenting code. With some guidance on what is confusing or helpful or not etc... I can upgrade the comments to be more helpful. Just let me know.

This PR is basically ready to go, but if it literally was merged now it would slow the benchmark down because an assert I've find useful is still compiled into release. But with the assert out, the speed up should be something like 30%.

I figured this PR should wait on #1223 and #1203 getting merged into main before final touches and squashing are done on this. I would also like to re-profile this code because I haven't yet and I was dreaming of much better than 30% speed up.

One other TOTALLY OPTIONAL related branch to think about while you're look at this one is a little experimenting I've been trying out:

https://github.com/castedo/msprime/compare/fast_search_lookup...castedo:nice_fast_search_lookup

This "nice_fast_search_lookup" branch is just like this one, but in it I use an experimental MSP_NICE_ASSERT macro. The roughly idea is for it to be as convenient as an assert, but with the proper graceful error handling of the usual "goto out" pattern. Not something critical to look at. I'm actually a bit on the fence but thought you might want to consider it.

lib/rate_map.c
grahamgower: Does this have the intended behaviour on a 32 bit platform? For 32 bit `size_t`, this will evaluate to `0`, right?  (Does msprime even try to support 32 bit platforms?)

castedo: Good catch @grahamgower! Indeed it should be a `1LL` shifted and not `1L`. It was causing Windows unit tests to fail. `long` is 32-bit on 64-bit Windows.

grahamgower: To be clear, I meant that on a 32 bit platform, `size_t max_input_size` will be a 32 bit variable. So,  `1LL << (53 - 1)` will be cast to a 32 bit variable. So on a 32 bit platform you'll have `max_input_size = 0` even if you change the right hand side from 1L to 1LL.

castedo: Oh, now I follow you. That makes sense. I find the out-of-date diffs that github shows in the conversation tab a bit confusing. This has got to be something like the fifth time that I have misunderstood comments because my brain seems to pay more attention C code than English. :)

I doubt anybody is using 32-bit and I suspect 32-bits can be dropped from support. I'll add an #ifdef #error so that it's clear. @jeromekelleher knows whether this blatant disrespect for 32-bit platforms is acceptable or not.

castedo:
OK, this PR is ready for review. In particular, I'm uncertain what level of comments/explanation is useful or not, clear or not, etc... I tend try and write code that is "self-documenting/self-explanatory" to avoid the problem of code and comments getting out-of-sync over time. And perhaps a tiny teeny bit of laziness. :) Old comments can sometimes increase confusion with since-changed code. But obviously English can explain many things better than code.

And if you have some spare cycles, check out
https://github.com/castedo/msprime/compare/fast_search_lookup...castedo:nice_fast_search_lookup
and share your thoughts.

lib/rate_map.c
```diff
@@ -206,3 +210,249 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+
+    const size_t max_input_size = 1LL << (DBL_MANT_DIG - 1); // 4096 terabytes
+
+    memset(self, 0, sizeof(*self));
+
+    if (!valid_sorted_nonempty_array(elements, n_elements)) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+    if (n_elements >= max_input_size) {
```
jeromekelleher, line 361: I think we can wrap this into ``MSP_ERR_BAD_PARAM_VALUE``.

lib/rate_map.c
```diff
@@ -206,3 +210,249 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    if (elements[0] != 0.0 || isnan(elements[0])) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+    max_element = elements[n_elements - 1];
+    if (max_element < 0.0 || !isfinite(max_element)) {
+        ret = MSP_ERR_BAD_PARAM_VALUE;
+        goto out;
+    }
+
```
jeromekelleher, line 374: A quick comment here on what this is doing would be helpful

lib/rate_map.c
```diff
@@ -206,3 +210,249 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    int ret = 0;
+    const double *ptr, *stop;
+    uint32_t idx;
+    double query;
+
+    self->lookups[0] = elements;
+    ptr = elements;
+    stop = elements + n_elements;
+    for (idx = 1; idx < self->num_lookups; idx++) {
+        query = idx / self->query_multiplier;
```
jeromekelleher, line 315: Another quick note here on what we're storing in self->lookups would be good.

lib/rate_map.c
```diff
@@ -206,3 +210,249 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+const double *
+fast_search_ptr_strict_upper(fast_search_t *self, double query)
+{
+    const double **lookups = self->lookups;
+    double fidx; // index that can be way larger than max size_t
+    size_t idx;
+    const double *ret;
+
+    assert(query >= 0.0);
+    fidx = query * self->query_multiplier;
```
jeromekelleher, line 444: An explanation of what this is doing would be good.

jeromekelleher (changes requested):
This is looking great, thanks @castedo!

I entirely agree that clear code is much better than lots of comments, and we should strive to write code that is so clear and obvious that it doesn't need any comments. However, what we're doing here is really quite subtle and so I think it's important that the actual algorithm is laid out and explained in some detail. I guess the most helpful thing would be to have a worked example in the comments, where we have an example list of values, show what the fast_search lookups generated are, and how an example query proceeds. 

This is an ingenious idea and will be really valuable to us, so I'm just anxious to capture the knowledge of how it works now so that we can maintain and develop it in the long term.

Another thing that helps with making code clear is to have less of it. For example: do we really need to have both fast_search_ptr_strict_upper and fast_search_ptr_upper? Presumably we're only using one of these in the actual simulations, right? So, why implement both? It's just dead code cluttering up the repo. 

We will need full test coverage on all the code paths except for malloc failures. It turns out we're not tracking coverage on the rate_map.c file, can you update the circle config [here](https://github.com/tskit-dev/msprime/blob/570a01892651215bfc9eeb1a6a670915e7358243/.circleci/config.yml#L85) please?

If it turns out we can't write tests to provoke the errors then I suggest dropping the checks or turning them into ``tsk_bug_assert`` calls.

jeromekelleher:
> And if you have some spare cycles, check out castedo/msprime@fast_search_lookup...castedo:nice_fast_search_lookup
and share your thoughts.

It's a nice idea @castedo, but I don't want to make a macro that does a ``goto out``. I've thought a lot about this in the past as a way to reduce the number of lines of code, but decided on balance it's not worth it. It would be very confusing for newcomers and I think would surely lead to unexpected consequences. I'm happy with our asserts now, I don't think we need to change anything else.

lib/rate_map.c
```diff
@@ -206,3 +210,249 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
     double result_mass = rate_map_position_to_mass(self, pos) + mass;
     return rate_map_mass_to_position(self, result_mass);
 }
+
+/* The `fast_search_t` structure and functions
+ *
+ * `fast_search_t` is a lookup table pointing to an array of `double`s
+ * (the "elements"). It does not own the array but does point to all its memory.
+ *
+ * It's purpose is speed up the search for a "query" value in that array of
```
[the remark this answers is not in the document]
castedo: Man, I keep doing this. I need a grammar checker extension for clang-format or something. Can you add one to pre-commit? :)

lib/rate_map.c
```diff
@@ -206,3 +210,249 @@ rate_map_shift_by_mass(rate_map_t *self, double pos, double mass)
...
+    const double *start, *stop;
+    size_t idx;
+
+    if (!(self->query_multiplier >= 0.0)) { // NaN not valid
+        return false;
+    }
+    if (self->num_lookups < 2) {
+        return false;
+    }
+    for (idx = 1; idx < self->num_lookups; idx++) {
```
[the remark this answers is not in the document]
castedo: Thx for sharing that term nominal-case. I'll use that term in a future communication.
Good point on this function and I now see it's an unholy mix of internal state-checking and nominal-case parameter input validation. So I'll separate the two. My thinking here was an independent internal state-checking function that is carved out and not cluttering the main code and also agnostic as to how it gets used in one of four different ways:

1. terminate the process, even for end users (in which can call it inside tsk_bug_assert)
2. "if not goto out" it with MSP_ERR_ASSERTION_FAILED
3. terminate the process in only debug builds (call it from inside assert())
4. only call it directly from unit tests (and not indirectly via debug builds in automated tests)

Obviously if the function is used only via 4. then the function can exist in only the unit test code. I've been thinking of starting an initial descriptive (not prescriptive) document, maybe in FAQ format, to collect your and other's thoughts on three inter-related concepts:
* what are the debug vs release builds
* use of tsk_bug_assert vs assert
* nominal/expected/unexpected/tolerated execution scenarios

I'm hoping it can helpful to folks contributing to the C codebase.

For now I'll just opt for 4, unless you prefer opting for one of the other three. I'd probably go for BOTH 2 and 3, but that reflects my habit of being 'assert happy', never crashing the process for users and writing proprietary code that few if any people will ever read. I'm OK commiting which ever of 1 through 4 you want. However as a user it makes me very unhappy when Python libraries terminate my process.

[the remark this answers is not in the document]
jeromekelleher: The way we've separated out the state-checking and input validation up to now has been to have a function ``x_verify`` on class x, e.g. [fenwick_verify](https://github.com/tskit-dev/msprime/blob/main/lib/fenwick.c#L49). We then call x_verify from the unit tests. It would be good if we could follow this pattern (so, your option 4).

We don't have any distinction between debug and release builds. There's no real point, with Python being in charge of the distribution and I'd rather not introduce this extra layer of complexity. I've tried fiddling with all sorts of flags for msprime in past and it really makes very little difference because it's fundamentally a memory-bound process. So, we build the tests with asserts turned on, and we do whatever Python does by default when we ship release binaries.

I absolutely agree 100% that Python libraries should never terminate the process. There must be 100s of millions or billions of individual msprime simulations run at this point and I would bet that < 100 of these have ended because of process termination. We have some very large number of "nines" of reliability. We are absolutely not coding for the common case here when we do a tsk_bug_assert - we're coding for the 1-in-a-hundred-billion case.

I think you're missing full context of how this C code is run: there are three layers of input validation performed before a simulation is run. Parameters are validated in the high-level Python, at the C Python layer and finally in the C code that you are working with. I feel confident in saying this is far stronger input validation that most software gets, and this frees us to write C code that is unencumbered by having to worry about all possible inputs. The inputs are *stringently* validated before the C code sees them, and this lets us write simpler, easier to understand, and more efficient C.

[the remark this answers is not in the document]
jeromekelleher: > How would you classify the required sortedness of array of elements/values ... that fast_search_t "points" to?


> Should fast_search_alloc() not check and assume callers will make this validation? (rate_map_alloc currently does make this validation)

That's a bit of a grey area, since we're only using fast_search in one place, and it's already validating this requirement. I guess it's no harm checking the sortedness on the way in for fast_search_alloc, if you don't mind writing test cases to cover this.

lib/util.h
```diff
@@ -129,6 +129,20 @@ size_t idx_1st_upper_bound(const double *values, size_t n_values, double query);
 size_t idx_1st_strict_upper_bound(
     const double *elements, size_t n_elements, double query);
 
+inline const double *
+ptr_1st_upper_bound(const double *start, const double *stop, double query)
```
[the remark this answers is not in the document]
castedo: I'm remembering now that in another PR you had a similar reaction to inline functions in headers. For this PR I'll just avoid inline functions in headers like I did in that other PR.

I'm guessing that, at least partly, one of the things you don't like is having function implementations mixed in with function declarations; breaking up what would otherwise be a single coherent block of just declarations.

I can make a branch, kind of like what I did with the macro, with a suggestion for keeping a "clean" section of just declarations, and then inline implementations in a separate section at the very end of the header file (or it can be a separate 'inline implementations' file). Or perhaps you already know about this approach and already know you do not like it. I can include some thoughts on the suggestion and also on how the inline keyword is currently being used in the codebase.

Regarding the difference it makes in terms of compiled code I have a complicated answer, but the simple answer is that it is not measurable in time with my current test setup (which has high variance in timings).

castedo:
> a macro that does a `goto out`

Yeah, it's a pretty "dramatic" (i.e. dangerous :)) statement to have behind a macro. I would not be surprised if a newcomer did not fully grasp that execution will jump to `out:` and program continue. I started out with the macro name THROW_ASSERT to make clear a change in execution path. But then it had a weird feel, as through if the condition were true, then it would "throw". I figure any word like "THROW" or "GOTO" or "GO_OUT" would have a similar weird feel.

Another thing that didn't work out so well after actually trying it out, was error codes. In the end, it seemed like I almost always should specify an error code, and so most of those single lines ended up being rather long and harder to read. And it makes sense that usually one would want to always specify the error code.

It's seems nicer to usually have the error code and condition on separate lines.

Oh well.

castedo:
> do we really need to have both fast_search_ptr_strict_upper and fast_search_ptr_upper?

Not really. `fast_search_ptr_upper` is getting unit tested but no production code is calling it. I was just thinking about the prospect that somebody wants to speed up the other kind of binary search and didn't realize this same lookup table can be used. But I can just say that in the additional comments ... that writing a non-strict version is easy, just copy the strict version and replace strict calls with non-strict. If this need arises I guess I will probably still be alive and online to help too. :)

lib/util.h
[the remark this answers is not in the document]
jeromekelleher: The main thing I'm aiming for is simplicity: someone who knows a little bit of C should be able to read this and understand it *unless* there's a good reason for doing tricky complicated things. If we can't measure the difference perf-wise, then that's not a good reason in my book.

lib/rate_map.h
```diff
@@ -23,13 +23,39 @@
...
+    fast_search_t position_lookup;
 } rate_map_t;
 
+int fast_search_alloc(fast_search_t *self, const double *values, size_t n_values);
+int fast_search_free(fast_search_t *self);
+const double *fast_search_ptr_upper(fast_search_t *self, double query);
+const double *fast_search_ptr_strict_upper(fast_search_t *self, double query);
+
+inline size_t
+fast_search_idx_upper(fast_search_t *self, double query)
```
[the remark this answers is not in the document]
castedo: Good question. Due to the compiler's freedom to move around what registers are used, I think it usually (depending on the caller) would result in fewer machine instructions. I'm hoping to do a quick investigation out of curiosity. Not that these few machine instructions are doing to make much diff.
And in my next commit I'll make the issue moot. The trade-off here doesn't need to be perf vs inlines in headers. All these ptr_* versions of the functions do not need to be available to multiple .c files. They only get used in one spot. So I've just moved this inline code to the .c files that use them. So it'll be just a hypothetical future trade-off in reduced code re-use.

castedo:
Thanks for the guidance of what extra explanation and comments to add. That was very helpful. The idea of a working example helped me focus on a specific idea to explain. I hope the comments and explanations are helpful. The explanation section is really long. It kind of feels weird having so much explanation in the middle of that file.

Let me know if you want this to be squashed.

lib/rate_map.h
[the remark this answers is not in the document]
castedo: I was half-incorrect. There is still a perf trade-off avoiding inline functions in headers. One of the functions (fast_search_idx_strict_upper) that was inlined is used across .c files, namely rate_map.c and the unit tests.

I did an investigation and confirm that the answer to your question is: yes, it does make a difference, no, you do not get the same thing. Specifically, by not having `rate_map_position_to_mass` call `fast_search_idx_strict_upper` as an inline function, `rate_map_position_to_mass` is spending 79 instructions per call. In contrast, letting `fast_search_idx_strict_upper` be an inline function the result is 74 instructions per call. So 6% fewer instructions.

The reason for this is complicated and really interesting. I can write a longer answer with the details.

Anyway, not a big difference but it is there. The current commit has no inline functions in headers. But that does make it slightly less performant. An inline version of this with the 6% boost and inline function implementations at the end of headers files is here
https://github.com/castedo/msprime/compare/fast_search_lookup...castedo:inline_fast_search_lookup
in case you want to use that.

castedo:
I've created a separate PR #1237 with branch 4G_fast_search which is an improvement on this PR/commmit/branch. It is derived from this commit so merging this commit I don't think will complicate merging the newer faster PR/branch.

The newer faster PR/branch clocks `rate_map_position_to_mass` at 38 machine instructions per call down from 79, so more than 50% drop.

More details to be written out in the new PR.

castedo:
This now obsoleted by merged #1237.

PR #1199, opened 2020-09-23
add time to slim metadata

### LINKED ISSUE #1188: use mutation.time to populate `origin_generation_` in slim mutation metadata
Once #1179 is in. Note that origin_generation_ is an *integer*, but following discussion with @bhaller, we decided it's still useful to put this in there, rounded. But it needs to be multiplied by -1, maybe offset by one, and either floor'ed or ceiling'ed. Details TBD.

Explanatory note: the "origin_generation_" entry in the metadata does *not* have to be consistent with anything in SLiM, since people can change it in the fly, use it as a tag, and other confusing things. So, we could just leave it set to 0 and nothing would break. But, we might as well put something useful in there. Note that under dtwf it will be consistent with 'time'.

petrelharp:
Closes #1188

### DIFF
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -146,15 +146,28 @@ class SLiMMutationModel(_msprime.SLiMMutationModel, MutationModel):
     """
     An infinite-alleles model of mutation producing "SLiM-style" mutations.
 
-    The ancestral state of each new site is set to the empty string,
-    and each derived state is produced by appending the "next allele" to the previous
-    state. The result is a comma-separated string of all mutations that have occurred
-    up to the root.
+    To agree with mutations produced by SLiM, the ancestral state of each new
+    site is set to the empty string, and each derived state is produced by
+    appending the "next allele" to the previous state. The result is that each
+    allele is a comma-separated string of all mutations that have occurred up
+    to the root.
+
+    Mutations produced by SLiM carry both a ``time`` attribute, in units of "time ago"
+    as usual for tskit, as well as a metadata attributed called "origin_generation",
+    in units of time since the start of the simulation. Adding these two together -
+    time since the start of the simulation plus time until the end - is equal to the
+    total number of generations of the simulation. The origin_generation is not
+    currently used by SLiM, but for consistency, the origin_genration attribute for
+    mutations produced by this model is set equal to ``slim_generation`` minus
+    ``floor(mut.time)``, where ``mut.time`` is the (tskit) time ago of the mutation.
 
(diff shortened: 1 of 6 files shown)


petrelharp:
This is waiting on #1179.

petrelharp:
Whoops - I need to finish this off.

jeromekelleher:
Bumping this one - it'd be nice to get this cleared away @petrelharp (but no rush, obvs)

petrelharp:
Ok, this should be good to go. I have no idea why the docs build is failing: they build fine locally, and it is failing while building `ancestry`, and there are no changes in this PR that affect `docs/ancestry.rst`, AFAICT. Doing a diff against #1346, which is passing the docs build, doesn't turn up anything either. If no-one else knows, I can do some more debugging.

petrelharp:
Ok, this should be good to go. I have no idea why the docs build is failing: they build fine locally, and it is failing while building `ancestry`, and there are no changes in this PR that affect `docs/ancestry.rst`, AFAICT. Doing a diff against #1346, which is passing the docs build, doesn't turn up anything either. If no-one else knows, I can do some more debugging.

msprime/mutations.py
```diff
@@ -146,15 +146,28 @@ class SLiMMutationModel(_msprime.SLiMMutationModel, MutationModel):
...
+    currently used by SLiM, but for consistency, the origin_genration attribute for
+    mutations produced by this model is set equal to ``slim_generation`` minus
+    ``floor(mut.time)``, where ``mut.time`` is the (tskit) time ago of the mutation.
 
     :param int type: The nonnegative integer defining the "type" of SLiM mutation
         that will be recorded in metadata.
     :param int next_id: The nonnegative integer to start assigning alleles from.
         (default: 0)
+    :param int slim_generation: The "SLiM generation" time used in determining the
+        "origin_generation" metadata attribute of mutations. This can safely be left
```
jeromekelleher, line 169: The "safely" reads a bit weirdly - maybe "this can usually be" instead? (If there's no good reason for anyone using this parameter, is there much point in exposing it?)

jeromekelleher (approved):
LGTM! One minor documentation suggestion.

The docs failure is because we disabled some stuff during December because of a security hole and haven't gotten around to fixing it yet, I think, nothing to worry about.

benjeffery:
Docs failure was due to an unpinned dependency breaking. I've pinned in #1359

petrelharp:
Ok, made that docs change!

Issue #1192, opened 2020-09-19
keep enabling asserts by default?


castedo:
As I'm improving the performance for some of these great benchmarks available, I have to ask whether enabling asserts by default should still be the case approaching v1.0.
https://github.com/tskit-dev/msprime/blob/96dc8d32528dfee218d8e7f6e53cfced9f797bdb/setup.py#L126

I usually would insert asserts often, but in some functions that should be particularly fast, i'm not putting them in. I could introduce a new C define like REALLY_DEBUG and then have a real_assert() function but that seems questionable.

What are people's thoughts? Seems to me there are two choices:

A) use NDEBUG and assert like it's usually used and define some assert() func alternative that is always enabled, say a function called kaboom() or abort_if() ...

B) keep NDEBUG undefined and keep assert running in production, but the having some other define like HARD_DEBUG and hard_assert that can be used in special cases ... like the amazyingly fast code i'm writting :)

molpopgen:
My two cents:

If you can't measure the asserts in a profile, then leave them in.  Yes, it is a little odd to have them enabled, but they have proven useful in catching corner case bugs and annoying numerical issues.  I think it is worth occasionally crashing the interpreter if it helps to squash another bug.

Your option A doesn't solve the "never ever abort" in Python issue.  As for option B, it only matters if you can measure it.

castedo:
Some more details to just make sure we're all on the same page:
1) there are currently two kinds of builds:
    - the "release" build, with the Python package, via msprime/Makefile, setup.py build_ext3
    - the "debug" build, via msprime/lib/meson.build, run by automated C unit tests
2) there are 174 calls to "assert()" in `msprime/lib/*.c`
3) at the moment there is no C define which allows C code to detect which of these two builds is being compiled

Just to clarify possibilities here, imagine option B with `fail_if()` defined just like `assert()` from 'assert.h', but unlike standard `assert.h` it always fails/aborts regardless of NDEBUG being defined or not.

In this scenario, @molpopgen recommendation can be "translated" (for option B) as "If you can't measure the `fail_if()` in a profile, use `fail_if()` instead of `assert()`. `fail_if()` is useful in catching corner case bugs and annoying
numerical issues."

I should share that as someone new to the code base I was slightly surprised `assert` is getting compiled into code that gets released to regular users. I imagine there is a great variety in how different programmers use `assert()` and in what situations they think the use of `assert()` is appropriate or not. I certainly have my own inclinations and opinions.

[...]

PR #1201, opened 2020-09-23
Make msp_binary_interval_search behave like standard binary search algos

### LINKED ISSUE #1195: Non-standard binary search semantics

castedo:
Closes #1195

  - msp_binary_interval_search has binary search semantics like:
      * numpy.searchsorted(..., side='left')
      * std::lower_bound() of the C++ <algorithm> library
  - also add and use new macro `expensive_assert()` per #1192

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -310,24 +310,29 @@ __msp_safe_free(void **ptr)
     }
 }
 
-/* Find the `index` of the interval within `values` the `query` fits, such that
- * values[index-1] < query <= values[index]
- * Will find the leftmost such index
- * Assumes `values` are sorted
+/* This function follows standard semantics of:
+ *   numpy.searchsorted(..., side='left') and
+ *   std::lower_bound() from the standard C++ <algorithm> library
+ * PRE-CONDITION:
+ *   1) `values` are sorted and not NaN
+ * RETURNS:
+ *   First (leftmost) `index` of upper bounds of `query`
+ *   **or** `n_values` if no such upper bound is in `values`
+ * POST-CONDITION:
+ *   If `query` is not strictly greather than all values, then::w
+ *     values[index-1] < query <= values[index]
  */
 size_t
 msp_binary_interval_search(double query, const double *values, size_t n_values)
 {
-    if (n_values == 0) {
-        return 0;
-    }
     size_t l = 0;
-    size_t r = n_values - 1;
+    size_t r = n_values;
     size_t m;
 
     while (l < r) {
         m = (l + r) / 2UL;
-
+        /* TODO: uncomment assert when #1203 done and this longer slows down release
+           assert(values[l] <= values[m]); */
         if (values[m] < query) {
             l = m + 1;
         } else {
--- a/lib/rate_map.c
+++ b/lib/rate_map.c
@@ -156,17 +156,17 @@ rate_map_position_to_mass(rate_map_t *self, double pos)
     double offset;
     size_t index;
 
-    /* TODO we should add an LRU cache with ~3 items here to cache recent
-     * hits and avoid looking up the map for them. This will make a difference
-     * in the large recombination map case.*/
-    if (pos == position[0]) {
+    assert(pos >= 0.0);
+    if (pos <= position[0]) {
         return 0;
     }
+    assert(pos <= position[self->size]);
     if (pos >= position[self->size]) {
         return self->cumulative_mass[self->size];
     }
-    /* TODO replace this with tsk_search_sorted */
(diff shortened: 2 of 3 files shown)


lib/util.c
```diff
@@ -307,24 +307,28 @@ __msp_safe_free(void **ptr)
...
-    }
     size_t l = 0;
-    size_t r = n_values - 1;
+    size_t r = n_values;
     size_t m;
 
     while (l < r) {
         m = (l + r) / 2UL;
-
+        expensive_assert(values[l] <= values[m]);
```
castedo, line 331: I should mention (probalby here?), that this assert is indeed expensive, it causes approx 10% slow down in the Hudson.time_human_chr22 benchmark.
But it's also a good check for development and testing because this is per-condition for a sane call, even though it might be harmlessly violated in some cases.

castedo:
I should mention that the assert is indeed expensive, it causes approx 10% slow down in the Hudson.time_human_chr22 benchmark. But it's also a good check for development and testing because this is a pre-condition for a sensible call. If it's getting violated, it's worth investigating.

lib/util.c
jeromekelleher, line 331: I'm not sure about this @castedo, see #1203. I think the simplest thing to do for now would be to put in your assert, but comment it out with a note saying "uncomment this once bug_assert has been implemented, #1203. "

jeromekelleher (commented):
LGTM, thanks @castedo, but see the comment about assertions. Basically I think you're right and we should use them in the standard way, but there's a bit of upstream infrastructure needs to be put in place first.

castedo:
@jeromekelleher I think this is now good to go.

BTW, also @ivan-krukov, I should mention, just because it seems interesting. Mainly out of curiosity, I investigated the perf compared to the implementation in the C++ standard algo library ported to C. I thought it really interesting that this implementation by I think @ivan-krukov was actually slightly FASTER (something like 5%, I don't remember the exact number) with x86_64. The reason was two fold:
1) x86_64 will do add "half distance" calc in fewer instructions when calculating with integer indexes rather than pointers
2) x86_64 has one single instruction that can do `array[index]`, in contrast to older microprocessors where this would need two instructions (one to add to a pointer and another to derference (load the memory contents to a register)).

jeromekelleher (approved):
LGTM, thanks @castedo

jeromekelleher:
That is interesting @castedo - good to know that the straightforward code using integer indexes is actually a little bit faster than fiddling with pointers!

PR #1203, opened 2020-09-24
Proposal for bug_assert macro

jeromekelleher:
Recent discussions with @castedo have highlighted that the current usage of ``assert`` in the msprime code is problematic (#1192). Here's a proposal for a new ``bug_assert`` macro, which is basically a drop-in replacement for our current usage of ``assert``, which is (a) more useful in that it tells the user what do do; and (b) solves some annoying distribution problems with compiler warnings.

Annoyingly, we can't turn off the NDEBUG flag without generating some compiler warnings from tskit, so I think we'll have to push this change through to tskit too, if we think it's a good idea. So, the proposal would be

1) Add a new ``tsk_bug_assert()`` macro to tskit, which has the same definition as this one. Update tskit to use this, and tag a C release.
2) Pull in the new tskit C release to msprime, and update this PR to use ``tsk_bug_assert``.

Thoughts? @molpopgen, @castedo, @benjeffery?

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -18,7 +18,6 @@
 */
 #include <stdio.h>
 #include <string.h>
-#include <assert.h>
 #include <float.h>
 #include <math.h>
 
@@ -667,12 +666,12 @@ msp_alloc_segment(msp_t *self, double left, double right, tsk_id_t value,
     if (seg == NULL) {
         goto out;
     }
-    assert(left < right);
+    bug_assert(left < right);
     if (self->recomb_mass_index != NULL) {
-        assert(fenwick_get_value(&self->recomb_mass_index[label], seg->id) == 0);
+        bug_assert(fenwick_get_value(&self->recomb_mass_index[label], seg->id) == 0);
     }
     if (self->gc_mass_index != NULL) {
-        assert(fenwick_get_value(&self->gc_mass_index[label], seg->id) == 0);
+        bug_assert(fenwick_get_value(&self->gc_mass_index[label], seg->id) == 0);
     }
     seg->prev = prev;
     seg->next = next;
@@ -708,7 +707,7 @@ msp_alloc(msp_t *self, tsk_table_collection_t *tables, gsl_rng *rng)
     /* Use the standard coalescent by default. */
     self->model.type = -1;
     ret = msp_set_simulation_model_hudson(self);
-    assert(ret == 0);
+    bug_assert(ret == 0);
     self->rng = rng;
     self->discrete_genome = true;
 
@@ -822,7 +821,7 @@ msp_free_pedigree(msp_t *self)
 
     ind = self->pedigree->inds;
     if (ind != NULL) {
-        assert(self->pedigree->num_inds > 0);
+        bug_assert(self->pedigree->num_inds > 0);
         for (i = 0; i < self->pedigree->num_inds; i++) {
             msp_safe_free(ind->parents);
             msp_safe_free(ind->segments);
@@ -941,8 +940,8 @@ msp_get_segment(msp_t *self, size_t id, label_id_t label)
 {
     segment_t *u = object_heap_get_object(&self->segment_heap[label], id - 1);
 
-    assert(u != NULL);
-    assert(u->id == id);
+    bug_assert(u != NULL);
+    bug_assert(u->id == id);
     return u;
 }
 
@@ -970,7 +969,7 @@ msp_insert_individual(msp_t *self, segment_t *u)
     int ret = 0;
     avl_node_t *node;
 
-    assert(u != NULL);
+    bug_assert(u != NULL);
     node = msp_alloc_avl_node(self);
     if (node == NULL) {
         ret = MSP_ERR_NO_MEMORY;
@@ -978,7 +977,7 @@ msp_insert_individual(msp_t *self, segment_t *u)
     }
     avl_init_node(node, u);
     node = avl_insert_node(msp_get_segment_population(self, u), node);
-    assert(node != NULL);
+    bug_assert(node != NULL);
 out:
     return ret;
 }
@@ -989,9 +988,9 @@ msp_remove_individual(msp_t *self, segment_t *u)
     avl_node_t *node;
     avl_tree_t *pop = msp_get_segment_population(self, u);
 
-    assert(u != NULL);
+    bug_assert(u != NULL);
     node = avl_search(pop, u);
-    assert(node != NULL);
+    bug_assert(node != NULL);
     avl_unlink_node(pop, node);
     msp_free_avl_node(self, node);
 }
@@ -1033,7 +1032,7 @@ msp_insert_breakpoint(msp_t *self, double left)
     m->value = 0;
     avl_init_node(node, m);
     node = avl_insert_node(&self->breakpoints, node);
-    assert(node != NULL);
+    bug_assert(node != NULL);
 out:
     return ret;
 }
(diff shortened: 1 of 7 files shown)


grahamgower:
You might consider calling this `BUG_ON()`, to match the name used for this in the Linux kernel. I find the convention of using upper case for preprocessor macros is useful too.

lib/util.h
```diff
@@ -37,6 +37,16 @@
...
+#define bug_assert(condition)                                                           \
+    if (!(condition)) {                                                                 \
+        fprintf(stderr,                                                                 \
+            "Bug detected in %s at line %d. Please open an issue on GitHub, "           \
+            "ideally with a reproducible example. "                                     \
+            "https://github.com/tskit-dev/msprime/issues\n",                            \
+            __FILE__, __LINE__);                                                        \
+        abort();                                                                        \
+    }
+
```
grahamgower, line 49: There's a convention for multiline c macros that uses do-while loop. This gives you an inner brace scope, but also catches if a user calls the macro without a semicolon.
```
#define my_thing(x) do { \
  // ... do something with x \
} while(0)
```

jeromekelleher: Ah, I was wondering if I needed that when I had an ``if`` in there. Thanks, I'll stick in the ``do`` loop.

jeromekelleher:
> You might consider calling this BUG_ON(), to match the name used for this in the Linux kernel. I find the convention of using upper case for preprocessor macros is useful too.

I was thinking about exactly this, but using ``BUG_ON`` would involve negating all the current conditions, which is a bit of a pain.

benjeffery:
@jeromekelleher I've worked this up for tskit at https://github.com/tskit-dev/tskit/pull/860 not sure if there are some that we should keep as plain `assert` - I guess the expensive ones if any?

castedo:
>  it tells the user what do do

@jeromekelleher good idea, nice touch. Totally extreme idea: makes me wonder whether there is some reliable web service out there such that the line number and the filename (or hash of) can be put into a URL and the user can just click on a URL and this bug event gets recorded without the user having to do more than click/copy base the URL into a brower. Probably not worth the effort, but that would be super slick.

castedo:
@grahamgower, thanks for the mention of `BUG_ON()`. New to me and stackoverflow has some interesting discussion about it.

I'm used to uppercase macros too. Here's a possible reason to make macros uppercase vs lowercase: if the "call" to the macro is not behaving like a function call, then is should be uppercase. This gives a reminder to the reader that something special is going on.
               
For something really fancy like a for loop helper, the choice is clear. In the case of `assert()` it seems like it's kind of in a
grey zone. One could make a reasonable case that it does more or less behave like a regular function call, ignoring that the macro is able to get a string of the code, line and function name.
    
So net-net, matching the case of `assert()` doesn't seem too bad. Although I think I'd go for uppercase TBH. Not that it matters much. :)

jeromekelleher:
Anyone think it's worth negating all the current conditions and doing ``TSK_BUG_ON``? It feels like the right thing to do, to me, but it's a fair bit of work.

benjeffery:
> Anyone think it's worth negating all the current conditions and doing `TSK_BUG_ON`? It feels like the right thing to do, to me, but it's a fair bit of work.

The ugly but quick way is to search and replace adding brackets and an `!`. But I think asserting what you expect feels more natural than what you don't. I wasn't aware of `BUG_ON`, interesting!

castedo:
The distinction between `bug_assert` and `assert` is not very suggestive of what's diff between them. Unless the negative/positive condition lines up with `BUG_ON`, it doesn't seem worth much to line up with it.

Here's some brainstorming on alternative names to `bug_assert`:
- die_if  (like Perl)
- halt_if
- abort_if
- panic_if
- ever_assert
- perm_assert

castedo:
I feel the same way as @benjeffery that asserting what you want feels more natural. But it's probably just because I'm used to using asserts a lot. But it does line up with how we think about preconditions for functions and valid calculations. And this seems to be a really common situation in tskit related code.

I can hypothesize that in the Linux kernel realm BUG_ON makes sense because usually they are weird scary situations that are detected and are so bad that it's worth a kernel panic. So the focus is more on the fact that "some really bad situation" has been detected. But the tskit realm is more in the nature of the pre-conditions of correct valid calculations have failed.

castedo:
Just throwing this out there, but perhaps it would not be that bad to actually have two complementary macros like:
* STOPIFNOT()  (consistent with R)
* BUG_ON()       (consistent with Linux)

grahamgower:
I think of `BUG_ON(x)` as being analogous to `if (x) { raise error }`. I think developers quickly adapt to either BUG_ON-style catching the case you think is in error, or assert-style statement of the case you think is correct. If its too much work to change the polarity of existing asserts, then keep it as is. The documentation describes the behaviour anyway, so there's not much cognitive overhead here. But definitely, I believe there should be only one such macro/function.

jeromekelleher:
I just had a go at changing over the current positive asserts here to ``TSK_BUG_ON`` and I found it a bit less natural to state the things that go wrong rather than stating the conditions I expect to be true. So, I think we should keep the asserts in the positive direction.

I agree with @grahamgower though, and there should only be one macro for this, and having two would just be overboard and confusing.

I quite like ``tsk_bug_assert`` as a name, but I guess I've been using asserts wrongly for years, so maybe my opinion should be discounted here.

jeromekelleher:
Just had a chat with @benjeffery there, and we think ``tsk_bug_assert`` is a good way to go. It's pretty clear that it's a custom thing that a bit like assert, and devs will get used to using it pretty quickly.

molpopgen:
> Just had a chat with @benjeffery there, and we think ``tsk_bug_assert`` is a good way to go. It's pretty clear that it's a custom thing that a bit like assert, and devs will get used to using it pretty quickly.
> 
> So, the plan is to finish up  tskit-dev/tskit#860, push out a tskit C API release and then update this PR to use the new tsk_bug_assert macro appropriately.

I was gonna ng to suggest a simple `tsk_assert` that behaves just like `assert` but cannot be compiler out.  Seems like you beat me to it!

castedo:
> I just had a go at changing over the current positive asserts here to `TSK_BUG_ON` and I found it a bit less natural to state the things that go wrong rather than stating the conditions I expect to be true. So, I think we should keep the asserts in the positive direction.

Another benefit to asserts in the positive, especially with scientific software, is that making floating point comparisons is more likely to result in what is intended. For instance:
```
tsk_bug_assert(x > 0)
```
will trigger abort when x is NaN which is likely the intention. In contrast
```
BUG_ON(x <= 0)
```
will NOT trigger when x is NaN ... the correct equivalent would have to be written
```
BUG_ON(x <= 0 || isnan(x))
```
Any comparison with NaN is false, even `x == x` is false when x is NaN.

benjeffery:
Great point @castedo! I hadn't thought of that one.

jeromekelleher:
OK, I think this is ready to go now.

lib/msprime.c
```diff
@@ -4363,7 +4356,7 @@ msp_pedigree_climb(msp_t *self)
                 if (u[j] == NULL) {
                     continue;
                 }
-                /* assert(u[j]->prev == NULL); */
+                /* tsk_bug_assert(u[j]->prev == NULL); */
```
benjeffery, line 4359: This looks like a case for a plain `assert`?

lib/mutgen.c
```diff
@@ -840,7 +839,7 @@ mutgen_init_allocator(mutgen_t *self)
     /* Need to make sure we have enough space to store sites and mutations. We
      * allocate ancestral and derived states, as well as a list of mutations
      * for each site. This ensures that we can always allocate the required amount.
-     * We need to add one because the assert trips when the block size is equal
+     * We need to add one because the tsk_bug_assert trips when the block size is equal
```
benjeffery, line 842: I'm not sure this is referring to `tsk_bug_assert`. I couldn't find which assertion it is referring to as removing the "plus ones" below had no adverse effect on the tests.

benjeffery (approved):

benjeffery:
@jeromekelleher LGTM - just two minor quibbles.

lib/msprime.c
jeromekelleher: Good call.

lib/mutgen.c
jeromekelleher: Ah yes, you're right. This was working around a bug in an older version of tsk_blkalloc which has been fixed. Good spot.

PR #1205, opened 2020-09-24
Refactor RateMap to use numpy functions

### LINKED ISSUE #1172: Refactor RateMap to use numpy functions and add functionality
The current [intervals.py](https://github.com/tskit-dev/msprime/blob/master/msprime/intervals.py#L29) is a minimal refactoring required to add a more sensible interface and deprecate the old RecombinationMap.  We should at least refactor the current code to use numpy APIs rather than (e.g.) Python bisect. Fleshing out some additional simple functionality would be good too (but optional).

@grahamgower, you've been involved in the recombination map code before - this is a clean slate now, any thoughts?

grahamgower:
This pr, plus #1185, should close #1172.

### DIFF
--- a/msprime/intervals.py
+++ b/msprime/intervals.py
@@ -19,8 +19,6 @@
 """
 Utilities for working with intervals and interval maps.
 """
-import bisect
-import gzip
 import warnings
 
 import numpy as np
@@ -37,11 +35,12 @@ class RateMap:
     genome.
     """
 
-    def __init__(self, position, rate):
+    def __init__(self, position, rate, map_start=0):
         # We take copies to make sure there are no unintended consequences
         # later when a user modifies the arrays in this map.
-        self.position = np.array(position, copy=True)
-        self.rate = np.array(rate, copy=True)
+        self.position = np.array(position, dtype=float, copy=True)
+        self.rate = np.array(rate, dtype=float, copy=True)
+        self.map_start = map_start
         size = len(self.rate)
         if size < 1:
             raise ValueError("Must have at least two positions")
@@ -73,6 +72,17 @@ class RateMap:
     def size(self):
         return self.rate.shape[0]
 
+    @property
+    def mean_rate(self):
+        """
+        Return the weighted mean across all windows of the entire map.
+        """
+        window_sizes = self.position[1:] - self.position[:-1]
+        weights = window_sizes / self.sequence_length
+        if self.map_start != 0:
+            weights[0] = 0
+        return np.average(self.rate, weights=weights)
+
     def slice(self, start=None, end=None, trim=False):  # noqa: A003
         """
         Returns a subset of this rate map between the specified end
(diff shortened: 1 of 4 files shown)


msprime/intervals.py
```diff
@@ -347,3 +310,65 @@ def get_length(self):
...
+        chr1	82571	        2.082414	0.080572
+        chr1	88169	        2.081358	0.092229
+        chr1	254996	        3.354927	0.439456
+        chr1	564598	        2.887498	1.478148
+        ...
+        chr1	182973428	2.512769	122.832331
+        chr1	183630013	0.000000	124.482178
+
+    :param str filename: The name of the file to be parsed. This may be
+        in plain text or gzipped plain text.
```
jeromekelleher, line 345: Need a return docstring here.

msprime/intervals.py
```diff
@@ -347,3 +310,65 @@ def get_length(self):
...
+    """
+    # fmt: on
+    positions = []
```
jeromekelleher, line 348: I was hoping we could use numpy.loadtxt here, and make this a little less crappy.

msprime/intervals.py
```diff
@@ -347,3 +310,65 @@ def get_length(self):
...
+            # Rate is expressed in centimorgans per megabase, which
+            # we convert to per-base rates
+            rates.append(rate * 1e-8)
+        if rate != 0:
+            raise ValueError(
+                "The last rate provided in the recombination map must be zero"
+            )
+    finally:
+        f.close()
+    return RecombinationMap(positions, rates, map_start=map_start)
```
jeromekelleher, line 374: We should return a RateMap here, rather than a RecombationMap (which is going for full deprecation)

msprime/intervals.py
```diff
@@ -81,60 +80,66 @@ def slice(self, start=None, end=None, trim=False):  # noqa: A003
...
+                if rate[-1] == 0:
+                    position[-1] = self.position[-1]
                 else:
-                    new_positions.insert(0, 0)
-                    new_rates.insert(0, 0.0)
-            if new_positions[-1] != positions[-1]:
-                new_positions.append(positions[-1])
-                new_rates.append(0)
-        return self.__class__(new_positions, new_rates[:-1])
+                    position.append(self.position[-1])
```
jeromekelleher, line 134: Can use np.append here.

msprime/intervals.py
```diff
@@ -81,60 +80,66 @@ def slice(self, start=None, end=None, trim=False):  # noqa: A003
...
+            # The simplest way to do this is to convert the position and
+            # rate arrays into regular lists, then prepend and/or append.
+            position = list(position)
+            rate = list(rate)
+            # Prepend or extend zero-rate region at start of map.
+            if position[0] != 0:
+                if rate[0] == 0:
+                    position[0] = 0
+                else:
+                    position.insert(0, 0)
```
jeromekelleher, line 127: Can use np.insert here.

msprime/intervals.py
```diff
@@ -81,60 +80,66 @@ def slice(self, start=None, end=None, trim=False):  # noqa: A003
...
         if trim:
-            new_positions = [pos - start for pos in new_positions]
+            position -= start
         else:
-            if new_positions[0] != 0:
-                if new_rates[0] == 0:
-                    new_positions[0] = 0
+            # For the trim=False case we need to zero out all rates outside the
+            # interval [start, end], but retain the sequence_length.
+            # The simplest way to do this is to convert the position and
```
jeromekelleher, line 118: I find it hard to believe we can't easily replicate with numpy operations. Isn't it just a straight replace of insert/append for numpy equivalents?

msprime/intervals.py
```diff
@@ -214,66 +219,24 @@ def uniform_map(cls, length, rate, num_loci=None):
...
-                )
-        finally:
-            f.close()
-        return cls(positions, rates, map_start=map_start)
+        warnings.warn(
+            "RecombinationMap.read_hapmap() is deprecated. "
+            "Use msprime.read_hapmap() instead.",
+            DeprecationWarning,
+        )
+        return read_hapmap(filename)
```
jeromekelleher, line 239: So, this would need to ``return RecombinationMap(rate_map.position, list(rate_map.rate) + [0])`` so we've got the right types and don't break old code.

jeromekelleher (commented):
Looks good, thanks @grahamgower. Some comments.

msprime/intervals.py
grahamgower: Ah, yes. Somehow I forgot about that.

msprime/intervals.py
grahamgower: There's the small issue of the `map_start` parameter. How do you want to handle that? Add that to the rate map?

msprime/intervals.py
grahamgower: Yes, that will work. But numpy's append and insert functions copy the arrays, and we're already doing so many copies around this code...

msprime/intervals.py
jeromekelleher: Ah, yes. Yes, I guess we should add it to the RateMap all right.

msprime/intervals.py
jeromekelleher: :shrug: it's a corner case and array copies are surprisingly fast.

grahamgower:
Ok, this is ready for another look @jeromekelleher.

grahamgower:
The test failure is a bit strange. Seems unrelated to the changes here though, as the failing test isn't using a recombination map.

jeromekelleher (approved):
Looks great, thanks @grahamgower!

jeromekelleher:
The test failure is weird all right. Let's see if it shows up in other PRs.

jeromekelleher:
Huh, weird. Travis issue didn't show up in #1210

grahamgower:
I squashed/rebased, so we can see if it happens again.

jeromekelleher:
Dunno what happened there. Probably some weird thing because we didn't specify a random seed somewhere. Not worth worrying about.

PR #1223, opened 2020-09-28
Add and use strict upper bound binary search function

castedo:
This is a stepping stone towards #1197 and #1159. Some more explanation is in #1197.

Add and use strict upper bound binary search function
      - add function for strict upper bound binary search
      - use strict upper bound search in rate_map_mass_to_position
      - streamline rate_map_mass_to_position to run a little faster
      - rename binary search functions to be consistent
      - expand binary search unit tests to include new variant
      - add extreme cases to unit test for rate_map_mass_to_position

I'm not sure about the new function names. They make sense to me, but maybe making then consistent with the C++ algo library makes more sense. I actually find the C++ algo library names kind of confusing. So I'm trying to do better than a standard (which is always suspect :)). This relates to #1195.

The `ptr_...` inline functions get used in #1197 and make that code easier to understand I think.

### DIFF
--- a/lib/rate_map.c
+++ b/lib/rate_map.c
@@ -58,7 +58,7 @@ rate_map_alloc(rate_map_t *self, size_t size, double *position, double *rate)
         ret = MSP_ERR_INTERVAL_MAP_START_NON_ZERO;
         goto out;
     }
-    self->rate = malloc(size * sizeof(*self->rate));
+    self->rate = malloc((size + 1) * sizeof(*self->rate));
     self->position = malloc((size + 1) * sizeof(*self->position));
     self->cumulative_mass = malloc((size + 1) * sizeof(*self->cumulative_mass));
     if (self->position == NULL || self->rate == NULL || self->cumulative_mass == NULL) {
@@ -89,6 +89,10 @@ rate_map_alloc(rate_map_t *self, size_t size, double *position, double *rate)
             sum += (position[j + 1] - position[j]) * rate[j];
         }
     }
+    /* Consider the rate after the max position as zero.
+       This trailing zero simplifies calcs like rate_map_position_to_mass by
+       gracefully handling positions past the max position */
+    self->rate[size] = 0.0;
 out:
     return ret;
 }
@@ -153,25 +157,22 @@ rate_map_position_to_mass(rate_map_t *self, double pos)
 {
     const double *position = self->position;
     const double *rate = self->rate;
-    double offset;
+    double base, offset;
     size_t index;
 
-    assert(pos >= 0.0);
-    if (pos <= position[0]) {
+    if (pos <= 0.0) {
         return 0;
     }
     assert(pos <= position[self->size]);
-    if (pos >= position[self->size]) {
-        return self->cumulative_mass[self->size];
-    }
-    /* search for upper bounds strictly before the max position
-       any position greather than or equal to max position returns self->size */
-    index = msp_binary_interval_search(pos, position, self->size);
+    index = idx_1st_strict_upper_bound(position, self->size + 1, pos);
     assert(index > 0);
     index--;
+    base = self->cumulative_mass[index];
     offset = pos - position[index];
-
-    return self->cumulative_mass[index] + offset * rate[index];
+    if (offset <= 0) {
+        return base;
+    }
+    return base + offset * rate[index];
 }
 
 /* Finds the physical coordinate such that the sequence up to (but not
@@ -191,7 +192,7 @@ rate_map_mass_to_position(rate_map_t *self, double mass)
     }
     /* search for upper bounds strictly before the final cum mass
        any mass greather than or equal to final cum mass returns self->size */
-    index = msp_binary_interval_search(mass, self->cumulative_mass, self->size);
+    index = idx_1st_upper_bound(self->cumulative_mass, self->size, mass);
     assert(index > 0);
     index--;
     mass_in_interval = mass - self->cumulative_mass[index];
--- a/lib/tests/test_rate_map.c
+++ b/lib/tests/test_rate_map.c
@@ -177,6 +177,12 @@ test_translate_position_and_recomb_mass(void)
     /* inside recombination interval */
     CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, 8), 18);
 
+    /* max valid position */
+    CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, 20), 25);
+
+    /* fishy input */
+    CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, -DBL_EPSILON), 0);
+
     rate_map_free(&map);
 }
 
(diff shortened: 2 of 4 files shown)


lib/rate_map.c
```diff
@@ -153,25 +157,24 @@ rate_map_position_to_mass(rate_map_t *self, double pos)
 {
     const double *position = self->position;
     const double *rate = self->rate;
-    double offset;
+    double base, offset;
     size_t index;
 
     assert(pos >= 0.0);
-    if (pos <= position[0]) {
+    if (pos <= 0.0) {
```
jeromekelleher, line 164: Is there much value in the assert here now if we're explicitly handling the < 0 case?

lib/rate_map.c
```diff
@@ -153,25 +157,24 @@ rate_map_position_to_mass(rate_map_t *self, double pos)
...
-    index = msp_binary_interval_search(pos, position, self->size);
+    index = idx_1st_strict_upper_bound(position, self->size + 1, pos);
     assert(index > 0);
     index--;
+    base = self->cumulative_mass[index];
     offset = pos - position[index];
-
-    return self->cumulative_mass[index] + offset * rate[index];
+    assert(offset >= 0);
+    if (offset <= 0) {
```
jeromekelleher, line 174: I don't see how offset could be < 0 here - aren't we checking that all rates and masses are >= 0 on the way in?

lib/tests/test_rate_map.c
```diff
@@ -177,6 +177,18 @@ test_translate_position_and_recomb_mass(void)
     /* inside recombination interval */
     CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, 8), 18);
 
+    /* max valid position */
+    CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, 20), 25);
+
+    /* NaNnoying cases */
+#ifdef NDEBUG // test via `meson -Db_ndebug=true ndebug_build`
+    CU_ASSERT(isnan(rate_map_position_to_mass(&map, NAN)));
```
jeromekelleher, line 185: Would it be simpler to test the ``isfinite(value)`` and check the bounds for the parameter? Seems like an equivalent amount of computation/branching, and it would simplify the logic a bit.

lib/util.c
```diff
@@ -344,6 +345,45 @@ msp_binary_interval_search(double query, const double *values, size_t n_values)
...
 }
 
+/* This function follows standard semantics of:
+ *   std::upper_bound() from the standard C++ <algorithm> library
+ *   and numpy.searchsorted(..., side='right') (except numpy NaN query differs)
+ * PRE-CONDITION:
+ *   1) `values` are sorted and not NaN
+ * RETURNS:
+ *   First (leftmost) `index` of strict upper bounds of `query`
+ *   **or** `n_values` if all `values` are lower bounds of `query` or `query` is NaN
```
jeromekelleher, line 355: I don't think we should both worrying about NaN semantics here - we can just rule it out on input. We have a pretty tightly circumscribed application here, and I'm sure we'll never need to deal with NaNs.

jeromekelleher (commented):
LGTM @castedo. The function names are at least clear about what they do, so I don't have any better suggestions. I'd rather just rule out non-finite values entirely as input though, rather than complicate the code in dealing gracefully with them.

lib/util.c
castedo: For my next commit I've simplified the function comments to not talk about NaN, except for one caveat at the end that the function does not match the NaN logic of numpy but rather matches std::upper_bound.
Unfortunately, simplifying `!(elements[mid] > query)` to `elements[mid] <= query` means the function will match the semantics of **neither** nump.searchsorted nor std::upper_bound, which seems kind of a pity.

The std::upper_bound/lower_bound semantics for NaN make a ton of sense to me. Granted I probably think about NaN an abnormal amount. :)

Don't write off NaN so fast! It's an elegant and powerful error handling option in the absence of exception handling and error codes. I'm not surprised it exists as the result of `INFINITY * 0` and 'x / 0'. In the future you might find NaN as a useful way to gracefully handling very tricky numerical challenges.

lib/tests/test_rate_map.c
castedo: In the implementation of rate_map_position_to_mass in this commit there isn't any special handling for NaN as input. It's as simply as it would be ignoring the possibility of NaN.

One of the properties of NaN that I find rather elegant and almost beautiful is how often NaN gracefully propagates a state of erroneous computation without any special logic or handling. One of the ways I like to think of NaN is as "Can Not Be Computed/Determined". So if a "can't-be-determined" is passed in then it makes sense that "can't-be-determined" is passed right back out.
But that wasn't really the motivation for the current impl of rate_map_position_to_mass in this commit. The test is just there to make sure something like crashing doesn't happen in the absence of NDEBUG disabled asserts.

lib/rate_map.c
castedo: I'm getting a little ahead of myself with #1203 in the pipeline but not yet merged. I'll take it out to make merging #1203 easier. I also don't see anyway offset >=0 isn't true in the current impl but that's with some assumptions about the other code and data structures (present and future). Once #1203 is merged there's another discussion which is when is being 'assert happy' too enthusiastically ... 'assertive' ha ha :)

lib/rate_map.c
castedo: With #1203 merged I would say yes it is of value. But I might be too 'assert happy' and 'assertive' :rofl: .
I'll just take it out pre #1203 to make merging #1203 easier.

The idea with a traditional assert.h assert() here is that if this condition is getting violated, even by the slightest rounding error, it's worth having debug builds crash to investigate, even though a release build will handle it in a reasonable and well defined way.

castedo: Crap, I just deleted my previous comment. Anyway, I was basically saying the same thing as the other assert. I'll remove it for this pre #1203 code.

castedo:
> I'd rather just rule out non-finite values entirely as input though, rather than complicate the code in dealing gracefully with them.

Apart from the `!(x > y)` vs `x <= y` in the `std::upper_bound` equivalent function, I'd say there isn't any special handling for non-finite values. Whether it's more complicated to use standard IEEE doubles vs turn on gcc `-ffinite-math-only` that might be an interesting question to consider. It almost sounds like you'd prefer to turn on `-ffinite-math-only` and not use standard IEEE doubles.

I've never used `-ffinite-math-only` and never really considered it. Using NaN and infinity seem like great tools to use. My hunch is non-finite IEEE values got standardized for a good reason. That code become LESS complicated and more robust in the end by embracing them rather than avoiding them. But that's really speculation and hunch on my part.

castedo:
OK, I've made tweaks from the review comments and squashed. This sucker is good to go I believe.

The code coverage complaint I think is worth ignoring. It's mainly happening because my new perf optimized code isn't calling one of the trivial helper inline functions.

lib/tests/test_rate_map.c
```diff
@@ -177,6 +177,18 @@ test_translate_position_and_recomb_mass(void)
     /* inside recombination interval */
     CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, 8), 18);
 
+    /* max valid position */
+    CU_ASSERT_EQUAL(rate_map_position_to_mass(&map, 20), 25);
+
+    /* NaNnoying cases */
+#ifdef NDEBUG // test via `meson -Db_ndebug=true ndebug_build`
```
jeromekelleher, line 184: I don't want to have conditional compiles in the test suite, everything is already quite complicated and this seems gratuitous.

lib/util.h
```diff
@@ -121,7 +122,24 @@ void __msp_safe_free(void **ptr);
 
 #define msp_safe_free(pointer) __msp_safe_free((void **) &(pointer))
 
-size_t msp_binary_interval_search(double query, const double *values, size_t n_values);
+size_t idx_1st_upper_bound(const double *values, size_t n_values, double query);
+size_t idx_1st_strict_upper_bound(
+    const double *elements, size_t n_elements, double query);
+
+inline const double *
```
jeromekelleher, line 129: Looks like these aren't being used at all in this PR - easiest to remove for now?

lib/util.c
```diff
@@ -369,7 +408,7 @@ doubles_almost_equal(double a, double b, double eps)
  *
  */
 static inline size_t
-positive_interval_select(double x, size_t num_intervals, double const *restrict lengths)
+positive_interval_select(double x, size_t num_intervals, const double *lengths)
```
jeromekelleher, line 411: Why remove the ``restrict`` here?

jeromekelleher (commented):
A few loose ends to tidy up I think.

jeromekelleher:
> Apart from the !(x > y) vs x <= y in the std::upper_bound equivalent function, I'd say there isn't any special handling for non-finite values. Whether it's more complicated to use standard IEEE doubles vs turn on gcc -ffinite-math-only that might be an interesting question to consider. It almost sounds like you'd prefer to turn on -ffinite-math-only and not use standard IEEE doubles.

All I'm saying is that nonfinite values will never be used as input to these functions - we will explicitly check that all the values are finite at some other layer of the code. Since this is not a public API we fully control the inputs.  There's no need for any special compiler flags or anything like that, we're just using the fact that we have control over the domain of input to reduce the complexity of the code. I want the code to be as simple and readable as it can be, and having tricky conditions to catch nonfinite values that will never be valid input makes it more difficult for people to understand.

lib/util.c
castedo: I was just making the parameter more readable and simple. As I was debating whether to put this `restrict` in other parameters of new function, I started learning more this keyword that is new to me. https://en.wikipedia.org/wiki/Restrict is one of the examples I was looking at to understand better.

In this particular case, given the function does not change the values of any memory locations and is always reading new memory locations. There is nothing a compiler could do to compile more optimal code given the `restrict` keyword.

I disassembled the function to confirm `restrict` has no effect (at least on x86_64). And looking at the compiled instructions and my current understanding I don't see how `restrict` could have any effect.

It's a really interesting keyword, especially since it's in C and not C++. It does seem like the same desired instructions can be generated with modern compilers by declaring and using temporary variables. I'll speculate this is part of the reason the keyword hasn't been added to C++.

lib/util.h
castedo: Well, not *easiest* ... easiest would be leave 'em in. :) `pr/fast_search_lookup` is calling `ptr_1st_strict_upper_bound`.

But this does relate to a slightly annoying issue that is in these lines. No code, not even `pr/fast_search_lookup`, is currently calling `ptr_1st_upper_bound`. And I didn't bother writing any unit test to call it.

So I've been wonder whether I should:
A) leave it in and not worry about it,
B) add a unit test case to call `ptr_1st_upper_bound` for code coverage or,
C) remove `ptr_1st_upper_bound`

I've been proceeding along A) but let me know if you prefer B) or C). And by C) I mean remove not just for this PR #1223 but all PRs, forever, until some "real" code is actually going to call it.

jeromekelleher: Well as a general rule I think we'd want tests for any code that's introduced in a PR. So, seems easiest to delete these functions for this PR?

lib/tests/test_rate_map.c
castedo: Good instinct to discourage #ifdef'd code sprinkled inside code. I imagine in open-source that can really encourage excessive #ifdef use. And with git, it seems like lots of traditional uses of #ifdef in development are better handled with git branches.

I'll replace all these NaNnoying cases and with just that one line testing for finite -1 without ifdef.

This relates to a deeper question which I think is not really resolved yet which is what are the desirable uses of NDEBUG conditioned assert.h assert() calls in msprime/tskit. Not sure what's the best forum or time to dig deeper into this question.

lib/util.h
castedo: I'll go ahead and move them from this PR and into `pr/fast_search_lookup`. Although the not worrying about it option would be the easiest option.

jeromekelleher:
OK, great! Just needs a squash and we're good to merge.

jeromekelleher (approved):

Issue #1216, opened 2020-09-26
multiple merger documentation and truncation point


eldonb:
In the documentation for  multiple-merger coalescents,  I would  state that  time is measured in units of  N^{alpha - 1} generations, rather than  Ne^{alpha - 1}, where N is the census  population size ( unless your notation for  census size  is  Ne ?).   The standard notation for  "effective size" is  Ne, which in this case would be (proportional to)  N^{alpha - 1}.   

The truncation point in the    incomplete  Beta-coalescent  will depend on the actual  underlying population model,  but assuming a simple version  can be taken as   K/( K + m(alpha)), where  m(alpha)  =   1 +  (2^{1 - alpha}) / (alpha - 1) in the haploid case (for 1 < alpha < 2), where K > 0  is the  proportionality constant of the cut-off (each individual can produce at most  KN  'juveniles' in the population model).    The point is  as alpha -> 1  the truncation point decreases.   
In the diploid case (working with the Xi-Beta(2-alpha, alpha, K) coalescent)  one can take   m(alpha) =  2 + (2^{alpha}3^{1 - alpha})/(alpha - 1)  (this can vary  depending on the underlying population model but we don't need to go into that here).

jeromekelleher:
I think the confusion here is that "population size" can be specified through the parameter ``Ne`` to the ``simulate`` function. However, with the new ``sim_ancestry`` interface, we've dropped ``Ne`` and are using ``population_size`` instead.

So, I think your idea is a good one @eldonb. @JereKoskela, thoughts?

JereKoskela:
I don't have strong views on N vs Ne. I tend to think of N as the census size, and Ne as the population size which one needs to plug in to match the observed diversity, but that's far from well-established. Do we need notation for population sizes and coalescence probabilities elsewhere in the documentation? I think the most important thing is to be consistent.

The current implementation of the truncation point is a little bit more naive than what @eldonb described: basically, the user inputs the compound parameter truncation_point = K/(K + m(alpha)) directly. The disadvantage is that it's less explicit, but the advantage is that the standard, untruncated Beta-coalescent can be represented exactly as truncation_point = 1. If we had the user specify K instead, then we would need something like `K = inf` as a default parameter to get K/(K + m(alpha)) = 1. @jeromekelleher, do you see any pitfalls with something like that?

[...]

PR #1224, opened 2020-09-28
Multiple merger

JereKoskela:
- Replaced references to Ne with N in the multiple merger documentation.
- Changed `truncation_point` for the Beta-coalescent as discussed in #1216. In particular, the range is now (0, inf] and the default is inf, which gives the same behaviour as the previous default. Some tests had to change too because they were testing for the validity of the previous (0,1] range, and were now expecting errors from valid parameters, or vice versa.
- Fixed a minor error in the mean juvenile count `m` under the Beta-coalescent, and simplified some computations by extracting out two new static functions to perform some brief computations.

Ping @eldonb - I can't request you as a reviewer for some reason.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -6355,22 +6355,49 @@ out:
  **************************************************************/
 
 static double
+beta_compute_juvenile_mean(msp_t *self)
+{
+    double alpha = self->model.params.beta_coalescent.alpha;
+    double m;
+    if (self->ploidy > 1) {
+        m = 2 + exp(alpha * log(2) + (1 - alpha) * log(3) - log(alpha - 1));
+    } else {
+        m = 1 + exp((1 - alpha) * log(2) - log(alpha - 1));
+    }
+    return m;
+}
+
+static double
+beta_compute_truncation(msp_t *self)
+{
+    double truncation_point = self->model.params.beta_coalescent.truncation_point;
+    double m;
+    if (truncation_point < DBL_MAX) {
+        m = beta_compute_juvenile_mean(self);
+        truncation_point /= (truncation_point + m);
+    } else {
+        truncation_point = 1.0;
+    }
+    return truncation_point;
+}
+
+static double
 beta_compute_timescale(msp_t *self, population_t *pop)
 {
     double alpha = self->model.params.beta_coalescent.alpha;
-    double truncation_point = self->model.params.beta_coalescent.truncation_point;
-    double m = 1 + exp(alpha * log(2) + (1 - alpha) * log(3) - log(alpha - 1));
+    double truncation_point = beta_compute_truncation(self);
+    double m = beta_compute_juvenile_mean(self);
     double pop_size = pop->initial_size;
+    double timescale;
     /* For ploidy > 1 we assume N/2 two-parent families, so that the rate
      * with which 2 lineages belong to a common family is based on "population size"
      * N/2 */
     if (self->ploidy > 1) {
         pop_size /= 2.0;
-        m += 1.0;
     }
-    double timescale = exp(alpha * log(m) + (alpha - 1) * log(pop_size) - log(alpha)
-                           - gsl_sf_lnbeta(2 - alpha, alpha)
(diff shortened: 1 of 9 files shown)


JereKoskela:
The lint / pre-commit complaint is interesting: there is a `numpy.inf` hidden in `obj_copy = eval(repr(obj), globals(), msprime.__dict__)` on line 1785 of tests/test_ancestry.py when `model = msprime.BetaCoalescent` with `truncation_point = inf`. If I don't tell it what `numpy.inf` is, then the test fails, but if I do, then pre-commit complains. Any suggestions?

lib/msprime.c
```diff
@@ -6355,22 +6355,49 @@ msp_dirac_common_ancestor_event(msp_t *self, population_id_t pop_id, label_id_t
...
+    return m;
+}
+
+static double
+beta_compute_truncation(msp_t *self)
+{
     double truncation_point = self->model.params.beta_coalescent.truncation_point;
-    double m = 1 + exp(alpha * log(2) + (1 - alpha) * log(3) - log(alpha - 1));
+    double m;
+    if (truncation_point < DBL_MAX) {
```
jeromekelleher, line 6375: If we set the input value to ``DBL_MAX`` then I think this condition isn't necessary. I think (but we'd need to verify, just t to be sure) that ``BDL_MAX / (DBL_MAX  + m)`` is 1 for any reasonable value of m.

The way to specify DBL_MAX in python is via ``sys.float_info.max``.

TPPSellinger:
I don't think I can help you with the  numpy.inf but I will check your modifications  in the c code and all the tests in verifications.py.

lib/msprime.c
```diff
@@ -6819,7 +6846,7 @@ msp_set_simulation_model_beta(msp_t *self, double alpha, double truncation_point
         goto out;
     }
 
-    if (truncation_point <= 0.0 || truncation_point > 1.0) {
+    if (truncation_point <= 0.0) {
```
jeromekelleher, line 6849: I think we should put a ``isfinite(truncation_point)`` check here as well. Dealing with floating point special classes is a PITA, it's much easier if we can use DBL_MAX all the way through.

jeromekelleher (commented):
LGTM @JereKoskela. I think we should change from ``inf`` to DBL_MAX though, and hopefully this gets rid of the special case in the computation.

lib/msprime.c
JereKoskela: Is there much harm in keeping it as is? It costs a few extra lines of code and an if-statement, but the value of `m` diverges as alpha approaches 1, so in principle `DBL_MAX / (DBL_MAX + m) == 1` should fail for sufficiently small but otherwise valid values of alpha.

jeromekelleher: Yeah, OK, it's probably safer to have the special case and I guess it all works either way.

lib/msprime.c
jeromekelleher: Although, I guess it's more mathematically satisfying to say ``inf`` rather than "arbitrary very big number". We'll have to sort out the annoying provenance stuff if we use ``inf``. I can put in a commit to do that if you prefer being able to use ``np.inf`` e.g. as the input value.

JereKoskela: I used the first thing I could make work. I'm still not very familiar with Python, and am completely unfamiliar with the C-Python interface. I'm happy to swap to `sys.foat_info.max` from above, and add the `isfinite` check.

jeromekelleher:
Looks good to me, thanks @JereKoskela. Good to go after a squash and we get @TPPSellinger's approval I think.

TPPSellinger:
Sorry I'm late. I went through all modifications and I didn't find anything. I also ran all beta related test in verification.py. Everything looks fine.

jeromekelleher (approved):

jeromekelleher:
Thanks very much @JereKoskela, this is starting to look very polished. 

ps. One tiny thing: If you add a line "Closes #1216" to the commit message GitHub will take care of the issue closing. I've linked the issue here manually, which will hopefully do the same thing.

PR #1226, opened 2020-09-29
Fix flakey tests by fixing RNG seed

### LINKED ISSUE #1225: test_debug_logging_dtwf randomly fails
I've noticed this a couple of random times on my local machine. Every time I've just rerun unit tests and it's passed fine. Just now I had Travis CI fail with what a recall as the same issue (below). It looks like logging is not deterministic. Perhaps this logging based unit test case should be removed? or perhaps there is a way to get it into a mode that is deterministic?

```
test_multiple_replicate_equality (tests.test_threads.TestSimulatorThreads) ... ok
test_single_replicate_equality (tests.test_threads.TestSimulatorThreads) ... ok
======================================================================
FAIL: test_debug_logging_dtwf (tests.test_ancestry.TestSimulator)
----------------------------------------------------------------------
  File "/Users/travis/build/tskit-dev/msprime/tests/test_ancestry.py", line 278, in test_debug_logging_dtwf
    self.assertGreaterEqual(len(log.output), 3)
AssertionError: 2 not greater than or equal to 3
```

benjeffery:
Fixes #1225, added a few more for good measure!

### DIFF
--- a/tests/test_ancestry.py
+++ b/tests/test_ancestry.py
@@ -219,14 +219,14 @@ class TestSimulator(unittest.TestCase):
             self.verify_simulation(n, m, r)
 
     def test_perf_parameters(self):
-        sim = ancestry._parse_simulate(10)
+        sim = ancestry._parse_simulate(10, random_seed=42)
         sim.run()
         self.assertGreater(sim.avl_node_block_size, 0)
         self.assertGreater(sim.segment_block_size, 0)
         self.assertGreater(sim.node_mapping_block_size, 0)
 
     def test_event_chunk(self):
-        sim = ancestry._parse_simulate(10)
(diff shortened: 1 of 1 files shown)


castedo (approved):
Excellent thanks!
That looks like a solution to the random test fails. Although I can't really confirm until I run the unit tests a billion more times. :)

jeromekelleher (approved):
Ooops, looks like my bad; thanks for the fix!

benjeffery:
> Excellent thanks!
> That looks like a solution to the random test fails. Although I can't really confirm until I run the unit tests a billion more times. :)

The initial failure was about 1-in-50 on my machine. I ran about 10,000 without failure after - totally overboard though cos I can't see how this fix won't work!

PR #1227, opened 2020-09-29
Refactor docs

jeromekelleher:
Here's a proposal for refactoring the documentation. Rather than the current approach of splitting things between "API docs" and a "tutorial", I think we should split by high-level subject ("simulating ancestry", "simulating mutations", "specifying demography", and other things we might put in).

Then, within each subject page we split things between "API documentation" which gives precise and concise definitions of what things are, and an "Examples" section which shows you how to use things via quick examples. The examples are given via [jupyter-sphinx](https://jupyter-sphinx.readthedocs.io/en/latest/) which looks quite handy, and should enable all sorts of things like [running examples in an external notebook environment](https://jupyter-sphinx.readthedocs.io/en/latest/#thebelab-support) and [downloading code chunks](https://jupyter-sphinx.readthedocs.io/en/latest/#downloading-the-code-as-a-script). The idea is then to link heavily to the Examples sections from the API docs sections.

This, coupled with some improvements in HTML rendering on the tskit side (e.g. https://github.com/tskit-dev/tskit/issues/872) should make things *much* easier to document.

This is what my initial example here for sim_ancestry looks like:

Screenshot from 2020-09-29 12-41-46

The big change this will need is we'd have to drop readthedocs and start building the documentation and hosting ourselves. I'm pretty happy with this, to be honest: we can easily build the docs in a GitHub action, and host in a number of different ways. It's just not worth the hassle of trying to get GSL installed on readthedocs and dealing with the quirks. 

Thoughts? Pinging @benjeffery @gtsambos @agladstein @molpopgen @petrelharp


molpopgen:
I knew you'd eventually come around to "live" examples! 😜

I don't know about the hosting, but I like the organization idea, and refactored my own along these lines recently.

I install GSL to build the fwdpy11 docs on RTD.  It has been pretty stable for a while.  You may consider looking at how I do the conda setup?

agladstein:
Off the bat, I really like this organization idea. It's more in line with how I want to learn how to use things. First, find out how the thing works and what it does, then look at the examples to actually understand how to use it. I think I've commented before, but I often find API docs hard to understand. So, having the examples right there, makes it much easier to figure out.

Do you want a review on the content of the docs in this PR or just the concepts?

What about tutorials? I still think there is a place for tutorials. When I start using a new program, I often go straight for the "quick start" and tutorials to get a feel for how it works, then I through the docs for the specific things I want to do.

petrelharp:
Sounds great! I'm with @agladstein, though - we should still keep the tutorials: common-ish use cases that might cover topics from more than one section. This is particularly important as most workflows will involve bits from all three sections.

Re: dropping rtd and hosting sounds great, and easy.  Maybe that's for a separate issue, though.

jeromekelleher:
Thanks for the input all. Yeah, I agree there's room for tutorials too, but that's probably a job for another day. No need for any detailed reviews here yet, I was just hoping to get some quick feedback on the overall outline before going any further. Looks like people are basically happy with the idea, so I'll go ahead and fill out the rest of the detail.

I'll ping back where there's something ready for review.

Re readthedocs - I honestly think it'll take less time to set up building and hosting ourselves that it would to get this working on readthedocs, even with a working example of compiling against GSL from @molpopgen!

molpopgen:
I'd drop RTD in a flash if I had a good alternative. But inertia is working against me in that regard. The setup took a fair bit of effort a couple of years ago, largely due to missing documentation, which I believe is still missing/insufficient.

benjeffery:
Loving the  `jupyter-sphinx` snippets - I'm assuming the doc build fails if the snippet code fails? I'm all for actually running the code in docs.

Happy to move off RTD, shouldn't be too bad to self-host on github pages. I start to worry locked-in we are to github, but I think it could all be repurposed else where with not-too-much work.

petrelharp:
>  I start to worry about how locked-in we are to github

Even self-hosting the docs would be super easy. And, gitlab does something similar to github pages: https://docs.gitlab.com/ee/user/project/pages/

jeromekelleher:
> I'm assuming the doc build fails if the snippet code fails? I'm all for actually running the code in docs.

Yep! It's a great feature - CI fails if your examples are broken.

jeromekelleher:
This is still a work in progress, but it could be merged if others wanted to work on it. It'll break the RTD build, but I think that's been silently broken for a while anyway, so that won't make much difference. The basic infrastructure for building the docs is in here, we'd just need to get a hosting solution into place (probably gh-pages?). So, my suggestion would be to merge this much, get the new docs site up, and then we can get more of the docs ported over to using jupyter. 

Once we have the new docs site fairly functional, we can set up a redirect from RTD (latest) so that it goes "live".

What do you think @benjeffery @petrelharp ?

benjeffery:
I agree with this approach. I will add commits to this PR to deploy to a pages site, you already have the action set up for building. This may also allow us to generate docs for a PR, so a reviewer can see them without fetching and building.

benjeffery:
This was superseded by #1238, closing.

PR #1234, opened 2020-09-30
Update benchmarks

jeromekelleher:
Couple of benchmark tweaks.

The GC one is going to fail on any older versions, I think that's OK @benjeffery?


benjeffery (approved):
Yep, that's fine! LGTM

PR #1237, opened 2020-10-04
Halve number of machine instruction running in bottleneck function

### LINKED ISSUE #1159: Add LRU cache rate_map_position_to_mass

castedo:
This is improvement on PR #1197. This new commit drops the number of instructions per call of the bottleneck function `rate_map_position_to_mass` by more than half. From 79 to 38 instructions per call. The `rate_map_position_to_mass` execution path runs an order of magnitude more than all other paths in the time_human_chr22 benchmark.

This new PR also has the lookup table use half as much memory, without any execution time trade-off.

More details to follow.

Closes #1159.

### DIFF
--- a/lib/tests/test_rate_map.c
+++ b/lib/tests/test_rate_map.c
@@ -339,6 +339,244 @@ test_binary_search_edge_cases(void)
 }
 
 static void
+verify_search(fast_search_t *fastie, const double *values, size_t n)
+{
+    int i;
+    double x;
+    size_t expect, got;
+
+    for (i = 0; i < n; i++) {
+        x = values[i];
+        expect = idx_1st_strict_upper_bound(values, n, x);
+        got = fast_search_idx_strict_upper(fastie, x);
+        CU_ASSERT_EQUAL(expect, got);
+
+        x = nextafter(x, INFINITY);
+        expect = idx_1st_strict_upper_bound(values, n, x);
+        got = fast_search_idx_strict_upper(fastie, x);
+        CU_ASSERT_EQUAL(expect, got);
+    }
+}
+
+static bool
+fast_search_verify(fast_search_t *self)
+{
+    size_t start, stop;
+    size_t idx;
+
+    if (!(self->query_multiplier >= 0.0)) { // NaN not valid
+        return false;
+    }
+    if (self->num_lookups < 2) {
+        return false;
+    }
+    for (idx = 1; idx < self->num_lookups; idx++) {
+        if (self->lookups[idx - 1] > self->lookups[idx]) {
+            return false;
+        }
+    }
+    start = self->lookups[0];
+    stop = self->lookups[self->num_lookups - 1];
+    if (stop <= start || start != 0 || self->elements[0] != 0.0) {
+        return false;
+    }
+    return true;
+}
+
+static void
+fast_search_alloc_verify(fast_search_t *self, const double *elements, size_t n_elements)
+{
+    CU_ASSERT_EQUAL_FATAL(0, fast_search_alloc(self, elements, n_elements));
+    CU_ASSERT(fast_search_verify(self));
+}
+
+static void
+test_fast_search_identity(void)
+{
+    size_t expect, got;
+    double p[] = { 0.0, 1.0, 2.0, 3.0, 4.0, 5.0 };
+    for (size_t n = 2; n <= 6; n++) {
+        fast_search_t fastie;
+        fast_search_alloc_verify(&fastie, p, n);
+        CU_ASSERT_EQUAL(1.0, fastie.query_multiplier);
+        CU_ASSERT_EQUAL(n + 1, fastie.num_lookups);
+        for (size_t i = 0; i <= n; i++) {
+            CU_ASSERT_EQUAL(i, fastie.lookups[i]);
+        }
+        verify_search(&fastie, p, n);
+        {
+            expect = idx_1st_strict_upper_bound(p, n, n);
+            got = fast_search_idx_strict_upper(&fastie, n);
+            CU_ASSERT_EQUAL(expect, got);
+
+            double x = nextafter(n, INFINITY);
+            expect = idx_1st_strict_upper_bound(p, n, x);
+            got = fast_search_idx_strict_upper(&fastie, x);
+            CU_ASSERT_EQUAL(expect, got);
+        }
+    }
+}
+
+static void
+test_fast_search_2powers(void)
+{
+    {
+        double p[] = { 0, 2 };
+        size_t n = 2;
+        fast_search_t fastie;
+        fast_search_alloc_verify(&fastie, p, n);
+        CU_ASSERT_EQUAL(0.5, fastie.query_multiplier);
+        CU_ASSERT_EQUAL(3, fastie.num_lookups);
+        CU_ASSERT_EQUAL(0, fastie.lookups[0]);
+        CU_ASSERT_EQUAL(n, fastie.lookups[n]);
+        verify_search(&fastie, p, n);
+        CU_ASSERT_EQUAL(n, fast_search_idx_strict_upper(&fastie, 2.1));
+        CU_ASSERT_EQUAL(n, fast_search_idx_strict_upper(&fastie, 4.0));
+        fast_search_free(&fastie);
+    }
+    {
+        double p[] = { 0, 0.25 };
+        size_t n = 2;
+        fast_search_t fastie;
+        fast_search_alloc_verify(&fastie, p, n);
+        CU_ASSERT_EQUAL(4.0, fastie.query_multiplier);
+        CU_ASSERT_EQUAL(3, fastie.num_lookups);
+        CU_ASSERT_EQUAL(0, fastie.lookups[0]);
+        CU_ASSERT_EQUAL(n, fastie.lookups[n]);
+        verify_search(&fastie, p, n);
+        fast_search_free(&fastie);
+    }
+    {
+        double p[] = { 0, 8 };
+        size_t n = 2;
+        fast_search_t fastie;
+        fast_search_alloc_verify(&fastie, p, n);
+        CU_ASSERT_EQUAL(0.125, fastie.query_multiplier);
+        CU_ASSERT_EQUAL(3, fastie.num_lookups);
+        CU_ASSERT_EQUAL(0, fastie.lookups[0]);
+        CU_ASSERT_EQUAL(n, fastie.lookups[n]);
+        verify_search(&fastie, p, n);
+        fast_search_free(&fastie);
+    }
+}
+
+static void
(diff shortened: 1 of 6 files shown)


castedo:
Here are the three main modifications I think worth mentioning on this PR
compared to the previous.

1) Rejection of rate maps with more than 4 billion distinct rates (or 4 Gibi, 1024^3, to be precise). Based on my
limited knowledge it seems ill-advised for any user to use a rate map with more than 4 billion distinct rates. If in theory a user did have a rate map for this many rates, the rate_map_t structure alone would be eating up 128GB of RAM. It
seems that if a user was attempting to do this, they should be taking a very different approach.  So it almost feels like this is not a limitation but a feature. :)

2) The fast_search_t code is a binary search improvement. It's not specific to rate maps or chromosome positions at all. So I moved all of it to util.h/.c to be with the other binary search functions.

3) util.h has inline functions

There appears an aversion to putting inline functions in headers. No surprise there's a trade-off between costs and benefits of doing so. The only costs I can see to inline functions in headers is the following:

A) It usually causes an increase in the size of the compiled library/executable. Usually executable size is not a problem with 21st century computers and networks. It can be, so it's worth not going totally inline crazy. For functions with only a few compiled lines this is unlikely to be much cost.

B) Some programmers are not used to or have not learned how to use inline functions.  This I suspect is a much thornier discussion, perhaps not worth completely diving into. Perhaps it's useful to frame it as whether a programming tool is "Good2Learn" or not.  Is using inline functions "Good2Learn"?  As you can guess, I think it is. It's a pretty basic C99/C++ language feature and not that hard or complicated to learn how to use.

Usage is as simple and easy as 1-2-3:

1) write the function declaration and body together, kind of like Python
2) put it in a header file with the `inline` keyword
3) in *any* C file, doesn't really matter, put `extern inline` plus the function 'declaration' without the body 

The rest is gravy. Beyond that 1-2-3 is optional extra complexity added by conventions and style. Put them only at the end of header files? Sure. Also declare the function without function body with other non-inline functions? Fine. Put `extern inline` only in certain parts of certain .c files? Sure, whatever file org you want. Never learn `static inline`? Sure, you can get by fine only learning the above 1-2-3 and never ever learning `static inline`.

In some ways it's strange that *only* `static inline` is used in msprime and `extern inline` *never* used. `extern inline` is "Good2Learn" not `static inline`.

benjeffery:
Hi @castedo, @jeromekelleher will be off for a few weeks. Is this PR ready for final review? If so I will try to get my head around it to review it. I haven't been following closely so far.

castedo:
Yep, it's ready. Thx for reviewing. It should be fine starting with this PR (vs main) instead of the previous one. The super long explanation and example in comments in this PR are updated to reflect one implementation change.

All issues in the previous PR were resolved. The most notable changes from the previous PR are the three points I mention above. There are other small changes related to perf but not nearly as notable as the 3 mentioned above.

castedo:
BTW, when does Codecov generate reports? does it eventually get around to doing it for every commit?

benjeffery:
It does it on every push that passes CI. For subsequent pushes it updates its comment rather than adding one for every push.

petrelharp:
This sounds really great, and is over my head C-wise (I will have to read about what inlining is), but I'm curious: what's the overall speedup? Have you tried out any of the benchmarks?

castedo:
> Have you tried out any of the benchmarks?

This reminds me, I should add that this closes #1159 which Jerome did a great job isolating with the Hudson.time_human_chr22 benchmark. I've been exclusively focusing on this benchmark.
https://tskit-dev.github.io/msprime-asv/#benchmarks.Hudson.time_human_chr22

> what's the overall speedup?

I'm really looking forward to seeing the official speed up %s on asv! I have a cheap approx copy of that benchmark on my computer that I have been running. It's rather noisy but this PR/branch compared to current main appears to have about a 1/3rd speed up in that benchmark.

Other code paths are now the bottleneck and not this `rate_map_position_to_mass` function. I've speed it up enough to be very "out of the way" now in terms of fraction of execution for the benchmark, despite being by far the most executed function.

One nice thing about this code (now in util.h/.c) is most of the perf improvements apply to any binary search (on re-used arrays of doubles). So hopefully this fast code can be deployed in the future to speed up other similar bottlenecks.

castedo:
I should mention inlining functions usually will not automatically make a notable perf improvement. I'm happy to write up more about inline functions ... suggestions on when to avoid (usually) or consider them (sometimes maybe), possible conventions for file organization, etc...

petrelharp:
> this PR/branch compared to current main appears to have about a 1/3rd speed up in that benchmark.

Wow, awesome!

lib/util.c
```diff
@@ -312,6 +312,58 @@ __msp_safe_free(void **ptr)
...
 
+bool
+doubles_almost_equal(double a, double b, double eps)
+{
+    return (fabs(a) < eps && fabs(b) < eps) || gsl_fcmp(a, b, eps) == 0;
+}
+
+/* Inputs:
+ *   - `lengths` is an array of the real number lengths of `num_intervals`
+ *      consecutive half-open intervals on the real number line start at zero,
```
benjeffery, line 323:
```suggestion
 *      consecutive half-open intervals on the real number line starting at zero,
```

lib/util.c
```diff
@@ -312,6 +312,58 @@ __msp_safe_free(void **ptr)
...
+ *      i.e. given lengths L_i, the intervals are:
+ *          [0.0, L_0), [L_0, L_0 + L_1), [L_0 + L_1, L_0 + L_1 + L_2), ...
```
benjeffery, line 325: Given that the return is zero if `x` is negative aren't the intervals actually
```suggestion
 *          [-INFINITY, L_0), [L_0, L_0 + L_1), [L_0 + L_1, L_0 + L_1 + L_2), ...
```
??

lib/util.c
```diff
@@ -361,69 +412,244 @@ idx_1st_upper_bound(const double *values, size_t n_values, double query)
...
 {
-    return (num_probs > 0 ? positive_interval_select(u, num_probs - 1, probs) : 0);
+    int ret = 0;
+    double max_element;
+
+    const size_t max_input_size = 1LL << (DBL_MANT_DIG - 1); // 4096 terabytes
+
+    memset(self, 0, sizeof(*self));
+
+    if (!valid_sorted_nonempty_array(elements, n_elements)) {
```
benjeffery, line 592: Maybe this has already been discussed - but should these parameter checks not be assertions? If they are driven by end-user input I can see that returning an error would be good. But if we are in control of the parameters then they should be assertions?

benjeffery (approved):
LGTM! I'm seeing `Hudson.time_human_chr22` benchmark go from 30s to 20s with this patch. 
Just a few minor quibbles.

lib/util.c
castedo: You're right. Conceptually the motivation was having intervals that have lengths equal to the inputs. And then a negative x or an x that is past all these intervals are a special case.
It probably would be more clear to just doc N+1 intervals extending all the way from -INF to +INF corresponding to all the x that can be passed in.

lib/tests/test_rate_map.c
```diff
@@ -338,6 +338,244 @@ test_binary_search_edge_cases(void)
     CU_ASSERT_EQUAL(idx, 3);
 }
 
+static void
+verify_search(fast_search_t *zoom, const double *values, size_t n)
```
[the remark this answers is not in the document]
castedo: fast/zoom/speed are too many diff words for the same idea. In hindsight I wish I had picked a more interesting structure type name than "fast_search_t". That type name I picked is so boring it's going to cause readers to fall asleep.
I'll rename the variables to something more readable ... like ... `readable`. ;-)

lib/util.c
castedo: My hunch it hasn't really been discussed. It's only recently there is a distinction between assert() vs tsk_bug_assert(). Thus not a lot of discussion yet about when one vs the other should be used.

With the introduction of tsk_bug_assert() vs assert(), me learning about the excellent API structure and layers in msprime, and other observations I have made, I've been thinking it would be useful for me to collect many ideas relating to this into a descriptive non-prescriptive doc for many of us to look at. My hope it will clarify the right trade-offs to make in cases such as in this one.

Maybe I can use this case as an example for how it can be improved.

castedo:
OK, I've updated the files based on review comments. Thank you!

Up to now I have had no issue merging and going dancy git maneuvers with many branches and merges. But now I seem to have tied myself up. I attempted a rebase and am getting errors. I'm trying to educated myself now as to why.

benjeffery:
> I attempted a rebase and am getting errors. I'm trying to educated myself now as to why.

I can rebase if git is having a bad day.

castedo:
I've squashed the entire history into one commit via `git merge --squash`. That worked without conflicts in contrast to `git rebase` but I didn't get a chance to fine tune the commit messages and how much squashing was done. `git rebase -t` seems better for that.
I was hoping `git rebase -i -r` would enable some glorious hybrid of the two but it didn't seem that smart at automatically resolving conflicts.

benjeffery:
@castedo You can now see the impact at https://tskit-dev.github.io/msprime-asv/#benchmarks.Hudson.time_human_chr22 AWESOME!

castedo:
NICE! Very satisfying to see!
Is the huge jump up a month ago due to a change in how the benchmark code is getting called? It kind of looks like that might be case looking at the Python code diff.

benjeffery:
The commit which caused that was https://github.com/tskit-dev/msprime/commit/598291de#diff-2ea61695ca9d1d0d62b3e9838a1a1ff9 which changed how Fenwick trees and rate maps worked. I'm not familiar with the details.

PR #1239, opened 2020-10-05
Refactor docs

benjeffery:
This is a copy of #1227 as for some reason I don't have the usual maintainer permissions to edit that PR. PR is from main fork as it needs secrets access.


AdminBot-tskit:
📕 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/1239/) 📕

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/1239/)

benjeffery:
I think this is ready to merge now. It builds the docs and places them at "https://tskit-dev.github.io/msprime-docs/XXXX/" where XXXX is either the tag, "main" or the PR number. This means we can preview doc changes without building locally. Later on I will add an index.html or other mechanism that allows switching between tagged versions, and direct "tskit.dev/msprime-docs" or similar URL.

petrelharp:
Very smooth deployment mechanism!) (I'll have a look at the actual docs next.)

petrelharp:
What sort of feedback are you looking for here, by the way? About the deployment to a separate website, as well as the big-picture reorganization?

One note is that I think the "tutorial" should come first, after "installation", rather than at the end, since people usually want to jump in and try out some common example workflows before reading either the in-depth explanation or the technical documentation.

benjeffery:
Quoting JK: 
"This is still a work in progress, but it could be merged if others wanted to work on it. It'll break the RTD build, but I think that's been silently broken for a while anyway, so that won't make much difference. The basic infrastructure for building the docs is in here, we'd just need to get a hosting solution into place (probably gh-pages?). So, my suggestion would be to merge this much, get the new docs site up, and then we can get more of the docs ported over to using jupyter."

So the idea would be to merge this then file issues for follow up work.

gtsambos:
I'll take a proper look at this tonight!

petrelharp:
I like the set-up. I guess the only question I have is: how will we keep the docs repo clean? We could have a script to delete PR directories when they are closed, I suppose? Otherwise, I'm good with merging this and continuing work.

benjeffery:
> I like the set-up. I guess the only question I have is: how will we keep the docs repo clean? We could have a script to delete PR directories when they are closed, I suppose? Otherwise, I'm good with merging this and continuing work.

As each PR is a non-conflicting commit we can have a process that prunes them after they are closed or after a period of time. Almost all the space is taken up by font files, which eventually we'll bring in from a CDN instead.

agladstein:
I'm actually a little bit confused about what we are doing here. Maybe someone can give an overview of how we are planning to change the docs workflow next tskit meeting?

gtsambos:
Sorry for the delay on this --  I think it's looking pretty good! I'm happy to help bring the existing content in line with this structure by creating examples, etc.

I echo @agladstein and @petrelharp's comments about the tutorial still being useful.

Also, I feel there needs to be some higher-level explanation of why we've structured this in the way we have, so that users know which sections they'll need to consult to see how to do what they want.

For instance, I know that in our heads, it's fairly obvious that 'simulating ancestry' refers to the actual topologies/genetic genealogies of the samples, while 'simulating mutations' refers to the genotypes in the samples. I'm not sure that this would be obvious to new users though, since most other genetic simulators only return genotype information, and this is sometimes all that users want from the simulation. It may not be clear that in order to simulate mutations, you need to know how to simulate ancestry first.

benjeffery:
Ok, I'll go ahead and merge this and we'll follow up with the changes that @gtsambos suggests above.

PR #1241, opened 2020-10-06
Use lcov for coverage

### LINKED ISSUE #1240: Upload lcov report data to codecov.io
The msprime code coverage infra, just today, help identify some needless code I wrote. It really did an awesome job making clear where there was some fishy code worth investigating.

Uploading lcov reports can make the coverage number even more useful and clear. Using lcov will make it possible to not have good error handling code lower the code coverage number, and reach 100% in many files. For example, the goto out handling for failed memory allocations. Specfically `// LCOV_EXCL_LINE` can be appended to these kinds of lines, that are not worth trying to hit from unit tests, but they do the right error handling for users.

Per https://docs.codecov.io/docs/supported-report-formats codecov.io will also accept "Lcov TXT" reports in addition to the usual gcov reports that are currently uploaded.

Using lcov I'm able to get lib/util.c to 100% coverage even though there are error handling lines for when too much memory is getting allocated.

I believe all that is needed is to do what is done today with gcov but then also run the command
```
lcov --capture --directory . --output lcov-data.txt
```
and then have `codecov` upload that data text file.

I'm running this manually and then running `genhtml` (from lcov) to generate an html file. But I think this last step is to be skipped when using codecov.io.

There is an `lcov` package on Debian.

I also tried out `gcovr` but after using both and reading online I have a strong hunch `lcov` will have less headaches at the moment... even though `gcovr` will probably be a better tool in the future.

benjeffery:
Fixes #1240 

Just trying this out to see the effect on CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
           name: Install dependencies and set path
           command: |
             sudo apt-get update
-            sudo apt-get install libgsl0-dev libcunit1-dev libconfig-dev ninja-build valgrind clang
+            sudo apt-get install -y libgsl0-dev libcunit1-dev libconfig-dev ninja-build valgrind clang lcov
             pip install --user -r requirements/development.txt
             pip install --user meson twine
             # way to set path persistently https://circleci.com/docs/2.0/env-vars/#setting-path
@@ -83,6 +83,7 @@ jobs:
             gcov -pb ./libmsprime.a.p/interval_map.c.gcno ../lib/interval_map.c
             gcov -pb ./libmsprime.a.p/util.c.gcno ../lib/util.c
             gcov -pb ./libmsprime.a.p/likelihood.c.gcno ../lib/likelihood.c
+            lcov --capture --directory . --output lcov-data.txt
             cd ..
             codecov -X gcov -F C
 


benjeffery:
@castedo It seems that no coverage for the c files is coming through after adding `lcov`.  Any ideas?

castedo:
I think when I ran it I did `--directory ..` when I was running lcov from within build-gcc. So my guess is either lcov needs to run after `cd ..` or the `..` needs to be with the `--directory` param.
Not super confident of this though. I'll look into this some more now.

castedo:
I have not made any promising discoveries here. One minor detail is that locally I'm using lcov version 1.14 from Ubuntu 20.04 whereas CircleCI is running Debian stretch which has lcov version 1.13. So maybe that version diff is the issue.

If I had to guess right now, I'd guess that codecov.io doesn't *really* support lcov and codecov.io has buggy handling of this lcov data.

One question I have is how the resulting `codecov -X gcov -F C --dump > dump.txt` differs from the `dump.txt` that I get in my development container `docker.io/castedo/tskit-nests:ubuntu-20.04-cdev`
(built from https://github.com/castedo/containers/tree/nests/nests/ubuntu-20.04-cdev)

I'm going to do a little experiment making a container based on Debian stretch that spits out lcov coverage reports. It can be a nice example of how to run different containers with different OS/software on the same local git clone of same source code.

benjeffery:
Closing this for now, but I think it is worth considering telling coverage to ignore the un-hit-able lines if it isn't too hard.

PR #1242, opened 2020-10-11
use gcovr to generate cobertura.xml for codecov

### LINKED ISSUE #1240: Upload lcov report data to codecov.io

castedo:
Fixes #1240

  - enable exclusion of lines via LCOV_EXCL_LINE (see util.c for example)
  - also generate C coverage data for Python/C layer

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
           name: Install dependencies and set path
           command: |
             sudo apt-get update
-            sudo apt-get install libgsl0-dev libcunit1-dev libconfig-dev ninja-build valgrind clang
+            sudo apt-get install libgsl0-dev libcunit1-dev libconfig-dev ninja-build valgrind clang gcovr
             pip install --user -r requirements/development.txt
             pip install --user meson twine
             # way to set path persistently https://circleci.com/docs/2.0/env-vars/#setting-path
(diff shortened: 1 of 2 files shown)


castedo:
@benjeffery Based on what I've learned so far, I recommend against using gcovr and uploading the generated Cobertura XML file. Bummer, I was hoping this would be a great new tool. So I'm closing this PR.

The fundamental problem I see is that gcovr is basically a tool to generate its own reports and the way it structures it's data (in XML) makes sense for the way _it_ displays results (like whether lines are green or not). Unfortunately, codecov.io does not use the Cobertura XML data in the same way. So for instance `assert()` lines show up as yellow and not green in codecov.io with the uploaded Covertura XML. This is not the case with gcovr's own generated HTML reports nor is this the case when just the "raw" gcov data files are uploaded.

So it seems like it's just one more layer and dependency that doesn't appear to have any gain but only downside at this stage.

PR #1252, opened 2020-10-22
bugfix too long continuous tract length

### LINKED ISSUE #1251: gene conversion tract lengths are slightly to large
For small gene conversion tract lengths the distribution of tract lengths is currently not the correct one. For example, the mean tract length is 2.54 if the tract length is set to 2. The effect vanishes for larger tract lengths. I already found the reason for this issue (switching from continuous to discrete coordinates is slightly biased) and am working on a PR that fixes this. Will give more details in the PR.

fbaumdicker:
This one fixes #1251.

We have unintentionally been simulating too long gene conversion tract lengths, when we switched to continuous coordinates. Tract lengths should be geometrically distributed with mean gc_tract_length. But since we are using continuous coordinates we are using the ceiling of an exponentially distributed random variable. So far we have been using an exponentially distributed variable with mean `gc_tract_length`. But the ceiling of this does not give the correct geometric distribution. This is most easily seen in the special case where `gc_tract_length` is 1. In this case, the discrete tract length should always be equal to 1, but the ceiing of the continuous tract length can be larger. Similary, a continuous mean tract length of 2 will lead to a discrete mean tract length of ~2.54.
The correct mean tract length in the continous case is given by 1/(log(tract_length) - log(tract_length - 1)).
I am defining this now as `self->continuous_gc_tract_length`.
In the special case where the discrete mean tract length is 1, I set `continuous_gc_tract_length` to a small enough number and do not generate any random tract lengths in the discrete case, as this is not necessary.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -1457,6 +1457,14 @@ class Simulator:
             ret = x, alpha
         return ret
 
+    def generate_gc_tract_length(self):
+        # generate tract length
+        if self.discrete_genome:
+            tl = np.random.geometric(1 / self.tract_length)
+        else:
+            tl = np.random.exponential(self.tract_length)
+        return tl
+
     def wiuf_gene_conversion_within_event(self, label):
         """
         Implements a gene conversion event that starts within a segment
@@ -1470,7 +1478,7 @@ class Simulator:
         )
         x = y.prev
         # generate tract_length
-        tl = np.random.geometric(1 / self.tract_length)
+        tl = self.generate_gc_tract_length()
         assert tl > 0
         right_breakpoint = left_breakpoint + tl
         if y.left >= right_breakpoint:
@@ -1589,7 +1597,7 @@ class Simulator:
         assert y is not None
 
         # generate tract_length
-        tl = np.random.geometric(1 / self.tract_length)
+        tl = self.generate_gc_tract_length()
 
         bp = y.left + tl
         while y is not None and y.right <= bp:
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -217,6 +217,18 @@ msp_get_num_gene_conversion_events(msp_t *self)
     return self->num_gc_events;
 }
 
+size_t
+msp_get_num_internal_gene_conversion_events(msp_t *self)
+{
+    return self->num_internal_gc_events;
+}
+
+double
+msp_get_sum_internal_gc_tract_lengths(msp_t *self)
+{
+    return self->sum_internal_gc_tract_lengths;
+}
+
 int
 msp_set_start_time(msp_t *self, double start_time)
 {
@@ -297,11 +309,13 @@ msp_set_gene_conversion_tract_length(msp_t *self, double tract_length)
 {
     int ret = 0;
 
-    if (tract_length < 0 || tract_length > self->sequence_length) {
+    if ((self->discrete_genome && tract_length < 1) || tract_length <= 0
+        || tract_length > self->sequence_length) {
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
     }
     self->gc_tract_length = tract_length;
+
 out:
     return ret;
 }
@@ -2631,6 +2645,29 @@ out:
     return ret;
 }
 
+static double
+msp_generate_gc_tract_length(msp_t *self)
+{
+    double tl;
+    int num_resamplings = 0;
+
+    /* generate tract length */
+    if (self->discrete_genome) {
+        tl = gsl_ran_geometric(self->rng, 1 / self->gc_tract_length);
+    } else {
+        do {
+            if (num_resamplings == 10) {
+                tl = -1;
+                goto out;
+            }
+            tl = gsl_ran_exponential(self->rng, self->gc_tract_length);
+            num_resamplings++;
+        } while (tl <= 0);
+    }
+out:
+    return tl;
+}
+
 static int MSP_WARN_UNUSED
 msp_gene_conversion_event(msp_t *self, label_id_t label)
 {
@@ -2638,10 +2675,10 @@ msp_gene_conversion_event(msp_t *self, label_id_t label)
     segment_t *x, *y, *alpha, *head, *tail, *z, *new_individual_head;
     double left_breakpoint, right_breakpoint, tl;
     bool insert_alpha;
-    int num_resamplings = 0;
 
     tsk_bug_assert(self->gc_mass_index != NULL);
     self->num_gc_events++;
+    self->num_internal_gc_events++;
     ret = msp_choose_uniform_breakpoint(
         self, label, &self->gc_map, self->gc_mass_index, true, &left_breakpoint, &y);
     if (ret != 0) {
@@ -2651,18 +2688,13 @@ msp_gene_conversion_event(msp_t *self, label_id_t label)
     x = y->prev;
 
     /* generate tract length */
-    do {
-        tl = gsl_ran_exponential(self->rng, self->gc_tract_length);
-        if (self->discrete_genome) {
-            /* We want the tract length to be at least 1 */
-            tl = ceil(tl);
-        }
-        if (num_resamplings == 10) {
-            ret = MSP_ERR_TRACTLEN_RESAMPLE_OVERFLOW;
-            goto out;
-        }
-        num_resamplings++;
-    } while (tl <= 0);
+    tl = msp_generate_gc_tract_length(self);
+    if (tl == -1) {
+        ret = MSP_ERR_TRACTLEN_RESAMPLE_OVERFLOW;
+        goto out;
+    }
+    tsk_bug_assert(tl > 0);
+    self->sum_internal_gc_tract_lengths += tl;
     right_breakpoint = left_breakpoint + tl;
 
     if (y->left >= right_breakpoint) {
@@ -3634,6 +3666,9 @@ msp_reset(msp_t *self)
     self->next_sampling_event = 0;
     self->num_re_events = 0;
     self->num_gc_events = 0;
+    self->num_internal_gc_events = 0;
+    self->sum_internal_gc_tract_lengths = 0;
+    self->num_noneffective_gc_events = 0;
     self->num_ca_events = 0;
     self->num_rejected_ca_events = 0;
     self->num_trapped_re_events = 0;
@@ -3968,25 +4003,19 @@ msp_gene_conversion_left_event(msp_t *self, label_id_t label)
     double h = gsl_rng_uniform(self->rng) * gc_left_total;
     double tl, bp;
     segment_t *y, *x, *alpha;
-    int num_resamplings = 0;
 
     y = msp_find_gc_left_individual(self, label, h);
     assert(y != NULL);
 
     /* generate tract length */
-    do {
-        tl = gsl_ran_exponential(self->rng, self->gc_tract_length);
-        if (self->discrete_genome) {
-            /* We want the tract length to be at least 1 */
-            tl = ceil(tl);
-        }
-        if (num_resamplings == 10) {
-            ret = MSP_ERR_TRACTLEN_RESAMPLE_OVERFLOW;
-            goto out;
-        }
-        num_resamplings++;
-    } while (tl <= 0);
+    tl = msp_generate_gc_tract_length(self);
+    if (tl == -1) {
+        ret = MSP_ERR_TRACTLEN_RESAMPLE_OVERFLOW;
+        goto out;
+    }
     tsk_bug_assert(tl > 0);
+    self->num_internal_gc_events++;
+    self->sum_internal_gc_tract_lengths += tl;
 
     bp = y->left + tl;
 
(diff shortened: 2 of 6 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/1252/)

fbaumdicker:
Hm, the test_gc_rate is failing now, which is maybe no surprise as we have changed the tract length distribution. However, I am not sure how to change the simulation parameters in this test. What is the intention behind this test?
https://github.com/tskit-dev/msprime/blob/2d482aff75dd2b033efb817ac2c28dd13c68d2a8/tests/test_ancestry.py#L1356

lib/msprime.c
```diff
@@ -2651,18 +2658,33 @@ msp_gene_conversion_event(msp_t *self, label_id_t label)
     x = y->prev;
 
     /* generate tract length */
-    do {
-        tl = gsl_ran_exponential(self->rng, self->gc_tract_length);
```
jeromekelleher, line 2655: Looks like there's some non-trivial logic here that we're duplicating. Perhaps we could have a function 
```c
int msp_generate_tract_length(self, double * ret_tract_length)
```
to do the work?

lib/msprime.c
```diff
@@ -297,11 +297,18 @@ msp_set_gene_conversion_tract_length(msp_t *self, double tract_length)
...
     int ret = 0;
 
-    if (tract_length < 0 || tract_length > self->sequence_length) {
+    if (tract_length < 1 || tract_length > self->sequence_length) {
         ret = MSP_ERR_BAD_PARAM_VALUE;
         goto out;
     }
     self->gc_tract_length = tract_length;
+    if (tract_length == 1) {
+        self->continuous_gc_tract_length = 0.001;
```
jeromekelleher, line 306: Is it really worth storing this? I doubt this computation will even show up on any profiles, so my vote would be to keep it simple and to compute this value on the fly in ``msp_generate_tract_length``

jeromekelleher (commented):
Looks good @fbaumdicker. I don't quite follow the logic, but hopefully it'll be clearer when we've pulled the tract length generation into its own function.

jeromekelleher:
> Hm, the test_gc_rate is failing now, which is maybe no surprise as we have changed the tract length distribution. However, I am not sure how to change the simulation parameters in this test. What is the intention behind this test @jeromekelleher?

This test was basically to make sure that the GC code is doing something, just to check the all the layers are plumbed together correctly. Feel free to change the parameters to something more sensible.

lib/msprime.c
fbaumdicker: I thought since potentially we might be generating lots of (noneffective) gc events, I do this computation only once.
However, I have no idea how much this influences the performance, so I agree and will compute it on the fly.

lib/msprime.c
fbaumdicker: Yes, the current way we simulate gc_left and gc_within means we are generating the tract lengths in the same way. Before the speed up it was different. Will add a function.

lib/msprime.c
jeromekelleher: Saving computations like this will only make a difference within really tight inner loops. Our runtime is dominated by the cost of maintaining the Fenwick trees, binary search tree lookups and the cost of iterating over segments in merge_ancestors. Something like this won't even show up.

fbaumdicker: I see. So adding the variable gc rates, where we need another Fenwick tree will be much more of an issue regarding performance. Will compute it on the fly.

fbaumdicker:
> > Hm, the test_gc_rate is failing now, which is maybe no surprise as we have changed the tract length distribution. However, I am not sure how to change the simulation parameters in this test. What is the intention behind this test @jeromekelleher?
> 
> This test was basically to make sure that the GC code is doing something, just to check the all the layers are plumbed together correctly. Feel free to change the parameters to something more sensible.

I was wrong, the test fails as we are no longer allowing tract_lengths below 1. Can I just remove the two tests that use a tract length of 0.5 here? https://github.com/tskit-dev/msprime/blob/1cd73ef3f498903e022ebb0b2f6cfb9fa9564ab0/lib/tests/test_ancestry.c#L1567

jeromekelleher:
> I was wrong, the test fails as we are no longer allowing tract_lengths below 1. Can I just remove the two tests that use a tract length of 0.5 here?

Sure, whatever makes sense.

fbaumdicker:
From my side, this is ready to merge. I do not see a way to improve the coverage further, as the parts that are not covered are extremely unlickely to happen at all. We could simply remove them, but this might lead to wierd bahviour in this really unlickely event.

jeromekelleher:
Great, thanks @fbaumdicker! I'll take a look later today and merge.

lib/msprime.c
```diff
@@ -2631,17 +2644,58 @@ msp_recombination_event(msp_t *self, label_id_t label, segment_t **lhs, segment_
     return ret;
 }
 
+static double
+msp_generate_gc_tract_length(msp_t *self)
+{
+    double tl, continuous_gc_tract_length;
```
jeromekelleher, line 2650: Sorry to be slow to get around to this and for dragging it out more @fbaumdicker, but this looks really weird and unnatural to me. I've spent some time playing around with the values and how this is done, and I'm really confused now.

I thought that the *floor* of an exponential RV is geometrically distributed (note we're taking the ceiling here), and so we could generate the tract len value as an exponential with mean ``tract_length`` and then take the floor to get the required geometric with mean ``tract_length``. 

But this doesn't seem to be the case... I'm confused.

If we can't document the difference between the discrete and continuous case in one sentence, then we should probably just stick to discrete-only for GC.

fbaumdicker: Hi @jeromekelleher,
this is slightly more complicated. In fact, the ceiling and the floor of an exponential RV are both geometrically distributed.
There are two types of geometric distributions. A) Number of trials until the first success appears and B) Number of trials before the first success. ceiling and floor result in A) and B), respectively.
For gene conversion tract length we want to use A) as each gene conversion tract should have at least length 1, and this is the distribution Wiuf and Hein used.
So this is why I am using the ceiling instead of the floor.

Furthermore, the ceiling of an exponential distributed RV with mean r is geometrically distributed but the mean is no longer given by r. See point 50 at https://www.randomservices.org/random/dist/Transformations.html .

In order to get the mean gc_tract_length for the ceiling of the exponential distributed rv we have to set the mean of the exponential distributed rv to
1 / (log(self->gc_tract_length) - log(self->gc_tract_length - 1))
(A similar formular holds for the floor)

The special limit case where gc_tract = 1, would lead to an exponential distributed RV with infinite rate, i.e. mean close to zero. In this case, the continuous gene conversion tract length is set to a very small value.

I have a branch with an updated gc documentation ready here https://github.com/fbaumdicker/msprime/commit/ca21a562a3c99b0f30dd90571dee217e220255f4. I was planning to create a documentation pull request after this one is merged.
However, currently, the docs are not explaining it as detailed as shown here. 
I will update the documentation to explain what is going on as detailed as here.
This should explain it properly. However, here is a one-sentence version:

_As a geometric distributed rv with mean r is equally distributed as the ceiling of an exponentially distributed rv with mean 1 / (log(r) - log(r - 1)), we are using exponentially distributed gene conversion tract lengths with the transformed mean when a continuous genome is simulated._


A related issue:
Are users often simulating genomes with a scaled length? I.e. using a genome length of 1 instead of N and scaling the recombination rate accordingly? Or is this a rare scenario?
I am asking as we currently do not support mean tract lengths below 1, which means the length of the continuous genome can currently not be scaled (together with the tract length) such that the tract length would be below 1.
However, I think it might be possible to allow this in the continuous case if we make a few more case distinctions, but personally I think it should suffice as it currently is, as you can easily simulate a continuous genome with length N without any disadvantages.

jeromekelleher: Thanks very much for the explanation @fbaumdicker, this is super helpful. Since the distinction between the cases for discrete and continuous is rather subtle, and we're already making a if/else in any case, why don't we do somehting like
```
if (self->discrete_genome) {
    tl = gsl_ran_geometric(self->rng, 1 / self->tract_len);
} else 
    tl = gsl_ran_exponetial(self->rng, 1 / self->tract_len); // or whatever the param is that gives us mean = tract_len
}
```
I.e., in the discrete case we generate a geometric RV and in the continuous we generate an exponential?

fbaumdicker: That is an excellent idea. This will make the code much more readable and should not make any difference. Will make the changes and push it in the next days.

jeromekelleher: Fantastic, thanks @fbaumdicker! Would you mind also making the changes to the Python algorithms.py version - we want to keep the implementations reasonably in sync.

fbaumdicker: Done!
Simplified the tract length generation and made the same changes to the python version.
Not sure about the lint error, I get now? What does it mean @jeromekelleher? Can we merge anyway?

benjeffery:
The lint error is due to github actions removing a feature - I'll get a fix done.

benjeffery:
The fix has been merged to main - you can rebase this branch to get it.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1252/)

fbaumdicker:
> The fix has been merged to main - you can rebase this branch to get it.

Great! Thanks. I rebased and this PR should be ready to merge now.

algorithms.py
```diff
@@ -1457,6 +1457,20 @@ def hudson_recombination_event(self, label, return_heads=False):
             ret = x, alpha
         return ret
 
+    def generate_gc_tract_length(self):
+        # generate tract length
+        if self.discrete_genome:
+            tl = np.random.geometric(1 / self.tract_length)
+        else:
```
jeromekelleher, line 1464: I'm being dense here @fbaumdicker, but why not use ``tl = np.random.exponential(self.tract_length)``?

```
>>> np.mean(np.random.exponential(10, size=10**7))
10.002678522591678
>>> np.mean(np.random.exponential(1, size=10**7))
0.9999924193218064
>>> np.mean(np.random.exponential(0.1, size=10**7))
0.10002349776505161
```

This takes away all the singularities, and gives a nice simple way to explain what it's doing: When we have ``discrete_genome=False`` the gene conversion tract lengths are drawn from an exponential distribution with rate equal to the ``tract_length`` parameter. When ``discrete_genome`` is True, the values are drawn from a geometric distribution with parameter 1 / tract_length. 

This way the expected length of gene conversion tracts is always equal to ``tract_length``, right?

fbaumdicker: Yes, that would be the case.

I am mainly thinking in discrete coordinates, so I am really unsure what the standard user would expect if he is simulating a continuous genome. My reason to transform the mean of the exponential is mainly based on the idea that I would transform the coordinates into discrete coordinates after the simulation such that the left and right gene conversion tract ends would need to be rounded (to the previous and next larger integer). I wanted the mean of the resulting rounded tracts to be the same as the supplied parameter.
But maybe that is not what other users would expect? How should users interpret continuous gene conversion tracts?
Does a tract from 4.234 to 6.421 correspond to a tract from 4 to 6 or from 4 to 7?
Is a tract from 4.234 to 4.83 a gene conversion event at all?
Or should we think of continuous coordinates as the limit for large genome sizes anyway, such that even 4.234 to 4.235 covers a bunch of bases?

if we use the original tract_length and if users transform their coordinates after the simulation, the results will differ compared to the discrete case. I wanted to avoid that. In particular, for tract_length = 1 this can lead to very different results.
The geometric distribution with mean 1 has zero variance, as the tract length always equals 1 in this case. For the exponential distribution tract lengths above 1 are possible. 
However, if users will not transform their coordinates then all of this is not an issue.

I would recommend making the decision, based on what you think people are using continuous coordinates for and whether you want the discrete and continuous case to be interchangeable.

jeromekelleher: We should view continuous coordinates as just that, and assume that users will leave the coordinates as-is afterwards for analysis. I would imagine the majority of people working in continuous coordinates would work in the unit interval.

Put it this way - recombination already acts on continuous coordinates, and we can have millions of recombinations in a 10bp interval. It would be very confusing and surprising if gene conversion behaved differently - there's nothing inherently discrete above GC above and beyond recombination. If recombination can be assumed to occur at fractional genome locations, then gc breakpoints should be assumed to occur at fractional genome locations, and there's certainly no reason why gc tract lengths would have be to of integer lengths (particuarly and especially since the breakpoints would start at continuous points in the first place).

Do you see where I'm coming from? The default will be to work in discrete space as you prefer, but there's no reason this can't also work in continuous space in an analogous way to recombination.

fbaumdicker: Ok, I am convinced. Will change the mean of the continuous tract lengths back to tract_length. I will also remove the requirement for the tract_length to be >= 1 in the continuous case. Larger than 0 should suffice in this case. Especially if people use the unit interval as their coordinates.

fbaumdicker:
Done.

jeromekelleher (approved):
Great! I'm glad we resolved this, thanks for your patience @fbaumdicker.

Ready to merge after a squash.

fbaumdicker:
Great! squash is done.
It was certainly worth taking the necessary time here @jeromekelleher.

jeromekelleher (approved):

PR #1260, opened 2020-10-29
Don't push docs for PRs with no doc changes, prevent doc push on forks.

benjeffery:
Diff the generated HTML to check if a doc preview is needed. Tags still always get built and deployed.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,21 +43,29 @@ jobs:
         run: make -C docs
 
       - name: Checkout docs site
+        if: github.repository_owner == 'tskit-dev'
         uses: actions/checkout@v2
         with:
           repository: tskit-dev/msprime-docs
           token: ${{ secrets.ADMINBOT_DOCS_TOKEN }}
           path: msprime-docs
 
+      - name: Check for diff
+        if: github.repository_owner == 'tskit-dev'
+        id: diff
+        #The html contains commit hashes which we don't want to count as a change
+        run: |
+          diff -x .buildinfo -r msprime-docs/main docs/_build/html | grep "^[<>]" | grep -v "<title>\|VERSION:" | grep . && echo "::set-output name=change_detected::true" || echo "NO CHANGES"
+
       - name: Copy our docs to the PR specific location
-        if: github.event.pull_request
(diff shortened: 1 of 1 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/1260/)

.github/workflows/docs.yml
```diff
@@ -43,28 +43,37 @@ jobs:
...
           repository: tskit-dev/msprime-docs
           token: ${{ secrets.ADMINBOT_DOCS_TOKEN }}
           path: msprime-docs
 
+      - name: Check for diff
+        if: github.repository_owner == 'tskit-dev'
+        id: diff
+        #The html contains commit hashes which we don't want to count as a change
+        run: |
+          diff -x .buildinfo -r msprime-docs/main docs/_build/html | grep "^[<>]" | grep -v "<title>\|VERSION:" | grep . && echo "::set-output name=change_detected::true" || echo "NO CHANGES"
```
grahamgower, line 58: If there are changes, they will be printed by the final grep command. Maybe this is clutters the logs and it should instead be `grep . > /dev/null`?

grahamgower (commented):
It's not clear to me what the value of `steps.diff.outputs.change_detected` will be in the event that there are no docs diffs. I also looked in the github docs and didn't see anything about how `if undefined-variable` is interpreted. Is this somehow well-defined behaviour? (It would be nice to know that typos in variable names are not silently ignored!)

jeromekelleher (commented):
LGTM, but I'm not really keeping up with this tbh

jeromekelleher (approved):

.github/workflows/docs.yml
benjeffery: I left this in as it might be useful to see the diff sometimes. Log space isn't a problem.

benjeffery:
> It's not clear to me what the value of `steps.diff.outputs.change_detected` will be in the event that there are no docs diffs.

It's all javascript under the hood so it will be `undefined` and falsey. Spelling mistakes will be false, not ideal, but I've tested both with and without changes.

grahamgower (approved):

grahamgower:
Ok, thanks for indulging my ignorance!

PR #1287, opened 2020-11-08
Seeds for replicates

### LINKED ISSUE #1257: Seed the RandomGenerator for each replicate in 1.0?
We currently put a lot of work and complexity into making sure that we use the same random generator instance and keep the state in between simulation replicates. It would be a lot simpler if we just generated a bunch of seeds instead, which would also open the door to providing library support for doing stuff in parallel.

Is there actually any efficiency gain in keeping the same RandomGenerator instance lying around? Even if there is, we could still expose an API to set the seed.

If this works, we can remove the ``replicate_index`` parameter and some of the ugly infrastructure around keeping track of this for provenance.

jeromekelleher:
Closes #1257 

The idea here is a use an instance of the GSL random generator to generate a random seed for each replicate. Currently, we use a single random generator for *all* the replicates. This was done in the first instance partly for performance reasons but mainly through paranoia about introducing some weird bias.

I think it must be OK,  though, right? That is,
```
rng = Random(seed)
replicates = [random_process(rng) for _ in range(num_replicates)]
```
should be equivalent in distribution to
```
rng = Random(seed)
seeds = [rng.uniform_int(1, 2**31 - 1) for _ in range(num_replicates)]
replicates = [random_process(Random(seeds[j]) for j in range(num_replicates)]
```

Right? So, I guess the assumption is that choosing seeds uniformly from the range of the RNG is the same thing as running the RNG for a long time.

@molpopgen, @petrelharp, am I doing something stupid here?

We can simplify things quite a bit and start offering meaningful APIs for doing simulations in parallel if we do it this way, so I think it's a good idea to make this change pre 1.0.

ps. Most of the code in here is for numpyifying the GSL random generator interface. We'll probably want this at some point if we do present APIs for doing simulations in parallel.

### DIFF
--- a/msprime/_msprimemodule.c
+++ b/msprime/_msprimemodule.c
@@ -255,14 +255,19 @@ RandomGenerator_init(RandomGenerator *self, PyObject *args, PyObject *kwds)
 {
     int ret = -1;
     static char *kwlist[] = {"seed", NULL};
-    PyObject *py_seed;
+    PyObject *py_seed = NULL;
 
     self->rng = NULL;
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &py_seed)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O", kwlist, &py_seed)) {
         goto out;
     }
     self->rng = gsl_rng_alloc(gsl_rng_default);
-    ret = RandomGenerator_parse_seed(self, py_seed);
+    if (py_seed != NULL) {
+        if (RandomGenerator_parse_seed(self, py_seed) != 0) {
+            goto out;
+        }
+    }
+    ret = 0;
 out:
     return ret;
 }
--- a/msprime/_msprimemodule.c
+++ b/msprime/_msprimemodule.c
@@ -306,15 +306,31 @@ RandomGenerator_flat(RandomGenerator *self, PyObject *args)
 {
     PyObject *ret = NULL;
     double a, b;
+    long n = 1;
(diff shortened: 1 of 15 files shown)


molpopgen:
This will be fine.  In fact, I almost always do this for msprime because I often use concurrent.futures to run replicates.

There's probably no real reason to do this for most use cases, though.  But it will allow you to store a seed in the tree sequence metadata for each replicate, which would be nice.

benjeffery:
I'm in no way an expert here, but it seems to me that the change here is from "pick one sequence of replicates from 2^32 possible replicate sequences" to "pick one sequence of replicates from 2^32 possible replicates".

So before this change there were many more possible replicates that could be generated. I'm not sure that is an issue - but it is a material change.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1287/)

jeromekelleher:
This is ready for review now. It ended up being a rag-bag bunch of cleanups and refactors, which didn't do the thing that the PR was initially supposed to do. Oh well.

The reason for going back on having a single seed per replicate is that we can't do this without either breaking reproducibility or killing performance in the lots-of-small-replicates case. This small bit of extra complexity in explaining how randomness works is worth the cost, I think.

jeromekelleher:
Closes #1297 
Closes #1288

benjeffery (approved):
Yikes, that's a lot of cleanup - I've had a quick scan through and LGTM.

Issue #1297, opened 2020-11-11
Remove parameters added to simulate after 0.7.4


jeromekelleher:
This will simplify the code and help to make sure that the simulate() interface stays stable in the long term (as we don't have to worry about adding new parameters).

Thus, if you want to take advantage of new features you'll need to modify your code to you sim_ancestry/sim_mutations, but your old code will continue to work indefinitely.

Parameters added since 0.7.4:

- discrete_genome
- gene_conversion_rate/gene_conversion_tract_length
- pedigree (but this isn't documented or feature complete anyway #1186 etc)
- record_provenance
- demography

Seems like it makes sense to keep ``record_provenance`` as this could be useful to existing workflows.

The ``demography`` option is less clear - I don't want to discourage people from using the new demography API. It's not a lot of effort to support this, so I'm inclined to keep it.

PR #1299, opened 2020-11-12
Docs for simulating with multiple chromosomes

### LINKED ISSUE #1005: [Docs] Update section on "simulating multiple chromosomes"
This is currently out of date and, should be using the ``discrete`` option to RecombinationMap.

Also some important details should be spelled out. @apragsdale, would you mind listing out what needs to be said here?

apragsdale:
Closes #1005

### DIFF
--- a/docs/ancestry.rst
+++ b/docs/ancestry.rst
@@ -760,49 +760,81 @@ rate along the genome can be combined in ``msprime``.
 Multiple chromosomes
 +++++++++++++++++++++
 
-.. todo:: This documentation was copied from the old tutorial and is
-    actively wrong now. Rewrite to use the new APIs.
-
-.. warning:: This approach is somewhat hacky; hopefully we will have a more
-    elegant solution soon!
-
 Multiple chromosomes can be simulated by specifying a recombination map with
-hotspots between chromosomes. For example, to simulate two chromosomes each 1
-Morgan in length:
+single-bp segments with recombination probability 1/2 separating adjacent
+chromosomes. By simulating using ``sim_ancestry`` and ``discrete_genome=True``
+(the default setting), we set the recombination rate so that chromosomes
+separated by such "hotspots" have a 50% chance of being inherited from the
+same parent.
 
-.. code-block:: python
+The probability that a recombination breakpoint occurs at a given base pair
+that has recombination rate :math:`r` is :math:`1-e^{-r}`. Thus, we set the
+recombination rate in the base pair segments separating chromosomes to
+:math:`\log(2)`. This ensures that the recombination probility each generation
+is 1/2.
 
-    rho = 1e-8
-    positions = [0, 1e8-1, 1e8, 2e8-1]
-    rates = [rho, 0.5, rho, 0]
-    num_loci = int(positions[-1])
+For example, the following defines a recombination map for three chromosomes
+each 1 cM in length:
 
-    recombination_map = msprime.RecombinationMap(
-        positions=positions, rates=rates, num_loci=num_loci)
+.. jupyter-kernel:: python3
 
-    tree_sequence = msprime.simulate(
-        sample_size=100, Ne=1000, recombination_map=recombination_map,
-        model="dtwf")
+.. jupyter-execute::
 
-Care must be taken when simulating whole genomes this way, as the rescaling
-required to model such large fluctuations in recombination rate can result in
-the following error: ``Bad edge interval where right <= left``
+    import msprime
+    import numpy as np
 
-This can be avoided by discretizing the genome into 100bp chunks by changing
-the above to:
+    r_chrom = 1e-8
+    r_break = np.log(2)
+    positions = [0, 1e6, 1e6 + 1, 2e6, 2e6 + 1, 3e6]
+    rates = [r_chrom, r_break, r_chrom, r_break, r_chrom]
+    rate_map = msprime.RateMap(positions, rates)
 
-.. code-block:: python
+The main purpose of simulating multiple chromosomes together (instead of
+running independent simulations for each chromosome) is to capture the correct
+correlation between trees on distinct chromosomes. (Note that for many purposes,
+this may not be necessary, and simulating chromosomes independently may be
+preferred for efficiency.)
+To get the correct correlation between chromosomes, we must use the
+:ref:`discrete-time Wright-Fisher <sec_ancestry_models_dtwf>` model in the
+recent past. Because correlations between chromosomes break down very rapidly,
+we only need to simulate using ``dtwf`` for roughly 10-20 generations, after which
+we can switch to the ``hudson`` model to finish the simulation more efficiently.
 
-    rho = 1e-8
-    positions = [0, 1e8-1, 1e8, 2e8-1]
-    rates = [rho, 0.5, rho, 0]
-    num_loci = positions[-1] // 100 # Discretize into 100bp chunks
+For example, simulating 10 sampled diploid individuals using the recombination
+map defined above for three chromosomes and switching from the ``dtwf`` to
+``hudson`` model 20 generations ago, we run
 
-Also note that recombinations will still occur in the gaps between chromosomes,
-with corresponding trees in the tree sequence. This will be fixed in a future
-release.
+.. jupyter-execute::
 
+    tree_sequence = msprime.sim_ancestry(
+        10,
+        population_size=1000,
+        recombination_rate=rate_map,
+        model=["dtwf", (20, "hudson")]
+    )
 
+To isolate tree sequences for each chromosome, we can trim the full tree sequence
+for each chromosome by defining the end points of each chromosome and using
+the ``keep_intervals`` fuction. It is important to specify ``simplify=False`` so
+that node indices remain consistent across chromosomes. The function ``trim()``
+is then used to trim the tree sequences to the focal chromosome, so that position
+indices begin at 0 within each chromosome tree sequence.
+
+.. jupyter-execute::
+
+    chrom_positions = [0, 1e6, 2e6, 3e6]
+    ts_chroms = []
+    for j in range(len(chrom_positions) - 1):
+        start, end = chrom_positions[j: j + 2]
+        chrom_ts = tree_sequence.keep_intervals([[start, end]], simplify=False)
+        ts_chroms.append(chrom_ts.trim())
+
+This gives us a list of tree sequences, one for each chromosome, in the order that
+they were stitched together in the initial recombination map.
+
+.. todo:: Add a section to the Tutorial showing how to access tree sequences across
+    chromosomes and showing the correlation of trees on different chromosomes
+    using the ``dtwf`` model.
 
 **********************
 Controlling randomness


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1299/)

jeromekelleher (commented):
Superb, thanks @apragsdale!

It would be good to change the ``code-block:: python`` sections into Jupyter-sphinx blocks, though, so we're running live code chunks. We can make the example much smaller if it's slow. I wonder if there's any way we could illustrate that the trees on the different chromosomes are correlated? Or perhaps illustrate how we'd access the trees on the different chromosomes?

apragsdale:
Ok, switched the code blocks to jupyter sphinx - which was a good call, since I caught some mistakes in the code blocks!

For accessing trees on different chromosomes, there's nothing in the tree sequence that specifies anything about which segments belong to which chromosome, so the user would need to know themselves where positions map to along their genome.

I'm not sure how we'd show correlations between trees without doing a bunch of simulations and plotting statistics as in Dominic's DTWF paper (but that isn't really feasible within some docs juptyper blocks). Did you have other thoughts here?

jeromekelleher:
> For accessing trees on different chromosomes, there's nothing in the tree sequence that specifies anything about which segments belong to which chromosome, so the user would need to know themselves where positions map to along their genome.

That's true, but I guess we could give a little recipe here to make that easier. We could put in the lengths as variables to make this possible, I guess, but it's not totally clear how to do this nicely.

> I'm not sure how we'd show correlations between trees without doing a bunch of simulations and plotting statistics as in Dominic's DTWF paper (but that isn't really feasible within some docs juptyper blocks). Did you have other thoughts here?

I was thinking more along the lines of a small example, where we show visually that the tree on the end of one chromosome is very similar to the start of another one, and that this wouldn't be true if we simulated them independently. Maybe this is the subject for a tutorial though, tbh.

apragsdale:
> That's true, but I guess we could give a little recipe here to make that easier. We could put in the lengths as variables to make this possible, I guess, but it's not totally clear how to do this nicely.

Yeah, I think it would be straightforward to have a few extra lines to explain how to get trees for one chromosome vs the other. Is there a simple way now to start a tree iterator with a specific start and end position in the tree sequence?

I suppose you could use the `keep_intervals` function to pull out each chromosome - in a similar vein, is there a way to do something like `keep_intervals` on each chromosome that resets the position indexes to zero and erases everything outside of that kept chromosome, instead of just removing branches and mutations? That would give you tree sequences, one per chrom, as though you simulated them independently, but they would then have the correct correlation structure across chroms.

> I was thinking more along the lines of a small example, where we show visually that the tree on the end of one chromosome is very similar to the start of another one, and that this wouldn't be true if we simulated them independently. Maybe this is the subject for a tutorial though, tbh.

I agree, that does sound a bit more like a tutorial section instead of a short illustration here. Can we punt this for now and come back to it later when msprime 1.0 is a bit more documented and filled in?

jeromekelleher:
> I suppose you could use the keep_intervals function to pull out each chromosome - in a similar vein, is there a way to do something like keep_intervals on each chromosome that resets the position indexes to zero and erases everything outside of that kept chromosome, instead of just removing branches and mutations? That would give you tree sequences, one per chrom, as though you simulated them independently, but they would then have the correct correlation structure across chroms.

Fantastic idea! Can't believe we didn't think of that before. So, we'd have something like

```python
ts_chroms = []
for j in range(len(recomb_map.position) - 1):
    start, end = recomb_map.position[j: j + 1]
    chrom_ts = ts.keep_intervals([start, end], simplify=False) # simplify is important so the nodes retain identity
    ts_chroms.append(chrom_ts.trim())
```

And that should do it.

> I agree, that does sound a bit more like a tutorial section instead of a short illustration here. Can we punt this for now and come back to it later when msprime 1.0 is a bit more documented and filled in?

Yes, let's do that. Can we add a ".. todo::" section here saying we need to link to a tutorial here explaining how to access

apragsdale:
> And that should do it.

Yes, great idea - I think this improves this section quite a bit.

> Can we add a ".. todo::" section here saying we need to link to a tutorial here explaining how to access

Yep, added the todo.

petrelharp:
Why aren't you using a discrete recombination map, btw?

apragsdale:
> Why aren't you using a discrete recombination map, btw?

My understanding (which could be wrong - I'm just figuring out the new methods this week) is that the recombination map has been replaced with a generic rate map, and discrete coordinates are used by default in `sim_ancestry`. So by default all recombination events occur at integer positions.

E.g., the following always creates two trees, even though the rate is very large separating the two segments:
```python
import msprime
positions = [0, 1000, 1001, 2000]
rates = [0, 2, 0]
rate_map = msprime.RateMap(positions, rates)
ts = msprime.sim_ancestry(
    2, population_size=1000,
    recombination_rate=rate_map, model="dtwf")
print(ts.num_trees)
print(ts.first().interval)
print(ts.last().interval)
```

petrelharp:
oh, right - so, you *are* using a discrete map! In that case, I don't think this is necessary any more:
```
Note that in defining the recombination map, the rates separating chromosomes are much larger than 0.5. ...
```
- you should just set the rate to be 0.5 on that interval (but it's worth pointing out that this is a discrete map...).

petrelharp:
A suggestion: instead of saying "effectively unlink" why not just say they have a 50% chance of being co-inherited? Like maybe:
```
 By simulating using sim_ancestry with discrete_genome=True (the default setting)
and a rate of 0.5 on a single-bp segment separating two regions,
we ensure that the two regions have a 50% chance of being both inherited from the same parent,
just as real chromosomes (that assort independently). 
```

Also, it might be helpful to explain what the correlations between chromosomes are, e.g. "For instance, in the continuous-time Hudson model, each recombination event creates a new ancestor; so that with 22 chromosomes, a genome might have tens of distinct ancestors going back only one or two time units. Simulating using the DTWF model allows chromosomes to be inherited together, and is completely equivalent to independent assortment." (this could be improved, but you get the idea?)

apragsdale:
> you should just set the rate to be 0.5 on that interval (but it's worth pointing out that this is a discrete map...).

This is something I'm not clear on. Is the rate still a Poisson rate on that interval? If that's the case, then a rate of 0.5 wouldn't end up giving us the probability of drawing odd number of recombination events close to 1/2. Or is it really that every bp position within that interval (which is just the single site) has probability 1/2 of having a recombination event per unit time? In which case, yes, rate of 0.5 would work.

> A suggestion: instead of saying "effectively unlink" why not just say they have a 50% chance of being co-inherited? Like maybe:

 Good suggestions! Thanks

petrelharp:
> This is something I'm not clear on. Is the rate still a Poisson rate on that interval?

Well, our [former selves](https://github.com/tskit-dev/msprime/issues/1005#issuecomment-691291236) didn't know either. It  [happens here](https://github.com/tskit-dev/msprime/blob/main/lib/msprime.c#L2071), and I *think* that with a discrete genome, the upshot is that a breakpoint in a region of length 1 with rate `r` has either 0 or 1 breakpoints, and has 1 breakpoint with probability `1-exp(-r)`. (But, I'm not sure: @DomNelson? @jeromekelleher?)

If so, we should set the rate so that `exp(-r) = 1/2`, i.e., `r = log(2)`.

Also note that if it's really a Poisson number, then there's a different equation to get it set to exactly 1/2 (details in the SLiM manual).

grahamgower:
My lazy unthinking brain says to just simulate it and calculate the LD.

apragsdale:
> Also note that if it's really a Poisson number, then there's a different equation to get it set to exactly 1/2 (details in the SLiM manual).

If it's a Poisson number, the probability that the number is odd is always less than 0.5, and only approaches 0.5 in the limit of the rate going to infinity. Though practically it converges quickly, so even with a rate of 2, the even/odd split is ~ 0.51/0.49.

But I think you're right that recombination events are drawn differently in the discrete genome, and they follow something like `1-exp(-r)`. Though from some experimenting, it looks like we get 1 breakpoint with probability `1-exp(-2*r)` instead. I checked this by running a single generation of "dtwf" with a single locus hotspot and no recombination on either side, with two samples. We can then check if neither sample had a recombination event in that one generation, which should be in 1/4 of the replicates if the probability of coinheritance on each side of the hotspot is 1/2. The value of `r` that gave that proportion was `log(sqrt(2))` (dashed red lines in the figure). Not sure what to make of this, if that is what the correct rate _should_ be.

rec_decay

petrelharp:
> If it's a Poisson number, the probability that the number is odd is always less than 0.5, and only approaches 0.5 in the limit of the rate going to infinity. 

Oh, sorry, you're right about that.

> 1 - exp(-2r)

Oh, nice experiment! I got `1 - exp(-r)`, doing something similar, though:
```
r = np.log(2)
n = 10000
ts = msprime.sim_ancestry(n, recombination_rate=r, population_size=1e10, model='dtwf', end_time=2, sequence_length=2)
print("expected proportion of recombinants:",  (1 - np.exp(-r)))
print('observed proportion of recombinants:', np.sum(ts.tables.edges.right == 1) / (2 * n))
# expected proportion of recombinants: 0.5
# observed proportion of recombinants: 0.50095
```

So, I think the recommendation is to set recombination rate to `log(2) = 0.69314718` for the base pairs between chromosomes, with an explanation that in the DTWF, the probability of recombination is `1 - exp(-r)`? This is a teeny bit problematic because when you switch to Hudson you'd really like a recombination rate of 0.5, but I can't imagine this would make a difference in practice? I guess this is implementing the "at least one recombination event creates a crossing over" instead of "odd number of recombination events".

apragsdale:
I see - that makes sense to me! I didn't realize that setting number of samples to 2 is actually simulating 4 samples, so I was dividing by the wrong number of samples. I'm not a fan that the default behavior for number of samples now doubles the number of sampled sequences (`ts = msprime.sim_ancestry(2, ...); ts.num_samples == 4`), but maybe there's a good reason for that? Anyway, not relevant to this PR

Interestingly, it looks to still be valid using the Hudson model:
```
r = np.log(2)
n = 10000
ts = msprime.sim_ancestry(n, recombination_rate=r, population_size=1e10, model='hudson', end_time=1, sequence_length=2)
print("expected proportion of recombinants:",  (1 - np.exp(-r)))
print('observed proportion of recombinants:', np.sum(ts.tables.edges.right == 1) / (2 * n))
# expected proportion of recombinants: 0.5
# observed proportion of recombinants: 0.4986
```
I'll update the PR with all this in mind - thanks for the help @petrelharp!

petrelharp:
> I'm not a fan that the default behavior for number of samples now doubles the number of sampled sequences (ts = msprime.sim_ancestry(2, ...); ts.num_samples == 4), but maybe there's a good reason for that?

Switching the focus to individuals! Since that's what people work/think with usually anyhow. This was discussed elsewhere extensively, anyhow, and is why we're switching from `simulate( )` (which will retain its old behavior) to `sim_ancestry( )`.

thank you!

apragsdale:
> Switching the focus to individuals! Since that's what people work/think with usually anyhow. This was discussed elsewhere extensively, anyhow, and is why we're switching from `simulate( )` (which will retain its old behavior) to `sim_ancestry( )`.
> 
> thank you!

I see - I think I missed those discussions while away for a month or two recently due to a move :( I think I've also been confused because my impression was that `sim_ancestry( )` is essentially replacing `simulate( )` in msprime 1.0, which would be deprecated eventually - but I guess both will continue to be supported? If that's the case, then I don't have as much to worry about I think!

jeromekelleher:
Just to reassure you @apragsdale - ``simulate()`` will be supported indefinitely. Nearly all existing code should continue working in the long term after 1.0. (The exceptions are where we were doing tricky stuff with the recombination maps with num_loci - we had no choice but to break this.)

apragsdale:
Ok, sorry for making all this noise about it after the discussions had all been settled! Like I said, I just got worried yesterday, and turns out I had no reason to be.

Because `simulate()` also supports `discrete_genome` with a discrete genetic map, and if those aspects of the simulation work the same was as here, we can also simulate multiple chromosomes with `simulate()` in the same way. Might be worth a mention in those docs with a Multiple Chromosome header, with just something short and the analogous `similate(...)` options as needed, and then point to this page for the discussion since it's more or less the same.

On Nov 14, 2020 5:01 AM, Jerome Kelleher <notifications@github.com> wrote:

Just to reassure you @apragsdale<https://github.com/apragsdale> - simulate() will be supported indefinitely. Nearly all existing code should continue working in the long term after 1.0. (The exceptions are where we were doing tricky stuff with the recombination maps with num_loci - we had no choice but to break this.)

jeromekelleher:
> Because `simulate()` also supports `discrete_genome` with a discrete genetic map, and if those aspects of the simulation work the same was as here, we can also simulate multiple chromosomes with `simulate()` in the same way. Might be worth a mention in those docs with a Multiple Chromosome header, with just something short and the analogous `similate(...)` options as needed, and then point to this page for the discussion since it's more or less the same.

The current plan is for ``simulate()`` to **not** support the ``discrete_genome`` option in an attempt to make the code simpler (#1297). I've already done a lot of the changes in #1287 

I'd rather move away from the language of a "discrete genetic map" now too, as the recombination map and whether or not we have a discrete genome are fully decoupled.

apragsdale:
> The current plan is for `simulate()` to **not** support the `discrete_genome` option in an attempt to make the code simpler (#1297). I've already done a lot of the changes in #1287

Ah-ha! Ok, I can understand that. In that case, this is probably good to go. I cleaned it up a bit more as well. It could maybe use one more quick read-through to make sure it all reads clearly.

> I'd rather move away from the language of a "discrete genetic map" now too, as the recombination map and whether or not we have a discrete genome are fully decoupled.

That makes sense. Thanks a lot for taking the time to clarify.

jeromekelleher (approved):
This looks great, thanks @apragsdale! Ready to merge after a squash.

jeromekelleher:
> That makes sense. Thanks a lot for taking the time to clarify.

Not at all - thanks for taking the time to try out the new APIs!

apragsdale:
> Not at all - thanks for taking the time to try out the new APIs!

It's really nice with some great new features! Ok - all squashed.

diegovelizo:
This was not immediately clear to me, so I thought I'd add that here in case it's useful for anyone (and also in case I made any mistake and other people can spot it).
When stitching together recombination map files in hapmap format, it's unnecessary to edit both the third and fourth columns. You can either add 69.314 cM to the genetic position, or set the recombination rate to 69314718.06. Either of those columns can be set to all zeroes. If both column are specified, the fourth column seems to take precedence, and the third column is ignored. 

So for instance, these two text files give the same recombination map:
```
data1 = io.StringIO("""\
Chromosome  Position(bp)  Rate(cM/Mb)  Map(cM)
chr10       48232         0       100.002664
chr10       48486         0       100.002705
chr10       50009         0        100.002947
chr10       52147         0       100.003287
chr10       52148         0       169.31800522
""")
rate_map1 = msprime.RateMap.read_hapmap(data1)

print(rate_map1)
┌───────────────────────────────────────────┐
│left   │right  │      mid│   span│     rate│
├───────────────────────────────────────────┤
│0      │48232  │    24116│  48232│  2.1e-05│
│48232  │48486  │    48359│    254│  1.6e-09│
│48486  │50009  │  49247.5│   1523│  1.6e-09│
│50009  │52147  │    51078│   2138│  1.6e-09│
│52147  │52148  │  52147.5│      1│     0.69│
└───────────────────────────────────────────┘
```



```
data2 = io.StringIO("""\
Chromosome  Position(bp)  Rate(cM/Mb)
chr10       48232         0.1614       0
chr10       48486         0.1589       0
chr10       50009         0.159        0
chr10       52147         69314718.06       0
chr10       52148         0.1574       0
""")
rate_map2 = msprime.RateMap.read_hapmap(data2)

print(rate_map2)
┌───────────────────────────────────────────┐
│left   │right  │      mid│   span│     rate│
├───────────────────────────────────────────┤
│0      │48232  │    24116│  48232│  2.1e-05│
│48232  │48486  │    48359│    254│  1.6e-09│
│48486  │50009  │  49247.5│   1523│  1.6e-09│
│50009  │52147  │    51078│   2138│  1.6e-09│
│52147  │52148  │  52147.5│      1│     0.69│
└───────────────────────────────────────────┘
```

PR #1308, opened 2020-11-16
Fixes for compile problems on 32bit i686

jeromekelleher:
Most of these changes are inconsequential I think, except dropping the requirement that size_t is 64 bit. Was there a specific reason for doing this @castedo? It all seems to work fine on my 32 bit machine.

### DIFF
--- a/lib/tests/test_core.c
+++ b/lib/tests/test_core.c
@@ -101,7 +101,9 @@ test_probability_list_select(void)
     }
     {
         double probs[3] = { short_half, DBL_EPSILON / 8.0, 0.5 };
+#ifdef MSP_TEST_EXACT_FLOAT_COMPARISONS
         CU_TEST(1.0 == probs[0] + probs[1] + probs[2]);
+#endif
         CU_ASSERT_EQUAL(0, probability_list_select(0.0, 3, probs))
         CU_ASSERT_EQUAL(1, probability_list_select(short_half, 3, probs))
         CU_ASSERT_EQUAL(2, probability_list_select(0.5, 3, probs))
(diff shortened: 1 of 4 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1308/)

castedo:
One reason was make sure that `1ULL` is 64-bit so that `1LL << (DBL_MANT_DIG - 1)` is the intended really big number. @jeromekelleher I suspect it's OK to assume that `1ULL` must be 64-bit and skip having any compile time check for that.

I'm now double checking some code and re-thinking another possible assumption.

jeromekelleher:
> One reason was make sure that 1ULL is 64-bit 

We can still check that, if you like, with something like ``ULLONG_MAX < UINT64_MAX`` (although, this seems pretty remote, and I think we'd probably get a compiler warning about the overflow, anyway?) That's nothing to do with the size of ``size_t``, though, is it?

castedo:
Actually I was confused about my first comment. Now I'm remembering what I was thinking when I put that compiler check in there:
1) cppreference.com docs that the C standard says long long must be 64-bit
2) I was using this as a cheap quick dirty check to ensure that size_t is at least 64-bit
3) I was checking size_t 64-bit because I wasn't going to bother thinking through the 32-bit scenario because "Who would be still be using 32-bit?!" :laughing: 

I've reviewed the code that I can think maybe is affected. It all looks good. I'm not totally clear why I put this in. I suspect an early version wasn't handling things so robustly and 32-bit might have been an issue. But now it looks all good.

castedo (approved):

lib/tests/test_core.c
```diff
@@ -101,15 +101,19 @@ test_probability_list_select(void)
     }
     {
         double probs[3] = { short_half, DBL_EPSILON / 8.0, 0.5 };
+#ifdef MSP_TEST_EXACT_FLOAT_COMPARISONS
```
castedo, line 104: Didn't you say you didn't want conditional compiles cluttering up the code? :smiling_imp:
(looks fine to me, i'm just teasing you on `#ifdef NDEBUG` comments long ago)

jeromekelleher: Well it was either that or delete the tests, and this is a good sanity check when we have proper maths on board. I guess we could do a custom macro, but comparing floats for exact equality is rarely a good idea anyway, so didn't seem worth it.

jeromekelleher:
Thanks for checking @castedo!

PR #1314, opened 2020-11-25
Updates to mutations API for 1.0

jeromekelleher:
This PR brings together the various changes needed to the mutate/sim_mutations API for the 1.0 release.

Basically, we revert mutate() back to what it looked like as of 0.7.4, and document it as deprecated-but-supported-forever, and rename the ``discrete`` argument to ``discrete_genome``.

This isn't quite finished, so I've marked it as draft for now.

@petrelharp, do you agree with this approach?

(I've also updated the mutation tests to use ``sim_ancestry`` where it makes sense as a dogfooding exercise)

### DIFF
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -1058,7 +1058,7 @@ def _simple_mutate(
         random_generator,
         rate_map.asdict(),
         model=BinaryMutationModel(),
-        discrete_sites=False,
+        discrete_genome=False,
     )
 
 
@@ -1070,7 +1070,7 @@ def mutate(
     keep=False,
     start_time=None,
     end_time=None,
-    discrete=False,
+    discrete_genome=False,
     kept_mutations_before_end_time=False,
 ):
     """
@@ -1095,7 +1095,7 @@ def mutate(
     mutations are simulated. Under the infinite sites mutation model, all new
     mutations generated will occur at distinct positions from each other and
     from any existing mutations (by rejection sampling). Furthermore, if sites
-    are discrete, trying to simulate mutations at time periods that are older
+    are discrete_genome, trying to simulate mutations at time periods that are older
     than mutations kept from the original tree sequence is an error, because
     this would create an extra transition (from the new allele to the old
     one below it) that may be incorrect according to the model of mutation.
@@ -1131,12 +1131,12 @@ def mutate(
         occur. (Default: no restriction.)
     :param float end_time: The maximum time ago at which a mutation can occur
         (Default: no restriction).
-    :param bool discrete: Whether to generate mutations at only integer positions
+    :param bool discrete_genome: Whether to generate mutations at only integer positions
         along the genome.  Default is False, which produces infinite-sites
         mutations at floating-point positions.
     :param bool kept_mutations_before_end_time: Whether to allow mutations to be added
         ancestrally to existing (kept) mutations. This flag has no effect
-        if either keep or discrete are False.
+        if either keep or discrete_genome are False.
     :return: The :class:`tskit.TreeSequence` object resulting from overlaying
         mutations on the input tree sequence.
     :rtype: :class:`tskit.TreeSequence`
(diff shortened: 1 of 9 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1314/)

jeromekelleher:
@petrelharp, mind taking a quick look here please? I'd like to get this shipped away soon.

msprime/mutations.py
```diff
@@ -1095,7 +1177,7 @@ def mutate(
     mutations are simulated. Under the infinite sites mutation model, all new
     mutations generated will occur at distinct positions from each other and
     from any existing mutations (by rejection sampling). Furthermore, if sites
-    are discrete, trying to simulate mutations at time periods that are older
+    are discrete_genome, trying to simulate mutations at time periods that are older
```
mufernando, line 1180: oops, unwanted replace of the word discrete here

mufernando (approved):
LGTM!

mufernando:
We may have to reconsider the statistical tests so that they work with discrete sites. I guess the approximations will only work for large chromosomes when sites are discrete.

jeromekelleher:
> We may have to reconsider the statistical tests so that they work with discrete sites. I guess the approximations will only work for large chromosomes when sites are discrete.

Ah, I hadn't considered that! Should I change them back to being on a continuous genome? Thanks for the review btw.

mufernando:
I think I'm wrong heh. We already had these tests working for discrete genome before, I think because stacking is allowed they should work fine.

tests/test_mutations.py
```diff
@@ -1007,22 +1085,36 @@ def sign_tst(self, diffs):
 
 class TestMutationStatistics(StatisticalTestMixin):
     def verify_model(self, model, verify_roots=True, state_independent=False):
-        ots = msprime.simulate(10, random_seed=5, recombination_rate=0.05, length=20)
+        # NOTE: this test is pretty brittle and some will fail if
```
mufernando, line 1088: I was thinking about this in my previous comment. Why is this brittle?

jeromekelleher: It's very sensitive to the random seed - sometimes it works, but more often than not it doesn't. So, even if it works on the current random seed, it will fail mysteriously at some point down the line when GSL changes the RNG a bit.

TBH, I wonder if these tests should be unit tests at all, and they would be more appropriate as part of the verification suite.

jeromekelleher:
Bumping this again @petrelharp - any chance you can take a quick look please?

msprime/mutations.py
```diff
@@ -1070,8 +1070,90 @@ def mutate(
...
+        however.
+
+    If a random seed is specified, this is used to seed the random number
+    generator. If the same seed is specified and all other parameters are equal
+    then the same mutations will be generated. If no random seed is specified
+    then one is generated automatically.
+
+    If the ``model`` parameter is specified, this determines the model under
+    which mutations are generated. Currently only the :class:`.InfiniteSites`
+    mutation model is supported. This parameter is useful if you wish to obtain
```
petrelharp, line 1094: This second sentence should be deleted (the rest is ok).

msprime/mutations.py
```diff
@@ -1171,15 +1253,24 @@ def mutate(
         end_time = float(end_time)
     if start_time > end_time:
         raise ValueError("start_time must be <= end_time")
-    keep = bool(keep)
-    discrete = bool(discrete)
-    kept_mutations_before_end_time = bool(kept_mutations_before_end_time)
+    if keep is None:
```
petrelharp, line 1256: What's the reason to have the default be `None` instead of `False`? It's easier to tell from the signature what the default is if it is `False`.

tests/test_mutations.py
petrelharp: Hm. Moving them to verification makes sense - as long as we make sure to run verification if we change the mutation code in the future.

But, we should be able to make these not brittle... oh! by switching from a p-value condition to a confidence interval condition!

petrelharp: https://github.com/tskit-dev/msprime/issues/1334

petrelharp (approved):
LGTM, after deleting that sentence! Sorry about the delay.

msprime/mutations.py
jeromekelleher: It's a general principle: unless you are absolutely sure that under all possible combinations of input paraters in the future the default will be False, it's better to use a sentinel value so you can always tell whether the user specifically asked for False, or if it was the default.

PR #1316, opened 2020-11-25
Demography metadata

jeromekelleher:
Changes needed to make demography useable for 1.0.0a1.

WIP

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -22,7 +22,6 @@ Module responsible for defining and debugging demographic models.
 import collections
 import inspect
 import io
-import json
 import logging
 import math
 import sys
@@ -30,6 +29,7 @@ import warnings
 
 import attr
 import numpy as np
+import tskit
 
 from . import ancestry
 
@@ -122,9 +122,41 @@ class Demography:
         Insert population definitions for this demography into the specified
         set of tables.
         """
+        metadata_schema = tskit.MetadataSchema(
+            {
+                "codec": "json",
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "description": {"type": ["string", "null"]},
+                },
+                # The name and description fields are always filled out by
+                # msprime, so we tell downstream tools this by making them
+                # "required" by the schema.
+                "required": ["name", "description"],
+                "additionalProperties": True,
+            }
+        )
         assert len(tables.populations) == 0
+        tables.populations.metadata_schema = metadata_schema
         for population in self.populations:
-            tables.populations.add_row(metadata=population.temporary_hack_for_metadata)
+            metadata = {
+                "name": population.name,
+                "description": population.description,
+            }
+            if population.extra_metadata is not None:
+                intersection = set(population.extra_metadata.keys()) & set(
+                    metadata.keys()
+                )
+                if len(intersection) > 0:
+                    printed_list = list(sorted(intersection))
+                    raise ValueError(
+                        f"Cannot set standard metadata key(s) {printed_list} "
+                        "using extra_metadata. Please set using the corresponding "
+                        "property of the Population class."
+                    )
+                metadata.update(population.extra_metadata)
+            tables.populations.add_row(metadata=metadata)
 
     def sample(self, *args, **kwargs):
         """
@@ -465,18 +497,7 @@ class Population:
     growth_rate = attr.ib(default=0.0)
     name = attr.ib(default=None)
     description = attr.ib(default=None)
-
-    # TODO add extra metadata when we bring in tskit 0.3. We should always
-    # have the name and description, and other metadata items can optionally
-    # be declared and added. For now, let's keep the metadata style introduced
-    # in 0.7.4 working.
-    temporary_hack_for_metadata = attr.ib(default=None)
-
-    def temporary_hack_for_encoding_old_style_metadata(self):
-        encoded_metadata = b""
-        if self.temporary_hack_for_metadata is not None:
-            encoded_metadata = json.dumps(self.temporary_hack_for_metadata).encode()
-        return encoded_metadata
+    extra_metadata = attr.ib(factory=dict)
 
     def asdict(self):
         return attr.asdict(self)
@@ -490,11 +511,11 @@ class Population:
         return Population(
             initial_size=pop_config.initial_size,
             growth_rate=pop_config.growth_rate,
-            temporary_hack_for_metadata=pop_config.metadata,
+            extra_metadata=pop_config.metadata,
         )
 
     def validate(self):
-        # TODO more checks, and put in population ID/names
+        # TODO more checks
         if self.initial_size < 0:
             raise ValueError("Negative population size")
         if self.name is None:
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -227,6 +227,8 @@ class Demography:
         else:
             for pop_config in population_configurations:
                 demography.populations.append(Population.from_old_style(pop_config))
+        for j, population in enumerate(demography.populations):
+            population.name = f"pop_{j}"
         if migration_matrix is None:
             migration_matrix = np.zeros(
                 (demography.num_populations, demography.num_populations)
@@ -237,7 +239,6 @@ class Demography:
         return demography
 
     # TODO give this a better name and document it.
-    # What about "isolated" as it gives an easy way of describing isolated pops?
     def simple_model(initial_size=1, growth_rate=None):
         """
         Returns a simple single-population model.
@@ -326,9 +327,7 @@ class Demography:
 
         model = Demography()
         model.populations = [
-            # TODO add names/metadata here.
-            Population(initial_size=Ne)
-            for _ in range(num_populations)
+            Population(initial_size=Ne, name=f"pop_{j}") for j in range(num_populations)
         ]
         model.migration_matrix = np.zeros((num_populations, num_populations))
         if num_populations > 1:
@@ -498,6 +497,10 @@ class Population:
         # TODO more checks, and put in population ID/names
         if self.initial_size < 0:
             raise ValueError("Negative population size")
+        if self.name is None:
+            raise ValueError("A population name must be set.")
+        if not self.name.isidentifier():
+            raise ValueError("A population name must be a valid Python identifier")
 
 
 # This was lifted out of older code as-is. No point in updating it
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -22,6 +22,7 @@ Module responsible for defining and running ancestry simulations.
 import collections.abc
 import copy
 import inspect
+import json
 import logging
 import math
 import struct
@@ -235,6 +236,7 @@ def _demography_factory(
 
 
 def _build_initial_tables(*, sequence_length, samples, ploidy, demography, pedigree):
+    # NOTE: this is only used in the simulate() codepath.
     tables = tskit.TableCollection(sequence_length)
 
     if pedigree is None:
@@ -261,9 +263,14 @@ def _build_initial_tables(*, sequence_length, samples, ploidy, demography, pedig
             for _ in range(ploidy):
                 tables.nodes.add_row(node_flags, time, population=0, individual=ind_id)
 
+    # This is for the simulate() code path so we don't add metadata schemas
+    # and insert the user metadata in directly as encoded JSON, as before.
     for population in demography.populations:
-        md = population.temporary_hack_for_encoding_old_style_metadata()
-        tables.populations.add_row(metadata=md)
+        encoded_metadata = b""
+        if population.extra_metadata is not None:
+            encoded_metadata = json.dumps(population.extra_metadata).encode()
+        tables.populations.add_row(encoded_metadata)
+
     return tables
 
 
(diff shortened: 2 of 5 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/msprime-docs/PR-1316/)

benjeffery:
Schema looks great, `description` is effectively optional, but always present.

jeromekelleher:
Here's a proposal for how we deal with metadata in msprime simulated *populations*. We're not talking about top-level metadata or any other entities yet, just populations.

For tree sequences produced by msprime.simulate(), the metadata will be the byte encoded JSON of the PopulationConfiguration object and *no* schema is set. That means that we don't break any existing code, and users can continue to use., e.g., ``md = json.loads(population.metadata)``.

For tree sequences produced by msprime.sim_ancestry(), we specify a simple and uncontroversial (hopefully) schema which requires a ``name`` and ``description`` attribute (but description can be None), which are filled in from the ``Population.name`` and ``Population.description`` attributes. Population.name is required to be a unique identifier to populations (e.g. "CEU").

Users are also allows to provide their own metadata that can be associated with populations. This is an arbitrary Python dictionary, which is then stored in the ``user`` key of the metadata.

Thus, *all* simulations produced by ``msprime.sim_ancestry`` will be guaranteed to have metadata that contains the population ``name``.  Here's an example test case, which hopefully illustrates the idea:
```python
        demography = msprime.Demography.simple_model(10)
        md = {"x": "y", "z": "z", "abc": 1234}
        demography.populations[0].user_metadata = md
        ts = msprime.sim_ancestry(2, demography=demography, random_seed=2)
        assert ts.num_populations == 1
        metadata = ts.population(0).metadata
        assert metadata["name"] == "pop_0" # automatically set to this unless overridden
        assert metadata["description"] is None
        assert metadata["user"] == md
```

(Don't pay too much attention to the first 3 lines, as the way that we create Demography objects is still under development and this is pretty clunky.)

This is not a Population schema that would be appropriate for other programs, but we might consider reserving things like ``name`` and ``description`` in the top-level of the namespace as something we could standardise on.

This is an important decision which will affect upstream and downstream tools, so I'm hoping to get some thoughts from a bunch of people here! Pinging @petrelharp @benjeffery @hyanwong @molpopgen @benjeffery @grahamgower @apragsdale

grahamgower:
Sounds like a reasonable proposal. So the suggestion for downstream applications would be to put application specific stuff in `user/<appname>` and leave the rest of the `user` namespace to the actual user?

> This is not a Population schema that would be appropriate for other programs, but we might consider reserving things like `name` and `description` in the top-level of the namespace as something we could standardise on.

Yes please!

jeromekelleher:
> Sounds like a reasonable proposal. So the suggestion for downstream applications would be to put application specific stuff in user/<appname> and leave the rest of the user namespace to the actual user?

Actually, no, I don't think we should do namespacing by application. I think downstream applications should add extra keys to the top-level dictionary for generally useful stuff (i.e., for things that *further* downstream applications might use). 

Which actually makes me wonder whether we should bother with the ``user`` key at all, and should instead update the top-level dictionary with the ``user_metdata`` field. If the user choses to overwrite the ``name`` field that's provided by default we can warn them, but we should probably let them.

I think we should be permissive when it comes to metadata, and should focus on what produces the easiest and nicest code for *end users*. I think we got too caught up in preventing rogue applications from clobbering existing metadata, but maybe we should follow a more Pythonic philosophy of making it easy for end users to write good code, rather than trying to prevent them writing bad code. To me ``population.metadata["msprime"]["some_useful_data"]`` is much worse than ``population.metadata["some_useful_data"]``

So yes, I think I've convinced myself now that we *shouldn't* have a ``user`` key, and that we should let up and downstream applications write whatever they want into the population metadata. (Contrary to the current code).

For reference, here's the schema I propose now:
```json 
           {
                "codec": "json",
                "type": "object",
                "properties": {
                    "name": {"type": "string"},
                    "description": {"type": ["string", "null"]},                
                },
                "required": ["name", "description"],
                "additionalProperties": True,
            }
```

hyanwong:
Could we have a convention that non-user data should have a key that e.g. starts with an underscore? That would roughly mirror the use of underscores in python variable names. Alternatively we could end the key with an underscore, or perhaps use an initial capital letter to distinguish non-user keys. Either way, it would just be a convention.

jeromekelleher:
Interesting idea... What are we trying to achieve by this though? if a user goes to the trouble of adding metadata into their simulations, they don't really make any distinction between "user data" and something like the population name. It would not improve their code in any way if some of the metadata keys that they use start with underscores and some don't - it would just be confusing.

jeromekelleher:
I really think we should focus on making end-user code as easy to read and follow as possible and not worry so much about clobbering. We can add some tooling to tskit which enables metadata updates that automatically warns on clobber. That's enough - it's really not that big a problem.

hyanwong:
I think I mean that publicly released software that adds metadata would be expected to use initial capital letters: that shifts the responsibility onto software devs to check they don't clobber each others' stuff. End users (like me) can then happily dump stuff into lowercase keys, without worrying that they might later run some software on the TS that accidentally clobbers the data that they carefully stuck there. It's simply an alternative to having a 'user' key.

As I said, it's only a helpful convention, not a strict rule. But it would be very easy to implement - it would just require a line of documentation.

jeromekelleher:
Here's my updated proposal: call the field ``Population.extra_metadata`` and just drop it right in to the top level:

```python
    def test_extra_metadata(self):
        demography = msprime.Demography.simple_model(10)
        md = {"x": "y", "z": "z", "abc": 1234}
        demography.populations[0].extra_metadata = md
        ts = msprime.sim_ancestry(2, demography=demography, random_seed=2)
        assert ts.num_populations == 1
        metadata = ts.population(0).metadata
        assert metadata["name"] == "pop_0"
        assert metadata["description"] is None
        assert len(metadata) == len(md) + 2
        for key, value in md.items():
            assert metadata[key] == value
```

I'll probably update with a warning if we try to overwrite the existing fields.

benjeffery:
I'm wary of solving problems we don't yet have with mechanisms to divide metadata in to categories such as capitals/underscores or by tool. I think this will discourage re-use of metadata and end up with downstream tools having to check a load of tool names or word formats. I don't think users need to worry as long as tools warn if they are overwriting keys, and we will encourage this by making it easy to use the tskit `update_metadata` method.

So in summary also think putting everything at the root of the metadata makes sense for now.

hyanwong:
> I'm wary of solving problems we don't yet have with mechanisms to divide metadata in to categories such as capitals/underscores or by tool. I think this will discourage re-use of metadata and end up with downstream tools having to check a load of tool names or word formats. 

Sure. You're the metadata guru, so whatever you say, I reckon.

> I don't think users need to worry as long as tools warn if they are overwriting keys, and we will encourage this by making it easy to use the tskit [`update_metadata`](https://github.com/tskit-dev/tskit/issues/1044) method.

As long as the tools warn *before* they do heavy lifting. I have some scripts that take e.g. 2 days to process a tree sequence. It would be pretty painful if an equivalent tool took 2 days to run, then aborted at the end because it didn't want to overwrite some metadata (I have hit this sort of issue many times over).

hyanwong:
p.s. @benjeffery - maybe I'm unusual, or misguided or (more probably) both, but if there's no explicit place in the metadata that's recommended for me to store my own user data, than I feel it might actively *discourage* me from using the metadata area to store my own stuff. And unless I had read this discussion, I'd probably revert to using csv files instead, because I'd be worried that I was sorting stuff in the wrong place, or that whatever keys I chose would conflict with some other tool. I think that's why I keep on bringing up the issue: there's something about the freedom that you propose that would make me, with my naive user hat on and hoping for some explicit guidanbce, steer away from actually using metadata.

A line in the docs saying something like "although any keys can be used, the 'user' key (or whether) is guaranteed not to be used in the official tskit toolchain" would, I suspect, go a long way to assuaging my (misplaced) worries in this regard.

I mean, now that you've reassured me, it's not an issue for me personally. But it would be my line of thinking if I had come to this from scratch, and so other users might be put off in the same way?

benjeffery:
> Sure. You're the metadata guru, so whatever you say, I reckon.

I can be wrong! Often! 

> As long as the tools warn _before_ they do heavy lifting. I have some scripts that take e.g. 2 days to process a tree sequence. It would be pretty painful if an equivalent tool took 2 days to run, then aborted at the end because it didn't want to overwrite some metadata (I have hit this sort of issue many times over).

Yeah I hate when that happens with long-running stuff  - one of the reasons I think a warning should be the default. At least then you can patch up the metadata by hand if you wanted to retain the metadata that has been overwritten, without re-running your job.

> I feel it might actively discourage me from using the metadata area

This is a good point - we should be sensitive to this in the docs.

jeromekelleher:
This is ready for review now I think. Parts of the Demography API are still pretty clunky, but I think the fundamental metadata question is settled here, which we should try to merge.

msprime/demography.py
```diff
@@ -122,9 +122,36 @@ def insert_populations(self, tables):
...
+                "description": population.description,
+            }
+            if population.extra_metadata is not None:
+                intersection = set(population.extra_metadata.keys()) & set(
+                    metadata.keys()
+                )
+                if len(intersection) > 0:
+                    warnings.warn(
+                        f"Overwriting standard metadata keys: {list(intersection)}; "
+                        "this may lead to compatibility problems."
```
petrelharp, line 151: Is this warning right? Isn't it "existing" keys, not necessarily "standard" keys? Also, "compatibility" problems is pretty vague and alarming - I'd say either say concretely what the problems might be, or else just warn "overwriting existing keys: { }".

petrelharp:
Sounds good! I agree with the conclusion.

msprime/demography.py
jeromekelleher: Well, we do hope to standardise these (``name`` and ``description``) - and they are definitely standard metadata for msprime. I take your point about the warning, but it's hard to pinpoint what might happen downstream if you change the name of the population to something else. Probably nothing, but then why would you want to overwrite the name or description when you can set these as properties on the Population object in the first place. So - maybe this should be an error then, since you can always set the ``name`` and ``description`` via the Population class and it's probably a mistake if you're overwriting these.

@grahamgower, any thoughts here? I'm imagining stdpopsim as being one of the key users of this metadata API. Should this be a warning or an error?

grahamgower: Probably it should be an error. I also can't see any reason to modify these, so it would most likely be done only accidentally.

grahamgower (approved):

msprime/demography.py
hyanwong: It might be because of my comment that it's a pain if a long-running script errors out after (say) 2 days or running, just before saving the final data files. I guess that's good reason to make these sorts of changes before you enter long-running loops, but that's not always the natural thing to do. 

I've definitely had problems with long running scripts failing at the very end due to things that were not vital to do, but which I would like to have been warned about. But I probably just have bad coding practices.

jeromekelleher: Right, but I don't see how this would happen for an msprime simulation. We're not talking about general rules here, just what to do in this specific situtation.

hyanwong: Ah, good point. Seems sensible to error out then.

jeromekelleher: OK, I'll change this to an error and then I think we're good to merge.

benjeffery: I agree - error in this case.

msprime/demography.py
```diff
@@ -122,9 +122,36 @@ def insert_populations(self, tables):
...
         """
+        metadata_schema = tskit.MetadataSchema(
+            {
+                "codec": "json",
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "description": {"type": ["string", "null"]},
+                },
+                "required": ["name", "description"],
```
benjeffery, line 133: Given the discussion around standardised schemas being flexible (and we intend this schema to be replaced with a reference to the standard one) - what's the thinking around `description` being required. It could be absent instead of `None`? We don't want all standardised keys to be required, so why this one?

benjeffery (approved):
LGTM - one question about `required`

msprime/demography.py
jeromekelleher: Yes, I agree. Initially I thought it would be nice if users of msprime could be sure that the ``description`` key was always present, even if equal to None. But, we can do that anyway and don't need to state it in the schema. So, I guess there's no real point in making ``name`` required here either, come to think of it.

benjeffery: Having thought about this further - any metadata that we feel strongly enough that it be `required` in the standard schema probably should be in the data model and not metadata, right?

jeromekelleher: Hah, actually no, I've had a (re)change of heart there. By making these fields required in the schema, we're passing on valuable information to downstream software: you're guaranteed that these values are defined in each bit of metadata. This isn't a general thing, just that in msprime's case we will *always* have these fields. The "standard" base schemas shouldn't make anything mandatory, but programs can choose to notify client code that the fields will always be present, if they wish.

benjeffery: Gotcha - msprime will modify the base schema to indicate what it always provides.

PR #1335, opened 2020-12-13
Improvements to RateMap

hyanwong:
This is the first part of what was previously all stuck into a single PR (#1185). This contains the changes to the underlying RateMap class, addressing review comments by @jeromekelleher - in particular we now have `get_cumulative_mass()` and `get_rate()` instead of `cumulative_at()` and `at()` respectively.

There was a question about whether we should allow a "name" (e.g. "chr1") to be associated with a RateMap as an (optional?) attribute. This would allow nicer outputting of a ratemap, and make it easier for users to check whether that have the right map after importing from a file. I have not added any such functionality yet: it could be appended as an optional attribute only if we make a RateMap using read_hapmap().

### DIFF
--- a/msprime/intervals.py
+++ b/msprime/intervals.py
@@ -33,35 +33,134 @@ class RateMap:
     """
     A class mapping a numeric value to a set of adjacent intervals along the
     genome.
+
+    :param list position: A list of :math:`n+1` positions, starting at 0, and ending
+        in the sequence length over which the RateMap will apply.
+    :param list rate: A list of :math:`n` positive rates that apply between each
+        position.
+    :param float start_position: When returning a mean_rate over the genome, ignore the
+        part of the map from zero up to ``start_position``, as long as the rate is 0 in
+        that region. If ``None`` (default), the ``start_position`` is 0.
+    :param float end_position: When returning a mean_rate over the genome, ignore the
+        part of the map from ``end_position`` onwards, as long as the rate is 0 in that
+        region. If ``None`` (default), the ``end_position`` is the end of the map.
     """
 
-    def __init__(self, position, rate, map_start=0):
-        # We take copies to make sure there are no unintended consequences
-        # later when a user modifies the arrays in this map.
-        self.position = np.array(position, dtype=float, copy=True)
-        self.rate = np.array(rate, dtype=float, copy=True)
-        self.map_start = map_start
-        size = len(self.rate)
-        if size < 1:
+    def __init__(
+        self,
+        position,
+        rate,
+        *,
+        start_position=None,
+        end_position=None,
+    ):
+        # Making the arrays read-only guarantees rate and cumulative mass stay in sync
+        self._position = np.array(position, dtype=float)
+        self._position.flags.writeable = False
+        size = len(self._position)
+        if size < 2:
             raise ValueError("Must have at least two positions")
-        if self.position.shape[0] != size + 1:
-            raise ValueError(
-                "Rate array must have one less entry than the position array."
-            )
-        if self.position[0] != 0:
+        if self._position[0] != 0:
             raise ValueError("First position must be zero")
-        if np.any(self.position[:1] >= self.position[1:]):
-            raise ValueError("Position values must be in increasing order")
-        if np.any(self.rate < 0):
-            raise ValueError("Rates must be non-negative")
+        if np.any(np.diff(self._position) <= 0):
+            bad_pos = np.where(np.diff(self._position) <= 0)[0] + 1
+            raise ValueError(
+                f"Position values not strictly increasing at indexes {bad_pos}"
+            )
 
-        self.cumulative = np.insert(np.cumsum(np.diff(self.position) * self.rate), 0, 0)
+        self._rate = np.array(rate, dtype=float)
+        self._rate.flags.writeable = False
+        if len(self._rate) != size - 1:
+            raise ValueError(
+                "Rate array must have one less entry than the position array"
+            )
+        if np.any(self._rate < 0):
+            bad_rates = np.where(self._rate < 0)[0]
+            raise ValueError(f"Rate values negative at indexes {bad_rates}")
 
-    def __len__(self):
+        self._cumulative_mass = np.insert(
+            np.cumsum(np.diff(self._position) * self._rate), 0, 0
+        )
+        self._cumulative_mass.flags.writeable = False
+
+        if start_position is None:
+            start_position = 0
+        elif self.get_cumulative_mass(start_position) > 0:
+            raise ValueError("Rates before the start_position must all be zero")
+        self._start_position = start_position
+
+        if end_position is None:
+            end_position = self.sequence_length
+        elif self.get_cumulative_mass(end_position) < self.total_mass:
+            raise ValueError("Rates after the end_position must all be zero")
+        self._end_position = end_position
+
+        # Make all of the internal arrays read-only, so they can't get out of sync
+
+    def __len__(self):  # TODO - remove this, as the length of a ratemap is not obvious
         return len(self.rate)
+
+    @property
+    def position(self):
+        """
+        The positions between which a constant rate applies. There will be
+        one more position than the number of rates, with the first position being 0
+        and the last being the :attr:`sequence_length`.
+        """
+        return self._position
+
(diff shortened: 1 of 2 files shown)


grahamgower (approved):
LGTM!

grahamgower:
> There was a question about whether we should allow a "name" (e.g. "chr1") to be associated with a RateMap as an (optional?) attribute. This would allow nicer outputting of a ratemap, and make it easier for users to check whether that have the right map after importing from a file. I have not added any such functionality yet: it could be appended as an optional attribute only if we make a RateMap using read_hapmap().

How about  just making the `write_hapmap()` function accept a `chromosome_name` parameter?

hyanwong:
> How about just making the `write_hapmap()` function accept a `chromosome_name` parameter?

Yes, now you mention it I think that's what @jeromekelleher suggested. It's certainly a start. Reading it during read_hapmap() seems sensible though, if nothing else to check that there aren't multiple chromosomes in there? So I guess saving it would seem reasonable too, although maybe we don't want to support that?

msprime/intervals.py
```diff
@@ -33,83 +33,232 @@ class RateMap:
...
-        if size < 1:
+    def __init__(
+        self,
+        position,
+        rate,
+        *,
+        start_position=None,
+        end_position=None,
+    ):
+        # Taking copies ensures there are no unintended consequences later if a user
```
jeromekelleher, line 57: I'm not sure we need to do this any more - we're going to some lengths to make sure the user can't write into the arrays, right? Numpy should manage the read-only view semantics for us - but we should set this flag immediately so this is clear.

msprime/intervals.py
```diff
@@ -33,83 +33,232 @@ class RateMap:
...
+        # modifies the input arrays.
+
+        # Save positions into a read-only array
+        self._position = np.array(position, dtype=float, copy=True)
+        size = len(self.position)
```
jeromekelleher, line 62: I find it confusing (and potentially error-prone; you'd normally assume the constructor as competed successfully when writing a property accessor) to use both the private variable and the public property within the constructor. I think it'd be most appropriate to just use ``self._position``, etc within the constructor

msprime/intervals.py
```diff
@@ -33,83 +33,232 @@ class RateMap:
...
-    def size(self):
+    def size(self):  # TODO - should we remove this, for the same reason as for __len__
         return self.rate.shape[0]
 
     @property
     def mean_rate(self):
         """
-        Return the weighted mean across all windows of the entire map.
+        The mean rate over this map, from :attr:`start_position` to
+        :attr:`end_position`, weighed by the span covered by each rate.
```
jeromekelleher, line 195: typo: weighted

msprime/intervals.py
```diff
@@ -33,83 +33,232 @@ class RateMap:
...
+        beyond the end  of the map will return the final rate value.
+
+        :param numpy.ndarray x: The positions for which to return values.
+
+        :return: An array of rates, the same length as ``x``.
+        :rtype: numpy.ndarray(dtype=np.float64)
+        """
+        loc = np.searchsorted(self.position, x, side="right") - 1
+        if np.any(loc < 0):
+            raise ValueError("No rate exists for positions below the minimum map pos")
```
jeromekelleher, line 213: Took me a minute to figure out what this was doing, this would be more obvious:
```python
loc = np.minimum(loc, len(self.rate) - 1)
return self.rate[loc]
```

msprime/intervals.py
```diff
@@ -33,83 +33,232 @@ class RateMap:
     """
     A class mapping a numeric value to a set of adjacent intervals along the
     genome.
+
+    :param list position: A list of :math:`n+1` positions, starting at 0, and ending
+        in the sequence length over which the RateMap will apply.
+    :param list rate: A list of :math:`n` positive rates that apply between each
+        position.
+    :param float start_position: When returning a mean_rate over the genome, ignore the
```
jeromekelleher, line 41: I wonder if there's much point in these start_position and end_position values, actually. Since we only use them in mean_rate calculations, wouldn't it be simpler to add an ``exclude_telomeres=True`` option there, which would disregard any flanking zero-rate regions?

It's not really helping the user at the moment, since they have to decide where the start and end positions of the map are, presumably doing the same numpy calculations as we'd do to find them.

jeromekelleher (commented):
This looks great, thanks @hyanwong.

My main comment would be, let's get rid of start_position and end_position if we can. @grahamgower, what do you think here? Would adding an ``exclude_telomeres=True`` option to ``mean_rate`` (and turning it into a function) do the trick for us upstream in stdpopsim?

msprime/intervals.py
hyanwong: Good point. Can probably remove

msprime/intervals.py
hyanwong: Yes, I wasn't sure which to use. I'll switch to `self._XXX`

msprime/intervals.py
hyanwong: They get passed in when reading a hapmap, so they do automatically get calculated/set for most users.

jeromekelleher: Excluding hapmap...

jeromekelleher: I mean, we compute the centromeres when reading hapmap, right? What happens when we start using some other (non dreadful) interchange format for recombination maps? Duplicate the code that's in hapmap?

grahamgower:
> My main comment would be, let's get rid of start_position and end_position if we can. @grahamgower, what do you think here? Would adding an `exclude_telomeres=True` option to `mean_rate` (and turning it into a function) do the trick for us upstream in stdpopsim?

Yeah, probably. I wonder why we didn't do that in the first place? Would this option just exclude the first and last positions with zero rates, or all rate entries that are zero at each end util a non-zero rate is found?

hyanwong:
I suggested something like this in the first PR, but I think we didn't do this because it makes it impossible to explicitly set a zero-rate region at the start.

The general idea was that we should remember if we magically inserted the 0-rate region without letting the user know, so that we can remove it as appropriate e.g. when reporting rates or when outputting the RateMap again. This requires some sort of state to be set in the RateMap to mark the "automatically included" region. I can't see how we can get around storing some state while still respecting the originally-input data.

msprime/intervals.py
hyanwong: Should I leave soemthing like `self._position = np.array(position, dtype=float)` but just remove `copy=True`, which would allow lists or numpy int arrays to be passed in as well as numpy float arrays,?

msprime/intervals.py
hyanwong: I don't think we "compute" any telomeres in HapMap files. We just allow them to start at a position > 0 and end at a position < sequence_length. Internally we store these as zero rate regions, which seems reasonable, I guess. It also allows us to "slice" a map while not altering the coordinate space (i.e. "trim=False"), and replace the sliced-out regions with zero rates, while still maintaining the mean rate in the region we keep. So it seems quite handy actually.

grahamgower: I also prefer the start_position/end_position, because this is explicit about how the chromosome coordinates differ from the rate map's coordinates. I can imagine these values being useful to rate map consumers, even if they are (mostly) unused internally. E.g. https://github.com/popsim-consortium/stdpopsim/issues/701 (although we have other issues in stdpopsim too, as our chromosome is actually shorter than the recombination map).

hyanwong: You could use `slice(trim=True)` to get down to the size of the recombination map. Which makes me thing, actually, should the `start` and `end` of slice default to the rate_map `.start_position` and `.end_position` respectively, rather than `0` and `sequence_length` as they do now? At the moment if we call slice with no parameters, it will reset the start and end positions, which seems a bit counterintuitive.

jeromekelleher (approved):
Looks good! I'm not convinced about the start/end positions, but let's merge this much. Thanks @hyanwong, this is a big improvement!

hyanwong:
I'm not sure why this is suddenly failing with "ValueError: signal only works in main thread" in `test_cli.py ` - clearly nothing to do with this PR, and it works on OS X w/ Python 3.8 and 3.9 for me

PR #1337, opened 2020-12-15
Remove __len__ and size() from RateMap

hyanwong:
As [suggested](https://github.com/tskit-dev/msprime/pull/1185#issuecomment-737279602) by @grahamgower, because it is unclear if we mean the number of rates or the number of positions. This will force users to do len(rate_map.rate) or len(rate_map.position) as appropriate.

I'm assuming that len(ratemap) and ratemap.size() were never previously released - if so this should not affect backwards compatibility.


jeromekelleher (approved):
LGTM

grahamgower (approved):
:+1:

Issue #1334, opened 2020-12-12
switch statistical tests of mutation to verification.py


petrelharp:
The tests that our mutation model are working are statistical, perhaps unavoidably, [and brittle](https://github.com/tskit-dev/msprime/pull/1314/files#r534381142). This is a consquence of using p-values - multiple testing and so forth. And there's no good way around this, really, since: suppose we do N tests, then setting the p-value threshold at 0.01/N should result in a fail only 1% of the time... but, suppose we change a bit of code that only affects 5 of those tests. Now our p-value threshold is far too small, since really we'd want to set it at 0.01/5 (say). Also, even a 1% failure is going to get triggered too often.

But! A much better test would be to ask if the statistics say the deviation from expected is no more than something small. That'd mean replacing e.g. [this bit](https://github.com/tskit-dev/msprime/blob/5ca4eb490d12203e0ec791f655ddeda2fae3e254/tests/test_mutations.py#L986):
```
            chisq = stats.chisquare(obs[not_zeros], exp[not_zeros])
            assert chisq.pvalue > p_th
```
with something like
```
            assert np.max(np.abs(obs[not_zeros] - exp[not_zeros])/exp[not_zeros] < 0.01)
```
to say that none of the categories are off by more than 1%. If a test fails, we would just need to increase the sample size (e.g., do a longer sequence).

petrelharp:
I've not been able to get this to work with a small enough tolerance and reasonable run times. I think that (a) these tests should go into verification, and (b) verification should not just rely on relative error being below some threshold, but should plot relative error as a function of sample size (number of mutations), and show that it's going down as expected.

jeromekelleher:
That sounds much more useful @petrelharp. In practise a whole bunch of problems will only show up if we run the verification checks, so I'm relaxed about this not getting run automatically as part of CI. The unit tests are pretty strong, and I run the verification tests before every release.

PR #1339, opened 2020-12-16
adjust p-thresholds

petrelharp:
This makes the 'fragile' tests from #1334 no longer fragile, by decreasing the p-value threshold. As discussed there, I think these tests should go into verification, but this should do for now.


jeromekelleher (approved):
LGTM!

Issue #1072, opened 2020-06-15
parse_species_tree should create new populations for internal nodes


jeromekelleher:
Since #1069, empty inactive populations cost nothing, so we should not use them sparingly. As such, it makes no sense to reuse population IDs for the internal nodes in the species trees that we read in. If these internal nodes have names, we can use them. Otherwise, I guess we call them ``ancestors_species_1``, etc.

PR #1340, opened 2020-12-16
Use recent features in species tree parsing code.

jeromekelleher:
See #958 and #1072. Also tidy up some documentation and metadata handling.

### DIFF
--- a/docs/demography.rst
+++ b/docs/demography.rst
@@ -47,12 +47,6 @@ API Documentation
 .. autoclass:: msprime.Demography
     :members:
 
-
-.. todo:: These functions should be static methods of Demography.
-
-.. autofunction:: msprime.parse_species_tree()
-.. autofunction:: msprime.parse_starbeast()
-
 ***********
 Populations
 ***********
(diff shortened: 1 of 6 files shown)


jeromekelleher:
I think this is mostly ready to go. I've updated the species tree parsing code so that it hooks in nicely to the new Demography class, and also changed the internal semantics so that we allocate one population per species tree node (there's no longer a cost to lots of populations in msprime, so we might as well). The last thing I did was change the Ne parameter to ``initial_size`` and changed it so that it can take either a single number or a mapping from species name to size value. I also added a growth_rate argument for completeness.

@mmatschiner, this does change the interface a bit, but I think it's all an improvement and we haven't actually released a version of msprime that includes the old versions, so I'm OK with making changes like this. If you'd like to keep a compatible version of the code in here though, we can do that easily enough.

mmatschiner:
Great work, thanks! All changes seem reasonable. And from my side, there's no need to keep a compatible version.

jeromekelleher:
Great, thanks @mmatschiner. Merging.

PR #1341, opened 2020-12-17
move to verification

### LINKED ISSUE #1334: switch statistical tests of mutation to verification.py

petrelharp:
Closes #1334. I think I've got all of @GertjanBisschop's great tests moved over.

Some of the plots are (unavoidably) not qq plots, but plots of relative error as a function of sample size, that we expect to be going down towards zero, like this one:
F84_continuous_transitions

How's this look?


verification.py
```diff
@@ -4087,6 +4087,285 @@ def test_simulate_from_demography(self):
...
+    p_threshold = 0.001
+
+    def chisquare(self, observed, expected, p_th=p_threshold):
+        obs = observed.reshape(np.product(observed.shape))
+        exp = expected.reshape(np.product(observed.shape))
+        assert np.all(obs[exp == 0] == 0)
+        not_zeros = exp > 0
+        if sum(not_zeros) > 1:
+            chisq = scipy.stats.chisquare(obs[not_zeros], exp[not_zeros])
+            assert chisq.pvalue > p_th
```
jeromekelleher, line 4101: What happens if this assertion fails? Does it bring everything down with a clear error message when running with multiple threads? I just want to be sure that I don't miss a problem if this ever fails.

jeromekelleher (commented):
Looks great, thanks @petrelharp. Would you mind putting a hline at y=0 in the left plot please (maybe the right one too), just so it's obvious what the values should be tending to? I need to scan hundreds of plots when doing these checks, so it really helps if they are "right" in as obvious a way as possible.

verification.py
petrelharp: Whoops - didn't even mean to leave this in here. I could put these p-value checks back in, but they *should* be redundant with the figures.

jeromekelleher: It's up to you - might be best to remove them and keep things simple.

petrelharp:
Ok, how's this?
F84_continuous_roots

I added the 1/sqrt(n) line on the right; the visual check here is that the blue line is "going down", and the dotted line is the guide for how much it should be going down. It's not expected to be always below it, or always on top of it or anything. Is this clear enough?

jeromekelleher (approved):
LGTM! Quick squasharoo and we're ready to merge.

petrelharp:
Whoops, sorry - I was just finishing up some minor fixing - I got rid of some divide-by-zero warnings, and realized one of the comparisons could be made better (the things we were comparing were in pairs, so a y=x plot is better than a qq plot). I'll put that in a new PR, I guess.

Issue #1338, opened 2020-12-15
Should we maintain 2 separate read_hapmap functions, and a single write_hapmap?


hyanwong:
Opening this to seek consensus between @jeromekelleher and @grahamgower. At the moment we have 2 functions:  `msprime.RecombinationMap.read_hapmap()` (deprecated but maintained) and `msprime.read_hapmap()`. Currently the first is a thin wrapper for the second. As far as I understand, @grahamgower suggested separating out the logic so that `msprime.RecombinationMap.read_hapmap()` carried on reading the "rate" column (not the recommended data source), whereas `msprime.read_hapmap()` would only read the "genetic position (cM)" position (the canonical data source). Additionally `msprime.read_hapmap()` would be more strict about the data format required. However, I think @jeromekelleher wasn't convinced by the code duplication that would entail.

As part of this, @grahamgower suggested that an `msprime.RecombinationMap.write_hapmap()` function would be useful to output the lower quality maps in `stdpopsim` to a format that would then be readable by `msprime.read_hapmap()`. @jeromekelleher thought that perhaps this would be useful for RateMaps in general, and so should be a method of the general RateMap class. However, a RateMap is more general than a recombination map, so the headers that would be output would probably not be suitable for general use. I guess we could have a `RateMap.write_text(name=XXX, headers=XXX)` function which required the user to provide headers - `RecombinationMap.write_hapmap()` could then call this with the appropriate headers.

Before coding this up, I'd like some clarity on what the approach should be - thanks!

jeromekelleher:
Thanks for this @hyanwong, good summary.

Ideally, I'd like to see the following:

1. An implementation RateMap.read_hapmap that can read from *either* the rate or map column, which looks like 

```
def read_hapmap(fileobj, position_col=None, rate_col=None, map_col=None):
     position_col = 1 if position_col is None else position_col
     if rate_col is None and map_col is None:
          # Default to map_col
          map_col = 3
    elif rate_col is not None and map_col is not None:
         raise ValueError("Cannot specify both rate_col and map_col")
   if map_col is not None:
         # read in the hapmap from the map col
   else:
         assert rate_col is not None
         # read in the hapmap from map_col
```

Then, RecombinationMap.read_hapmap() uses this implementation to continue supporting the existing functionality.

We can add RateMap.write_hapmap(). This would just use the format - if people happen to use this for non recombination map data, then so be it. If we want a more general write_text method we can add it later.

grahamgower:
[...]

PR #1346, opened 2020-12-18
Read hapmap reworking

hyanwong:
This addresses the first part of https://github.com/tskit-dev/msprime/issues/1338 as suggested by @jeromekelleher, with the addition of reading in the 0th column as the name, if appropriate. I've covered a few other edge cases such as the user accidentally specifying the same pos & map cols. 

After this, the only remaining part to do is to implement `write_hapmap`, I think.

### DIFF
--- a/msprime/intervals.py
+++ b/msprime/intervals.py
@@ -384,7 +384,7 @@ class RecombinationMap:
             "Use msprime.read_hapmap() instead.",
             FutureWarning,
         )
-        rate_map = read_hapmap(filename)
+        rate_map = read_hapmap(filename, position_col=1, rate_col=2)
         rate = np.append(rate_map.rate, 0)
         return cls(rate_map.position, rate, map_start=rate_map.start_position)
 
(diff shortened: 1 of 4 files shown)


grahamgower (commented):
Looks good, thanks @hyanwong. I had a couple of additional thoughts...

* It seems strange to read in the rate column, convert to genetic map units, then convert back to rates in order to create the RateMap object. One of your arguments in favour of focusing on the genetic map units was to avoid potential error accumulation when using the RateMap's map units. Doesn't this exacerbate the problem?
* Maybe the function name should be changed to suggest the information being read, rather than the originating dataset that used the format? E.g. `s/read_hapmap/read_genetic_map/`?

hyanwong:
> Looks good, thanks @hyanwong. I had a couple of additional thoughts...
> 
> * It seems strange to read in the rate column, convert to genetic map units, then convert back to rates in order to create the RateMap object. One of your arguments in favour of focusing on the genetic map units was to avoid potential error accumulation when using the RateMap's map units. Doesn't this exacerbate the problem?

Yes, I agree it seems weird. I did originally allow the RateMap `__init__` function to take a `cumulative_rate` parameter as well as a `rate` parameter, but @jeromekelleher didn't like the added complexity, so I argued for a bit to just have the `cumulative_rate` value, but eventually just left it alone. I can reinstate if you and @jeromekelleher agree what is best. I assumed that Jerome wasn't worried about precision errors because numpy presumably stored numbers internally to a much higher precision than the values listed in the file, meaning that switching between rate and cumulative_mass is probably not going to result in much accumulation of error.

> * Maybe the function name should be changed to suggest the information being read, rather than the originating dataset that used the format? E.g. `s/read_hapmap/read_genetic_map/`?

Good point. Happy to change this if we decide which is preferable. I have no strong opinion. Not sure about @jeromekelleher ?

msprime/intervals.py
```diff
@@ -449,50 +449,151 @@ def asdict(self):
...
-        in plain text or gzipped plain text.
+        chr10	    133762002     3.358        181.129345
+        chr10	    133766368     0.000        181.144008
+
+    :: note..
+        If neither ``position_col``, ``rate_col``, or ``map_col`` are 0, then the
+        initial column is assumed to be the chromosome name, and a warning given if
+        rows do not contain identical names.
+
+    :param str fileobj: File, filename, or generator to read. If the filename
```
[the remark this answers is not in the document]
hyanwong: I just copied the text basically verbatim from the `loadtxt` docs (as this is passed straight on to `numpy.loadtxt()`), rather than make the decision myself. But happy to remove this bit: as you say, it complicates the matter.

msprime/intervals.py
```diff
@@ -367,7 +367,7 @@ def uniform_map(cls, length, rate, num_loci=None):
         return cls([0, length], [rate, 0], num_loci=num_loci)
 
     @classmethod
-    def read_hapmap(cls, filename):
+    def read_hapmap(cls, filename, position_col=1, rate_col=2):
```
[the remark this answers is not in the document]
hyanwong: Good point. I should remove these.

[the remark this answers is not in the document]
hyanwong: Removed in lastest push

tests/test_intervals.py
```diff
@@ -358,3 +363,163 @@ def test_getitem_slice_errors(self):
...
+    def test_read_hapmap_nonzero_map_start(self):
+        hapfile = io.StringIO(
+            """\
+            HEADER
+            chr1 1 x 0.000001
+            chr1 2 x 0.000001 x
+            chr1 3 x 0.000006 x x x"""
+        )
+        rm = msprime.read_hapmap(hapfile)
+        print(rm.position, rm.rate)
```
jeromekelleher, line 465: print

tests/test_intervals.py
```diff
@@ -358,3 +363,163 @@ def test_getitem_slice_errors(self):
             recomb_map[-101:]
         with pytest.raises(IndexError):
             recomb_map[:-101]
+
+
+class TestReadHapmap:
```
jeromekelleher, line 368: We're only testing fileobjs here - it would be nice to verify that we can use a filename as well. Writing a file to a temporary directory would be a good way to check this.

tests/test_intervals.py
```diff
@@ -358,3 +363,163 @@ def test_getitem_slice_errors(self):
...
+    def test_bad_sequence_length(self):
+        hapfile = io.StringIO(
+            """\
+            HEADER
+            chr1 0 x 0
+            chr1 1 x 0.000001 x
+            chr1 2 x 0.000006 x x x"""
+        )
+        with pytest.raises(ValueError, match="sequence_length"):
+            msprime.read_hapmap(hapfile, sequence_length=1.999)
```
jeromekelleher, line 525: It would be good to add a test in which we have a fragment of a real hapmap file  (e.g., from HapMap, say 10 lines) and we verify that we get approximately the same values for the rate in the returned RateMap object when we use either the rate or the map columns.

msprime/intervals.py
```diff
@@ -449,50 +449,151 @@ def asdict(self):
...
         ...
-        chr1	182973428	2.512769	122.832331
-        chr1	183630013	0.000000	124.482178
-
-    :param str filename: The name of the file to be parsed. This may be
-        in plain text or gzipped plain text.
+        chr10	    133762002     3.358        181.129345
+        chr10	    133766368     0.000        181.144008
+
+    :: note..
```
jeromekelleher, line 484: What if the column at position zero contained something completely different, wouldn't this warning be unnecessary and annoying then? There's no need for this check, it's not doing anything useful *unless* we actually use the value at some point.

msprime/intervals.py
```diff
@@ -449,50 +449,151 @@ def asdict(self):
...
+            "Cannot specify the same columns for position_col and rate_col or map_col"
+        )
+    column[position_col] = ("pos", int)
+    if 0 not in column:
+        column[0] = ("chrom", object)
+
+    colnames = [c[0] for c in column.values()]
+    data = {
+        name: col
+        for name, col in zip(
```
jeromekelleher, line 538: I think you can just call ``dict(zip())`` here,  i.e,
```
col_data = np.loadtxt(..)
data = dict(zip(col_names, col_data))
```

jeromekelleher (commented):
Looks good, thanks @hyanwong. Needs a bit more testing, but generally very clean.

jeromekelleher:
@hyanwong, I'm keen to get 1.0.0a1 released today. Do you want to get these changes in, or should we wait until a2 (which I'd imagine won't be long following, since there's bound to be issues).

hyanwong:
Sorry. I let this slip. I have some time now, so I'll just do it in the next hour or so.

msprime/intervals.py
hyanwong: That's a good point. The reason I want to keep the check is that (a) it is common to munge multiple chromosomes together in the same file, and we should check that the user is not doing this by mistake (b) it is useful to have the chromosome name, and we had thought that we might use it in future (although we are not doing so at the moment) and (c) @grahamgower suggests (and I agree) that we should try to be pretty strict about the format we are defaulting to in read_hapmap(), so that we don't get into the bind we had before of making files that will "work" in stdpopsim, but are not in any conventional format.

My inclination therefore is to keep it, but maybe to have a keyword only param that either specifies the column containing the name (default=0) or if a string, the chromosome name. I suspect that you won't like this though?

msprime/intervals.py
```diff
@@ -449,50 +449,151 @@ def asdict(self):
...
+        physical position in the file, and the last value in the rate column must be
+        zero.
+    :param str map_col: The zero-based index of the column in the file specifying
+        the genetic map position in centiMorgans. If ``None`` (default), assume an index
+        of 3 (i.e. the fourth column). If the first genetic position is 0, set the
+        :attr:`RateMap.start_position` of the returned map to the first physical
+        position in the file. Otherwise, act as if an additional row, specifying
+        physical position 0, genetic position 0, was inserted at the start of the file.
     :return: A RateMap object.
+    :rtype: RateMap.
```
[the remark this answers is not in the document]
hyanwong: Good spot

msprime/intervals.py
jeromekelleher: Let's drop if for now, please. It'll break anyway if someone puts in multiple chrs because of the ordering requirement of positions.

hyanwong: OK. I'll open up an issue and paste the code & test in there so we don't forget it, though.

hyanwong:
OK - I've dealt with all your comments @jeromekelleher - thanks for those. Because I put in an instersphinx link to `np.loadtxt()` I also reactivated the inter sphinx link to numpy in `docs/conf.py`, which has been removed at some point because it was 404ing (but isn't any more). That's isolated as the first commit in this PR.

hyanwong:
P.s. 2 comments brought up by @grahamgower which were not addressed above:

1. should this be called `read_genetic_map` rather than `read_hapmap`?
2. should we allow the map pos value to be passed directly into the RateMap constructor (@jeromekelleher wanted me to remove it, which I did, but @grahamgower suggested it should be in there)?

msprime/intervals.py
hyanwong: Now in https://github.com/tskit-dev/msprime/issues/1363

jeromekelleher:
Thanks @hyanwong, looks good. I think this should be a static method of RateMap, but I'll do this in a follow up now.

> 1. should the function be called msprime.read_genetic_map rather than msprime.read_hapmap?

I think RateMap.read_hapmap is pretty descriptive. Something like ``read_genetic_map`` seems more general than is warranted (i.e., this is designed specifically to work with the quirks of hapmap)

> should we allow the map_pos value to be passed directly into the RateMap constructor, or should we use it within read_hapmap to construct a rate array which we pass in?

Possibly, but I doubt it'll make any real difference and it the RateMap constructor is already quite complicated.

jeromekelleher (approved):

PR #1347, opened 2020-12-22
Continue simulation

jelliotthiggins:
Moving this here from pyslim at the suggestion of @petrelharp 

In using pyslim, I found the process described in "[Vignette: Following up with more coalescent simulation](https://pyslim.readthedocs.io/en/latest/vignette_continuing.html)" incredibly useful. To make it more accessible for simulations instead of just slim, I added it to msprime as a method called `continue_simulation`. It is all contained in the high level python right now, although a C implementation could be useful long term. 

It is a companion to `recapitate` / `simulate(from_ts=)`, in that it continues the simulation but on the young end of a tree sequence instead of on the ancient side. It takes in the initial tree sequence, the length of time to simulate in front of the slim simulation, a sample size for the continued simulation, and any msprime kwargs. It outputs a complete tree sequence, containing the input tree sequence as well as the continued simulation. It uses one private helper method, `_adjust_tables_time`, which performs the time adjustment on the tables - @petrelharp noted that this may be better suited as a tskit method but suggested I leave it here for now. 

The initial use described in the example taking a tree sequence produced by SLiM, recapitating, adding mutations, and continuing the simulation. This could be done to any tree sequence, with all times simply moved back as much as needed. 

It is possible more work is needed in marking / unmarking samples, I'm not quite sure how pyslim and msprime differ on that front. I am also asserting that the number of roots in the new simulation must be fewer than the number of samples in the old simulation - this is to allow random matching of samples between the two sequences without duplication, i.e. all new samples are matched randomly to an old sample, leaving out some old samples. I am very open to suggestions on that implementation - perhaps in that case (which seems fairly unlikely in most uses) the matching process could be reversed, i.e. all old samples are attached randomly to new samples, leaving out some new samples. 

Much of this code came directly from the example, with modifications to make it work for all tree sequences instead of just slim tree sequences. Below I'll demonstrate a very basic example. The ability to combine slim simulations with msprime simulations on either end, using this technique and recapitation or from_ts, is so powerful and I hope this tool can be of use to others. Tests are coming soon so this will be kept as a draft for now. 

Jack Elliott-Higgins


petrelharp:
Thanks, this looks great!

A related thing: not all the samples might be alive at the end of the simulation, so I think there should be another argument - maybe `match_samples`? Or, `continue_nodes`? - that takes a vector of node IDs that are the ones that will be matched up to the lineages in the msprime simulation. It could default to the samples.

petrelharp:
>  I am also asserting that the number of roots in the new simulation must be fewer than the number of samples in the old simulation - this is to allow random matching of samples between the two sequences without duplication, i.e. all new samples are matched randomly to an old sample, leaving out some old samples. I am very open to suggestions on that implementation - perhaps in that case (which seems fairly unlikely in most uses) the matching process could be reversed, i.e. all old samples are attached randomly to new samples, leaving out some new samples.

Conclusion from [the pyslim thread](https://github.com/tskit-dev/pyslim/pull/118#discussion_r546984298): the number of lineages (i.e., [new_nodes, here](https://github.com/tskit-dev/msprime/pull/1347/files#diff-633c4e6b2a4aab70e9ffa4f341a814d7fd50c4354dc01109ee7cf85741cb0836R1190)) needs to be less than or equal to the number of nodes at the end of the SLiM simulation, and an error will be thrown otherwise. This is guaranteed if the number of samples requested is less than or equal to the number of nodes at the end of the SLiM simulation. Then matching will be done by sampling without replacement.

jelliotthiggins:
> A related thing: not all the samples might be alive at the end of the simulation, so I think there should be another argument - maybe `match_samples`? Or, `continue_nodes`? - that takes a vector of node IDs that are the ones that will be matched up to the lineages in the msprime simulation. It could default to the samples.

Great idea, added this functionality. Cleaned up a good amount and have pretty comprehensive testing, so this seems ready to review to me (only failing test is Build Docs which seems to be failing in all pull requests right now).

msprime/ancestry.py
```diff
@@ -1146,6 +1146,128 @@ def sim_ancestry(
...
+):
+    """
+    Returns a full tree sequence, by continuing the simulation using msprime
+    after an earlier simulation has been performed. The msprime simulation and slim
+    simulation are joined using :meth:`tskit.tables.TableCollection.union()`.
+
+    This method executes code from the pyslim tutorial  'Following up with more
+    coalescent simulation'
+
+    ``sample_size`` defaults to the number of samples present in the input tree sequence.
```
petrelharp, line 1164: Documentation needs some work (e.g., remove reference to SLiM, since this is generic).

msprime/ancestry.py
```diff
@@ -1146,6 +1146,128 @@ def sim_ancestry(
...
+    if sample_size is None:
+        sample_size = len(continue_nodes)
+
+    new_ts = simulate(sample_size=sample_size, end_time=time, **kwargs)
+
+    new_nodes = np.where(new_ts.tables.nodes.time == time)[0]
+
+    if len(new_nodes) > len(continue_nodes):
+        raise RuntimeError(
+            f"""More lineages present in new simulation samples than in the old.. \
```
petrelharp, line 1199: Something more precise would be good here. This is tricky, but how about: "More lineages present at the longest-ago end of the 'continued' portion of the simulation than there are nodes in 'continue_nodes'."

msprime/ancestry.py
```diff
@@ -1146,6 +1146,128 @@ def sim_ancestry(
...
+        oldest_node_time = max(
+            [
+                new_ts.node(new_ts.first().roots[i]).time
+                for i in range(0, len(new_ts.first().roots), 1)
+            ]
+        )
+        raise RuntimeError(
+            f"""No surviving lineages in new simulation at merge time. \
+            Depth of new simulation is {oldest_node_time} - alter parameters \
+            to extend simulation or reduce the simulation time."""
```
petrelharp, line 1215: Oh, right! Forgot about this. So, if there's no nodes with time equal to `time`, that doesn't mean there aren't lineages, it just means that everything's coalesced. There is one lineage still at each position (there is always at least one, since the samples had to inherit from *someone*), but its the root of the tree. What we're doing currently - leaving these out - only gives us the variants that are segregating within the sample at the end. We'd be missing the scribbled edge in this picture:
trees


My first thought is to just add the roots of any trees that have coalesced to `new_nodes`, like so:
```
roots = []
for t in new_ts.trees():
    if t.num_roots == 1:
        roots.append(t.root)
new_nodes.extend(roots)
```
But let's see: then the two nodes we're telling union are the same would have different times, and we'd end up with the node time from `tables`, thus wrongly increasing the time of the roots of those trees. So, that's not right. Really, we need to add on new nodes above these roots, something like:
```
new_tables = new_ts.tables
for t in new_ts.trees():
    if t.num_roots == 1:
         n = new_tables.nodes.add_row(time=time)
         new_tables.edges.add_row(parent=n, child=t.root, left=t.left, right=t.right)
         new_nodes.append(n)
```

Do you see any issues with this? (It helps to draw the pictures.

msprime/ancestry.py
```diff
@@ -1146,6 +1146,128 @@ def sim_ancestry(
...
+
+    tables = _adjust_tables_time(ts, time)
+    tables.union(
+        new_ts.tables, node_map, add_populations=False, check_shared_equality=False
+    )
+
+    return tables.tree_sequence()
+
+
+def _adjust_tables_time(ts, time):
```
petrelharp, line 1231: two suggestions:
- let's just move this in to `continue_simulation`, without defining a new method, with a node saying that we can replace it with a call to the tskit method, when available
- this should act on `tables` (which are edited in place), rather than on a tree sequence

msprime/ancestry.py
```diff
@@ -1146,6 +1146,128 @@ def sim_ancestry(
...
+    from a tree sequence's tables and modifying the time in the nodes and
+    mutations tables.
+
+    This method executes code from the pyslim tutorial  'Following up with more
+    coalescent simulation'
+
+    :type ts: :class`tskit.TreeSequence`
+    :param ts: tree sequence used as basis for new tables
+    :param time int: The time to adjust tables by,
+        in units of generations
```
petrelharp, line 1243: this should be a float (times in tskit are not necessarily integers)

msprime/ancestry.py
```diff
@@ -1146,6 +1146,128 @@ def sim_ancestry(
...
+    :return: The :class:`tskit.TableCollection` object with modified time
+    :rtype: :class:`tskit.TableCollection`
+    """
+    tables = ts.tables
+    tables.nodes.set_columns(
+        flags=ts.tables.nodes.flags & ~np.uint32(tskit.NODE_IS_SAMPLE),
```
petrelharp, line 1249: oh, right - this code is doing more than shifting the times, anyhow - it's also removing sample flags. We should actually remove not *all* the sample flags, but only the flags from `continue_nodes`.

tests/test_continue_simulation.py
```diff
@@ -0,0 +1,148 @@
...
+#
+"""
+Test cases for continue_simulation.
+"""
+import pytest
+
+import msprime
+
+
+class TestContinueSimulation:
```
petrelharp, line 27: I think we should write a verification function, that would
- check that subsetting to all nodes present before `time` gets back the original tables
- check that all roots of all trees are older than `time`

petrelharp:
Looks great! One tricky thing in the comments, there. And: in the upcoming msprime, we're switching from using `msprime.simulate( )` to two method: `msprime.sim_ancestry` and `msprime.sim_mutations`. The docs don't seem to be updating at the moment, so maybe the best way to catch up on this is to build them locally (run `make` in the `docs` directory; then navigate to `docs/_build/html`). For our purposes here, the upshot is that I think we should:
- change `simulate( )` to `sim_ancestry( )`
- note in the docs that this won't put mutations on the new ancestry, so it should be followed up with a call to `msprime.sim_mutations(..., end_time=time)`

msprime/ancestry.py
jelliotthiggins: This looks right to me by inspection, will have to verify with some testing

msprime/ancestry.py
jelliotthiggins: That makes more sense - I wrote this up in a way that makes sense to me but if this is something that is done elsewhere in a specific way I'd be happy to plug that in!

tests/test_continue_simulation.py
jelliotthiggins: On that first condition, which properties could be appropriate and sufficient to do the comparison on? We can get back aspects of the original tables, but with altered flags and times in the nodes and mutations tables. An edge table comparison could be fairly comprehensive and feasible.

petrelharp: Good question, and this can be pretty tricky. ButI think if you re-mark the nodes present at `time` as samples, and then simplify to them, you should get back exactly the original tables?  Might anything else change?

petrelharp: Oh - sites and mutations shouldn't have changed.

tests/test_continue_simulation.py
```diff
@@ -25,124 +25,149 @@
...
-        return max(
-            [
-                ts.node(ts.first().roots[i]).time
-                for i in range(0, len(ts.first().roots), 1)
-            ]
-        )
+    def verify(self, ts, new_ts, time, continue_nodes=None):
+        assert round(new_ts.max_root_time - time, 10) == round(ts.max_root_time, 10)
+        assert ts.num_mutations == new_ts.num_mutations
+        assert ts.num_sites == new_ts.num_sites
```
petrelharp, line 31: Shouldn't we have
```
assert ts.tables.sites == new_ts.tables.sites
assert ts.tables.mutations == new_ts.tables.mutations
```
... ?

tests/test_continue_simulation.py
```diff
@@ -25,124 +25,149 @@
...
+        assert list(ts.tables.sites.position) == list(new_ts.tables.sites.position)
+        if continue_nodes is None:
+            continue_nodes = ts.samples()
+        section = list(set(continue_nodes).intersection(new_ts.samples()))
+        assert len(section) == 0
+        for tree in new_ts.trees():
+            assert new_ts.node(tree.root).time > time
+        # check that subsetting to all nodes present before time gets back orig. tables
```
petrelharp, line 39: I'd say first mark `continue_nodes` as samples, and then to
```
sub_tables = new_ts.simplify(continue_nodes).tables
sub_tables.provenances.clear()
orig_tables = ts.simplify(continue_nodes).tables
orig_tables.provenances.clear()
assert sub_tables == orig_tables
```
There can be tricky sorting issues, but I think this should work.

petrelharp:
Looks good - a couple of suggested changes to the tests. Let me know if you'd like me to jump in to debug things.

jeromekelleher:
Hi @jelliotthiggins, and welcome to tskit-dev! Thanks for doing this, it looks great and will be a very useful operation. I think the functionality is definitely something that we should provide. However, I'm not sure that the current strategy is the best approach.

Really, I think, the actual tricky bit here is a *join* operation between two tree sequences: we glue the roots of one tree sequence to the samples of another. Let's take a look at one of your examples:
```python
      ts = msprime.sim_ancestry(
￼            samples=10000,
￼            population_size=5000,
￼            recombination_rate=1e-8,
￼            sequence_length=1e4,
￼            random_seed=72,
￼        )￼       
￼        continue_nodes = ts.samples()[0::2]
￼        time = 1000
￼        cts = msprime.continue_simulation(
￼            ts,
￼            time,
￼            continue_nodes=continue_nodes,
￼            population_size=5000,
￼            recombination_rate=1e-8,
￼            sequence_length=1e4,
￼            random_seed=72,
￼        )
```
I think it would be clearer to write it like this:
```python
       ancient_ts = msprime.sim_ancestry(
￼            samples=10000,
￼            population_size=5000,
￼            recombination_rate=1e-8,
￼            sequence_length=1e4,
￼            random_seed=72,
￼        )￼   
        recent_ts = msprime.sim_ancestry(
            samples=2,
            end_time=1000,       
￼            population_size=5000,
￼            recombination_rate=1e-8,
￼            sequence_length=1e4,
￼            random_seed=72,
￼        )
        join_nodes = ancient_ts.samples()[0: recent_ts.num_samples]
        joined_ts = ancient_ts.join(recent_ts, nodes=join_nodes)
```

There's a few advantages I think:
- The process is entirely generic - we can join arbitrary tree sequences, not just glueing an msprime simulation onto another tree sequence
- We avoid API duplication in msprime: the ``continue_simulation`` function here basically just has the same arguments as ``sim_ancestry``, and we also avoid an awkward naming problem. What would we call this in order to keep the ``sim_X`` pattern going? Maybe ``sim_extra_ancestry``?
- Arguably this is easier to understand.

@petrelharp, what do you think? This would mean moving the PR to tskit and rejigging things a bit. Also, I'm not sure ``join`` is the right name, perhaps this is too general a name, and we might want to reserve this for cases where we're doing other composition operations (like a "horizontal" join, as opposed to this "vertical" join).

jeromekelleher:
@jelliotthiggins - if you're not a member of the tskit-dev Slack workspace already, please send me an email (jerome.kelleher@bdi.ox.ac.uk) and I'll send an invite. It'd be great to have you as a member of the community!

petrelharp:
Hm, I'm not sure: isn't this "join" just `union ( , check_shared_overlap=False)`?  This seemed like a good idea to me since it's conceptually easier and more findable. Even the fact that you can do this is not obvious to people new to coalescent simulations. This seemed like the natural complement to `from_ts`.

jelliotthiggins:
This operation seems to strike a tricky balance between a core functionality and standardizing something anyone could do with tree sequences. It's definitely limiting what could be an entirely generic process to a more specific function, but the limit in place makes it ideal for the specified functionality and prevents most nonsensical uses. I think either approach could work well assuming it's properly document and I'm happy to implement whichever makes more sense to the team. I'm going to finish implementing the tests as is for now to make sure there is either a finished product to merge with msprime or a strong test base to make the move to tskit.

jeromekelleher:
> Hm, I'm not sure: isn't this "join" just union ( , check_shared_overlap=False)? This seemed like a good idea to me since it's conceptually easier and more findable. Even the fact that you can do this is not obvious to people new to coalescent simulations. This seemed like the natural complement to from_ts.

I'm really not keen on having another function in here which is almost but not quite identical to sim_ancestry. It seems like a maintenance burden, and I'm sure they're be unexpected consequences down the line, as people want to do more complicated things with the "recent" simulation that is being done here. For (contrived) example, what if someone wants to do three simulations in paralell and chain them together afterwards?

I think it's conceptually easy for you as you're used to thinking about mushing together back and forward simulations, but it took me a good bit to figure out what this was doing and I find separating it out into an "ancient" and "recent" simulation is much easier to grasp. If we can't come up with a descriptive name for a function, then this is a good indication to me that it shouldn't *be* a function. ("continue" doesn't work for me - continue in which direction? We're working backwards in time here, so I would assume that this means continue backwards in time, which is what ``initial_state`` is for. )

We could add some extra arguments to ``sim_ancestry``, but really if all it's going to be doing is calling ``union``, is it worth the complexity and additional API surface to an already extremely complicated function?

Could we add some arguments to ``union`` to make the process of combining the recent and ancient simulations a one-liner? I think it's about the same amount of documentation effort to write up how to use union in this way, as it would be to write up how (and why you would want to) combine multiple simulations like this here.

petrelharp:
I think @jelliotthiggins would be a good judge of what'd help users out most, as it's fresh - what motivated you to write the PR? Now that you've thought through the details, would a thoughtful example in the docs be as good or better?

jelliotthiggins:
I think as a pyslim-focused function it made the most sense this way, but I do agree that the terminology is confusing with back vs. forwards and that the maintenance of this within msprime may get tricky. I think the best thing to do for now is to implement this in tskit so we can do an actual comparison - I will take a stab at that. I'm struggling with how to incorporate all of it into `union` to facilitate a one-liner, especially given that much of this operation is creating the node map. So I see two options:

1. Incorporate a time offset into union to shift `self` to the appropriate time. Document extensively how to feed in the correct arguments (i.e. how to construct the node map, appropriate tree sequences to join together)
2. Shift all of the work into a new method, such as `join` or `time_offset_union` or something, which does all of the parameter construction necessary and then calls `union` internally.  This would facilitate a one-liner but runs into the issue of maintaining that call to `union`. 

Another consideration is whether this should be a Tables function, a TreeSequence function, or both?

Either way by moving `sim_ancestry` out of the method, we lose some control over what kind of tree sequences get joined together, so we will have to be more careful in catching inappropriate use. Documentation is going to be key on this for sure. Which path seems more appropriate for now to you both, for comparison's sake?

jeromekelleher:
This is really helpful @jelliotthiggins, and thanks for your patience! Some thoughts:

- Things are a bit trickier when dealing with an arbitrary tree sequence rather than one you've just gotten from ``sim_ancestry``, but it's not too bad I think. We would have to worry about how to combine sites and mutations. I'm sure we're doing this somewhere else, though, so that shouldn't be too bad. We won't be able to find the nodes to join by using time like you're doing here, but we can use the roots instead, like ``set().union(*[set(t.roots) for t in ts.trees()])``. Then, you take ``time = np.max(tables.nodes.time))`` and add unary edges out to ``time``, like you're doing here, but checking if the root's time == ``time`` rather than whether it's a single root.
- I'm not worried at all about any extra maintenance in tskit from adding new functionality - we want more functionality in tskit! The concern here is that passing arguments through via kwargs can get quite tricky, and front-ending another function isn't usually a good pattern, unless there's a compelling reason. Providing the tools to *compose* the results of functions is usually a much better and more flexible pattern, which long term leads to less maintenance.

So, I propose we add a new method ``TableCollection.time_join`` to tskit which implements this functionality? (We can add a TreeSequence equivalent if we want, but the TableCollection method is where it will be implemented). 

I think ``join`` is a better name than ``union``, because we are doing a bit more than unioning based on nodes - we're also adding unary edges from the roots up to the samples. ``time_join`` seems a bit better than ``vertical_join``, as it's more descriptive. Definitely open to better ideas though!

How does this sound @jelliotthiggins, @petrelharp?

petrelharp:
Hah, I see what you mean now, @jeromekelleher - that makes a lot of sense. Sounds good to me.  It *would* be useful to put the time shift part into `union`, by the way, as our use-case for union requires a time shift also.

jelliotthiggins:
All sounds good to me, thank you for all the input @petrelharp and @jeromekelleher. Will take a stab at this soon!

jelliotthiggins:
>  Things are a bit trickier when dealing with an arbitrary tree sequence rather than one you've just gotten from `sim_ancestry`, but it's not too bad I think. We would have to worry about how to combine sites and mutations. I'm sure we're doing this somewhere else, though, so that shouldn't be too bad. We won't be able to find the nodes to join by using time like you're doing here, but we can use the roots instead, like `set().union(*[set(t.roots) for t in ts.trees()])`. Then, you take `time = np.max(tables.nodes.time))` and add unary edges out to `time`, like you're doing here, but checking if the root's time == `time` rather than whether it's a single root.

@jeromekelleher I'm getting a bit confused by this, so I'm going to walk through how I'm thinking about it to make sure we're on the same page. Any time I use `time` I'm referring to the join time. 

- by an arbitrary tree sequence, I'm taking that to mean as opposed to one that has coalesced. So that leaves us with 3 use cases I think
1. TreeSequence from `sim_ancestry` with an imposed `end_time`, such that coalescence is not complete and we have nodes at `time` to use as the nodes to join by
2. TreeSequence from `sim_ancestry` where coalescence is complete before the merger time, in which case we add unary edges out to `time` before merging
3. An arbitrary TreeSequence, where we cannot necessarily use `time` to find the nodes and where trees can have multiple roots
- This third case is the case you're referring to above and the one I'm confused about. So to handle this we first get all the roots using the set union and get the max node time in the tree sequence, and then I get lost. When you set `time` to be that max node time, do you mean that we're now merging the tree sequences on that time? Or that we're just going to use that time as a parameter to determine which roots to use to extend unary edges to the merge time? If `time` is referring to two separate time points (the oldest point in `recent_ts` and the merge time) then I follow, but if not then I'm not understanding why we would conflate the two. 
- I'm trying to determine whether these three use cases are mutually exclusive and all need to be implemented or whether we can combine at least the second and third somehow. 
- I also will probably have farther questions on combining sites and mutations but trying to implement this functionality first!



> It _would_ be useful to put the time shift part into `union`, by the way, as our use-case for union requires a time shift also.

@petrelharp just making sure I'm interpreting this right, are you suggesting performing the actual time shift portion within `union` instead of within `time_join`? Should be similar implementation either way, just want to be certain!

petrelharp:
One note: in case (1) you'll have to add edges out to `time` at any trees where everyone has coalesced. If `time` is very small this might never happen, but in general I don't really see a hard distinction between the three cases.  I think that you'll need to look at each root, and if that root's time is greater than `time`, throw an error, and if it's less than time, add an edge. (and, we have to figure out what to do here - is there a new node added?).

petrelharp:
> @petrelharp just making sure I'm interpreting this right, are you suggesting performing the actual time shift portion within union instead of within time_join? Should be similar implementation either way, just want to be certain!

Yep - I'm suggesting a `shift_time = 0.0` argument to `union`.

jeromekelleher:
I think the best way to approach this is first try joining two msprime simulations that have fully coalesced. So, each tree in recent ts will have one root, but all the roots will be at different times.  We would have had to deal with this case anyway even if we front-ended ``sim_ancestry`` like you're doing here, because it's perfectly possible everything will coalesce before your join time.

Dealing with multiple roots per tree is a straightforward update once that works.

jelliotthiggins:
@petrelharp @jeromekelleher Moved the PR and conversation to [tskit](https://github.com/tskit-dev/tskit/pull/1117) so I'm going to close this out!

PR #1359, opened 2021-01-04
Pin sphinx due to error

benjeffery:
An error in sphinx is occurring on CI that I don't think is worth looking into till it persists or we need to upgrade. Pinning the version for now.


jeromekelleher (approved):
LGTM. Might be worth adding a comment to development.txt to note why we're pinning here? This will probably crop up again when people are trying to build locally, and would be good to have a reminder of what the problem was.

PR #1360, opened 2021-01-11
Finalise sampling for sim_ancestry

jeromekelleher:
This is a proposal to finalise how we draw samples in the ``sim_ancestry`` function. The standard ``msprime.sim_ancestry(n)`` form will still work. More elaborate examples follow.

The ``samples`` parameter can be interpreted as a mapping, in which the keys are populations and the values are the number of samples to draw from the population
```python
demography = msprime.Demography.island_model([100, 100], migration_rate=0.1)
# Take 10 haploid samples from each population
ts = msprime.sim_ancestry({0: 10, 1: 10}, demography=demography, ploidy=1)
```

The keys can also be population ids:
```python
demography = msprime.Demography.from_species_tree(
    "(((human:5.6,chimpanzee:5.6):3.0,gorilla:8.6):9.4,orangutan:18.0)",
    branch_length_units="myr",
    initial_size=10**4,
    generation_time=20)
ts = msprime.sim_ancestry(samples={"gorilla": 2, "human": 4}, demography=demography)
```

~We can also specify a list of Samples in the same way as before so we can control. the sampling time and node IDs, like~
We can specify a list of SampleSets (a bit like the old Samples object, but more efficient/simpler/more flexible) so that we have more control:
```python
N = 100
ts = msprime.sim_ancestry(
    demography=msprime.Demography.island_model([N, N], migration_rate=1),
    samples=[msprime.SampleSet(1, population=0), msprime.SampleSet(1, population=1, time=N)],
)
```

I think this is a good combination of making the things people commonly want to do (sampling N individuals from a given population) easy and efficient, and also providing full flexibility for those who want to do more esoteric things. If we think this general approach is good, I have some ideas for adding more attributes to the Sample object that will give us more control of how sampling is done (e.g., add a ``num_nodes`` argument which lets us override the default behaviour of adding ``ploidy`` sample nodes for each sample).

There's also a bunch of things to finish up here, so don't bother doing a full review at this point.

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2015-2020 University of Oxford
+# Copyright (C) 2015-2021 University of Oxford
 #
 # This file is part of msprime.
 #
@@ -25,6 +25,7 @@ import inspect
 import json
 import logging
 import math
+import numbers
 import struct
 import sys
 
@@ -177,7 +178,7 @@ def _samples_factory(sample_size, samples, population_configurations):
                 "Cannot specify sample size and population_configurations "
                 "simultaneously."
             )
-        s = demog.Sample(population=0, time=0.0)
+        s = Sample(population=0, time=0.0)
         the_samples = [s for _ in range(sample_size)]
     # If we have population configurations we may have embedded sample_size
     # values telling us how many samples to take from each population.
@@ -187,7 +188,7 @@ def _samples_factory(sample_size, samples, population_configurations):
             the_samples = []
             for j, conf in enumerate(population_configurations):
                 if conf.sample_size is not None:
-                    the_samples += [demog.Sample(j, 0) for _ in range(conf.sample_size)]
+                    the_samples += [Sample(j, 0) for _ in range(conf.sample_size)]
         else:
             for conf in population_configurations:
                 if conf.sample_size is not None:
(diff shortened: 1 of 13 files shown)


grahamgower:
Fantastic! The population id form is a welcome addition!

petrelharp:
> ts = msprime.sim_ancestry(samples={"gorilla": 2, "human": 4}, demography=demography)`

Names are *great*! And, what about being able to do, say
```
samples={"gorilla": 20, "human": 4000, "human": (2, 12000)}
```
where if you're passing in a tuple, the second number in the tuple is the time ago?

petrelharp:
So, with this change, what are all the different ways to specify samples? Does `demog.sample( )` still work? (I hope not? if so we might have too many ways to do the same thing.)

jeromekelleher:
Woo, glad you like it @petrelharp!

> samples={"gorilla": 20, "human": 4000, "human": (2, 12000)}

I like the tuple-value form too: I had implemented *exactly* this, but I thought I should reign things in a bit for the first version. We'd still need the overall list form to specify arbitrary samples, so I thought it best to make sure we can do the most general thing before filling out usability gaps like this. (What you have here won't work though, as you have duplicate keys in the dictionary).

> So, with this change, what are all the different ways to specify samples? Does demog.sample( ) still work? (I hope not? if so we might have too many ways to do the same thing.)

Agreed, ``demog.sample`` is dead. This change fully decouples demography from sampling, which is good, I think.

petrelharp:
> We'd still need the overall list form to specify arbitrary samples, so I thought it best to make sure we can do the most general thing before filling out usability gaps like this.

Makes sense, but we should leave the way open for improving.

>  (What you have here won't work though, as you have duplicate keys in the dictionary).

Hah, whoops. =) So, what's the extension, then? Like
```
samples={"gorilla": 20, "human": {0: 4000, 2: 12000}}
```
... ?

> Agreed, demog.sample() is dead. This change fully decouples demography from sampling, which is good, I think.

I agree, it's conceptually simpler.  So, the ways to specify samples are

0. a list of integers of length equal to the number of populations
1. a list of Sample objects
2. a dict with keys equal to population IDs
3. a dict with keys equal to population labels

... right? And that's all?

jeromekelleher:
No quite, no - there's no list of integers form. I'd like to reserve that for simulating pedigrees, and I'm not sure it's all that useful anyway. The ways are:

1. A single integer (can only be used in 1-population models)
2. A list of Sample objects
3. Mapping populations to sample numbers. Keys can be either integer population IDs or string population names, and can be mixed (but, can't specify the same population twice).

msprime/ancestry.py
```diff
@@ -793,59 +756,69 @@ def _insert_samples_from_map(samples_map, demography, ploidy, tables):
...
-        raise TypeError("bad samples type")
-    if samples_map is not None:
-        assert samples_list is None
-        _insert_samples_from_map(samples_map, demography, ploidy, tables)
-    else:
-        assert samples_list is not None
-        _insert_samples_from_list(samples_list, demography, ploidy, tables)
+        raise TypeError(
+            f"The value '{samples}' cannot be interpreted as sample specification. "
+            "Samples must either be a single integer, a mapping of populations "
```
petrelharp, line 815: maybe "a dict that maps" rather than "a mapping"

petrelharp:
That sounds like a good plan. Nice idea about SampleSet, and also about the initial tables. (That's what I'd end up using!)

Well, I guess my gut is suggesting that it might be too many ways, but I'm not seeing which to cut out, so maybe it's actually just right.

msprime/ancestry.py
jeromekelleher: Well it doesn't have to be a dict (could be, say, a collections.defaultdict), but you're right it's easier to follow if it's concrete.

PR #1362, opened 2021-01-14
Quickstart docs for 0.x users.

jeromekelleher:
Adds a some quickstart docs for 0.x users and puts in some quickref tables in various places. It's all still pretty messy, but at least this gives experienced users somewhere to start.

I don't know if this is where this information will ultimately live, this is really just "good enough for an alpha", hopefully.


docs/quickstart.rst
```diff
@@ -107,5 +107,124 @@ to describe more complex demographic histories.
...
+the :ref:`updated recipe <sec_ancestry_multiple_chromosomes>`.
+
+++++++++
+Ancestry
+++++++++
+
+The new :func:`.sim_ancestry` function replaces the 0.x :func:`.simulate`
+function and is very similar. There are some important differences though:
+
+* Samples are now **individuals** rather **nodes** (i.e. monoploid
```
petrelharp, line 148: This is a bit confusing - in the *tree sequence*, "being a sample" is still a property of a node, not an individual! How about
```
* The `samples` argument now refers to the **number of individuals** rather than **the number of nodes**
```

docs/quickstart.rst
```diff
@@ -107,5 +107,124 @@ to describe more complex demographic histories.
...
++++++++++
+Mutations
++++++++++
+
+* For symmetry with the :func:`.sim_ancestry` function, there is now a :func:`.sim_mutations`
+  function. The 0.x :func:`.mutate` function is **deprecated**.
+
+* The :func:`.sim_mutations` function works on a **discrete** genome by default.
+
+* There are now also many new mutation models which support;
```
petrelharp, line 218: "new mutation models supported by :func:`.sim_mutations`"

docs/quickstart.rst
jeromekelleher: Done.

docs/quickstart.rst
jeromekelleher: done.

jeromekelleher:
Thanks @petrelharp, updates in #1366.

PR #1369, opened 2021-01-16
Remove attrs and Python 3.6

### LINKED ISSUE #1361: Increase min version to Python 3.7 and replace attrs with dataclasses
Python 3.6 is pretty old now and will be officially end-of-lifed at the end of 2021. It's likely to be dropped in things like conda-forge sooner than that. Since we're doing a big disruptive update, it makes sense to take the opportunity to hike ourselves up a Python version.

The big advantage of 3.7 here is that we get dataclasses, which has all the features of attrs that we need and we can drop the dependency.

jeromekelleher:
Closes #1361 

Half hearted effort to do annotations, will fill out more in follow up.

Might as well get this in ahead of alpha2

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python: [ 3.6, 3.7, 3.8 ]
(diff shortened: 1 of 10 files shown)


jeromekelleher:
Woot, all works!

PR #1385, opened 2021-01-20
Jupyterbook experiment

jeromekelleher:
This is an initial pass at converting the documentation to [jupyterbook](https://jupyterbook.org/intro.html)

I've converted the files that currently use jupyter-sphinx to execute code cells to the markdown format. It looks great, there's so many nice new features that we can now access.

The rest of the files are left as restructured text for now, but we can convert them as we go.

I've not made any effort to put in a meaningful top-level structure, just tried to get the content rendering so we can try it out.

~Also all the section links are broken - the markdown conversion seems to have change all underscores in the targets to hyphens, which is annoying. I'll see if this is a real problem or not in a follow up.~ [Edit] I fixed this

There's also a lot of unnecessary ``{eval-rst}`` blocks in there; these can be called directly from markdown. I didn't bother with them for now, as I thought we could get the site up and running and fix these issues later with the right regexp magic.

For now, I suggest we merge this so we can see it in action.


jeromekelleher:
Going to merge this, let's see how it looks when deployed.

Issue #1403, opened 2021-01-26
Make nucleotides the default output for sim_mutations


jeromekelleher:
I don't think there's any good reason to output 0/1 any more - it causes many more problems than it solves, and the information about ancestral and derived states is maintained within the table structure.

So, we keep ``mutate()`` as it is, outputting 0/1 alleles, but we change ``sim_mutations`` to do what the vast majority of people want, by default.

(I thought there was another issue relating to this, but couldn't find it)

petrelharp:
Sounds good.

Issue #1399, opened 2021-01-24
Set mutation times to null for 0.x APIs


jeromekelleher:
The main thing that's breaking code in tskit's test suite when using msprime 1.0.0a3 is the fact that ``simulate()`` and ``mutate()`` now return times for mutations. On the one hand, it's clearly and improvement that we now have times for mutations, but on the other since it's looking like the 0.x APIs are really just going to be kept for supporting legacy code (I can't see any good reason for recommending new users to use them; it's just too confusing from a docs perspective).

The reason the tskit code is breaking is because of a defect in the tables APIs that means we can't add entire rows to a table (https://github.com/tskit-dev/tskit/issues/1111). When we do things like 

```python
copy = tables.copy()
do_something_to_copy(copy)
assert copy == tables # Fails because copy doesn't have mutation times
```

So, in the interest of being as compatible as possible, I suggest we set the mutation times for these legacy APIs to null. If someone wants to use the times, then this is a good motivation to upgrade their code to 1.0.

Any objections? cc @petrelharp @benjeffery

petrelharp:
Hm, I'm not so sure - would this be a problem for anything other than testing for equality in our unit tests? It's hard to imagine how providing some extra information would break anyone's code otherwise. Leaving out times would make it harder for people to, say, do minor upgrades to previous scripts to take advantage of mutation times. Maybe that's good because they get motivated to switch to 1.0 for reals, but it kinda seems like an unnecessary barrier.

A stop-gap would be to just assign the `time` column to be null before problematic asserts like that in the tskit test suite.

jeromekelleher:
I don't think there's an obvious right or wrong answer here @petrelharp, so I'm happy to go with your take. My feeling was that we should do our best to minimise breakage though (as being more important that letting people do new things with the 0.x code). I agree it's hard to imagine how people might get caught out by this, but code breakage does tend to be unexpected.

I don't mind breaking the tskit tests, that's easy to fix, I just don't want to break other people's code, who might be doing similar things.

petrelharp:
Well, that's still my gut feeling, but happy to go the other way.

It also seems like a confusing bit of complexity in the documentation.

benjeffery:
This is an instance of a larger question of whether the legacy APIs are "frozen", or just a thin compatibility wrapper around the current, evolving, methods. A similar question might come up with pedigree recording for example?

[...]

PR #1405, opened 2021-01-26
Mutation updates

jeromekelleher:
Some updates to the mutations API for the next 1.0 alpha.

- Sets the default models for sim_mutations to JC69. I really can't see a good reason for defaulting to 0/1 now when we have a chance to fix this. (#1403)
- ~More controversially, discards the mutations times for the deprecated APIs. (#1399).~

~There we some unexpected hiccups with this, where we mix the output msprime.simulate() with the ``keep`` option, which now requires that times be set. So, I *think* this means that old code that does ``msprime.simulate()``, ``msprime.mutate(keep=True)`` will break, which is much worse than the vaguely forseen compatibility problems we might get from having times assigned where there used to be none. So, likely the best thing to do is drop the last commit in this PR, and make some notes on potential compatibility problems.~

~It was a good exercise to go through anyway, I think.~

~@petrelharp, what do you think?~

[EDIT] I've updated this PR to drop the commit in question and tack in another one with some notes to the effect that we can't remove the ``time`` column, and moved it do it's own PR for the record (#1406). Ready for review and merge now I think.

### DIFF
--- a/docs/v0/upgrading.md
+++ b/docs/v0/upgrading.md
@@ -86,6 +86,12 @@ function and is very similar. There are some important differences though:
 * There are now also many new mutation models supported by :func:.sim_mutations;
   see {ref}`sec_mutations` for details. These are *not* supported in the deprecated
   {func}`.mutate` function.
+* The simulated mutations now have a simulated ``time`` value, which specifies the 
+  precise time that the mutation occurred. Note that this value is also provided in the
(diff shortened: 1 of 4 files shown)


petrelharp (approved):
LGTM!

benjeffery (approved):

PR #1407, opened 2021-01-26
Emulate v0.x discrete genome simulations.

jeromekelleher:
Should help to reduce breakage of old code.

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -340,6 +340,7 @@ def _parse_simulate(
                 "equal to the number of populations in from_ts"
             )
 
+    discrete_genome = False
     if recombination_map is None:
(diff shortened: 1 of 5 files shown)


petrelharp:
So you're putting the `discrete_genome` flag back in to simulate, for the pre-alpha users?

jeromekelleher:
> So you're putting the discrete_genome flag back in to simulate, for the pre-alpha users?

No, it's to support old code that made discrete genome simulations by using the trick of making a rate map in which the ``num_loci`` is equal to ``sequence_length``.

petrelharp:
oh! that makes more sense.

jeromekelleher:
@mergifyio rebase

PR #1433, opened 2021-01-29
document population_size

### LINKED ISSUE #1384: population_size parameter is not documented in sim_ancestry
This is particularly confusing as it used to be called Ne, I presume. As well as adding to the doctoring, might it be an idea to allow Ne as a (deprecated) parameter, or at least if Ne is provided, to raise an error to say "use the `population_size` parameter instead"?

awohns:
Will close #1384.

Documents `population_size` in `sim_ancestry()` and notes that `Ne` now has a new name.

Perhaps a bit nitpicky, but I noticed that legacy.md [uses the term "argument' when referring to "parameters,"](https://docs.python.org/3.3/faq/programming.html#what-is-the-difference-between-arguments-and-parameters), so I've changed these

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1072,6 +1072,11 @@ def sim_ancestry(
         greater than or equal to 1. For continuous genomes the tract lengths are
         exponentially distributed with mean ``gene_conversion_tract_length``,
         which must be larger than 0.
+    :param population_size: The size of the default single population
+        :class:`.Demography`. If not specified, defaults to 1. Cannot be specified
+        along with the ``demography`` parameter. See the :ref:`sec_demography`
+        section for more details on demographic models and population sizes
+        and the :ref:`sec_ancestry_population_size` section for usage examples.
     :param int random_seed: The random seed. If this is not specified or `None`,
         a high-quality random seed will be automatically generated. Valid random
         seeds must be between 1 and :math:`2^{32} - 1`.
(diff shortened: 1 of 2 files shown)


msprime/ancestry.py
```diff
@@ -1072,6 +1072,9 @@ def sim_ancestry(
         greater than or equal to 1. For continuous genomes the tract lengths are
         exponentially distributed with mean ``gene_conversion_tract_length``,
         which must be larger than 0.
+    :param population_size: The effective (diploid) population size for the reference
```
jeromekelleher, line 1075: This isn't quite right I think. How about 

> The size of the default single population :class:`.Demography. If not specified, defaults to 1. Cannot be specified along with the ``demography`` parameter. See the :ref:`sec_demography` section for more details on demographic models and population sizes.

jeromekelleher (changes requested):
Looks good, but we need to update the text a bit from the old version. See suggestion above.

msprime/ancestry.py
jeromekelleher: Also we should refer to ``sec_ancestry_population_size`` for examples.

awohns: Ok, so how's this:
> :param population_size: The size of the default single population :class:`.Demography`. If not specified, defaults to 1. Cannot be specified along with the ``demography`` parameter. See the :ref:`sec_demography`section for more details on demographic models and population sizes and the :ref:`sec_ancestry_population_size` section for usage examples.

Should it read `the default single population _of_  :class:`.Demography``?

jeromekelleher: yep, LGTM. I think "single population Demography" is ok, "single population of Demography" wouldn't make sense? (The :class: things are markup)

jeromekelleher (approved):

PR #1434, opened 2021-01-29
Update API for kept mutations in sim_mutations.

### LINKED ISSUE #1206: rename "kept"-related options in sim_mutations
As discussed in #1179, the current API around mutation on top of existing mutations is a bit confusing. One issue is that if you mutate a tree sequence that already has mutations, those mutations silently disappear unless you say `keep=True`, which is easy to forget. Also: it does not make sense for many mutation models to put new mutations *ancestral* to existing ones, so if that could happen then we throw an error, unless the option `kept_mutations_before_end_time` is `True`. But, the name of this option is confusing and unwieldy.

Proposal:
- change `mutate` so that if it is provided with tables that have any mutations in, it throws an error, unless `add=True`
- rename the `kept_mutations_before_end_time` option to `add_ancestral`.

Thoughts, @mufernando? @GertjanBisschop?

jeromekelleher:
Closes #1206

This threw up a problem I hadn't previously considered: what happens when we generate mutations when ``add_ancestral`` is True that won't decode in tskit? It feels like we should be able to rejection sample (or something) so that we don't do this, but I haven't thought hard about it. We should at least detect the problem and raise an error before the users calls ``variants()`` anyway.

@petrelharp, what do you think?

### DIFF
--- a/lib/tests/test_mutations.c
+++ b/lib/tests/test_mutations.c
@@ -1,5 +1,5 @@
 /*
-** Copyright (C) 2016-2020 University of Oxford
+** Copyright (C) 2016-2021 University of Oxford
 **
 ** This file is part of msprime.
 **
@@ -508,7 +508,6 @@ static void
 test_single_tree_mutgen_keep_sites_many_mutations(void)
 {
     int ret = 0;
-    int j;
     gsl_rng *rng = gsl_rng_alloc(gsl_rng_default);
     tsk_table_collection_t tables;
     mutgen_t mutgen;
@@ -520,23 +519,17 @@ test_single_tree_mutgen_keep_sites_many_mutations(void)
     ret = matrix_mutation_model_factory(&mut_model, ALPHABET_NUCLEOTIDE);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    for (j = 0; j < 8192; j++) {
-        ret = tsk_mutation_table_add_row(
-            &tables.mutations, 0, 0, -1, 0.0, "C", 1, NULL, 0);
-        CU_ASSERT_EQUAL_FATAL(ret, j + 1);
-    }
-
     gsl_rng_set(rng, 2);
     ret = mutgen_alloc(&mutgen, rng, &tables, &mut_model, 1);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     ret = mutgen_set_rate(&mutgen, 10);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = mutgen_generate(&mutgen, MSP_DISCRETE_SITES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    for (j = 0; j < 10; j++) {
-        ret = mutgen_generate(&mutgen, MSP_KEEP_SITES);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
-    }
-    CU_ASSERT_TRUE(tables.sites.num_rows > 2);
+    ret = mutgen_generate(&mutgen,
+        MSP_DISCRETE_SITES | MSP_KEEP_SITES | MSP_KEPT_MUTATIONS_BEFORE_END_TIME);
+    CU_ASSERT_EQUAL_FATAL(ret, MSP_ERR_BAD_ANCESTRAL_MUTATION);
 
     mutgen_free(&mutgen);
     mutation_model_free(&mut_model);
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -1201,11 +1201,17 @@ def sim_mutations(
     keep=True)` might insert an X->Z mutation above the existing mutation, thus
     implying the impossible chain X->Y->Z.)  For this reason, if this method
     attempts to add a new mutation ancestral to any existing mutation, an error
-    will occur, unless ``add_ancestral=True``. The ``add_ancestral`` parameter
-    has no effect if ``keep=False``. In summary, to add more mutations to a
-    tree sequence with existing mutations, you need to either ensure that no
-    new mutations are ancestral to existing ones (e.g., using the ``end_time``
-    parameter), or set ``add_ancestral=True``.
+    will occur, unless ``add_ancestral=True``. Furthermore, even if
+    ``add_ancestral`` is True, this will throw an error if there are any
+    mutations that result in a silent transition (e.g., placing a mutation to A
+    above an existing mutation to A), since silent transitions are not allowed
+    in tskit. The ``add_ancestral`` parameter has no effect if ``keep=False``.
+
+    In summary, to add more mutations to a tree sequence with existing
+    mutations, you need to either ensure that no new mutations are ancestral to
+    existing ones (e.g., using the ``end_time`` parameter), or set
+    ``add_ancestral=True`` and be prepared for ocassional errors when silent
+    mutations arise.
 
     :param tskit.TreeSequence tree_sequence: The tree sequence onto which we
         wish to throw mutations.
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -1148,6 +1148,7 @@ def mutate(
         keep=keep,
         start_time=start_time,
         end_time=end_time,
+        add_ancestral=True,
         discrete_genome=False,
     )
 
@@ -1158,11 +1159,11 @@ def sim_mutations(
     *,
     random_seed=None,
     model=None,
-    keep=None,
     start_time=None,
     end_time=None,
     discrete_genome=None,
-    kept_mutations_before_end_time=None,
+    keep=None,
+    add_ancestral=None,
 ):
     """
     Simulates mutations on the specified ancestry and returns the resulting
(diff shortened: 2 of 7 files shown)


petrelharp:
> what happens when we generate mutations when add_ancestral is True that won't decode in tskit? 

What do you mean?

petrelharp:
```
E       _tskit.LibraryError: Inconsistent mutations: state already equal to derived state
```
Oh, I see - this, right?

jeromekelleher:
Yeah - we can add in "A" -> "A" mutations, which will give this error. (Sorry, could have explained that better)

petrelharp:
We definitely thought about this. I think we left it like this because it's not reasonable to do it 'right' in general, so if a user ran into this problem then they got what they asked for and would have to sort it out themselves.

This seems a little unfair - it seems like someone ought to be able to add more parent-independent mutations (eg from JC69) on top of old ones.  This would be trivial if we simulated and allowed non-state-changing mutations: we'd get JC69 by throwing down a Poisson process of ACGT on the tree, and so doing it twice with rate 1 is the same as doing it once with rate 2. However, what *we* do is equivalent to that but followed up by erasing non-state-changing mutations; so if we threw down mutations with rate 1 and got this on a branch:
```
   A --------> A ------> C
```
and then erased the derived A:
```
   A -------------------> C
```
and then threw down more mutations with rate 1:
```
   A --> G ------------> C
```
this would be *incorrect* (ie we'd be under-simulating mutations), because we *should* have
```
   A --> G --> A -----> C
```
So, that rules out "just put down more JC69 and erase mutations with state equal to their parent".

But, rejection sampling? It is true that if you wanted to add more mutations to
```
   A -------------------> C
```
with JC69, then you should (a) place mutations
```
   A ------> ? --------> C
```
and then (b) choose an allele, rejection sampling until you get a G or a T:
```
   A ------> G --------> C
```

However, this isn't always possible: in a binary alleles model, you just can't insert another (single) state change between two alleles:  `0 ------> ? -------> 1`.

So, I think the previous decision about rejection sampling was fine: we'd just need to let the user do it themselves. It wouldn't be hard to go back and fix up the tables for JC69, for instance. (And if we were to do rejection sampling, this would probably be the most efficient way to do it, since usually the situation will be rare.)

But, you're right, we should alert the user. We could easily throw an error or warning [here](https://github.com/tskit-dev/msprime/blob/main/lib/mutgen.c#L1071) when populating the tables, or [here](https://github.com/tskit-dev/msprime/blob/main/lib/mutgen.c#L1221) when generating mutations. How about an error (warning?) that says "Mutating on top of existing mutations resulted in a derived state is equal to ancestral state: the resulting tree sequence will not be valid. If you meant to do this, please file an issue." We could then provide the "go back and rejection sample alleles to fix up the tables" method if someone does need it.

jeromekelleher:
Sounds good @petrelharp, I vote we throw an error with a message saying "Please open an issue if you want us to fix this". I think a warning would be bad because someone could generate a bunch of tree sequences, try to analyse them later and then have no idea why some of them are mysteriously broken (they might not see the warnings: suppose they ran the simulations on a cluster and didn't go to the trouble of looking at the ``.err`` files).

This will be one of those annoying "fails occasionally" problems though...

I'll update the PR with a proposal.

jeromekelleher:
Here's a proposal @petrelharp. It seems a bit weak to me though, just throwing an error, and we'll end up having to do this properly. Up to you whether we should tackle this now or later, though.

jeromekelleher:
Your rejection sampling proposal above seems logical enough @petrelharp - how about we do this in the cases where it makes sense and throw an error if the user tries to put down ancestral mutations in a non-parent independent model (or one with unsufficient alleles)? They're almost certainly doing something wrong, and we can relax the blanket error into a warning ``force_ancestral`` parameter later if it's needed.

So, I think I'm suggesting that we do the right thing by default here with JC69 (rejection sample mutations, and guarantee no exceptions ever occur), get rid of the ``add_ancestral`` option, and error out in any case where bad things might happen. If the need arises, we can implement less stringent criteria and allow the user to shoot themselves in the foot if they *really* want to. 

The vast majority of users will just want to throw JC69 mutations onto their forward simulation though, I'd imagine, which we should make simple and bullet proof.

(Maybe we should move this discussion to an issue, though?)

petrelharp:
Sounds good, with a few caveats:
- we can only check if everything is OK with mutation matrix models
- and so we'll still need the `add_ancestral` option, to e.g. add SLiM mutations above existing ones
- this isn't at the top of my list of things to do

**OH WAIT:** problem. Suppose we have mutations on each of three descendant branches
```
derived_state parent
A  -1
C   0
G   0
T    0
```
and then add a mutation to the parent branch:
```
derived_state parent
A  -1
?   0
C   1
G   1
T    1
```
... there's *no* legal choice for `?`.  The right way to do this is pretty involved: we'd have to also possibly add a reverse mutation (i.e., ?=T but then put in a T->A mutation later) and/or remove the later, pre-existing mutation.

jeromekelleher:
How about we merge this much first here @petrelharp? We can take the follow up to an issue (I'll copy in the relevant comments so we know where we are).

lib/tests/test_mutations.c
```diff
@@ -532,11 +532,8 @@ test_single_tree_mutgen_keep_sites_many_mutations(void)
     ret = mutgen_set_rate(&mutgen, 10);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    for (j = 0; j < 10; j++) {
-        ret = mutgen_generate(&mutgen, MSP_KEEP_SITES);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
-    }
-    CU_ASSERT_TRUE(tables.sites.num_rows > 2);
+    ret = mutgen_generate(&mutgen, MSP_KEEP_SITES);
+    CU_ASSERT_EQUAL_FATAL(ret, MSP_ERR_BAD_ANCESTRAL_MUTATION);
```
petrelharp, line 536: seems like we should still be testing the ability to add mutations - we could do this by adding new mutations with a different alphabet

msprime/mutations.py
```diff
@@ -1183,26 +1184,29 @@ def sim_mutations(
...
+    mutations to a tree sequence with existing mutations must be done with
+    caution, since it can lead to incorrect or nonsensical results if mutation
+    probabilities differ by ancestral state. (As an extreme example, suppose
+    that X->Y and X->Z are allowable transitions, but Y->Z is not. If a branch
+    already has an X->Y mutation on it, then calling `sim_mutations(...,
+    keep=True)` might insert an X->Z mutation above the existing mutation, thus
+    implying the impossible chain X->Y->Z.)  For this reason, if this method
+    attempts to add a new mutation ancestral to any existing mutation, an error
+    will occur, unless ``add_ancestral=True``. The ``add_ancestral`` parameter
+    has no effect if ``keep=False``. In summary, to add more mutations to a
```
petrelharp, line 1205: Edits:
```
Furthermore, even if ``add_ancestral`` is True, this will throw an error if there are any mutations that result in a silent transition (e.g., placing a mutation to A above an existing mutation to A), since silent transitions are not allowed in tskit.
[In summary ... or set ``add_ancestral=True``]
and ensure that silent mutations cannot occur.
```

tests/test_lowlevel.py
```diff
@@ -2562,8 +2562,9 @@ def test_keep_mutations_block_size_mutations(self):
         tables = tskit.TableCollection(1)
         tables.nodes.add_row(0)
         tables.sites.add_row(0, "a")
-        for _ in range(8192):
```
petrelharp, line 2565: I don't really know what this block size thing is doing, but note it's changed from 8192 mutations to three. (If 8192 is important, you could do
```
         for _ in range(4096):
             tables.mutations.add_row(0, node=0, time=0, derived_state="b")
             tables.mutations.add_row(0, node=0, time=0, derived_state="c")
```

petrelharp:
LGTM; some minor suggestions. We could more robustly test adding new mutations with the infinite alleles mutation model, that lets us ensure we don't use the same allele again.

petrelharp:
So - happy to merge when you are.

lib/tests/test_mutations.c
jeromekelleher: We're testing it in other places all right. I tried using a different alphabet here, but we get MSP_ERR_UNKNOWN_ALLELE.

msprime/mutations.py
jeromekelleher: Done, but I don't know how a user could ensure that silent mutations cannot occur (since we require the same alphabet)

jeromekelleher: I've changed the last bit to "and be prepared for ocassional errors when silent mutations arise".

I think would lead to a lot of confusion, but let's get this merged and follow up.

jeromekelleher:
OK, thanks @petrelharp, I've updated with your suggestions. It all seems quite fragile now, tbh, but I'll open the issue to follow up.

(ps. It's helpful if you use the "approve" button on the review thing, if you're happy to merge. That means I can use Mergify stuff rather than needing rebase manually and wait until CI finishes)

msprime/mutations.py
petrelharp: >  I don't know how a user could ensure that silent mutations cannot occur (since we require the same alphabet)

It'd have to be an infinite alleles model (that doesn't look at the parent allele).

Issue #1390, opened 2021-01-21
Add example pedigree files as test data


jeromekelleher:
It would be good to have a set of standard (small) pedigree files that we can use as test data. These can go in ``tests/data/pedigrees``.

@ivan-krukov I think you have some of these lying around?

Issue #1389, opened 2021-01-21
Pedigree.from_ped should add null individual to avoid confusion


jeromekelleher:
When getting pedigree data from a PED file (or other source with 1-based IDs) we should add in a null individual to the tskit individual table to avoid needing to translate back and forth between zero and 1-based IDs.

PR #1439, opened 2021-01-29
Added test pedigrees, with imputed time checks

ivan-krukov:
This adds test pedigrees, resolving #1390.

I think more tests can be added, but these should cover some of the basic structure variations.


jeromekelleher:
Looks good, thanks @ivan-krukov. The only issue here is that the IDs are one-based and we're using 0 as null. I think this could just end up being confusing as test data, since we'll be using 0-based IDs internally.

These aren't strictly like Plink PED files anyway, so I woudn't worry too much about compatibility there. We should certainly provide the functionality to work with 1-based pedigrees (#1389), but we don't need that immediately.

PR #1447, opened 2021-01-31
added information about GSL installation in development section

cdquinto:
There was a to-do box in the development section about instructions to install GSL but those are already in the Installation section. So I added a link to this section in development. Not sure if that's enough.


jeromekelleher (approved):
Yep, LGTM. Thanks @cdquinto!

PR #1466, opened 2021-02-03
Update development.md

hyanwong:
Put ref on one line (it doesn't format correctly otherwise)


jeromekelleher (approved):
Good spot, thanks.

PR #1468, opened 2021-02-04
Clarify reason for full_arg identical topologies

hyanwong:
To a newcomer, it might be a bit confusing that the topology of these 2 trees looks the same, the only difference seemingly being the node labels. Hopefully this text clarifies it a bit?

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -801,10 +801,13 @@ the other identifying the individuals providing the genetic material to the
 right. The effect of this extra node can be seen in the trees: node 9 is
 present as a 'unary' node in the left hand tree and node 10 in the right.
 
-Node 11 has a flags value of 262144 ({data}`.NODE_IS_CA_EVENT`), which
-tells us that it is an ARG common ancestor event that *did not* result
-in marginal coalescence. This class of event also results in unary nodes
-in the trees, which we can see in the example.
+In this particular case, the first thing that happens to these two lineages
+as we go up the tree sequence is that they coalesce straight back together again
+at node 11, forming a (normally undetectable) "diamond event" in the ARG, and
+explaining why the topology of both these trees appears the same. In a full ARG,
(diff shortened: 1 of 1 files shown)


docs/ancestry.md
```diff
@@ -830,10 +830,13 @@ the other identifying the individuals providing the genetic material to the
...
 present as a 'unary' node in the left hand tree and node 10 in the right.
 
-Node 11 has a flags value of 262144 ({data}`.NODE_IS_CA_EVENT`), which
-tells us that it is an ARG common ancestor event that *did not* result
-in marginal coalescence. This class of event also results in unary nodes
-in the trees, which we can see in the example.
+In this particular case, the first thing that happens to these two lineages
+as we go up the tree sequence is that they coalesce straight back together again
+at node 11, forming a (normally undetectable) "diamond event" in the ARG, and
+explaining why the topology of both these trees appears the same. This illustrates
```
jeromekelleher, line 836: This seems to imply to me that all CA nodes are involved in diamonds, though, which isn't true...

hyanwong: That's why I wrote "this illustrates that it is possible ...". But perhaps that's not clear enough. How about "This is one of several ways that a coalescent event can occur which does not result in a marginal coalescence, creating instead an additional unary node in the trees..."?

hyanwong: Changed to a similar wording. Let me know if it's not enough.

jeromekelleher: It still sounds to me like your conflating the diamond event and CA nodes though. CA nodes are recorded for *every* common ancestor event, most of which don't result in coalescence. Occasionally, this will result in a diamond, because it happens to bring back together two lineages that have not experienced any other events since they were split by a recombination.

So, I'd phrase it the other way around. We get the CA event and this *happens* to give us a diamond.

jeromekelleher (approved):

PR #1471, opened 2021-02-06
Multiple mergers note in docs

JereKoskela:
Added a line to the multiple chromosome docs about multiple merger models, which behave quite differently to the standard coalescent in this respect.


jeromekelleher (approved):
LGTM, thanks.

PR #1472, opened 2021-02-06
Demography docs definitely this time

jeromekelleher:


jeromekelleher:
This is ready to go in I think if anyone wants to take a look. The demography API still isn't complete or fully documented, but this is a big step along the way. We're now able to use population name references throughout, and can make more demes-like models. Worth merging this so folks can see how the docs are shaping up while I get the PopulationSplit stuff sorted out.

PR #1481, opened 2021-02-14
Demes lite

jeromekelleher:
Make the msprime demography model more like demes. WIP


jeromekelleher:
Dropping this because we want to keep compatibility with old code as the top priority. Demes will fit in seamlessly, when it's ready.

PR #1482, opened 2021-02-15
Change initial_size -> start_size

jeromekelleher:
This is for compatability with Demes. Get this noisy change out of the way before looking at more interesting stuff like start_time, etc.

### DIFF
--- a/msprime/_msprimemodule.c
+++ b/msprime/_msprimemodule.c
@@ -934,7 +934,7 @@ Simulator_parse_population_configuration(Simulator *self, PyObject *py_pop_confi
 {
     int ret = -1;
     Py_ssize_t j;
-    double initial_size, growth_rate;
+    double start_size, growth_rate;
     int err;
     PyObject *item, *value;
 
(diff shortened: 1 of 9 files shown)


grahamgower:
I wonder if this just adds more confusion, because in Demes, the start_time/end_time terms are using a forwards time convention.

jeromekelleher:
Hmm, good point. We could flip the high-level demography API around so that it sees time going in the same direction as demes? This will be a bit confusing internally for developers in msprime, but would probably be a lot less confusing for users. Keeping things in sync with Demes will probably reduce the sum confusion in the long term, even if it hurts a bit now.

We can still keep old code working, that's not an issue.

jeromekelleher:
But yet, this change is bad then, as we should be changing ``initial_size`` to ``end_size``, not ``start_size``. Thanks for the heads-up!

grahamgower:
I would favour either adopting the Demes conventions/API wholesale, or keeping things largely as they were in msprime 0.7.x. Both of these are going to exist, and both are going to be confusing in different ways (particularly for users already familar with one API). So adding some third system, which cherry picks bits from each side, is probably a bad idea.

petrelharp:
Just a plug for terminology that doesn't depend on which direction in time you are thinking. e.g., `ancestral_size`, or `older_size`, or `longer_ago_size`. (but, consistently using forwards-time words in demes is a great choice)

grahamgower:
> Just a plug for terminology that doesn't depend on which direction in time you are thinking. e.g., `ancestral_size`, or `older_size`, or `longer_ago_size`. (but, consistently using forwards-time words in demes is a great choice)

We had several discussions on this for Demes (during calls, but also across multiple issues and PRs, including https://github.com/popsim-consortium/demes-python/issues/2). I don't think we ever found anything we all liked better than `start_time` and `end_time`using a forwards-time convention. And having `start_size`, `end_size` to match seemed most logical.

jeromekelleher:
I'm going to back out of this change. We want to keep things reasonably compatible with old code (which there is a lot of out there). I'm thinking specifically of stdpopsim. We want to provide a reasonable upgrade path here for models defined in the old style way of doing things. Changing terminology this like is just gratuitous, from that perspective.

PR #1484, opened 2021-02-15
Basic Demes support.

jeromekelleher:
Quick mock-up of what a demes-dependent msprime would look like. I think the proposal would be to make the demography.Demography class an internal API, which basically just existed as a bridge between demes and msprime. We then save some effort by not documenting this class, and instead say "go look at demes" for people wanting to define demographic models.

Any thoughts @grahamgower?

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -33,6 +33,7 @@ import sys
 from typing import ClassVar
 from typing import Union
 
+import demes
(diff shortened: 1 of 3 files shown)


grahamgower:
Looks great!

jeromekelleher:
Closing in favour of #1656

PR #1487, opened 2021-02-16
Link to underlying code

hyanwong:
This is pretty helpful when looking through the docs for stuff that need documenting. We do it in tskit. Is there any reason *not* to show the code in msprime?

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -29,6 +29,7 @@ sphinx:
     - sphinx.ext.autodoc
     - sphinx.ext.autosummary
     - sphinx.ext.todo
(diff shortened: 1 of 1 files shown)


hyanwong:
Note that there is also https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html if we want to link to the github codebase rather than generate the code listing on the fly. But this presumably would mean that "stable" docs would link to the "wrong" (i.e. "latest") code listing, which is not ideal, I presume.

jeromekelleher (approved):

jeromekelleher:
Thanks, this is a handy feature.

PR #1490, opened 2021-02-16
More demography tidyups

jeromekelleher:
This has ended up being quite a big change, but I think the new demography API makes it much harder to make mistakes, and provides a reasonable transition path from old 0.x style demography towards Demes. (I'm thinking in particular of stdpopsim, where we need to provide a reasonably smooth upgrade path).

The new API is best explained with an example. This is the Browning et al model, derived from the [Demes example](https://popsim-consortium.github.io/demes-docs/main/tutorial.html#admixture-events):

```python
T_OOA = 920
# We need to work out the growth rates based on the start and end sizes provided.
N_EUR = 34039
r_EUR = math.log(N_EUR / 1000) / T_OOA
N_EAS = 45852
r_EAS = math.log(N_EAS / 510) / T_OOA

T_ADMIX = 12
N_ADMIX = 54664
r_ADMIX = math.log(N_ADMIX / 30000) / T_ADMIX

demography = msprime.Demography()
demography.add_population(
    name="ANC", description="Ancestral equilibrium population", initial_size=7310
)
demography.add_population(
    name="AMH", description="Anatomically modern humans", initial_size=14474
)
demography.add_population(
    name="OOA", description="Bottleneck out-of-Africa population", initial_size=1861
)
demography.add_population(
    name="AFR", description="African population", initial_size=14474
)
demography.add_population(
    name="EUR",
    description="European population",
    initial_size=N_EUR,
    growth_rate=r_EUR,
)
demography.add_population(
    name="EAS",
    description="East Asian population",
    initial_size=N_EAS,
    growth_rate=r_EAS,
)
demography.add_population(
    name="ADMIX",
    description="Admixed America",
    initial_size=N_ADMIX,
    growth_rate=r_ADMIX,
)
demography.add_admixture(
    T_ADMIX,
    derived="ADMIX",
    ancestral=["AFR", "EUR", "EAS"],
    proportions=[1 / 6, 2 / 6, 3 / 6],
)
demography.set_symmetric_migration_rate(["AFR", "EUR"], 2.5e-5)
demography.set_symmetric_migration_rate(["AFR", "EAS"], 0.78e-5)
demography.set_symmetric_migration_rate(["EUR", "EAS"], 3.11e-5)

demography.add_population_split(T_OOA, derived=["EUR", "EAS"], ancestral="OOA")
demography.add_symmetric_migration_rate_change(
    time=T_OOA, populations=["AFR", "OOA"], rate=15e-5
)
demography.add_population_split(2040, derived=["OOA", "AFR"], ancestral="AMH")
demography.add_population_split(5920, derived=["AMH"], ancestral="ANC")
print(demography.debug())
```
gives:
```
╠═══════════════════════════════╗                                                              
║ Epoch[0]: [0, 12) generations ║                                                              
╠═══════════════════════════════╝                                                              
╟    Populations (total=7 active=4)                                                            
║    ┌───────────────────────────────────────────────────────────────────────────────────┐     
║    │       │     start│       end│growth_rate  │   AFR   │   EUR    │   EAS    │ ADMIX │     
║    ├───────────────────────────────────────────────────────────────────────────────────┤     
║    │    AFR│   14474.0│   14474.0│ 0           │    0    │ 2.5e-05  │ 7.8e-06  │   0   │     
║    │    EUR│   34039.0│   32508.3│ 0.00383     │ 2.5e-05 │    0     │ 3.11e-05 │   0   │     
║    │    EAS│   45852.0│   43238.8│ 0.00489     │ 7.8e-06 │ 3.11e-05 │    0     │   0   │     
║    │  ADMIX│   54664.0│   30000.0│ 0.05        │    0    │    0     │    0     │   0   │     
║    └───────────────────────────────────────────────────────────────────────────────────┘     
╟    Events @ generation 12                                                                    
║    ┌──────────────────────────────────────────────────────────────────────────────────┐      
║    │  time│type       │parameters            │effect                                  │      
║    ├──────────────────────────────────────────────────────────────────────────────────┤      
║    │    12│Admixture  │derived=ADMIX         │Moves all lineages from admixed         │      
║    │      │           │ancestral=[AFR, EUR,  │population 'ADMIX' to ancestral         │      
║    │      │           │EAS]                  │populations. Lineages move to 'AFR'     │      
║    │      │           │proportions=[0.17,    │with proba 0.167; 'EUR' with proba      │      
║    │      │           │0.33, 0.50]           │0.333; 'EAS' with proba 0.5. Set        │      
║    │      │           │                      │'ADMIX' to inactive, and all migration  │                                                                                                     
║    │      │           │                      │rates to and from 'ADMIX' to zero.      │      
║    └──────────────────────────────────────────────────────────────────────────────────┘      
╠═════════════════════════════════╗                                                            
║ Epoch[1]: [12, 920) generations ║                                                            
╠═════════════════════════════════╝                                                            
╟    Populations (total=7 active=3)                                                            
║    ┌─────────────────────────────────────────────────────────────────────────┐               
║    │     │     start│       end│growth_rate  │   AFR   │   EUR    │   EAS    │                                                                                                              
║    ├─────────────────────────────────────────────────────────────────────────┤               
║    │  AFR│   14474.0│   14474.0│ 0           │    0    │ 2.5e-05  │ 7.8e-06  │               
║    │  EUR│   32508.3│    1000.0│ 0.00383     │ 2.5e-05 │    0     │ 3.11e-05 │               
║    │  EAS│   43238.8│     510.0│ 0.00489     │ 7.8e-06 │ 3.11e-05 │    0     │               
║    └─────────────────────────────────────────────────────────────────────────┘               
╟    Events @ generation 920
║    ┌────────────────────────────────────────────────────────────────────────────────────┐    
║    │  time│type            │parameters           │effect                                │    
║    ├────────────────────────────────────────────────────────────────────────────────────┤    
║    │   920│Population      │derived=[EUR, EAS],  │Moves all lineages from derived       │    
║    │      │Split           │ancestral=OOA        │populations 'EUR' and 'EAS' to the    │    
║    │      │                │                     │ancestral 'OOA' population. Also set  │    
║    │      │                │                     │the derived populations to inactive,  │    
║    │      │                │                     │and all migration rates to and from   │    
║    │      │                │                     │the derived populations to zero.      │    
║    │┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈│    
║    │   920│Symmetric       │populations=[AFR,    │Sets the symmetric migration rate     │    
║    │      │migration rate  │OOA], rate=0.00015   │between AFR and OOA to 0.00015 per    │    
║    │      │change          │                     │generation                            │    
║    └────────────────────────────────────────────────────────────────────────────────────┘    
╠═══════════════════════════════════════╗
║ Epoch[2]: [920, 2.04e+03) generations ║
╠═══════════════════════════════════════╝
╟    Populations (total=7 active=2)
║    ┌─────────────────────────────────────────────────────────────┐                           
║    │     │     start│       end│growth_rate  │   OOA   │   AFR   │                           
║    ├─────────────────────────────────────────────────────────────┤                           
║    │  OOA│    1861.0│    1861.0│ 0           │    0    │ 0.00015 │                           
║    │  AFR│   14474.0│   14474.0│ 0           │ 0.00015 │    0    │                           
║    └─────────────────────────────────────────────────────────────┘                           
╟    Events @ generation 2.04e+03
║    ┌────────────────────────────────────────────────────────────────────────────────┐        
║    │  time│type        │parameters           │effect                                │        
║    ├────────────────────────────────────────────────────────────────────────────────┤        
║    │  2040│Population  │derived=[OOA, AFR],  │Moves all lineages from derived       │        
║    │      │Split       │ancestral=AMH        │populations 'OOA' and 'AFR' to the    │        
║    │      │            │                     │ancestral 'AMH' population. Also set  │        
║    │      │            │                     │the derived populations to inactive,  │        
║    │      │            │                     │and all migration rates to and from   │        
║    │      │            │                     │the derived populations to zero.      │        
║    └────────────────────────────────────────────────────────────────────────────────┘        
╠════════════════════════════════════════════╗
║ Epoch[3]: [2.04e+03, 5.92e+03) generations ║
╠════════════════════════════════════════════╝
╟    Populations (total=7 active=1)
║    ┌─────────────────────────────────────────┐                                               
║    │     │     start│       end│growth_rate  │                                               
║    ├─────────────────────────────────────────┤                                               
║    │  AMH│   14474.0│   14474.0│ 0           │                                               
║    └─────────────────────────────────────────┘                                               
╟    Events @ generation 5.92e+03
║    ┌───────────────────────────────────────────────────────────────────────────┐             
║    │  time│type        │parameters      │effect                                │             
║    ├───────────────────────────────────────────────────────────────────────────┤             
║    │  5920│Population  │derived=[AMH],  │Moves all lineages from the 'AMH'     │             
║    │      │Split       │ancestral=ANC   │derived population to the ancestral   │             
║    │      │            │                │'ANC' population. Also set 'AMH' to   │             
║    │      │            │                │inactive, and all migration rates to  │             
║    │      │            │                │and from the derived population to    │             
║    │      │            │                │zero.                                 │             
║    └───────────────────────────────────────────────────────────────────────────┘             
╠═══════════════════════════════════════╗
║ Epoch[4]: [5.92e+03, inf) generations ║
╠═══════════════════════════════════════╝
╟    Populations (total=7 active=1)
║    ┌───────────────────────────────────────┐
║    │     │    start│      end│growth_rate  │
║    ├───────────────────────────────────────┤
║    │  ANC│   7310.0│   7310.0│ 0           │
║    └───────────────────────────────────────┘
```

Things to note:
- Rather than work with objects like Population() and various Event classes, we now call methods on the Demography object. This both reduces boilerplate and makes the model easier to program with.
- We have two new event types PopulationSplit and Admixture. These should be much less confusing an error prone than manually putting things together via MassMigrations (which are still there, if you need them)
- Populations now have a "state machine" where they are active or inactive. It is an error to either sample lineages or move lineages into (via migration or events) a population is not active. This allows us to refine the output in the Debugger and also should make it much harder to make mistakes. The details of the actual state-machine were tedious to work out, but I think it's pretty solid now.

There's quite a few things we could improve and make more "demes-like", but I think changing things any more would move the model too far. We want to provide existing users an opportunity to upgrade their code in a relatively painless way, and ultimately a path to converting to demes.

The PR isn't quite ready to merge as I need to work out the details of how to compare Demographys for equivalence, and in particular how to compare an "new style" model with an old style one (which will be super useful in stdpopsim). It would be good to get some feedback on the overall design (and review, if anyone is up for it), though.

Pinging @grahamgower @apragsdale @molpopgen @petrelharp @gtsambos

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -514,6 +514,7 @@ msp_alloc_populations(msp_t *self)
         self->initial_populations[j].growth_rate = 0.0;
         self->initial_populations[j].initial_size = 1.0;
         self->initial_populations[j].start_time = 0.0;
+        self->initial_populations[j].state = MSP_POP_STATE_ACTIVE;
     }
 out:
     return ret;
@@ -559,8 +560,8 @@ out:
 }
 
 int
-msp_set_population_configuration(
-    msp_t *self, int population_id, double initial_size, double growth_rate)
+msp_set_population_configuration(msp_t *self, int population_id, double initial_size,
+    double growth_rate, bool initially_active)
 {
     int ret = MSP_ERR_BAD_POPULATION_CONFIGURATION;
 
@@ -574,6 +575,8 @@ msp_set_population_configuration(
     }
     self->initial_populations[population_id].initial_size = initial_size;
     self->initial_populations[population_id].growth_rate = growth_rate;
+    self->initial_populations[population_id].state
+        = initially_active ? MSP_POP_STATE_ACTIVE : MSP_POP_STATE_INACTIVE;
     ret = 0;
 out:
     return ret;
@@ -1626,6 +1629,7 @@ msp_print_state(msp_t *self, FILE *out)
     fprintf(out, "]\n");
     for (j = 0; j < self->num_populations; j++) {
         fprintf(out, "pop[%d]:\n", (int) j);
+        fprintf(out, "\tstate        = %d\n", self->populations[j].state);
         fprintf(out, "\tstart_time   = %.14g\n", self->populations[j].start_time);
         fprintf(out, "\tinitial_size = %.14g\n", self->populations[j].initial_size);
         fprintf(out, "\tgrowth_rate  = %.14g\n", self->populations[j].growth_rate);
@@ -1858,6 +1862,11 @@ msp_move_individual(msp_t *self, avl_node_t *node, avl_tree_t *source,
     int ret = 0;
     segment_t *ind, *x, *y, *new_ind;
     double recomb_mass, gc_mass;
+
+    if (self->populations[dest_pop].state != MSP_POP_STATE_ACTIVE) {
+        ret = MSP_ERR_POPULATION_INACTIVE_MOVE;
+        goto out;
+    }
 
     ind = (segment_t *) node->item;
     avl_unlink_node(source, node);
@@ -3418,9 +3427,15 @@ msp_insert_sample(msp_t *self, tsk_id_t node)
 {
     int ret = 0;
     segment_t *root_seg;
+    population_t pop;
 
     root_seg = self->root_segments[node];
-    if (self->populations[root_seg->population].initial_size == 0) {
+    pop = self->populations[root_seg->population];
+    if (pop.state != MSP_POP_STATE_ACTIVE) {
+        ret = MSP_ERR_POPULATION_INACTIVE_SAMPLE;
+        goto out;
+    }
+    if (pop.initial_size == 0) {
         ret = MSP_ERR_BAD_SAMPLES;
         goto out;
     }
@@ -3642,10 +3657,11 @@ msp_reset(msp_t *self)
     for (population_id = 0; population_id < (population_id_t) N; population_id++) {
         pop = self->populations + population_id;
         /* Set the initial population parameters */
-        initial_pop = self->initial_populations + population_id;
+        initial_pop = &self->initial_populations[population_id];
         pop->growth_rate = initial_pop->growth_rate;
         pop->initial_size = initial_pop->initial_size;
         pop->start_time = self->time;
+        pop->state = initial_pop->state;
     }
     /* Reset the tables to their correct position for replication */
     ret = tsk_table_collection_truncate(self->tables, &self->input_position);
@@ -3713,9 +3729,14 @@ msp_initialise_simulation_state(msp_t *self)
             min_root_time = GSL_MIN(node_time[root], min_root_time);
         }
     }
-    /* Make sure that the start-time is no less than the time of the
-     * youngest root */
-    self->start_time = GSL_MAX(self->start_time, min_root_time);
+    if (self->input_position.nodes == 0) {
+        /* When we're initialising for debugging we provide no samples. */
+        self->start_time = 0;
+    } else {
+        /* Make sure that the start-time is no less than the time of the
+         * youngest root */
+        self->start_time = GSL_MAX(self->start_time, min_root_time);
+    }
 
     /* Anything that has min_root_time is an initial sample, otherwise we
      * register them as sampling events */
@@ -5467,8 +5488,8 @@ msp_get_num_migration_events(msp_t *self, size_t *num_migration_events)
 }
 
 int MSP_WARN_UNUSED
-msp_get_population_configuration(
-    msp_t *self, size_t population_id, double *initial_size, double *growth_rate)
+msp_get_population_configuration(msp_t *self, size_t population_id, double *initial_size,
+    double *growth_rate, int *state)
 {
     int ret = 0;
     population_t *pop;
@@ -5480,6 +5501,7 @@ msp_get_population_configuration(
     pop = &self->populations[population_id];
     *initial_size = pop->initial_size;
     *growth_rate = pop->growth_rate;
+    *state = pop->state;
 out:
     return ret;
 }
@@ -5823,6 +5845,43 @@ out:
     return ret;
 }
 
+static void
+msp_deactivate_population(msp_t *self, int population_id)
+{
+    population_t *pop = &self->populations[population_id];
+
+    tsk_bug_assert(pop->state == MSP_POP_STATE_ACTIVE);
+    tsk_bug_assert(msp_get_num_population_ancestors(self, population_id) == 0);
+    pop->state = MSP_POP_STATE_PREVIOUSLY_ACTIVE;
+    /* Set these to zero for tidyness sake */
+    pop->initial_size = 0;
+    pop->growth_rate = 0;
+}
+
+static int
+msp_activate_population(msp_t *self, int population_id)
+{
+    int ret = 0;
+    population_t *pop = &self->populations[population_id];
+
+    /* It's not currently possible to do this because we're only calling
+     * msp_activate_population from population split, where we only activate
+     * if it's not already active. But, leaving this here as a reminder
+     * in case we ever do call activate from somewhere else.
+     */
+    /* if (pop->state == MSP_POP_STATE_ACTIVE) { */
+    /*     ret = MSP_ERR_POPULATION_CURRENTLY_ACTIVE; */
+    /*     goto out; */
+    /* } */
+    if (pop->state == MSP_POP_STATE_PREVIOUSLY_ACTIVE) {
+        ret = MSP_ERR_POPULATION_PREVIOUSLY_ACTIVE;
+        goto out;
+    }
+    pop->state = MSP_POP_STATE_ACTIVE;
+out:
+    return ret;
+}
+
 /* Population split */
 
 static int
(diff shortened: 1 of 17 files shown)


jeromekelleher:
I've ended up putting an ``active`` flag on Populations. The idea is that the population split events tell us when populations should and shouldn't have lineages in them, and we set a flag accordingly during the simulation. Then, if the user tries to either migrate lineages or take some samples from an inactive population, an error is raised.

Seems like this will help reduce errors?

@grahamgower, any thoughts? The idea is that old code will continue to work as before, but the reward for using population splits is that your code is more robust. This should fit well with Demes integration too.

I'm hoping to use the "active" status of populations to make the debugger output more helpful, also help with the ``is_equivalent`` operation (which I've started here). The next step after once this is all working is to try out the new code in stdpopsim, and see how much breaks or not.

grahamgower:
This will definitely reduce user errors, so I wholeheartedly support this change! And I agree that the debugger output could also be made less confusing using this. I'm curious though, why did you choose to add an `active` field rather than basing this purely off the population size (`active == N > 0`)?

jeromekelleher:
> This will definitely reduce user errors, so I wholeheartedly support this change! And I agree that the debugger output could also be made less confusing using this.

Great!

> I'm curious though, why did you choose to add an active field rather than basing this purely off the population size (active == N > 0)?

I tried initially, but I couldn't figure out a way to give good error messages. It seemed less fiddly to just add a flag than to try and reason about zero-growth-rate + zero-pop-size.

msprime/demography.py
```diff
@@ -1282,17 +1639,38 @@ def __post_init__(self):
...
         }
 
+    def _assert_equivalent(
+        self,
+        other: StateChangeEvent,
+        *,
+        rel_tol: float,
+        abs_tol: float,
+    ):
+        super()._assert_equivalent(other, rel_tol=rel_tol, abs_tol=abs_tol)
```
grahamgower, line 1660: Is this supposed to be here? I might be missing something, but I didn't see this method in the super class hierarchy.

gtsambos:
The example looks great, especially with the debugger output; it's clear and readable. I'll have a closer look at the code now.

msprime/demography.py
jeromekelleher: No, good spot. I've gone back and forth on how to do the ``equivalent`` stuff. Will finalise today, hopefuly.

lib/msprime.c
```diff
@@ -5928,9 +5994,121 @@ msp_add_population_split(
     de->params.population_split.num_derived = num_derived;
     de->change_state = msp_population_split;
     de->print_state = msp_print_population_split;
+    /* Set the ancestral population state to not active before the event. */
+    self->initial_populations[ancestral].state = MSP_POP_STATE_INACTIVE;
```
gtsambos, line 5998: I'm a bit confused about why the initial state of the ancestral population is set to be inactive only at this point in the simulation. Once the split has been simulated, the ancestral population becomes active, so it's too late to find any bugs to do with the initial inactive state, right?

lib/msprime.c
```diff
@@ -5823,6 +5843,18 @@ msp_add_mass_migration(
     return ret;
 }
 
+static void
+msp_deactivate_population(msp_t *self, int population_id)
```
gtsambos, line 5847: Just thinking out loud -- are there any 'global' checks that need to be done before you deactivate a population, perhaps to do with the populations that the extant lineages belong to? If so they'd go in here, right?

lib/msprime.c
```diff
@@ -514,6 +514,7 @@ msp_alloc_populations(msp_t *self)
         self->initial_populations[j].growth_rate = 0.0;
         self->initial_populations[j].initial_size = 1.0;
         self->initial_populations[j].start_time = 0.0;
+        self->initial_populations[j].state = MSP_POP_STATE_ACTIVE;
```
gtsambos, line 517: I take it from this that all of the specified populations are initially active by default? What would happen if we specified some initial samples from a population that we think is an ancestral one (because, eg., it is the parent pop in a split event later on). From what I can see this wouldn't lead to an error even though it probably should -- the initial state of the ancestral pop is set to INACTIVE in the split event, but it doesn't retroactively check that there aren't already samples from the ancestral pop. (I've just skimmed the code so might be missing something here.)

lib/msprime.c
gtsambos: Also, is there a reason why there isn't also an an msp_activate_population function? Because I can imagine some general assertions that might need to be checked before doing this too -- for example, if there are already samples in the simulation belonging to the population to be 'activated'

lib/msprime.c
jeromekelleher: This is done *before* the simulation is actually run. All the events are added before we run ``initialise()``, and run later.

jeromekelleher:
Thanks @gtsambos - I'll go through your comments tomorrow AM

lib/msprime.c
gtsambos: Ah, okay!

lib/msprime.c
jeromekelleher: Good ideas - I've added a bug_assert for the number of lineages in the deactivated pop == 0 and added a "activate" population. In the interest of keeping things simple I've focused on just maintaining the state machine here - we catch all the other errors if we try to migrate/sample any lineages into a populations where state != active.

lib/msprime.c
jeromekelleher: Yes, all populations are initially active by default. If we try to take samples at a time when a populations is inactive we get the MSP_ERR_POPULATION_INACTIVE_SAMPLE error. Rather than try to catch these errors at the start of the simulation we just catch them when they happen - that's why there's no logic here trying to think about samples etc.

jeromekelleher:
Thanks for the comments @gtsambos, very helpful. (I deleted one comment as it was a dupe of another one)

jeromekelleher:
This has ended up being a massive code drop, but I wanted to make sure that we had all the tools we need for implementing/converting stdpopsim models as we transition to Demes there. The idea is that we now have a path from the "old style" models to much more readable new style models, and a reasonably robust method of comparison using ``is_equivalent``. The ``population_map`` argument in ``Demography.from_old_style`` let's us define a mapping from the implicit populations in the old style definition to the explicit ones in the "demesish" model.

I don't think anyone is going to want to go through all of this, but I'll ping @petrelharp, @apragsdale and @grahamgower here just in case they really hate the changes. If no one complains I'll merge this tomorrow, and push out the next alpha and see how stdpopsim integration goes.

If it's looking OK, then I think it's just documentation and we're good to go beta.

apragsdale (approved):
Nothing pops out to me, though I skimmed quite a bit of it.. great work here, these will be awesome changes!

PR #1512, opened 2021-03-05
Demography numerical updates

jeromekelleher:
Brings the ``lineage_movements`` method up to date. Pinging @apragsdale and @petrelharp for review here.

Does the new sample specification make sense to you? What about the semantics of the no-parameters case? Do we need to update the other methods, or are they all OK?


petrelharp (approved):
Looks great, and makes sense! Two minor suggestions.

Issue #1077, opened 2020-06-16
Add model argument to msp mutate CLI


jeromekelleher:
We added a CLI for mutate in #1076. This doesn't have any way to specify the model. Once #1070 is done as we have a way to create non-parametric models from strings like "HKY", etc, we can add a model argument here.

We should also think about whether we want to support the more complex parametric models (probably not?)

jeromekelleher:
@winni2k, can we sign you up for this?

winni2k:
Hehe, sure ;)

jeromekelleher:
@winni2k, any update on when you might get to this?

winni2k:
Thanks for the ping. I'll give it a shot this weekend.

jeromekelleher:
@winni2k, hows this looking? We're going to feature freeze at the end of next week, so we need to get it done soon.

winni2k:
Right-o. I'll sort this out tomorrow then.
________________________________

PR #1519, opened 2021-03-08
Add CLI hook for new mutation models

winni2k:
Implements #1077

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1240,75 +1240,100 @@ class TestMspArgumentParser:
         assert args.random_seed == 123
         assert args.compress
 
+    def test_mutate_default_values(self):
+        parser = cli.get_msp_parser()
+        cmd = "mutate"
+        args = parser.parse_args([cmd, "out.trees", "out2.trees"])
+        assert args.tree_sequence == "out.trees"
+        assert args.output_tree_sequence == "out2.trees"
+        assert args.mutation_rate == 0
+        assert args.random_seed is None
+        assert args.start_time is None
+        assert args.end_time is None
+        assert args.model == "jc69"
 
-class TestMspSimulateOutput(unittest.TestCase):
+    def test_mutate_raises_with_invalid_model_name(self):
+        parser = cli.get_msp_parser()
+        cmd = "mutate"
+        with pytest.raises(SystemExit):
+            parser.parse_args(
+                [cmd, "--model", "not-a-model", "out.trees", "out2.trees"]
+            )
+
+    def test_mutate_model_is_changeable(self):
+        parser = cli.get_msp_parser()
+        cmd = "mutate"
+        for model_arg in ["--model", "-m"]:
+            args = parser.parse_args(
+                [cmd, "out.trees", "out2.trees", model_arg, "blosum62"]
+            )
+            assert args.model == "blosum62"
+
+
+class TestMspSimulateOutput:
     """
     Tests the output of msp to ensure it's correct.
     """
 
-    def setUp(self):
-        fd, self._tree_sequence = tempfile.mkstemp(prefix="msp_cli", suffix=".trees")
-        os.close(fd)
-
-    def tearDown(self):
-        os.unlink(self._tree_sequence)
-
-    def test_run_defaults(self):
+    def test_run_defaults(self, tmp_path):
         cmd = "simulate"
         sample_size = 10
+        tree_sequence_file = str(tmp_path / "out.ts")
         stdout, stderr = capture_output(
-            cli.msp_main, [cmd, str(sample_size), self._tree_sequence]
+            cli.msp_main, [cmd, str(sample_size), tree_sequence_file]
         )
         assert len(stderr) == 0
         assert len(stdout) == 0
 
-        tree_sequence = tskit.load(self._tree_sequence)
+        tree_sequence = tskit.load(tree_sequence_file)
         assert tree_sequence.get_sample_size() == sample_size
         assert tree_sequence.get_sequence_length() == 1
         assert tree_sequence.get_num_mutations() == 0
 
-    def test_simulate_short_args(self):
+    def test_simulate_short_args(self, tmp_path):
         cmd = "simulate"
+        tree_sequence_file = str(tmp_path / "out.ts")
         stdout, stderr = capture_output(
             cli.msp_main,
-            [cmd, "100", self._tree_sequence, "-L", "1e2", "-r", "5", "-u", "2"],
+            [cmd, "100", tree_sequence_file, "-L", "1e2", "-r", "5", "-u", "2"],
         )
-        tree_sequence = tskit.load(self._tree_sequence)
+        tree_sequence = tskit.load(tree_sequence_file)
         assert len(stderr) == 0
         assert len(stdout) == 0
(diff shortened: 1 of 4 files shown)


winni2k:
@jeromekelleher please have a look. This only implements support for models that do not take any parameters. 

If we want models with parameters as well, then I suggest that we don't specify the parameters on the command line, as it looks like some models take more than just one or two parameters (correct me if I'm wrong). Instead, it might be better to supply a yaml file with the model specification. We would then need a way to validate the model parameters, ideally through introspection so that we can get useful error messages.

winni2k:
Is there a way to kick off the tests again? I don't think the failure has anything to do with the last commit.

tests/test_cli.py
```diff
@@ -1240,6 +1240,28 @@ def test_simulate_long_args(self):
...
 
+    def test_mutate_default_values(self):
+        parser = cli.get_msp_parser()
+        cmd = "mutate"
+        args = parser.parse_args([cmd, "out.trees", "out2.trees"])
+        assert args.tree_sequence == "out.trees"
+        assert args.output_tree_sequence == "out2.trees"
+        assert args.mutation_rate >= 0
+        assert args.random_seed is None
+        assert args.keep in (True, False)
```
jeromekelleher, line 1251: We should take a position here --- the default should be True

tests/test_cli.py
```diff
@@ -1240,6 +1240,28 @@ def test_simulate_long_args(self):
         assert args.random_seed == 123
         assert args.compress
 
+    def test_mutate_default_values(self):
+        parser = cli.get_msp_parser()
+        cmd = "mutate"
+        args = parser.parse_args([cmd, "out.trees", "out2.trees"])
+        assert args.tree_sequence == "out.trees"
+        assert args.output_tree_sequence == "out2.trees"
+        assert args.mutation_rate >= 0
```
jeromekelleher, line 1249: Should test for exactly what the default value is.

tests/test_cli.py
```diff
@@ -1240,6 +1240,28 @@ def test_simulate_long_args(self):
...
+        assert args.random_seed is None
+        assert args.keep in (True, False)
+        assert args.discrete_genome in (True, False)
+        assert args.start_time is None
+        assert args.end_time is None
+        assert args.model == "jc69"
+
```
jeromekelleher, line 1256: Another quick test here would be helpful:
```
def test_mutate_model(self):
     parser = cli.get_msp_parser()
￼    cmd = "mutate"
    args = parser.parse_args([cmd, "out.trees", "out2.trees" --model=blosum62])
    assert args.model == "blosum62"
    # also test the short form

**Review by jeromekelleher — COMMENTED** [review]
Looks great, thanks @winni2k!

Needs a couple more tests in the (badly named) TestMspConversionOutput class I think:
```
def test_mutate_binary_model(self):
    # looks much like test_mutate_discrete_start_end_time but we test if we get binary
    # mutations, i.e. ``set(tables.sites.ancestral_state) == {'0'}`` and set(tables.mutations.derived_state) == {'1'}
```

It would be nice to test that we simulate discrete sites ACGT mutations by default too.

The test class should really be renamed to TestMspMutateOutput, and the setup ts changed to use the new APIs so that we make a discrete tree sequence by default. There's no need for recombination in this ts either, it's a waste of time simulating recombination in this case.

If you're feeling super motivated you could refactor this small corner of the test suite to use pytest idiomatically, but that's only if you *really* don't like doing unittest stuff.

Sound OK?

**Comment by winni2k** [comment]
Thanks for the comments! 

Is the purpose of fixing the default arguments of the CLI in a test to make sure that any changes to default arguments have to be made in two places? That seems reasonable to me, but I had not considered this before.

Regarding https://github.com/tskit-dev/msprime/pull/1519/commits/531f6acaafdb93ba5bdd9cc1d2fae5bd3ce64f1e: I would expect a factory to always generate new objects. Right now, depending on the input, the mutation model factory might instantiate a new object or just return an already existing one. That seems like surprising behavior. Is there a particular reason for this? Would the model factory be able to instantiate every type of mutation model right now?

**Comment by jeromekelleher** [comment]
> Is the purpose of fixing the default arguments of the CLI in a test to make sure that any changes to default arguments have to be made in two places? That seems reasonable to me, but I had not considered this before.

Exactly - it's to prevent accidental changes to the defaults.

> Regarding 531f6ac: I would expect a factory to always generate new objects. Right now, depending on the input, the mutation model factory might instantiate a new object or just return an already existing one. That seems like surprising behavior. Is there a particular reason for this? Would the model factory be able to instantiate every type of mutation model right now?

"Factory" might not be the best name. The function is really just a way of implementing the type polymorphism needed for accepting ``model="jc69"`` and ``model=JukesCantor69()``. In the "instance of MutationModel" case, we should probably be returning a copy of the model, just in case we modify anything about it later (which we probably don't, though). It would probably be a good idea to just do a ``model = copy.deepcopy(model)`` here, just for consistency.

Would that make more sense?

File: tests/test_cli.py (hunk 1 above)
Comment (winni2k) [reply tests/test_cli.py]: Are you sure? The current default is definitely False. Changing it to true would mean making "--keep" a no-op and introducing a new flag to turn off keeping -- i.e something like "--discard" or "--no-keep"

Comment (jeromekelleher) [reply tests/test_cli.py]: Sorry, I meant --discrete

**Comment by jeromekelleher** [comment]
Ah, I see now that the defaults are actually the wrong way around!

So, I think the simplest thing to do is to drop the ``--keep`` and ``--discrete`` options - by default we want the simulation to keep existing mutations and to run on a discrete genome. We can add options to do the opposite later, but for now we just want to make sure that the defaults are right for 1.0.

**Comment by winni2k** [comment]
Alright, I think this is ready for a final review.

File: tests/test_cli.py [hunk 4]
```diff
@@ -1240,79 +1240,105 @@ def test_simulate_long_args(self):
...
         assert tree_sequence.get_sample_size() == 10
 
 
-class TestMspConversionOutput(unittest.TestCase):
-    """
-    Tests the output of msp to ensure it's correct.
-    """
+@pytest.fixture
+def unmutated_tree_sequence():
+    return msprime.simulate(
```
jeromekelleher, line 1318: might as well make this ``sim_ancestry``, since we're living in the future

jeromekelleher (approved):
Looks great @winni2k, thanks! Ready to merge after the sim_ancestry nitpick and a squash.

tests/test_cli.py
winni2k: Is this by any chance related to also setting `discrete=True` by default? 

The doc string for `sim_mutations` honestly reads like the `keep=True` argument is broken: It might work, but it might not, and if it does not work you'll get an exception, but there won't be much you can do to make it work.

winni2k: I have undone a few changes that I made previously. Test fixtures with mutations now still have the mutations. Four tests fail with errors about adding mutations, but I have marked them as xfail. I think that makes more sense?

jeromekelleher: Yep, perfect. Ready to merge after a squahs.

winni2k:
Squash complete. Ready to merge when tests complete.

jeromekelleher (approved):

PR #1521, opened 2021-03-09
Change end time inequality

### LINKED ISSUE #1304: Behavior of `end_time` when using "dtwf" model
I'm confused about the behavior of `end_time` when the model is set to "dtwf". When simulating using `sim_ancestry`, if I set the `end_time=1`, I figured the simulation would run for a single generation. I can see this in the "hudson" model, running `sim_ancestry` with `end_time=1` gives me multiple trees. However, simulating with "dtwf" and `end_time=1` results always in a single tree, so that nothing has occurred. Instead, specifying `end_time=2` with the "dtwf" model results in roughly the same distribution of the number of trees as with the "hudson" model and `end_time=1`.

Is this the intended behavior, or is the "dtwf" model ending a generation too early when `end_time` is specified?

Edit, as an illustration, with 2 samples, Ne=1000, a segment of 1 Morgan, with 20,000 replicates:
num_trees

apragsdale:
Closes #1304

Some tests are still failing here.. trying to figure out what's going on

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@
 - The `msp` CLI has been stripped of all sub-commands except for
   `simulate` and `mutate`. These sub-commands are provided by the `tskit`
   CLI or the `TreeSequence` API in `tskit`.
+- The `end_time` option now allows events up to and including the specified
+  max time. Previously, events occurred strictly before the max time.
+  ({user}`apragsdale`, {issue}`1304`, {pr}`1521`)
 
 **Deprecations**:
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -4229,7 +4229,7 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
         if (sampling_event_time < t_temp
             && sampling_event_time < demographic_event_time) {
             se = &self->sampling_events[self->next_sampling_event];
-            if (se->time >= max_time) {
+            if (se->time > max_time) {
                 ret = MSP_EXIT_MAX_TIME;
                 break;
             }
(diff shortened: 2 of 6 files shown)


apragsdale:
I fiddled with tests a bit to get them to pass. There were a few testing the timing of some events when it equaled the set end time, of course, so would be good to check that I haven't made those tests useless.

Also, two tests were failing in lib/tests/ancestry.c, and I'm wondering if lines 373 and 489 should be changed to `ret = msp_run(&msp, 10.0-1e-6, ULONG_MAX);`..?

jeromekelleher:
Excellent, thanks @apragsdale! Yes, I think you're right we can change those C tests. This does mean that our semantics wrt events like sampling etc is slightly different than before, I guess it's unlikely anyone is actually depending on this. This is the sort of stuff we want to break for 1.0 anyway, while we're breaking stuff.

We should add a note to the CHANGELOG to make sure we don't forget to tell people about it, though.l

apragsdale:
Ok, changelog updated. Let me know if you want any changes made here, or if I should rebase/etc

jeromekelleher:
Looks great, thanks @apragsdale. I think we just need to update the documentation for the ``end_time`` parameter, squash and we're good to merge.

apragsdale:
Is this ok to merge? Not sure what to do about the one codecov failuer here.

jeromekelleher:
@mergifyIO please rebase

jeromekelleher:
@Mergifyio  rebase

jeromekelleher:
@Mergifyio rebase

jeromekelleher (approved):
LGTM, thanks @apragsdale!

apragsdale:
Thanks Jerome - glad to get this sorted!

jeromekelleher:
Thanks for sorting it out @apragsdale!

PR #1522, opened 2021-03-10
Add missing matplotlib dev requirement to compile docs locally

winni2k:
I had to do this extra step to get the docs compiled locally on my machine.

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -4,6 +4,7 @@ codecov
 coverage
 daiquiri
 flake8
+matplotlib
 mock
 mypy
 numpy


docs/development.md
```diff
@@ -128,7 +128,15 @@ and the local development version of msprime.
 (sec_development_documentation_building)=
 ### Building
 
-To build the documentation locally, go to the `docs` directory and run `make`.
+To build the documentation locally, follow the 
+{ref}`Quickstart <sec_development_quickstart>`. Then install the python
+requirements for building the documentation:
+
+```bash
+pip install -r requirements/CI-docs/requirements.txt
```
jeromekelleher, line 136: Hmm, not sure we want to do this. The CI requirements are all pinned, which will cause issues for some people.

We should just add the packages needed to the ``requirements/development.txt`` - which ones were missing?

jeromekelleher (commented):
LGTM, but see comment.

docs/development.md
winni2k: matplotlib was the one I saw. There might be more. I'll debug and find out.

winni2k:
Ok, it looks like adding matplotlib to the dev requirments is sufficient to build all the docs on my machine.

jeromekelleher (approved):
LGTM, thanks.

Issue #1207, opened 2020-09-24
deprecation warnings don't use the DeprecationWarning category


grahamgower:
Deprecation warnings from, e.g. the DeprecationFacade and recombination map's num_loci, use `warnings.warn`, but don't set `category=DeprecationWarning`. This is a fairly easy change, but maybe there's a reason they don't use this warning category? An argument in favour of using the category is for easy filtering in downstream applications (it's nice to see these as an application developer, but less nice if users start seeing them when they upgrade a dependency).

petrelharp:
I ended up using FutureWarning instead of DeprecationWarning in pyslim because the latter didn't actually appear. This is by design: they differ in [intended audience](https://docs.python.org/3/library/warnings.html); if you want a warning that makes users change their scripts, you want FutureWarning, I believe. DeprecationWarning and FutureDeprecationWarning are both ignored by default.
![Screenshot from 2020-09-24 10-12-33](https://user-images.githubusercontent.com/1046249/94177578-9d688b80-fe4e-11ea-8d82-a0f6518d4099.png)

jeromekelleher:
#1194 is related.

jeromekelleher:
Closed in #1536

PR #1536, opened 2021-03-15
Fix error message on missing msprime attribute

### LINKED ISSUE #1194: Reinstate specific errors for old classes
The current fall-back of looking up tskit where there's a name error in msprime leads to some confusing error messages. For example, if I write
```
>>> msprime.simlate
  File "/home/jk/work/github/msprime/msprime/__init__.py", line 61, in __getattr__
    result = getattr(tskit, name)
AttributeError: module 'tskit' has no attribute 'simlate'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/home/jk/work/github/msprime/msprime/__init__.py", line 67, in __getattr__
    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
AttributeError: module 'msprime' has no attribute 'simlate'
```

This will catch people out, as they'll look at top error message first and then wonder what this has to do with tskit (it's certainly tripped me up a couple of times now).

We should explicitly list the things we expect to deprecate and keep the name resolution at the top-level straightforward.

benjeffery:
Fixes #1194,#1207

### DIFF
--- a/msprime/__init__.py
+++ b/msprime/__init__.py
@@ -51,20 +51,61 @@ old_module = sys.modules["msprime"]
 
 # Many attributes were moved to tskit, we use a facade here so we don't break old
 # code, but do emit a warning
+msprime_names_now_in_tskit = [
+    "Edge",
+    "EdgeTable",
+    "Edgeset",
+    "FORWARD",
+    "Individual",
+    "IndividualTable",
+    "LdCalculator",
+    "Migration",
+    "MigrationTable",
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM. Should we fix #1207 here too, or is that separate thing?

benjeffery:
Good shout, I've added that in.

jeromekelleher (approved):
Excellent!

PR #1539, opened 2021-03-15
Add Demography.from_tree_sequence.

### LINKED ISSUE #1528: `sim_ancestry(initial_state=ts)` does not use population names
Right now, it looks like `sim_ancestry(initial_state=ts, demography=d, ...)` matches up the populations in the Demography object, in order, to the populations in the initial state. The population table in the output is the same as in the initial state (this is good!). However, there's the possibility for confusion here, because names are not matched. For instance:
```
t = tskit.TableCollection(sequence_length=1)
t.populations.metadata_schema = tskit.MetadataSchema({ "codec": "json" })
t.populations.add_row(metadata={'name' : 'A'})
t.populations.add_row(metadata={'name' : 'B'})
t.nodes.add_row(time=0, flags=1, population=0)
t.nodes.add_row(time=0, flags=1, population=0)
t.nodes.add_row(time=0, flags=1, population=1)
t.nodes.add_row(time=0, flags=1, population=1)

d = msprime.Demography()
d.add_population(initial_size=0.001, name='B')
d.add_population(initial_size=1e6, name='A')

ts = msprime.sim_ancestry(
        initial_state=t.tree_sequence(),
        demography=d,
        end_time=1)

print(ts.first().draw(format='unicode'))
```
shows that 0 and 1, the samples in `A`, have coalesced, and so have been assigned to what we called `B` in the Demography.
```
 5  7 6
 ┃  ┃ ┃
 4  ┃ ┃
┏┻┓ ┃ ┃
0 1 2 3
```
(and, the populations are still labeled `A` and `B`, as before).

BTW, if the demography has one more population than the tables, we get this error:
```
msprime._msprime.InputError: Input error in population configuration: Population ID out of bounds.
```
and if the tables have one more population than the demography, we get:
```
ValueError: migration matrix must be a N x N square matrix encoded as a list-of-lists or numpy array, where N is the number of populations defined in the population_configurations. The diagonal elements of this matrix must be zero. For example, a valid matrix for a 3 population system is [[0, 1, 1], [1, 0, 1], [1, 1, 0]]
```

jeromekelleher:
Closes #1528

WIP, but the basic stuff is there. Will need to figure out how to deal with metadata when appending to the population table.

@petrelharp, what do you think?

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -958,58 +958,228 @@ not ancestral to these census nodes.
 
 ## Specifying the initial state
 
+By default `msprime` simulations are initialised by specifying a set of
+{ref}`samples<sec_ancestry_samples>`,
+which sets up the simulation with segments of ancestral material covering the
+whole sequence. Internally, this initial state is implemented by
+setting up a {class}`tskit.TableCollection` which contains the
+{ref}`population<tskit:sec_population_table_definition>`,
+{ref}`node<tskit:sec_node_table_definition>`,
+and {ref}`individual<tskit:sec_individual_table_definition>` information.
+The low-level simulation code then reads these tables and sets up the
+simulation accordingly. We can see what this initial state looks like
+by setting up a simple simulation and stopping it immediately:
 
 ```{code-cell}
-:tags: [remove-cell]
+initial_ts = msprime.sim_ancestry(1, end_time=0)
+initial_ts
+```
+We can see that the returned tree sequence has one population,
+two sample nodes and one individual (since sample individuals are
+{ref}`diploid<sec_ancestry_ploidy>` by default). There are no
+{ref}`edges<tskit:sec_edge_table_definition>` because the simulation
+didn't run for any period of time: we're essentially just returning
+the initial state used to start the simulation.
 
-    msprime.core.set_seed_rng_seed(42)
+Then, running a simulation in which the ``initial_state`` is equal to
+``initial_ts`` is *precisely* the same as running the simulation
+with the same arguments as we used above:
+```{code-cell}
+ts1 = msprime.sim_ancestry(initial_state=initial_ts, population_size=100, random_seed=2)
+ts2 = msprime.sim_ancestry(1, population_size=100, random_seed=2)
+assert ts1.equals(ts2, ignore_provenance=True)
 ```
 
-```{eval-rst}
-.. todo:: Go through an example where we specify the initial state
-    and explain that specifying the samples is actually the same thing.
+:::{warning}
+Note that the ``initial_state`` tree sequence only contains the
+information about the ancestral histories, and not any of the
+simulation parameters. In the previous example we used default
+parameters to keep things simple, but in general the full
+simulation model will need to be specified as well as the initial state.
+In particular, see the {ref}`sec_ancestry_initial_state_demography` section.
+:::
 
+While it is possible to specify the initial state manually by
+constructing the input tree sequence,
+it is an advanced topic and rarely necessary. More often we are
+interested in specifying the initial state of an ``msprime``
+simulation using the output of *another* simulation.
+
+:::{note}
+The ``initial_state`` tree sequence is quite literally the initial
+state of the tree sequence that is returned at the end of a simulation,
+and that any information present (including metadata) will not be
+altered.
+:::
+
+### Combining backward-time simulations
+
+Suppose that we wished to simulate a scenario
+in which we have recombination happening at some rate in the recent past,
+and in the more distant past there is no recombination.
+We can do this by combining the simulations,
+using one as the ``initial_state`` for the other.
+
+We first run the simulation of the recent past:
+
+```{code-cell}
+ts1 = msprime.sim_ancestry(
+    2, recombination_rate=0.01, sequence_length=10, end_time=10,
+    population_size=100, random_seed=1)
+SVG(ts1.draw_svg(y_axis=True))
 ```
 
-```{eval-rst}
-.. todo:: Port this old documentation to the new format. The sectioning
-    is definitely not right here also, as we've just pasted in the old
-    tutorial content directly in here.
+Some recombination and coalescence has happened and we have three
+marginal trees along the genome. None of these trees has fully coalesced,
+and so each of the trees has multiple roots.
+
+:::{important}
+The set of nodes at time 10 connecting to all the internal nodes in the
+trees are very important: without these nodes and unary edges, the
+statistical properties of the combined simulations would not be correct!
+(See also section below on
+{ref}`forward simulations<sec_ancestry_initial_state_forward_simulations>`.)
+:::
+
+We then run the next phase of the simulation in which there is
+*no* recombination by passing ``ts1`` as the argument to ``initial_state``:
+
+```{code-cell}
+ts2 = msprime.sim_ancestry(initial_state=ts1, population_size=100, random_seed=3)
+SVG(ts2.draw_svg(y_axis=True))
 ```
 
-By default `msprime` simulations are initialised by specifying a set of samples,
-using the `sample_size` or  `samples` parameters to {func}`.simulate`. This
-initialises the simulation with segments of ancestral material covering the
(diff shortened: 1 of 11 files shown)


grahamgower:
In some cases, I think the user will have the demography that was used to simulate `from_ts` already. That should work already though, right? (But maybe there should be a test for this?)

jeromekelleher:
> In some cases, I think the user will have the demography that was used to simulate from_ts already. That should work already though, right? (But maybe there should be a test for this?)

Yep, that should work fine. Definitely will be tests for that.

jeromekelleher:
Just added a couple of tests @grahamgower, works as you would expect when you don't provide a demography as a parameter.

msprime/demography.py
```diff
@@ -1501,6 +1515,58 @@ def from_old_style(
                 population_map,
             )
 
+    @staticmethod
+    def from_tree_sequence(
+        ts: Union[tskit.TreeSequence, tskit.TableCollection]
+    ) -> Demography:
```
petrelharp, line 1521: I don't really want to start making *all* methods take tree sequences *or* table collections interchangeably - is there a particular reason to do it for this one? It's easy enough to do `from_tree_sequence(tables.tree_sequence())` if that's what you've got.

petrelharp:
Looks good! Taking parameters (initial_size, growth_rate) from metadata seems like a big can of worms - e.g., that's the "initial size" at what time ago?

msprime/demography.py
jeromekelleher: Yep, that's a fair point. Since we need a TreeSequence here, it's probably a bad idea to go converting a set of tables into a tree sequence behind the scenes, since this can be expensive. Converting the other way around seems like a reasonable convenience in the ``initial_state`` argument to sim_ancestry though.

jeromekelleher:
@petrelharp, any chance you could take a look at this and comment on how it's going? I didn't get a chance to finish it, but you'll see how it's going.

I've decided against allowing ``demography=None`` and no ``population_size`` when initial_state is specified I think - it's hard to imagine when this is actually a good idea. These are still in the code, I just convinced myself it's a bad idea while writing the docs.

docs/ancestry.md
```diff
@@ -958,59 +958,179 @@ not ancestral to these census nodes.
...
+setting up a {class}`tskit.TableCollection` which contains the
+{ref}`population<tskit:sec_population_table_definition>`,
+{ref}`node<tskit:sec_node_table_definition>`,
+and {ref}`individual<tskit:sec_individual_table_definition>` information.
+The low-level simulation code then reads these tables and sets up the
+simulation accordingly. We can see what this initial state looks like
+by setting up a simple simulation and stopping it running immediately:
+
+```{code-cell}
+initial_ts = msprime.sim_ancestry(1, end_time=0, random_seed=2)
```
petrelharp, line 974: nice!

docs/ancestry.md
```diff
@@ -958,59 +958,179 @@ not ancestral to these census nodes.
...
+We can see that the returned tree sequence has one population,
+two sample nodes and one individual (since sample individuals are
+{ref}`diploid<sec_ancestry_ploidy>` by default). There are no
+{ref}`edges<tskit:sec_edge_table_definition>` because the simulation
+didn't run for any period of time: we're essentially just returning
+the initial state used to start the simulation.
+
+Then, running a simulation in which the ``initial_state`` is equal to
+``initial_ts`` is *precisely* the same as running the simulation
+with the same arguments as we used above:
```
petrelharp, line 986: but, this is only true because all the all the arguments above were set to defaults, right? - saying it this way might lead someone to think that all other arguments to `sim_ancestry( )` were somehow recorded in the tables and would be passed through?

docs/ancestry.md
```diff
@@ -958,59 +958,179 @@ not ancestral to these census nodes.
...
+ts1 = msprime.sim_ancestry(initial_state=initial_ts, random_seed=2)
+ts2 = msprime.sim_ancestry(1, random_seed=2)
+assert ts1.equals(ts2, ignore_provenance=True)
+```
+
+While it is possible to specify the initial state manually by
+constructing the input tree sequence,
+it is an advanced topic and rarely necessary. More often we are
+interested in specifying the initial state of an ``msprime``
+simulation as the output of *another* simulation.
```
petrelharp, line 997: "as the output" -> "using the output"

petrelharp: you could link to pyslim, `sec_tutorial_recapitation` here

docs/ancestry.md
```diff
@@ -958,59 +958,179 @@ not ancestral to these census nodes.
...
+
+### Combining backward-time simulations
+
+Suppose that we wished to simulate a scenario
+in which we have recombination happening at some rate in the recent past,
+and in the more distance past there is no recombination.
```
petrelharp, line 1003: "distance" -> "distant"

docs/ancestry.md
petrelharp: oh, never mind, you link to the "forward simulation" section below

docs/ancestry.md
```diff
@@ -958,59 +958,179 @@ not ancestral to these census nodes.
...
+
+```{code-cell}
+demography = msprime.Demography()
+demography.add_population(name="A", initial_size=100)
+demography.add_population(name="B", initial_size=200)
+demography.add_population(name="C", initial_size=300)
+demography.add_population_split(time=100, ancestral="C", derived=["A", "B"])
+demography
+```
+Then, we run a simulation of this model for 5 generations:
```
petrelharp, line 1068: the code says `end_time=50`

docs/ancestry.md
```diff
@@ -958,59 +958,179 @@ not ancestral to these census nodes.
...
+object which we can use to model the entire simulation from end-to-end.
+
+:::{todo}
+Keep going with this example, and show that if we use
+Demography.from_tree_sequence we  get the populations
+and their names, but not the events. This is a big gotcha
+if you forget to specify the demography because it'll
+automaticaly call from_tree_sequence for you. Maybe we
+shouldn't do this? Hard to imagine actually that creating a
+default demography with pop_sizes of 1 being a good idea.
```
petrelharp, line 1101: I agree.

docs/ancestry.md
```diff
@@ -1020,25 +1140,16 @@ corresponding to the initial generation. Furthermore, for every sample in the
...
-<!---
-(Please see the :ref:`tutorial
--->
-
-<!---
-<sec_tutorial_simulate_from>` for further explanation of this point and an
--->
+:::{todo}
+Link to a tutorial that'll explain how to combine forward time and msprime
+simulations better. Also link to pyslim.
```
petrelharp, line 1145: pyslim: sec_tutorial_recapitation

tests/test_simulate_from.py
```diff
@@ -882,18 +882,191 @@ def test_population_mismatch_no_population_configs(self):
...
+        d1 = msprime.Demography()
+        d1.add_population(name="A")
+        d1.add_population(name="B")
+        # All lineages are in B, so we will coalesce
+        ts1 = msprime.sim_ancestry(
+            {"B": 100}, demography=d1, end_time=0.1, random_seed=2
+        )
+        assert ts1.first().num_roots > 1
+
+        ts2 = msprime.sim_ancestry(initial_state=ts1, random_seed=3)
```
petrelharp, line 896: So, you're proposing this will raise an error, since no population_size is specified, right? That makes sense. But I guess that since there's more than one population (even though only one is used) that the user will need to make a demography object.

petrelharp:
Looks good! Only minor comments.

docs/ancestry.md
jeromekelleher: Good point - I've added a warning block.

jeromekelleher:
I think this is ready to go - @grahamgower it would be great to get your eyes on this, if you can. I'm mainly trying to make it less easy to make mistakes.

There's a follow-up in #1546, but I'll let that go until this is merged as it'll be noisy.

grahamgower (approved):
Looks excellent!

PR #1548, opened 2021-03-17
Use silent mutations

jeromekelleher:

The verification tests are perhaps a bit weak, should I be doing more than just looking at the number of mutations?

### DIFF
--- a/lib/mutgen.c
+++ b/lib/mutgen.c
@@ -1069,13 +1069,6 @@ mutgen_populate_tables(mutgen_t *self)
                     parent_id = TSK_NULL;
                 } else {
                     parent_id = m->parent->id;
-                    if (m->derived_state_length == m->parent->derived_state_length
-                        && memcmp(m->derived_state, m->parent->derived_state,
-                               m->derived_state_length)
-                               == 0) {
-                        ret = MSP_ERR_BAD_ANCESTRAL_MUTATION;
-                        goto out;
-                    }
                     tsk_bug_assert(parent_id != TSK_NULL);
                 }
                 mutation_id = tsk_mutation_table_add_row(mutations, site_id, m->node,
--- a/lib/tests/test_mutations.c
+++ b/lib/tests/test_mutations.c
@@ -529,7 +529,7 @@ test_single_tree_mutgen_keep_sites_many_mutations(void)
 
     ret = mutgen_generate(&mutgen,
         MSP_DISCRETE_SITES | MSP_KEEP_SITES | MSP_KEPT_MUTATIONS_BEFORE_END_TIME);
-    CU_ASSERT_EQUAL_FATAL(ret, MSP_ERR_BAD_ANCESTRAL_MUTATION);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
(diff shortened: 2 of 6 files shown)


benjeffery (approved):
LGTM

petrelharp:
Looks good! I think we also need to make it so that at least parent-independent mutation models include silent mutations. We need to do this so that we can correctly stack mutations. But, maybe people won't expect this, so maybe it should be optional, e.g. `silent=False`? I guess the options are

1. default to `silent=True` and have some people confused about why there are silent mutations
2. default to `silent=False` and have to explain that if you're going to (correctly) stack on more mutations in the future you need to set `silent=True`

What do you think? @GertjanBisschop?

I suppose we might also want a `TableCollection.remove_silent_mutations` method.

jeromekelleher:
> I think we also need to make it so that at least parent-independent mutation models include silent mutations.

What do we need to do this? Should I try and do it here, or create an issue for a follow-up?

> default to silent=True and have some people confused about why there are silent mutations

I think we should default to silent=True. We should be statistically correct by default. In fact, I'm not particularly enthusiastic about making ``silent`` an option right now - maybe we could do this after 1.0, if we decide we need it?

jeromekelleher:
> I suppose we might also want a TableCollection.remove_silent_mutations method.
Added an issue on tskit for this

petrelharp:
Ah, right - silent mutations literally won't affect anything: not genotypes, not statistics calculations. Only `ts.num_mutations`. I agree then - no need to make `silent` an option.

I got confused for a bit trying to work out what the silent mutation rates should be. But then I realized that the silent mutations are *already there* - notice for instance in the docs on [HKY](https://tskit.dev/msprime/docs/latest/api.html#msprime.HKYMutationModel) - it explains that this is implemented at a MatrixMutationModel with P_ii set to something so the rows all sum to the same thing. This is exactly what we want - usually, HKY is not-parent independent, but if for instance pi_A = pi_C = pi_T = pi_G then it is, and with this choice for P_ii, applying HKY multiple times will be correct (and approximately correct if the pi's aren't all equal).

> What do we need to do this? Should I try and do it here, or create an issue for a follow-up?

So, all we have to do is to go back in to the mutation generation code and ask it to actually record the silent mutations, for instance [here](https://github.com/tskit-dev/msprime/blob/main/lib/mutgen.c#L238) and [here](https://github.com/tskit-dev/msprime/blob/main/lib/mutgen.c#L1248).  And, fix up the tests to account for this change, heh. I can probably do that sufficiently soon, unless you're willing to have a go at it, @GertjanBisschop?  And, I think that should be a different PR.

jeromekelleher:
OK, great, thanks @petrelharp. I'll open an issue to track updating the mutation models, and an issue to document the fact that we produce silent mutations (we just need to put in a section about it in the docs, show some examples and refer to it in a few places).

I'll tidy this up and merge now.

GertjanBisschop:
Quickly skimming through this, I can see that I am simply not quick enough currently to contribute in a meaningful way. I wish I could have a go at the stats tests but have to prioritise.

PR #1556, opened 2021-03-18
set extra metadata in demography

### LINKED ISSUE #1555: need a way to specify metadata for populations in Demography objects
In trying to switch over `pyslim.recapitate`, I'm encountering the following conundrum. The population table has a metadata schema with required properties. So, suppose I do
```
demog = msprime.Demography.from_tree_sequence(slim_ts)
derived_names = [pop.name for pop in demog.populations]
demog.add_population(name='ancestral', initial_size=100)
demog.add_population_split(100, derived=derived_names, ancestral='ancestral')
ts = msprime.sim_ancestry(initial_state=slim_ts, demography=demog)
```
However, this fails in [`demography.Demography.insert_extra_populations`](https://github.com/tskit-dev/msprime/blob/ef46e2d8bb4dc4e12ee0d8fc88e2e99b8dcecc5c/msprime/demography.py#L817) on validating the metadata (as it tries to set it to `{}`). The solution, I think, is helpfully given in [this TODO](https://github.com/tskit-dev/msprime/blob/ef46e2d8bb4dc4e12ee0d8fc88e2e99b8dcecc5c/msprime/demography.py#L853), to use `extra_metadata` for this purpose.

The other option would be to fall back to `None` for the metadata when appropriate, except for https://github.com/tskit-dev/tskit/issues/1256 . And, I *think* that this will work for recapitate, except that I can't check yet because of https://github.com/MesserLab/SLiM/issues/168 .  <swearing>

petrelharp:
This might close #1555 but there's no tests yet.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -938,8 +938,11 @@ class Demography(collections.abc.Mapping):
         schema = tables.populations.metadata_schema.schema
         if schema is not None:
             properties = schema["properties"]
-            name_in_metadata = "name" in properties
-            description_in_metadata = "description" in properties
+            additional_properties = schema["additionalProperties"]
+            name_in_metadata = "name" in properties or additional_properties
+            description_in_metadata = (
+                "description" in properties or additional_properties
+            )
             if not name_in_metadata:
                 warnings.warn(
                     "The metadata schema does not have a 'name' property; "
@@ -954,16 +957,26 @@ class Demography(collections.abc.Mapping):
                     "tree sequence",
                     IncompletePopulationMetadataWarning,
                 )
+            left_out = []
             for population in self.populations[len(tables.populations) :]:
                 md = {}
                 if name_in_metadata:
                     md["name"] = population.name
                 if description_in_metadata:
                     md["description"] = population.description
-                # TODO we could also try to merge in the ``extra_metadata``
-                # depending on whether `additional_properties` is set,
-                # but it doesn't really seem worth it.
-                tables.populations.add_row(md)
+                for k in population.extra_metadata:
+                    if k in properties or additional_properties:
+                        md[k] = population.extra_metadata[k]
+                    else:
+                        left_out.append(k)
+                tables.populations.add_row(metadata=md)
+            if len(left_out) > 0:
+                warnings.warn(
+                    "The metadata schema does not allow for all properties specified "
+                    "in extra_metadata: these keys have not been recorded in the "
+                    f"output tree sequence: {', '.join(left_out)}",
+                    IncompletePopulationMetadataWarning,
+                )
         else:
             warnings.warn(
                 "No metadata schema present in population table, not recording "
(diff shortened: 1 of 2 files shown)


msprime/demography.py
```diff
@@ -850,10 +850,19 @@ def insert_extra_populations(self, tables):
                     md["name"] = population.name
                 if description_in_metadata:
                     md["description"] = population.description
-                # TODO we could also try to merge in the ``extra_metadata``
```
jeromekelleher, line 853: We should keep the TODO I think, it's not quite the same thing as you're doing here. Adding in stuff from extra_metadata if it's present in ``properties`` is definitely the right thing to do.

msprime/demography.py
```diff
@@ -850,10 +850,19 @@ def insert_extra_populations(self, tables):
...
-                # depending on whether `additional_properties` is set,
-                # but it doesn't really seem worth it.
-                tables.populations.add_row(md)
+                for k in population.extra_metadata:
+                    if k in properties:
+                        md[k] = population.extra_metadata[k]
+                # if we've still got invalid metadata and None is allowed then
+                # fall back to None
+                try:
+                    tables.populations.add_row(metadata=md)
```
jeromekelleher, line 859: Hmm, not sure what's motivating this. We should have valid metadata, shouldn't we? Is it that we're missing required properties? Silently storing *nothing* then seems like bad behaviour, wouldn't an informative error be better?

jeromekelleher (commented):
Basic idea looks good, not sure about the catch-all storing nothing behaviour though.

msprime/demography.py
petrelharp: What's motivating this is the SLiM example of populations, where `None` is valid metadata. Originally I put this in as
```
if 'type' in schema and 'null' in schema['type'] and (('required' in schema) and (len(set(schema['required']) - set(md.keys())) == 0)):
    md = None
```
... but the `try... except` seemed nicer.

petrelharp: But, yes - an informative error would be better! I'll omit this bit. (There is no way to ask that the metadata be None, but perhaps that's OK...)

msprime/demography.py
petrelharp: Ah, right. I've gone ahead and done the TODO.

petrelharp:
Ok, how's this?

msprime/demography.py
```diff
@@ -954,16 +957,26 @@ def insert_extra_populations(self, tables):
...
                     md["description"] = population.description
-                # TODO we could also try to merge in the ``extra_metadata``
-                # depending on whether `additional_properties` is set,
-                # but it doesn't really seem worth it.
-                tables.populations.add_row(md)
+                for k in population.extra_metadata:
+                    if k in properties or additional_properties:
+                        md[k] = population.extra_metadata[k]
+                    else:
+                        left_out = True
```
jeromekelleher, line 971: It would probably be helpful to tell the user what the left out fields are, how about ``left_out.append(k)`` and then insert ``left_out`` into the warning message?

petrelharp:
Passed tests, and rebased; should be good now.

jeromekelleher (approved):
Looks great, thanks @petrelharp!

Issue #1365, opened 2021-01-15
Fix MsMs sweep verification tests


jeromekelleher:
I disable the MsmsSweeps tests because they were unreliable (looks like some simulations were ocassionally returning no data).

We probably just need to make these more robust.

jeromekelleher:
What are the thoughts on the MsmsSweeps tests @gbinux @andrewkern? We need to get the stats suite robust now in the run up to release.

andrewkern:
i don't think we need the msms tests-- we've got discoal and SLiM agreeing with the msprime implementation

jeromekelleher:
@gbinux, any thoughts?

gbinux:
@jeromekelleher I can spend some time on this to see whether it fits well in the comparisons. If it does, it is good to have it; if not, we can get rid of this.

jeromekelleher:
Closed in #1561

Issue #1286, opened 2020-11-06
Compare command line parameters of msprime with other coalescent simulators


gbinux:
As many other coalescent simulators use the command line to specify parameters, maybe we could do a comparison of command-line parameters between `msprime` with other simulators, such as `macs`, `msms`, `discoal` (there are more to add to the list). I would suggest to:

1) Use several examples to show how to do a set of similar simulations with different software. I think this will make it easier for new users to switch from other software to msprime.  In the meantime, this can help transition to functionalities/models that other simulators do not have.

2) Use a table to compare definitions of parameters, such as coalescent rate, recombination rate, and time units (1, 2N, or 4N generations) used in `msprime` vs other software. I was confused every time a simulator uses a different scale factor or time units.

As @jeromekelleher suggested, this is likely to be in the quickstart page.

jeromekelleher:
@gbinux, the 1.0 release is getting closer and there's still a lot of stuff to do. Would you like to volunteer to do this, or can I take it out of the 1.0 Milestone?

jeromekelleher:
One place this could go is on the [notes for ms users page](https://tskit.dev/msprime/docs/latest/notes_ms_users.html) @gbinux, although we'd need to come up with a new name for the page.

gbinux:
I was thinking about this as well. How about `Notes for users of ms and other coalescent simulators`?

jeromekelleher:
Closed in #1624

PR #1561, opened 2021-03-19
fix msms sweep test in verification.py

gbinux:
To address #1365 

For neutral conditions, `msms` and `msprime` correspond nicely.
For selective sweeps,  `msms` and `discoal` are pretty consistent. There are some discrepancies between `msms` and `msprime`, and between `discoal` and `msprime`, but I think it is reasonable as the selective sweep simulation is implemented in a new framework.  Please help take a look @andrewkern .

[MsmsSweeps.zip](https://github.com/tskit-dev/msprime/files/6174514/MsmsSweeps.zip)

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -1128,13 +1128,7 @@ class SweepVsSlim(Test):
         self._run(10, 1e6, 1e3, 1e-7, 0.25, 5000, num_replicates=10)
 
 
-# FIXME disabling these for now because they are unreliable and result in
-# errors (root: EXCEPTION:No columns to parse from file). This is probably
-# harmless as it means there's no data simulated, but it stops the rest
-# of the tests from running so we need to deal with it somehow.
-
-
-class MsmsSweeps:
+class MsmsSweeps(Test):
     """
     Compare msms with msprime/discoal for selective sweeps.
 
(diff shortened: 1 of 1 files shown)


verification.py
```diff
@@ -1269,30 +1263,39 @@ def _msms_params_to_run_msp(self, params):
...
-            sample_size=params["nsam"],
-            Ne=0.25,
-            length=params["num_sites"],
-            mutation_rate=mutation_rate,
-            recombination_rate=recombination_rate,
+        repeats = msprime.sim_ancestry(
+            samples=params["nsam"],
+            population_size=params["refsize"],
             model=model,
+            recombination_rate=recombination_rate,
```
jeromekelleher, line 1285: For a comparison with msms I think we should probably use a continuous genome - set ``discrete_genome=False`` here and in the ``sim_mutations`` below.

jeromekelleher (commented):
Thanks @gbinux - let's try the continuous genome and see if the discrepancies remain.

verification.py
gbinux: just added this option and the result is similar  @jeromekelleher

verification.py
```diff
@@ -1269,30 +1263,40 @@ def _msms_params_to_run_msp(self, params):
...
+            discrete_genome=False,
+            sequence_length=params["num_sites"],
+            num_replicates=params["nrep"],
         )
 
-        return ts
+        # Critical to use BinaryMutationModel and get ancestral and derived alleles
+        mutated_repeats = [
+            msprime.sim_mutations(
+                ts, rate=mutation_rate, model=msprime.BinaryMutationModel()
```
jeromekelleher, line 1294: Need discrete_genome=False here too @gbinux

jeromekelleher (commented):
STill need discrete_genome for mutations @gbinux - because the sequence is so short this will make a difference.

verification.py
gbinux: Oops, missed the one in sim_mutations. Just added it as well. @jeromekelleher 

[MsmsSweeps_2.zip](https://github.com/tskit-dev/msprime/files/6175848/MsmsSweeps_2.zip)

verification.py
```diff
@@ -1269,30 +1263,43 @@ def _msms_params_to_run_msp(self, params):
...
 
-        ts = msprime.simulate(
-            sample_size=params["nsam"],
-            Ne=0.25,
-            length=params["num_sites"],
-            mutation_rate=mutation_rate,
-            recombination_rate=recombination_rate,
+        repeats = msprime.sim_ancestry(
+            samples=params["nsam"],
+            population_size=params["refsize"],
```
jeromekelleher, line 1283: Ah - these are diploid sample **individuals** now, we need ``params["nsam"] / 2`` here.

If this number is odd, then you can do
```
samples=msprime.SampleSet(params["nsam"], ploidy=1)
```
to achieve the same effect (i.e., keeping the timescale diploid but sampling haploid individuals)

jeromekelleher (commented):
These are major differences in my view @gbinux - but I think the samples thing might fix it.

verification.py
andrewkern: yeah this will probably account for a lot of the differences in these tests @gbinux

gbinux: Good catch! I tried `params["nsam"] / 2`  and avoided using odd numbers. It seems the discrepancies are still there. 
[MsmsSweeps_3.zip](https://github.com/tskit-dev/msprime/files/6190809/MsmsSweeps_3.zip)

Before I keep pushing the code, I want to show my understanding of the relevant parameters/options. Maybe listing them out could help find the issue if it is obvious. The following linked docs is also a draft related to   #1286 

[Compare args/options between different coalescent simulators](https://docs.google.com/document/d/1TmkfWN5hREOj0wy_B6K3OALAYIydgdPTaulImKheQ18/edit?usp=sharing)

Are there additional misunderstandings about the parameters? Is this kind of information useful to be included in the docs? If the discrepancies persist after we excluded the parameter scaling issue, are we going to keep this Test class within the msprime 1.0 release? @jeromekelleher @andrewkern

jeromekelleher: Hmm, these still look like pretty big differences to me. Your scalings look about right to me in the doc, but I'm not that well up on these other programs any more.

Can you summarise what you think the differences are, and where we're seeing deviations between the programs please?

andrewkern: Hi @gbinux-- we already know that discoal, msprime, and slim give the same results from my additions to `verification.py`. you might just look in the `DiscoalSweep` and `SweepVsSlim` test classes in there to see where your parameterization conversion differs.

verification.py
```diff
@@ -1269,30 +1263,43 @@ def _msms_params_to_run_msp(self, params):
                     position=params["sel_pos"],
```
gbinux, line 1263: Just realized this is another source of differences: for selection position, msms/discaol use a value between 0-1, msprime use position in bp.
After correcting this, it seems much better. This is `test_selective_discoal_vs_msp_use_discoal_paper_param` result:

before:
image

and  after scaling the position

image

gbinux:
I think the new results are good. Here are representative comparisons.
1. `test_selective_discoal_vs_msp_use_discoal_paper_param`
image

2.  `test_selective_msms_vs_discoal_use_discoal_paper_param`
image

3. `test_selective_msms_vs_msp_use_discoal_paper_param`
image
 
@jeromekelleher @andrewkern

andrewkern:
looks great to me @gbinux!

andrewkern (approved):

jeromekelleher:
These look much better thanks @gbinux! I'm still a little bit worried about the wobble we're seeing on D and pi, but maybe that's something to do with mutation models.

Can you bring this PR up to date with a rebase please @gbinux? I'll merge then.

gbinux:
> These look much better thanks @gbinux! I'm still a little bit worried about the wobble we're seeing on D and pi, but maybe that's something to do with mutation models.
> 
> Can you bring this PR up to date with a rebase please @gbinux? I'll merge then.

Yep. I used the new model syntax and rebased it already before I did the push, so it should be good to go @jeromekelleher 

I observed that the D and Pi differences occur mainly in `msms` vs `discoal` and `msms` vs `msp` but not obviously in `discoal` vs `msp` and so I think the difference due to the internal logics of how `msms` vs `discoal`(`msp`) simulate this type of selection.

andrewkern:
yes it looks like `msms` is the outlier here.

jeromekelleher (approved):

Issue #1530, opened 2021-03-13
Update Dockerfile for 1.0


jeromekelleher:
The Dockerfile is a bit out of date now. I just updated it quickly to use Ubuntu 20.04 so it at least works, but I think we could make it more useful:

1. Base it off one of the standard Python data science images (so it's more useful)
2. Change it so it installs the latest release rather than from github (``pip install --pre msprime``)

Because of the binary wheels now, this means that we don't need any of the GSL stuff and it should all work fine. It should literally be one line ``pip install --pre msprime``.

This should make things like Binder (#1413) work better.

@agladstein, any chance you could take a look at this?

agladstein:
Yes, I can work on it.

agladstein:
I vote going with https://hub.docker.com/r/jupyter/scipy-notebook as a base stack.

> jupyter/scipy-notebook includes popular packages from the scientific Python ecosystem.
Everything in jupyter/minimal-notebook and its ancestor images
dask, pandas, numexpr, matplotlib, scipy, seaborn, scikit-learn, scikit-image, sympy, cython, patsy, statsmodel, cloudpickle, dill, numba, bokeh, sqlalchemy, hdf5, vincent, beautifulsoup, protobuf, xlrd, bottleneck, and pytables packages
ipywidgets and ipympl for interactive visualizations and plots in Python notebooks
Facets for visualizing machine learning datasets

PR #1566, opened 2021-03-21
base docker image off of data science image and intall latest release

agladstein:
I used the jupyter/scipy-notebook base. One thing to consider is that it makes it a fairly large image (~2.69G). I personally, would probably prefer keeping the base msprime image lightweight and only using the Ubuntu base. The current docker image with the Ubuntu base is only 565MB.
And then I'd say make an additional image for a whole tskit ecosystem based on jupyter/scipy-notebook base.

Jupyter lab can be run as follows using Docker:
```
docker run -it -v ${PWD}:/app --rm -p 8888:8888 tskit/msprime:<<tag>> jupyter-lab
```
And you need to take care to not be in the msprime directory, or else you will get a conflict when trying to import `msprime`.

closes https://github.com/tskit-dev/msprime/issues/1530

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,25 +2,10 @@
 #organization tskit-dev
 #application "Msprime: A reimplementation of Hudson's classical ms simulator for modern data sets."
 
-FROM ubuntu:20.04
+FROM jupyter/scipy-notebook:latest
 
 # Set the working directory to /app
 WORKDIR /app
 
-# Copy the current directory contents into the container at /app
-COPY . /app
-
-RUN apt-get update && apt-get install -y --no-install-recommends \
-            python3 \
-            python3-dev \
-            python3-pip \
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Thanks @agladstein, that's very helpful. I appreciate your point about the size - I guess it depends on what you're using the Docker for. If it's online demos like Binder, we want something like this version here. If we want it just for running simulations, I guess the minimal image is what you want.

I guess there's less reason for needing a base image now that we have binary wheels - once there's a recent version of Python, we should now work without issues on any reasonable platform.

jeromekelleher:
Let's merge this one and see if it fixes our issues with Binder. I'm happy to talk about a more minimal image too if it's something you think we want.

jeromekelleher (approved):

PR #1573, opened 2021-03-23
New model specification syntax.

jeromekelleher:
The model specification syntax that we ended up with was a compromise, and ended up being quite convoluted. I started trying to explain it for the docs, and quickly realised that I was confusing myself trying to write it, as there were so many different ways to do things,  and needing to use callables to specify things that happen at random times.

This is a new proposal. As before, we can have things like ``model="dtwf"``, ``model=SmcAppoxCoalescent()`` etc. Now, if we want to specify multiple models we need to assign them a **duration**, for example:
```python
ts = msprime.sim_ancestry(10, model=[DiscreteTimeWrightFisher(duration=10), "hudson"])
```
this will run the DTWF model for 10 generations, then switch to the standard coalescent for the rest of time. Note that this is a **duration** rather than an absolute time. This is to allow for things like
```python
sweep_model = msprime.SweepGenicSelection(# some params)
ts = msprime.sim_ancestry(10, 
    model=[sweep_model, msprime.StandardCoalescent(duration=1)] * 100
)
```
which gives lots of sweeps separated in time by 1 generation of the coalescent. Expressing something like this would be super convoluted under the current approach, which assigns absolute **times** to the models.

What do you think @andrewkern, @gbinux, @petrelharp and others?


gbinux:
I like this syntax @jeromekelleher

jeromekelleher:
OK, going to merge this much. Will follow up with any more changes needed for verification.

petrelharp:
Specifying the `duration` seems sensible!

andrewkern:
this looks neat Jerome!

PR #1581, opened 2021-03-25
Change the example to 3 samples

hyanwong:
It's a bit confusing that the example code has lots of "1"s (e.g.  N - 1); migration_rate=0.1. I think it might be worth changing the number so it's immediately obvious which the sample number is.


jeromekelleher (approved):

jeromekelleher:
@mergifyio rebase

PR #1592, opened 2021-03-26
Edit multiple chromosomes section

### LINKED ISSUE #1565: Clarify multiple chromosomes section
#1564 added an update to the [multiple chromosomes](https://tskit.dev/msprime/docs/latest/ancestry.html#multiple-chromosomes) section. There's a few outstanding questions for me:

1. How do we integrate the information about expected number of recombs? (see note)
2. When do we actually need to do this?

I think we need a clear message at the top saying "You don't need to do this unless you're explicitly modelling the relationship **between** individual chromosomes in a single simulation" (or whatever the right answer is). We're assuming a very high level of understanding at the moment, and I think people are going to google directly into this section and either be very confused, or do some very expensive simulations for no reason.

Could someone who understands this better than me have a pass through please? Preferably after reading through the rest of the [genome properties](https://tskit.dev/msprime/docs/latest/ancestry.html#genome-properties) section to get the right level.

@petrelharp, @JereKoskela or @apragsdale, any takers? (Or anyone else?)

apragsdale:
Closes #1565

Before polishing up, wanted to get some feedback if this helps with the flow of the section and addresses some of the concerns from #1565.

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -649,25 +649,6 @@ positions. Simulating a continuous genome sequence can be useful for
 theoretical work, but we recommend using discrete coordinates for most
 purposes.
 
-(sec_ancestry_multiple_chromosomes)=
-
-### Multiple chromosomes
-
-Msprime does not directly support simulating multiple chromosomes simultaneously,
-but we can emulate it using a single linear genome split into chromosome
-segments. Multiple chromosomes are modelled by specifying a recombination map with
-single base-pair segments with recombination probability 1/2 separating adjacent
-chromosomes. The probability that a recombination breakpoint occurs
-at a given base pair
-that has recombination rate {math}`r` is {math}`1-e^{-r}`. Thus, we set the
-recombination rate in the base pair segments separating chromosomes to
-{math}`\log(2)`. This ensures that the recombination probability each generation
-is 1/2.
-
-:::{todo}
-Tie this next bit of information here into the derivation above.
-:::
-
 :::{note}
 With a discrete genome and a recombination rate of {math}`r`,
 the expected number of recombinations per unit time and per unit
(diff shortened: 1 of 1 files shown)


docs/ancestry.md
```diff
@@ -675,8 +656,48 @@ is {math}`L (1 - e^{-r})`
...
+correct distribution of the number and total length of shared IBD segments. We
+similarly require multiple chromosomes and the DTWF model to get the correct
+ancestry patterns for very recently admixed populations.
+
+Additionally,
+{ref}`Multiple merger coalescents <sec_ancestry_models_multiple_mergers>`
+result in positively correlated
+ancestries between unlinked chromosomes. This correlation does not break down
+in time and multiple chromosomes should not be simulated independently under
+the multiple merge model.
```
jeromekelleher, line 695: "multiple merger models" I think

docs/ancestry.md
```diff
@@ -675,8 +656,48 @@ is {math}`L (1 - e^{-r})`
...
+probability 1/2 separating adjacent chromosomes. The probability that
+a recombination breakpoint occurs at a given base pair that has recombination
+rate {math}`r` is {math}`1-e^{-r}`. Thus, we set the recombination rate in the
+base pair segments separating chromosomes to {math}`\log(2)`. This ensures that
+the recombination probability each generation is 1/2.
+
+#### Do I need to do this?
+
+The main purpose of simulating multiple chromosomes together (instead of
+running independent simulations for each chromosome) is to capture the correct
```
jeromekelleher, line 676: "correct" might be a bit misleading here. Maybe just "capture the correlation between"?

jeromekelleher (approved):
Looks good to me, thanks @apragsdale. Minor comments.

jeromekelleher:
@JereKoskela, can you take a quick look please? Is what we're saying abou the lambda coalescents correct? Is there an interplay with ploidy here in the lambda coalescents?

JereKoskela:
The multiple merger paragraph looks accurate to me. Ploidy doesn't enter the picture here; large, multiple merger events can cause simultaneous mergers in unlinked loci regardless. I have two pretty minor comments.

- We could throw in a citation to https://www.genetics.org/content/193/1/255 to justify the claim that unlinked loci are positively correlated.
- We could expand a bit on the fact that unlinked loci on the same individual recombine apart in 0 time under a coalescent model, so we would like `r` to be a big number (to approximate an infinite rate) rather than log(1/2). Though that is already pretty big in comparison to typical values of `r` and hence not a bad approximation. 

I'm happy to address those myself in a separate PR once this has been merged if that would make life easier.

docs/ancestry.md
```diff
@@ -675,8 +656,48 @@ is {math}`L (1 - e^{-r})`
...
+
+### Multiple chromosomes
+
+Msprime does not directly support simulating multiple chromosomes
+simultaneously, but we can emulate it using a single linear genome split into
+chromosome segments. Multiple chromosomes are modelled by specifying
+a recombination map with single base-pair segments with recombination
+probability 1/2 separating adjacent chromosomes. The probability that
+a recombination breakpoint occurs at a given base pair that has recombination
+rate {math}`r` is {math}`1-e^{-r}`. Thus, we set the recombination rate in the
```
petrelharp, line 669: "The probability that a recombination breakpoint occurs *over one unit of time*"?

docs/ancestry.md
```diff
@@ -675,8 +656,48 @@ is {math}`L (1 - e^{-r})`
...
+in time and multiple chromosomes should not be simulated independently under
+the multiple merge model.
+
+#### Example
+
+The following defines a recombination map for three chromosomes each
+1 centiMorgan in length:
 
 ```{code-cell}
 import math
```
petrelharp, line 703: Heh, whoops: this should actually be:
```
positions = [0, 1e6 - 1, 1e6, 2e6 - 1, 2e6, 3e6]
```

docs/ancestry.md
```diff
@@ -730,14 +739,13 @@ ts = msprime.sim_ancestry(
...
+To isolate tree sequences for each chromosome, we can trim the full tree
+sequence for each chromosome by defining the end points of each chromosome and
+using the {meth}`~tskit.TreeSequence.keep_intervals` method. It is important to
+specify `simplify=False` so that node indices remain consistent across
+chromosomes. The {meth}`~tskit.TreeSequence.trim` method is then used to trim
+the tree sequences to the focal chromosome, so that position indices begin at
+0 within each chromosome tree sequence.
 
 ```{code-cell}
 chrom_positions = [0, 1e6, 2e6, 3e6]
```
petrelharp, line 751: I know this isn't code you touched, but... perhaps `chrom_positions` should be moved above and used to define `positions`.

docs/ancestry.md
```diff
@@ -730,14 +739,13 @@ ts = msprime.sim_ancestry(
...
 
-To isolate tree sequences for each chromosome, we can trim the full tree sequence
-for each chromosome by defining the end points of each chromosome and using
-the {meth}`~tskit.TreeSequence.keep_intervals` method.
-It is important to specify `simplify=False` so
-that node indices remain consistent across chromosomes. The
-{meth}`~tskit.TreeSequence.trim` method
-is then used to trim the tree sequences to the focal chromosome, so that position
-indices begin at 0 within each chromosome tree sequence.
+To isolate tree sequences for each chromosome, we can trim the full tree
```
petrelharp, line 742: Maybe "To put each chromosome in its own separate tree sequence"? At first when I read "isolate" I thought that we were just going to cut out the inter-chromosome spacing (ie mark the data as missing there) but keep it all in one tree sequence.

petrelharp:
LGTM. Very clear.

apragsdale:
Excellent, thanks for the helpful comments - I’ll have another pass at cleaning it up more.

apragsdale:
> * We could expand a bit on the fact that unlinked loci on the same individual recombine apart in 0 time under a coalescent model, so we would like `r` to be a big number (to approximate an infinite rate) rather than log(1/2). Though that is already pretty big in comparison to typical values of `r` and hence not a bad approximation.

Right - and with the standard coalescent model that would be equivalent to running independent simulations of the separate chromosomes, which would be much faster and what we should advise in any case. Stitching them together and simulating with r = log(1/2) makes sense for DTWF instead of Hudson. But in the example we also suggest switching models from DTWF to Hudson after some 20 generations, so the recombination map with that large r should still be "good enough" I'd guess.

Not sure how much detail this needs in the docs, or if that is a bit more advanced than what we'd like to get into here?

docs/ancestry.md
apragsdale: Strangely, I think what is currently there puts the break point in the right place. The `[..., 1e6-1, 1e6, ...]` puts a break point at the positions `1e6 - 1`, so that you'd end up with the last base pair of the first chromosome (positions 999999) being unlinked from the chromosome it's supposed to be in and instead linked to the next chromosome. This can be seen by setting `r_chrom` in this example to zero, and checking number of trees per chromosome and their intervals.

It's a bit confusing to me, since I'd think that it would put the break point _between_ those positions, but the left inclusive interval setup and discrete genome means the break point is effective a base pair farther left than you'd expect.

docs/ancestry.md
apragsdale: This was actually an example I put together before, so happy to clean it up so it makes more sense! I moved a few things around so hopefully it helps a bit.

docs/ancestry.md
petrelharp: Ah, ok! Nice catch.

petrelharp:
> Not sure how much detail this needs in the docs, or if that is a bit more advanced than what we'd like to get into here?

I think "recommended best practices" with a bit of justification will always be appreciated.

jeromekelleher:
I agree; this is an advanced topic and we're not going to discuss it anywhere else, so this the place to get the knowledge "down on paper".

JereKoskela:
> > Not sure how much detail this needs in the docs, or if that is a bit more advanced than what we'd like to get into here?
> 
> I think "recommended best practices" with a bit of justification will always be appreciated.

Would recommended best practice be to stick to `log(1/2)` if you're using DTWF (even if that gets changed to a Hudson or multiple merger coalescent later), and to use a larger constant if you're only simulating a coalescent model (which you should never do for Hudson, but which can arise under multiple merger models)?

If that's right, then it raises the question of what the large constant should be. Is there a sensible way to answer that, e.g. find out what the biggest exponential rate that the RNG can handle?

apragsdale:
I think recommended best practices would be to use `log(2)` when using DTWF, and to run separate replicates for each chromosome independently if using the Hudson model. But I really don't know the multiple merger model at all to be able to say what to do there.

This is making me realize that I should be more clear that this only works with a discrete genome (I think). If `discrete_genome=False`, you'd end up with a ton of trees in those 1 bp intervals, and they are shifted by one bp as well so you'd need to set up the chromosomes and rate map differently if you did want to use a continuous genome. Something like:

```Python
import math
r_chrom = 1e-8
r_break = math.log(2)
positions = [0, 1e6, 1e6 + 1, 2e6 + 1]
rates = [r_chrom, r_break, r_chrom]
rate_map = msprime.RateMap(positions, rates)

ts = msprime.sim_ancestry(
    10,
    population_size=1000,
    recombination_rate=rate_map,
    discrete_genome=False,
    model=[
        msprime.DiscreteTimeWrightFisher(duration=20),
        msprime.StandardCoalescent(),
    ],
    random_seed=1234,
)
print("entire length:", ts.sequence_length)
print("total num trees:", ts.num_trees)

ts_chroms = []
for j in range(len(positions) // 2):
    start, end = positions[2 * j: 2 * j + 2]
    print("start, end:", start, end)
    chrom_ts = ts.keep_intervals([[start, end]], simplify=False).trim()
    ts_chroms.append(chrom_ts)
    print("chrom length:", chrom_ts.sequence_length)
    print("num trees:", chrom_ts.num_trees)
```

But that's pretty ugly, and probably should be recommended against?

jeromekelleher:
I think we can just say as an ``important`` that this will only work for discrete genomes @apragsdale --- nobody in their right mind will do this for continuous genomes, and if they do they can figure it out themselves.

apragsdale:
> I think we can just say as an `important` that this will only work for discrete genomes @apragsdale --- nobody in their right mind will do this for continuous genomes, and if they do they can figure it out themselves.

Ok, done. Think this is about ready to go.

jeromekelleher:
LGTM - can you squash please @apragsdale?

apragsdale:
Ok great, squashed!

Issue #1551, opened 2021-03-17
Update mutation models to generate silent mutations and document


jeromekelleher:
As discussed here: https://github.com/tskit-dev/msprime/pull/1548#issuecomment-801305017

> I got confused for a bit trying to work out what the silent mutation rates should be. But then I realized that the silent mutations are already there - notice for instance in the docs on HKY - it explains that this is implemented at a MatrixMutationModel with P_ii set to something so the rows all sum to the same thing. This is exactly what we want - usually, HKY is not-parent independent, but if for instance pi_A = pi_C = pi_T = pi_G then it is, and with this choice for P_ii, applying HKY multiple times will be correct (and approximately correct if the pi's aren't all equal).

> So, all we have to do is to go back in to the mutation generation code and ask it to actually record the silent mutations, for instance [here](https://github.com/tskit-dev/msprime/blob/main/lib/mutgen.c#L238) and [here](https://github.com/tskit-dev/msprime/blob/6f31a5359e858c5505af136f8073e62cc86b8640/lib/mutgen.c#L1241).  And, fix up the tests to account for this change, heh. I can probably do that sufficiently soon, unless you're willing to have a go at it, @GertjanBisschop? And, I think that should be a different PR.

So, we want to
1. Update the models so that they generate silent mutations
2. Update the stats tests to take this into account and verify
3. Document the fact that we do generate silent mutations. We'd to this by making a "Silent mutations" section at the end of the page, showing an example of a silent mutation happening, explain why we do this, and reassure the reader that this is fine, nothing will explode.

Sound good @petrelharp? I could do this (well, except for 2, maybe), but I don't think I'm the best person to...

petrelharp:
Yes! Sounds good!

petrelharp:
Note: I'm removing the functionality that [if a transition( ) method returns a nonzero value the mutation is discarded](https://github.com/tskit-dev/msprime/blob/6f31a5359e858c5505af136f8073e62cc86b8640/lib/mutgen.c#L1241), since we were no longer using it. I'm sure we could imagine a mutation model that used this functionality, but I don't really want to make one up just to keep that code in there. But if someone can think of a really good use case, say so.

petrelharp:
[...]

PR #1595, opened 2021-03-27
remove add_ancestral option and associated errors

### LINKED ISSUE #1553: remove add_ancestral?
In #1440:

> Does this mean that we can remove the add_ancestral parameter, since we expect the output to be "right" unless someone is doing something pretty obscure?

That'd be nice! Let's see:
- the upside would be that users adding more mutations to a forwards sim with selected mutations wouldn't have to think about "add_ancestral"
- the downside is that users who mutate more than once with non-parent-independent models would be less likely to be aware that this is statistically wrong.

So, we'd be reducing cognitive burden significantly on maybe a common use case, in exchange for avoiding some very, very slight statistical wrongness. It sounds like a great trade to me: not worrying about the precise mutation model in those rare cases of multiple hits is totally the right thing to do, I think.

I guess that also `add_ancestral` helps avoid bugs where someone means to mutate only the more recent part of the tree but forgets to specify an `end_time`. Or means to only mutate a segment of the genome. But this also seems pretty rare.

petrelharp:
This would close #1553. We'll get some more tests in there with #1551 (but this is just removing an optional error, so not much to do here).


jeromekelleher (approved):
LGTM!

docs/mutations.md
```diff
@@ -257,28 +257,25 @@ controls whether existing mutations are kept or discarded (the default is `keep=
 For instance, in final code block in {ref}`sec_mutations_time_span`, mutations were
 progressively added to a simulated tree sequence, beginninng with the oldest time period.
 
-While it is simple to add younger mutations to a tree sequence which already contains
-mutations, adding a mutation ancestral to an existing mutation will result in
-an error unless `add_ancestral=True` is specified, as in the following code block:
+While it is more natural to add younger mutations to a tree sequence which already contains
+mutations, you can also add mutations ancestral to an existing mutations,
```
winni2k, line 261: There's a typo here. "an" disagrees with "mutations". I am not sure whether "an existing mutations" should be plural or singular. @petrelharp?

jeromekelleher: It's a typo all right. I've created an issue #1641

PR #1596, opened 2021-03-27
Include silent mutations

### LINKED ISSUE #1551: Update mutation models to generate silent mutations and document

petrelharp:
This will close #1551. I need to have a look at verification.py yet.

Note that this includes #1595.

### DIFF
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -15,6 +15,7 @@ kernelspec:
 :tags: [remove-cell]
 
     import msprime
+    import tskit
     from IPython.display import SVG
     import numpy as np
     # Doing this to make the notebook outputs deterministic.
@@ -190,7 +191,7 @@ When we use the same seed, the resulting tree sequence is identical.
 
 As in {func}`.sim_ancestry` (see {ref}`sec_ancestry_discrete_genome`),
 the `discrete_genome` parameter controls whether mutations should be placed at
-discrete, integer coordinates or continously at floating point positions.
+discrete, integer coordinates or continuously at floating point positions.
 `discrete_genome=True` is the default:
 
 ```{code-cell}
@@ -200,7 +201,7 @@ mts.tables.sites.position
 ```
 
 Specifying `discrete_genome=False` places mutations at floating point positions
-continously along the genome.
+continuously along the genome.
 
 ```{code-cell}
 mts = msprime.sim_mutations(ts, rate=0.1, random_seed=5, discrete_genome=False)
@@ -229,7 +230,7 @@ mts = msprime.sim_mutations(ts, rate=2, start_time=1, random_seed=1)
 mts.tables.mutations.time
 ```
 
-Note, however, that the child node of the edge where the mutation occured can be younger
+Note, however, that the child node of the edge where the mutation occurred can be younger
 than `start_time`.
 
 ```{code-cell}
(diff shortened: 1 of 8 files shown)


petrelharp:
Well, this is a bit unfortunate: to make Jukes-Cantor parent independent requires 25% of the mutations being silent. This is not so great because people will get 75% the mutation rate that they expect. We could get around this by (a) letting each mutation model have a "rate multiplier" that we set by default to 1/(1- prob(silent mutation)), thus making the rate of non-silent mutations what people expect. This would be straightforward, if a bit annoying. Other options would be (b) making silent mutations optional, and asking that people turn them on if they want to stack mutations, or (c) just documenting that the mutation rate will be lower than they think.

I guess I'm leaning torward (b). This would happen at the python level, in e.g., the definition of the JC69 mutation model.

Another somewhat unfortunate thing is that the first example on the mutation docs page goes from this:
Screenshot from 2021-03-27 22-03-33
to this:
Screenshot from 2021-03-27 22-03-41
... so we end up needing to explain the silent mutation thing right off the bat.

petrelharp:
Ah-ha: another unfortunate thing about the change is that the Binary model, which is called by `msprime.mutate( )` now produces silent mutation, and is therefore a change from the previous behavior. So at the very least we need to modify `msprime.mutate( )` to use a "binary model without silent mutations".

jeromekelleher:
@mergifyio update

jeromekelleher:
Thanks for digging in to this @petrelharp, this is great. I think option 2 is the right way to go, given that we'd expect the "rate" to be what we want and it would be surprising if (by default) JC69 did produce loads of silent mutations, and we have a hard requirement that the old ``mutate`` and ``simulate`` code path can't produce silent mutations. We can make a big prominent warning about needing to allow silent mutations if someone wants to run multiple rounds of mutate.

What do we call the option? It's kind of more like a property of the model than a parameter to the function, isn't it? Could we have ``JC69MutationModel(parent_independent=False)`` or something?

petrelharp:
> What do we call the option? It's kind of more like a property of the model than a parameter to the function, isn't it? Could we have JC69MutationModel(parent_independent=False) or JC69MutationModel(silent=False) or something?

Right; depending on the value of the argument you'll get a different MatrixMutationModel out. The argument could be called `silent` or `parent_independent` for JC69 - same thing - but neither are perfect for the other models: HKY and F84 *always* have some silent mutations (unless they've reduced to JC69) and cannot be completely parent-independent (if kappa != 1). What the argument will do for them is (a) add *more* silent mutations, and (b) make them closer to parent-independent (completely, if kappa=1). Let's see:
- downsides of `silent`: people will think that `silent=False` always gives no silent mutations. This will be mildly confusing but mostly harmless as silent mutations don't affect almost anything.
- upsides of `silent`: it actually *will* remove silent mutations for JC69.
- downsides of `parent_independent`: people will think that `parent_independent=True` guarantees a parent-independent model, which is not the case - it just makes it *closer* to parent-independent.
- upsides of `parent_independent`: it better connotes why people would actually want to use it.

Let's see: how about `mostly_silent`? Or, `minimize_silent_mutations`? Or, `close_to_parent_independent`? Or, `parent_independent` but throw a warning if the model isn't *actually* parent-independent? The warning seems the most correct, but also a bit nitpicky.

jeromekelleher:
We could have slightly different option names depending on the model? So, ``JC69(parent_independant=False)``, but something slightly different for HKY etc which is accurate? I don't see any problem with having different parameter names for the different models - better to have meaningful names that can be interpreted meaningfully in isolation rather than try shoehorn one concept in everywhere.

petrelharp:
I've just put in `state_independent` for all of them, documenting appropriately. I think it's fine, and makes the testing and documentation easier. I've also documented:
- the fact there's silent mutations sometimes, but that this doesn't affect most things
- what "state independence" means and why you need it if you're going to mutate multiple times
- how to adjust 'mutation rate' for silent mutations

I think this is ready to go? The docs could use a proofread, at least.

docs/mutations.md
```diff
@@ -269,13 +348,127 @@ print(f"Before: {mts.num_mutations} mutations.\n"
...
-Adding ancestral mutations can result in confusing situations, so should be
-done with care. For instance, new mutations added to an already mutated site
+This might be done, for instance, if a tree sequence is the result of a forwards
+simulation that has non-neutral variation, and you'd like to add neutral mutations.
+
+(sec_mutations_state_independence)=
+
+### Silent mutations and state-independence
+
+However, adding mutations in a way that might result in a new mutation being
```
jeromekelleher, line 358: A minor nit here: I'm imagining each of these sections as being something you might land into from a completely different part of the documentation by linking in, and not necessarily as part of a linear narrative. Starting with "However" like this is jarring from that perspective, implying that you need to have read all the preceeding stuff to be able to follow what's going on.

docs/mutations.md
```diff
@@ -269,13 +348,127 @@ print(f"Before: {mts.num_mutations} mutations.\n"
...
+A mutation model has this property if rows of its transition matrix are the same,
+and is called *state independent* (or, *parent independent*).
+For instance, mutations under the {class}`.JC69MutationModel`
+choose *any* derived allele with equal probability.
+This means that the *result* of a mutation can be chosen without knowing
+what the previous state at that site was ("state independence"),
+which implies that adding new mutations above a given mutation does not affect
+the probability of that mutation ocurring.
+However, this is only true if "any derived allele" includes the previous allele,
+i.e., if we include *silent mutations*.
```
jeromekelleher, line 389: good to link back to the silent mutations section, {ref}`silent mutations<sec_mutations_silent_mutations>`. As I say, we the sections to be fairly independent, so linking out is good.

docs/mutations.md
```diff
@@ -269,13 +348,127 @@ print(f"Before: {mts.num_mutations} mutations.\n"
...
+print("Mutated once:")
+count_transitions(mts1, pi_jc69.alleles)
+# or, mutate twice:
+mts2 = msprime.sim_mutations(ts, rate=alpha, model=pi_jc69, random_seed=7)
+mts2 = msprime.sim_mutations(mts2, rate=beta, model=pi_jc69, random_seed=8)
+print("Mutated twice:")
+count_transitions(mts2, pi_jc69.alleles)
+```
+
+These are not identical, but similar.
```
jeromekelleher, line 437: "because we used different {ref}`random seeds<sec_mutations_controlling_randomness>`" (or whatever it is).

docs/mutations.md
```diff
@@ -269,13 +348,127 @@ print(f"Before: {mts.num_mutations} mutations.\n"
...
+print(f"              parent-independent JC69 with rate {rate}: {mts1.diversity():.2f}")
+print(f"              parent-independent JC69 with rate {rate * 4/3}: {mts_with_silent.diversity():.2f}")
+```
+
+```{important}
+1. If you're going to add mutations to the same tree sequence more than once,
+    you should use the ``parent_independent`` mutation models
+    (or carefully consider the consequences).
+2. If you use a parent-independent model, you need to make your mutation rate higher,
+    to account for the silent mutations. See :ref:`sec_mutations_adjusting_for_silent`
```
jeromekelleher, line 469: {ref} rather than :ref: (I'm considering adding some rules in my vimrc to do this automatically when I'm in a markdown file!)

docs/mutations.md
```diff
@@ -483,6 +624,55 @@ Pt = scipy.linalg.expm(t * rate * Q)[i,j]
 If the top of a branch of length {math}`t` has allele {math}`i`,
 the bottom of the branch has allele {math}`j` with probability `Pt[i,j]`.
 
+
+(sec_mutations_adjusting_for_silent)=
+
+#### Adjusting mutation rates for silent mutations
+
+As discussed in :ref:`sec_mutations_silent`, matrix mutation models can insert *silent* mutations,
```
jeromekelleher, line 632: :ref: again

docs/mutations.md
```diff
@@ -483,6 +624,55 @@ Pt = scipy.linalg.expm(t * rate * Q)[i,j]
...
+P = hky.transition_matrix
+u = 2.5e-7
+mu = u / (1 - np.sum(pi * np.diag(P)))
+print(f"Mutation rate should be increased by a factor of {mu/u:.2f}.")
+```
+
+Now, let's verify we get this nucleotide diversity:
+
+```{code-cell}
+ts = msprime.sim_ancestry(1000, population_size=1000, sequence_length=1e7, recombination_rate=1e-8, random_seed=5)
```
jeromekelleher, line 667: Is the recombination rate important here? I'm trying to leave it out where it doesn't matter. I guess if we don't include it here we will get a more skewed estimate because we're only using one tree.

jeromekelleher (approved):
LGTM! This is great, thanks @petrelharp. Some minor doc comments, but good to squash n merge when you're ready.

docs/mutations.md
petrelharp: Changed to "These are very similar - any differences are due to statistical noise.", since hopefully people wouldn't expect them to be *identical*. (Alternatively, I could explain that even setting random seeds we can't expect putting down mutations twice to be the same as putting down mutations once.)

docs/mutations.md
petrelharp: ah right. my least favorite thing about jupyterbook is how many different syntaxes there are.

docs/mutations.md
petrelharp: Right; that's the idea.

petrelharp:
I've checked the Pyvolve and MutationStatsTest in verification.py; any other verification that needs doing?

petrelharp:
I've cleaned up the docs; will squash & rebase now.

petrelharp:
Go ahead and merge when tests pass! (unless there's more verification I forgot about)

PR #1598, opened 2021-03-28
HTML display for ratemap

### LINKED ISSUE #1590: HTML and text rendering for RateMap
It would be nice to have good tabular output for the RateMap class for user and docs.

@hyanwong, fancy taking a look?

hyanwong:
Fixes #1590 - slightly complicated so that we can show the start and end positions when they are not the same as the first and last pos in the positions array.

### DIFF
--- a/msprime/intervals.py
+++ b/msprime/intervals.py
@@ -23,6 +23,8 @@ import warnings
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Superseded by #1599.

Issue #1212, opened 2020-09-25
Support arbitrary rate maps for gene conversion


jeromekelleher:
sim_ancestry supports providing an arbitrary rate map as input, but the underlying infrastructure currently assumes a single rate and will trip some assertions.

Waiting on #1196 first.

jeromekelleher:
We're not going to get this implemented in time for the 1.0 release so I'm pushing it to 1.1. It's not critical, as it's a straightforward extension API wise.

andrewkern:
reviving this old issue because i think we could use this in a project i was dreaming up today... in looking through the code, it looks like a lot of the implementation is ready to go. from what I see reading over the current state, what would need to happen would be like:

1. In`msprime/ancestry.py`
   - Remove assertion at line 1426: `assert len(gene_conversion_map.rate) == 1`
   - Remove line 1427 that extracts only first rate value
   - Pass full gene conversion map to C layer
2. In`msprime/_msprimemodule.c`
   - Change from accepting a scalar gene_conversion_rate to
   accepting a dictionary gene_conversion_map
   - Add a Simulator_parse_gene_conversion_map() function
3. Verify stuff works!

@jeromekelleher would it be okay if I picked this up?

jeromekelleher:
Sounds good to me I guess. Perhaps worth having a first pass at the low-level C stuff first without changing the Python/C layer? So, write a test in test_ancestry.c that actually passes a non-trivial GC rate map to msp_set_gene_conversion_map and test that it does something sensible?

@fbaumdicker do you have any recollections on what happened with the arbitrary gene conversion map?

fbaumdicker:
Yes, I remember quite well. :) 
It depends on how correctly you want to simulate gc.

There are two types of gc events we simulate, depending on how we determine the starting point of the gc tract. For both types, once the starting point is selected, we add the second crossover after some randomly drawn tract length. ( I do not see any problem when adding the second crossover.)
 
The first type `wiuf_gene_conversion_within_event`, basically acts exactly like recombination to determine the initial position of the gc event. Here, using a map instead of a constant rate is straightforward. Everything should already be there, as @andrewkern pointed out.

[...]

PR #1599, opened 2021-03-28
RateMap API update

jeromekelleher:
Thinking about how best to display this it seemed a good idea to introduce some ``left`` and ``right`` arrays for the intervals. I considered ``start`` and ``end`` as well, but these are a bit overloaded with the map_start etc stuff already, so thought left and right are simplest. Plus, fits well with the tskit conventions.

What do you think @hyanwong?

```
┌─────────────────────┐
│left  │right  │  rate│
├─────────────────────┤
│0     │ 1     │   0.1│
│1     │ 2     │   0.2│
└─────────────────────┘
```


hyanwong:
I had the same thought in #1598, but used "from" and "to". I guess "left" and "right" are fine too. I slightly prefer "from" and "to" because I associate "left" and "right" with edges, but perhaps that's not a problem. Either is fine by me.

hyanwong:
Here's what mine looks like:

Screenshot 2021-03-28 at 18 07 08

hyanwong:
Note that I think we should use (at least) `"{:.9g}"` for the positions, which should display the full position in non-scientific notation up to 100Mb. Actually, I think I should probably use `"{:.10g}"` at a  minimum.

jeromekelleher:
OK, thanks @hyanwong. Let's use ``left`` and ``right`` - I'll add the properties here and the text display, and then we can rebase your HTML version on this. I'll also add a ``mid`` property, which will be handy for plotting.

jeromekelleher:
This has morphed into a substantial update of the RateMap API, but a necessary one I think.

The idea is generalise from the current approach of treating the flanking left and right regions specially using the ``map_start`` and ``map_end`` values to introducing the idea of having "unknown" rates anywhere in the map. These are indicated by NaN values. This is much better because it allows us to deal with centromeres etc in a general way, without having to layer on more and more complexity.

Initially here we just mark the start and end of chromosomes read by ``read_hapmap`` and the flanks of ``slice`` as unknown, but we can imagine adding methods later (or others doing so) that could look at recombination maps and make decisions about which bits look fishy and should be marked "unknown".

The plan then would be to follow up with a parameter ``trim_unknown=True`` to ``sim_ancestry``. This would then snip out the regions of the input recombination map that were marked as unknown, unless told not to. This would generally be quite useful, and, more fundamentally, is the correct thing to do. 

The one remaining question would be what recombination rate one should use in the unknown regions for simulation. Zero is the obvious and easy choice, but I guess you could argue it might be free recombination too. I guess the right interface for specifying that is something we need to think about.

hyanwong:
Ha. I actually wondered if we should have some sort of marker for unknown flanks rather than using 0. But I hadn't though about extending the idea to regions in the middle of the chromosome. I think this is a great idea.

The only think that would make me think otherwise is if we actually wanted to provide telomere/centromere functionality in a completely different way, by having a "mask" that we overlay onto the TS somehow. That would make it easier to "mask out" regions in general (e.g. for recombination, gene conversion, and mutations, as well as for statistical calculations), rather than having to implement something specific to recombination rate maps, then again to mutation rate maps, then again to stats calculations.

I guess if `sim_ancestry` called `delete_intervals` on any regions that had NaN in the recombination map (or if there is NaN in the gene conversion map, [when that gets implemented](https://github.com/tskit-dev/msprime/issues/1212)), then there won't be any edges in those regions for `sim_mutate` to use, and (hopefully) the stats calculations will work sensibly too, so maybe what @jeromekelleher is suggesting is akin to a mask anyway.

hyanwong:
> The one remaining question would be what recombination rate one should use in the unknown regions for simulation. Zero is the obvious and easy choice, but I guess you could argue it might be free recombination too. I guess the right interface for specifying that is something we need to think about.

It doesn't matter at the ends of the chromosomes. So you could easily implement the NaN behaviour to work for the ends of the chromosomes, and for the moment raise a NotImplementedError when there's a NaN in the middle of the chromosome (until we decide what the right behaviour is).

Incidentally, if it's free recombination (which I suspect is the right thing), then that provides an easy way to create multiple chromosomes, assuming that we can add a NaN region of negligible width between one integer site and the next.

jeromekelleher:
Ah, I see that my earlier text wasn't clear:

> The plan then would be to follow up with a parameter trim_unknown=True to sim_ancestry. This would then snip out the regions of the input recombination map that were marked as unknown, unless told not to. This would generally be quite useful, and, more fundamentally, is the correct thing to do.

I meant to say "snip out the regions of the **output tree sequence** in the regions of the input recombination map that were marked unknown".

This is basically what you're saying above, isn't it? So, by default we snip out any regions that are unknown from the output tree sequence before returning it to the user. This means that everything downstream works like it should - if the region was marked as unknown, then it becomes missing data in the output ts.

In terms of combining different unknown regions from recomb and gene conversion, I'd imagine we just take the union. If the regions is unknown for any of the processes, then it's missing data in the output.

hyanwong:
> This is basically what you're saying above, isn't it? So, by default we snip out any regions that are unknown from the output tree sequence before returning it to the user. This means that everything downstream works like it should - if the region was marked as unknown, then it becomes missing data in the output ts.

Right. This does sound sensible. One thing: in the stats documentation, it says

>For site statistics, any windows without any sites in them never get touched, so they will have a value of 0. For branch statistics, any windows with no branches will similarly remain 0. That said, you should not rely on the specific behavior of whether 0 or nan is returned for “empty” cases like these: it is subject to change.

So if we intend this to work for stats, we should ensure the behaviour in the "empty" cases works as expected.

> In terms of combining different unknown regions from recomb and gene conversion, I'd imagine we just take the union. If the regions is unknown for any of the processes, then it's missing data in the output.

Yes, I guess so. What happens when a GC event extends into an unknown region? There may be various oddities here that we need to consider. And we should explain this all somehow to the user. It might be that when explaining how regions are "masked out" like this, some use-cases emerge that we haven't thought about.

Having said which, I can't see a problem with replacing the current behaviour, invisibly, with the NaN behaviour. So perhaps better to do it now and iron out wrinkles later.

petrelharp:
> The one remaining question would be what recombination rate one should use in the unknown regions for simulation. Zero is the obvious and easy choice, but I guess you could argue it might be free recombination too. 

I think zero is the right default: it's closer to correct for centromeres, and it results in less work. As implemented, "free recombination" would be a rate of log(2) per bp, right? which would tremendously slow down simulations if any missing gaps were longer than 1bp.

hyanwong:
> > The one remaining question would be what recombination rate one should use in the unknown regions for simulation. Zero is the obvious and easy choice, but I guess you could argue it might be free recombination too.
> 
> I think zero is the right default: it's closer to correct for centromeres, and it results in less work. As implemented, "free recombination" would be a rate of log(2) per bp, right? which would tremendously slow down simulations if any missing gaps were longer than 1bp.

Or, fo recombination, we could fill in the "missing" region with the average rate for the time being? So that's three options!

hyanwong:
> I think zero is the right default: it's closer to correct for centromeres, and it results in less work. As implemented, "free recombination" would be a rate of log(2) per bp, right? which would tremendously slow down simulations if any missing gaps were longer than 1bp.

Oh, I just thought. We *do* usually have a sensible recombination rate in centromeric sections, don't we (which might be different from the average)? That's because we can work it out using normal genetic mapping tools. In fact, the centromeric region is always present (with reliable, if quite averaged-out, values) in genetic maps.  So this way of "snipping out" a region isn't quite right. Or at least, we might want an alternative way of marking that an intermediate (e.g. centromeric) region should be excluded from the resulting tree sequence, while still keeping the passed-in recombination rates in that region.

Hmm, this is tricky.

jeromekelleher:
OK great - looks like we all think this is the right way to go. Regarding the correct default rate within unknown regions, let's make it zero for the initial pass, and defer discussion as to what the correct value is to an issue. (I think probably free recombination is probably the right option, but it's not as simple as filling in the array with log(2) or whatever. We want a one-base pair segment of free recombination *at the end* of the unknown region.)

I'll open an issue.

hyanwong:
> OK great - looks like we all think this is the right way to go. Regarding the correct default rate within unknown regions, let's make it zero for the initial pass, and defer discussion as to what the correct value is to an issue. 

As I mentioned above, I think we should probably raise an error for the moment if there are NaNs within the map, until we decide what to do with these. After all, we don't support this at the moment anyway.

hyanwong:
> > OK great - looks like we all think this is the right way to go. Regarding the correct default rate within unknown regions, let's make it zero for the initial pass, and defer discussion as to what the correct value is to an issue.
> 
> As I mentioned above, I think we should probably raise an error for the moment if there are NaNs within the map (but not if there are NaNs at either end), until we decide what to do with these. After all, we don't support this at the moment anyway.

hyanwong:
And I've just realised the (seemingly hacky but) probably correct way to do centromeres. We keep the recombination map with its (correct) values in centromeric regions, but we set the *gene conversion map* (when it is implemented) to NaN there. After all, we definitely don't know the GC rate within those regions. Then when we take the union, the centromeric region will be end up being deleted from the TS.

This does mean that we need to check that recombination can occur normally, as specified by the recombination map, in regions marked as NaN by the GC map. And the interval deletion comes later. So it's mainly a question of doing things in the right order.

The reason why the recombination_map is special is that its effects are not local, but across the whole simulated chromosome. So we need to be careful when rubbing out regions (which is the point about what rate to use in a snipped out section).

jeromekelleher:
Soungs good @hyanwong, but let's move that discuss to the issue. Thing will get lost, and this thread is getting distracted from the main issue. I've opened #1604.

msprime/intervals.py
```diff
@@ -168,134 +160,175 @@ def sequence_length(self):
...
+            return self.slice(key.start, key.stop)
+        if isinstance(key, numbers.Number):
+            index = self.find_index(key)
+            return self.rate[index]
+        raise KeyError("Key {key} not in map")
+
+    def __str__(self):
+        data = [
+            [
+                [f"{left:.10g}"],
```
hyanwong, line 312: I assume this just prints out the unknown regions as "NaN", which is nice?

jeromekelleher:
This isn't quite finished, but it could use some review please @hyanwong, @petrelharp, @grahamgower.

The basic idea is that NaN values in the ``rate`` array define "holes" of missing data in the map. The rest is following through on the consequences of this for the semantics of the different operations. Most of them are straightforward enough, I think. The tricky bit will be explaining that the numpy array interface doesn't mask out the missing intervals, but the mapping interface does. We'll also need to provide some interface for accessing the missing/non-missing via numpy arrays, but I think we can push that to future work. What we have here is enough for what we need in the first instance: marking the flanks as missing data and pushing that through to the simulation.

msprime/intervals.py
jeromekelleher: Yep.

docs/rate_maps.md
```diff
@@ -1,37 +1,342 @@
...
+See the {ref}`sec_rate_maps_creating` section for different ways of
+creating {class}`.RateMap` instances with various properties.
+:::
+
+To create the {class}`.RateMap` instance we called the contructor
+with two lists, ``position`` and ``rate``. These define the
+intervals along the genome and the rates in each interval. Because
+we are interested in nonoverlapping contiguous intervals, we can
+define all the {math}`n` intervals with {math}`n + 1` positions
+and correspoding {math}`n` rate values.
```
hyanwong, line 119: typo "corresponding"

hyanwong:
"Missing" sounds good to me. I would be (slightly) disinclined to use "missing data", as we could easily have genetic data present in those regions, it's just that the *rate* is missing. 

As I said elsewhere, if the rate data is calculated from a cumulative measure, like recombination, then I don't think we will usually have missing rates in the middle of a sequence. But it's as well to allow for this possibility in the design for other cases when rate is calculated in a different manner.

Otherwise, this all looks good to me. It's nice that the missing flanks of the map end up as empty regions of the tree sequence. We should have a word or phrase for an empty region in a tree sequence (one with no edges), that is subtly different from the case when we have edges but not any that are connected to samples (see https://github.com/tskit-dev/tskit/issues/1285). As you can see, I've been using "empty regions" for this.

hyanwong:
P.s. I do think, in the long term, it would be tidier to have `intervals.py` as part of tskit rather than msprime, so it might be useful to check that the RateMap code is as "msprime-independent" as possible (while, of course, allowing msprime to depend extensively on RaeMap functionality). I do think this is the case at the moment, but it's useful to bear in the back of our minds.

jeromekelleher:
> P.s. I do think, in the long term, it would be tidier to have intervals.py as part of tskit rather than msprime, so it might be useful to check that the RateMap code is as "msprime-independent" as possible (while, of course, allowing msprime to depend extensively on RateMap functionality). I do think this is the case at the moment, but it's useful to bear in the back of our minds.

Yep, I agree.

jeromekelleher:
Nobody has complained about this, so I'm going to go ahead and merge. We can still tweak the semantics of things a bit, but I think the overall picture is the right way to go.

hyanwong:
> Nobody has complained about this, so I'm going to go ahead and merge. We can still tweak the semantics of things a bit, but I think the overall picture is the right way to go.

Yes, I agree. Well done on pushing this through under time pressure!

grahamgower (commented):
I'm coming late to this, sorry, but it LGTM. I'm curious though, is there a specific use case in mind that wants/needs the full generality of the mapping protocol?

Issue #1419, opened 2021-01-28
Examples section for rate maps


jeromekelleher:
We have a new rate_maps.md file in the utilities section, which currently is a stub. We should fill this out with examples, discussing an illustrating the various features.

This isn't about running simulations here, just making, reading in and manipulating rate maps.

awohns:
I've been writing an example of `sim_mutations` using a RateMap and it seems that ratemaps for mutation rates work differently than for recombination rates. I haven't seen a test of non-uniform RateMap for `sim_mutations`, but rates should be considered the same as in a recombination rate map, right?

jeromekelleher:
> I haven't seen a test of non-uniform RateMap for sim_mutations, but rates should be considered the same as in a recombination rate map, right?

Yes, definitely. What's the difference you've seen?

jeromekelleher:
Actually, getting in to the details of this I realise that the RateMap should be a collections.abc.Mapping from Intervals -> rates and the API will need to be updated slightly to accommodate.

I'll do the write up at the same time. Apologies if you've already made a start on this @hyanwong.

hyanwong:
I have a section of text written already. Shall I just draft PR what I have and you can steal bits?

jeromekelleher:
That would be good, thanks.

hyanwong:
> That would be good, thanks.

It's at https://github.com/tskit-dev/msprime/pull/1602

It assumes a nice html display for a RateMap (https://github.com/tskit-dev/msprime/pull/1598), though.

In it I do actually run a simulation with a sliced RateMap, to show why `slice()` is useful. If that's a bit too much detail, perhaps it could be moved somewhere else.

hyanwong:
It's currently failing because of Matplotlib < 3.4.0. We could either update the required version, or use `plt.steps` instead of the slightly nicer `plt.stairs` which seems to be a newly introduced MPL function

jeromekelleher:
OK, thanks. Leave it with me - I've realised that we actually need to generalise some APIs around RateMaps as well. Will have a proposal later today.

PR #1602, opened 2021-03-29
RateMap doc text

hyanwong:
First start at addressing #1419


jeromekelleher:
I've cherry picked this commit into #1599, thanks @hyanwong. I'll get merged in with that PR as one of a number of commits.

Issue #1600, opened 2021-03-28
Need to clarify in docs that log_mutation_likelihood only applies to a specific model


hyanwong:
I suspect that the entire section https://tskit.dev/msprime/docs/latest/api.html#msprime.log_mutation_likelihood was written before we had flexible mutation models. I assume the single mutation rate parameter means that ~~the model is the JC69MutationModel~~ this applies to any model with a continuous genome and constant mutation rate between alleles. We should say that somewhere.

jeromekelleher:
@JereKoskela, thoughts?

JereKoskela:
Only the infinite sites model is supported, and only on continuous genomes. I can add a warning to that effect at the top of the docstring.

hyanwong:
(and only the JC69MutationModel within the infinite sites model, right @JereKoskela ?)

JereKoskela:
I think that's probably right, but I'm not very familiar with the nomenclature of the various mutation models, so I'll spell this out to avoid being confused. Since each site can mutate at most once, there are essentially only two alleles (ancestral and mutant). Only the ancestral-to-mutant mutation is permitted, and happens with constant rate.

hyanwong:
> I think that's probably right, but I'm not very familiar with the nomenclature of the various mutation models, so I'll spell this out to avoid being confused. Since each site can mutate at most once, there are essentially only two alleles (ancestral and mutant). Only the ancestral-to-mutant mutation is permitted, and happens with constant rate.

Yep, I get that, but I assume that this applies even to models that allow a single mutation to multiple possible alleles  at a site (like the JC69MutationModel), as long as the infinite sites assumption is met (i.e. only one mutation per site). So I think that we need to avoid talking about number of alleles allowed at a site (e.g. I presume all the single base DNA mutation models allow for 4 alleles: ATCG, even in continuous genomes), and just say that this applies to genomes where only one mutation can occur per site (i.e. discrete_genome=False), with a constant mutation rate. It's just a matter of careful wording I think.

JereKoskela:
> > I think that's probably right, but I'm not very familiar with the nomenclature of the various mutation models, so I'll spell this out to avoid being confused. Since each site can mutate at most once, there are essentially only two alleles (ancestral and mutant). Only the ancestral-to-mutant mutation is permitted, and happens with constant rate.
> 
[...]

PR #1603, opened 2021-03-29
Change the infinite sites wording

hyanwong:
While discussing https://github.com/tskit-dev/msprime/issues/1600 I noticed that the text describing infinite sites could be a bit confusing to the uninitiated, as it refers to the "infinite sites model", which is not one of the "mutation model"s that we actually list. So perhaps we should refer to it as the "infinite sites assumption" instead (and indeed, this makes it a little more obvious that any of the mutation models we have can be combined with an IS assumption?

I've also put it in a `{note}` for greater emphasis, as I think it's pretty important. But we could even put it in an `{important}` block instead?

### DIFF
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -207,9 +207,13 @@ mts = msprime.sim_mutations(ts, rate=0.1, random_seed=5, discrete_genome=False)
 mts.tables.sites.position
 ```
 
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
LGTM, thanks,

PR #1609, opened 2021-03-31
Update notes for 0.x users.

### LINKED ISSUE #1301: Guide to updating code to use 1.0 APIs
We should make it easy for existing users to update their code to use the 1.0 APIs. There's a number of important gotchas:

1. Samples are for k-ploid *individuals* now, with ploidy=2 by default. This *will* catch people out (has already caught me several times!)
2. Discrete genome by default
3. You have to simulate mutations separately
4. Can't use old style demography descriptions (but you can use the Demography.from_old_style())

### LINKED ISSUE #1425: Review legacy code section
Needs an update and finalisation (now that docs structure is settled)

See also #1301

jeromekelleher:
Closes #1425
Closes #1301

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -393,6 +393,8 @@ Note that this is the **default**, and can be overridden by the
 {attr}`.SampleSet.ploidy` value, as described in the
 {ref}`sec_ancestry_samples_sampling_details` section.
 
+(sec_ancestry_ploidy_coalescent_time_scales)=
+
 ### Coalescent time scales
 
 The ploidy argument also affects the time scales of
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -397,9 +397,10 @@ Note that this is the **default**, and can be overridden by the
 
 ### Coalescent time scales
 
-The ploidy argument also affects the time scales of
-coalescent models, in a similar way to
-{ref}`population size<sec_ancestry_demography>`.
+The ploidy argument also
+affects the time scale on which coalescence occurs,
+since it affects the total number of genomes in the population
+(see also {ref}`population size<sec_ancestry_demography>`).
 For example, consider the following two simulations, which
 are identical except for different ``ploidy`` values
 (note we use the {class}`.SampleSet` object to make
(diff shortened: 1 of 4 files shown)


docs/legacy.md
```diff
@@ -75,15 +82,14 @@ function and is very similar. There are some important differences though:
   two effects:
 
   1. Sets the default number of sample nodes per *individual*
-  2. Changes the timescale of the coalescent process (TODO link to a section
-     that explains this effect.) By default `ploidy` is 2 and
+  2. Changes the {ref}`timescale<sec_ancestry_ploidy_coalescent_time_scales>`
+     of the coalescent process.  By default `ploidy` is 2 and
      time is scaled scaled in units of 4N generations, which is the same as
      msprime 0.x.
```
petrelharp, line 88: I think this is the wrong way to say it, because time is not scaled (at least nowhere visible to the user). How about:
```
2. Changes the {ref}`time scale<sec_ancestry_ploidy_coalescent_time_scales>` over which coalescence occurs.
    By default `ploidy` is 2, and so mean time to common ancestor in a population of size `N` is `2N` generations,
    which is the same as msprime 0.x.
```

docs/ancestry.md
```diff
@@ -393,6 +393,8 @@ Note that this is the **default**, and can be overridden by the
 {attr}`.SampleSet.ploidy` value, as described in the
 {ref}`sec_ancestry_samples_sampling_details` section.
 
+(sec_ancestry_ploidy_coalescent_time_scales)=
+
```
petrelharp, line 397: suggested changes below: 
- "affects the time scale on which coalescence occurs, since it affects the total number of genomes in the population
   (see also {ref}`population size<sec_ancestry_demography>`)."
- "The resulting tree sequences are identical, except that node time differ by a factor of 2, because in the underlying coalescent algorithm the total number of genomes in the population acts as a time scaling."

docs/legacy.md
jeromekelleher: done.

docs/ancestry.md
jeromekelleher: done

PR #1611, opened 2021-03-31
Detect unsorted tables on output.

### LINKED ISSUE #1606: in some cases sim_ancestry with initial state produces unordered tables
I ran across this when trying to recapitate something and forgot to `keep_input_roots`.

Here is a MWE:
```
import tskit, msprime

t = tskit.TableCollection(sequence_length=2)
t.populations.add_row()
t.nodes.add_row(time=0, flags=tskit.NODE_IS_SAMPLE, population=0)
t.nodes.add_row(time=1, flags=tskit.NODE_IS_SAMPLE, population=0)
t.nodes.add_row(time=2, population=0)
t.edges.add_row(left=0, right=1, parent=2, child=0)

ts = t.tree_sequence()

msprime.sim_ancestry(initial_state=ts, population_size=1, random_seed=5)
```
produces
```
_tskit.LibraryError: Edges must be listed in (time[parent], child, left) order; time[parent] order violated
```
The error occurs *after* simulation, when trying to turn the tables that result from simulation back into a tree sequence.

jeromekelleher:
Closes #1606

@petrelharp - this feels like a hack, but I don't see much point in cluttering up the low-level code trying to detect the edge case.

What do you think about emitting a warning? Is this likely to be something that only happens when someone makes a mistake?


petrelharp:
Sure, why not. Recapitating SLiM simulations all the lineages have the same time, but I could also imagine someone inserting ancient samples by sticking some other sampled lineages at older times into the initial state. But then they should be doing this with demography instead of initial state.  I guess I do think that this is much more likely to be due to forgetting `keep_input_roots` than anything else, so a warning is called for.

Are there tests for correctness of initial_state with situations where not all the samples are at the same time, btw?

jeromekelleher:
I've left out the warning for now @petrelharp and left the TODO as I wasn't sure what the warning should say. Happy to insert one if you like.

I just checked and we have pretty good tests for correctness on initial state with ancient samples, as that is how all simulation initialisation is done now.

petrelharp:
Probably best not to warn.

jeromekelleher:
Looks like MacOS builds are stuck. I'm going to merge as this is all working and I'd like to release.

PR #1613, opened 2021-04-01
Docs update to multiple chromosomes and likelihoods

### LINKED ISSUE #1600: Need to clarify in docs that log_mutation_likelihood only applies to a specific model

JereKoskela:
Some brief clarifications to the section on simulating multiple chromosomes under multiple merger coalescents, and to the mutation likelihood docstring.

Ping @jeromekelleher, @hyanwong, @apragsdale.

Closes #1600.

### DIFF
--- a/msprime/likelihood.py
+++ b/msprime/likelihood.py
@@ -28,9 +28,19 @@ def log_mutation_likelihood(ts, mutation_rate):
     """
     Returns the unnormalised log probability of the stored pattern of mutations
     on the stored tree sequence, assuming infinite sites mutation. In particular,
-    each stored site must only contain a single mutation. The omitted normalising
-    constant depends on the pattern of mutations, but not on the tree sequence or
-    the mutation rate.
+    each stored site must only contain a single mutation, and all mutant alleles
+    are treated interchangeably as having arisen from a single mutation rate.
+    This is automatically true for some msprime mutation models, such as JC69.
+    The omitted normalising constant depends on the pattern of mutations, but not
+    on the tree sequence or the mutation rate.
+
+    .. warning::
+        The infinite sites assumption means that this method is only valid for
+        continuous genomes. It can also be run on discrete genomes, for which it
+        may provide a reasonable approximation when the number of sites is large
+        and the mutation rate is low. See :ref:`sec_ancestry_discrete_genome`
+        for how continuous genomes can be specified when simulating tree sequences
+        using msprime.
 
     The function first computes the probability of the overall number of mutations
     :math:`M` from the Poisson probability mass function
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -696,9 +696,18 @@ ancestry patterns for very recently admixed populations.
 
 Additionally,
 {ref}`Multiple merger coalescents <sec_ancestry_models_multiple_mergers>`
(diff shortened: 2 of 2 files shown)


msprime/likelihood.py
```diff
@@ -28,9 +28,15 @@ def log_mutation_likelihood(ts, mutation_rate):
     """
     Returns the unnormalised log probability of the stored pattern of mutations
     on the stored tree sequence, assuming infinite sites mutation. In particular,
-    each stored site must only contain a single mutation. The omitted normalising
-    constant depends on the pattern of mutations, but not on the tree sequence or
-    the mutation rate.
+    each stored site must only contain a single mutation, and all mutant alleles
+    are treated interchangeably as having arisen from a single mutation rate.
```
hyanwong, line 32: Do we want to say e.g. "(this is automatically true for some msprime mutation models, such as JC69)"

msprime/likelihood.py
```diff
@@ -28,9 +28,15 @@ def log_mutation_likelihood(ts, mutation_rate):
...
+    The omitted normalising constant depends on the pattern of mutations, but not
+    on the tree sequence or the mutation rate.
+
+    .. warning::
+        This method only supports continuous genomes.
```
hyanwong, line 37: How about "The infinite sites assumption means that this method is only valid for continuous genomes"?

We could also say "(and an error will be output if it is attempted on discrete genomes)", if that's true (I don't know if it is!)

msprime/likelihood.py
JereKoskela: Good idea, I'll add that in.

msprime/likelihood.py
JereKoskela: I like the rewording.

An error hasn't been implemented, but it would be a good idea to add it to both the likelihood methods if it is a simple thing to do. I guess we could either use MSP_ERR_BAD_PARAM_VALUE, or define a whole new error.

hyanwong: Thanks. I suspect it's difficult to detect if you are in a (mutationally) continuous genome or not, other than by checking if all the site positions are integers.

I guess people might want to calculate the likelihood even on a discrete genome, if they think the mutation rate is very low, so maybe it should be a warning, not an error?

JereKoskela: Are there any warnings (other than MSP_WARN_UNUSED) in the source code as it is? I have no idea how to generate one correctly, and could use an example.

jeromekelleher: Let's not bother with this - a warning to the user and an explanation what will happen if you run it on discrete coordinates will do for now.

jeromekelleher: As in, in the documentation.

jeromekelleher (approved):
LGTM, thanks.

Good to merge after squash

jeromekelleher:
CI failures are unrelated, GitHub's Macs are running very well.

jeromekelleher:
@mergifyio rebase

PR #1617, opened 2021-04-01
small clarification

petrelharp:
Reading the docs just now it was not easy for me to tell which was `source` and which was `dest` in setting migration rates. This short note I think should help.


jeromekelleher (approved):
LGTM, thanks!

Issue #1615, opened 2021-04-01
Examples for coalescence rates and mean times


jeromekelleher:
#1614 added some content for explaining the numerical methods on the demography debugger, and added a TODO section for the  coalescence_rate_trajectory() and mean_coalescence_time() methods. I decided I wasn't the right person to explain these.

@petrelharp, @apragsdale, would one of you be able to take this up? We can probably reuse the examples from whichever paper we used this on, right?

petrelharp:
I'll have a go at it.

petrelharp:
Starting [here](https://github.com/tskit-dev/msprime/pull/1614/files#diff-5ec23b83ce8afddf382b03ff5258476aded99dca85b8c21f6551ba74f86891dfR1653), right?

jeromekelleher:
Yep - are you OK to merge that other PR first? No major hurry with this, it's no going to block the beta.

Issue #1618, opened 2021-04-01
discrepancy between coalescence rates for msprime and stdpopsim OOA model


petrelharp:
This is probably a bug in how we compute coalescence rates with the new demographic models, but could conceivably be a bug in `msprime._ooa_model()`:
```
import msprime
import stdpopsim
import numpy as np

msp = msprime.Demography._ooa_model().debug()

species = stdpopsim.get_species("HomSap")
model = species.get_demographic_model("OutOfAfrica_3G09")
sps = msprime.Demography.from_old_style(
        population_configurations=model.population_configurations,
        migration_matrix=model.migration_matrix,
        demographic_events=model.demographic_events).debug()

T = np.concatenate([
            np.linspace(0, 1000, 2001),
            np.linspace(1000, 1e4, 401)[1:]
])

msp_R, _ = msp.coalescence_rate_trajectory(T, [0, 2, 0, 0, 0, 0])
sps_R, _ = sps.coalescence_rate_trajectory(T, [0, 2, 0])

print("msprime:", 1/(2*msp_R[-3:]))
print("stdpopsim:", 1/(2*sps_R[-3:]))
```
produces:
```
msprime: [-5.37590936e+257 -3.17744309e+267 -1.87803475e+277]
stdpopsim: [7300. 7300. 7300.]
```
These should be identical. *Note:* there's a warning about possible numerical instability, but that's not what's going on here.

*Edit:* actually using the current msprime, I get
```
msprime: [31995.80756988 32072.04220953 32148.51218145]
stdpopsim: [7300. 7300. 7300.]
```
which is much better but still not identical.

petrelharp:
The problem with OOA is fixed in #1620 :tada:; so this now depends on that.

jeromekelleher:
Closed in #1620

PR #1619, opened 2021-04-01
examples of coalescence rate calculations

petrelharp:
Here's some draft docs. Note that I'm using the stdpopsim OOA model instead of the built-in one because of #1618.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -3298,7 +3298,7 @@ class DemographyDebugger:
             t_j = all_steps[jj]
 
             # apply any mass migration events to P
-            # so if we sample at this time, we do no account for the instantaneous
+            # so if we sample at this time, we do not account for the instantaneous
             # mass migration events that occur at the same time. that will show up
             # at the next step
             if t_j > sample_time:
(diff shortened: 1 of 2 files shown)


petrelharp:
Ok, this is failing for lack of `stdpopsim`. It'd be nice to use stdpopsim; but it's still old-style demography for now, so maybe we don't want to. Or, we could fix #1615 first and then use the same OOA implementation from earlier in this document.  Here's the plot this produces, FYI:
Screenshot from 2021-04-01 13-00-02

petrelharp:
The problem with OOA is fixed in #1620 tada; so this now depends on that.

petrelharp:
Ok, I think this is ready to go, except for a squash.

docs/demography.md
```diff
@@ -1652,9 +1685,149 @@ distributed in space.
...
+```
+
+```{code-cell}
+# first define Gutenkunst et al model as above
+debug = demography.debug()
+
+T = np.concatenate([
+            np.linspace(0, 1000, 2001),
+            np.linspace(1000, 1e4, 401)[1:]
+])
```
jeromekelleher, line 1789: There's  a lot going on in this code cell, it'd be nice to break it up and explain the different chunks as we go.

docs/demography.md
```diff
@@ -1652,9 +1685,149 @@ distributed in space.
...
+
+R = np.zeros((len(T), 3))
+R[:,0], _ = debug.coalescence_rate_trajectory(T, [2, 0, 0, 0, 0, 0])
```
jeromekelleher, line 1792: Hmm, so this indicates to me that the interface isn't great. Maybe we could should reuse the ``samples`` specification from ``sim_ancestry`` instead, so you can do things like ``{"A" : 2}`` rather than having to make an array of zeros? This'll get problematic for big models, where we've got 10s populations. Also seems like quite an error prone way of doing it.

jeromekelleher (commented):
Looks great, thanks @petrelharp. See some notes about the interface - I'm wondering if we can tweak it a bit so that we have to remember population indexes less?

It might be nice to add a note somewhere about how we depend on model assumptions here. I assume this should basically hold for any WF-type situation? We are assuming ploidy=2 though, right?

docs/demography.md
petrelharp: Yep - let's push that to a different issue, though?

petrelharp:
> It might be nice to add a note somewhere about how we depend on model assumptions here. I assume this should basically hold for any WF-type situation? We are assuming ploidy=2 though, right?

I'm not totally sure which bit you're referring to, but I've added a bit of a note about this?

docs/demography.md
jeromekelleher: OK

jeromekelleher:
Looks good to me, ready to go after a rebase.

jeromekelleher (approved):

jeromekelleher:
Not sure what happened here @petrelharp  - did this get merged? If not, maybe open a different PR on a new branch or something with a fresh rebase?

petrelharp:
I think I know what happened - I accidentally added the AUTOMERGE label to this PR instead of the other, but then removed it 20 seconds later; but this confused AdminBot.

PR #1646, opened 2021-04-13
Numerics refactor

jeromekelleher:
I think most of this is good, but I'm confused about ``mean_coalescence_time`` now and what it's actually computing. 

@petrelharp, need your input here.

### DIFF
--- a/docs/demography.md
+++ b/docs/demography.md
@@ -1795,16 +1795,16 @@ demography.set_migration_rate(source="B", dest="A", rate=0.02)
 
 debug = demography.debug()
 T = np.linspace(0, 500, 51)
-RAA, _ = debug.coalescence_rate_trajectory(T, [2, 0])
-RBB, _ = debug.coalescence_rate_trajectory(T, [0, 2])
-RAB, _ = debug.coalescence_rate_trajectory(T, [1, 1])
+RAA, _ = debug.coalescence_rate_trajectory(T, {"A": 2})
+RBB, _ = debug.coalescence_rate_trajectory(T, {"B": 2})
+RAB, _ = debug.coalescence_rate_trajectory(T, {"A": 1, "B": 1})
 
 plt.plot(T, RAA, label="two A samples")
 plt.plot(T, RBB, label="two B samples")
 plt.plot(T, RAB, label="one of each")
 plt.legend()
 plt.xlabel("time ago")
-plt.ylabel("coalecence rate")
+plt.ylabel("coalecence rate");
 ```
 
 The coalescence rate of lineages sampled from different populations begins at zero,
@@ -1822,9 +1822,9 @@ change as we make the migration rate less asymmetric:
 def get_mean_coaltimes(m):
     demography.set_migration_rate(source="A", dest="B", rate=m)
     debug = demography.debug()
-    TAA = debug.mean_coalescence_time([2, 0])
-    TBB = debug.mean_coalescence_time([0, 2])
-    TAB = debug.mean_coalescence_time([1, 1])
+    TAA = debug.mean_coalescence_time({"A": 2})
+    TBB = debug.mean_coalescence_time({"B": 2})
+    TAB = debug.mean_coalescence_time({"A": 1, "B": 1})
     return [TAA, TBB, TAB]
 
 M = np.linspace(0.001, 0.02, 11)
@@ -1882,14 +1882,14 @@ demography = msprime.Demography._ooa_model()
 debug = demography.debug()
 
 T = np.concatenate([
-        np.linspace(0, 1000, 2001),
-        np.linspace(1000, 1e4, 401)[1:]
+    np.linspace(0, 1000, 2001),
+    np.linspace(1000, 1e4, 401)[1:]
 ])
 
 R = np.zeros((len(T), 3))
-R[:,0], _ = debug.coalescence_rate_trajectory(T, [2, 0, 0, 0, 0, 0])
-R[:,1], _ = debug.coalescence_rate_trajectory(T, [0, 2, 0, 0, 0, 0])
-R[:,2], _ = debug.coalescence_rate_trajectory(T, [0, 0, 2, 0, 0, 0])
+R[:,0], _ = debug.coalescence_rate_trajectory(T, {"YRI": 1})
+R[:,1], _ = debug.coalescence_rate_trajectory(T, {"CEU": 1})
+R[:,2], _ = debug.coalescence_rate_trajectory(T, {"CHB": 1})
 ```
 
 The time values look a bit weird because when at first we set them to have
@@ -1916,14 +1916,14 @@ shared "OOA" population for both the non-African populations).
 ```{code-cell}
 N = debug.population_size_trajectory(T)
 # CEU-CHB merge
-which_T = (T >= debug.epoch_times[1])
+which_T = (T >= debug.epoch_start_time[1])
 for j in (1, 2):
     N[which_T, j] = N[which_T, 3]
 # OOA-Afr merge
-which_T = (T >= debug.epoch_times[2])
+which_T = (T >= debug.epoch_start_time[2])
 for j in (0, 1, 2):
     N[which_T, j] = N[which_T, 4]
-which_T = (T >= debug.epoch_times[3])
+which_T = (T >= debug.epoch_start_time[3])
 for j in (0, 1, 2):
     N[which_T, j] = N[which_T, 5]
 
--- a/docs/demography.md
+++ b/docs/demography.md
@@ -1776,6 +1776,8 @@ distance, but as we go further back into the past
 the possible location of lineages becomes more evenly
 distributed in space.
 
+(sec_demography_numerical_coalescence)=
+
 ### Coalescence rates and mean times
 
 Lineage probabilities tell us where lineages might possibly be,
@@ -1846,6 +1848,8 @@ and that diversity within "B" goes from a level close to that found between popu
 to that within "A", as the "A→B" migration rate approaches the "B→A" rate.
 
 
+(sec_demography_numerical_trajectories)=
+
 ### Inverse instantaneous coalescence rates
 
 Coalescence rates are analogous to effective population sizes,
@@ -1887,9 +1891,9 @@ T = np.concatenate([
 ])
 
 R = np.zeros((len(T), 3))
-R[:,0], _ = debug.coalescence_rate_trajectory(T, {"YRI": 1})
-R[:,1], _ = debug.coalescence_rate_trajectory(T, {"CEU": 1})
-R[:,2], _ = debug.coalescence_rate_trajectory(T, {"CHB": 1})
+R[:,0], _ = debug.coalescence_rate_trajectory(T, {"YRI": 2})
+R[:,1], _ = debug.coalescence_rate_trajectory(T, {"CEU": 2})
+R[:,2], _ = debug.coalescence_rate_trajectory(T, {"CHB": 2})
 ```
 
 The time values look a bit weird because when at first we set them to have
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -3228,9 +3228,9 @@ class DemographyDebugger:
         return pop_size
 
     @property
-    def epoch_times(self):
+    def epoch_start_time(self):
         """
-        Returns array of epoch times defined by the demographic model
+        The array of epoch start_times defined by the demographic model.
         """
         return np.array([x.start_time for x in self.epochs])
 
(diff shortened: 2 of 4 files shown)


docs/demography.md
```diff
@@ -1822,9 +1822,9 @@ change as we make the migration rate less asymmetric:
 def get_mean_coaltimes(m):
     demography.set_migration_rate(source="A", dest="B", rate=m)
     debug = demography.debug()
-    TAA = debug.mean_coalescence_time([2, 0])
-    TBB = debug.mean_coalescence_time([0, 2])
-    TAB = debug.mean_coalescence_time([1, 1])
+    TAA = debug.mean_coalescence_time({"A": 2})
+    TBB = debug.mean_coalescence_time({"A": 1})
```
petrelharp, line 1826: Why the change? To match what it was before we need 
```
     TBB = debug.mean_coalescence_time({"B": 2})
```

jeromekelleher: Yep, sorry. Will fix.

docs/demography.md
```diff
@@ -1882,14 +1882,14 @@ demography = msprime.Demography._ooa_model()
...
-        np.linspace(1000, 1e4, 401)[1:]
+    np.linspace(0, 1000, 2001),
+    np.linspace(1000, 1e4, 401)[1:]
 ])
 
 R = np.zeros((len(T), 3))
-R[:,0], _ = debug.coalescence_rate_trajectory(T, [2, 0, 0, 0, 0, 0])
-R[:,1], _ = debug.coalescence_rate_trajectory(T, [0, 2, 0, 0, 0, 0])
-R[:,2], _ = debug.coalescence_rate_trajectory(T, [0, 0, 2, 0, 0, 0])
+R[:,0], _ = debug.coalescence_rate_trajectory(T, {"YRI": 1})
```
petrelharp, line 1890: These should have `2` instead of `1` because a single lineage can't coalesce with itself. To help out the user we could emit an error if the number of samples is not at least two?

msprime/demography.py
```diff
@@ -3478,17 +3478,26 @@ def mean_coalescence_time(
             import daiquiri
 
             daiquiri.setup(level="DEBUG")
-            debugger.mean_coalescence_time([2])
+            debugger.mean_coalescence_time(1)
```
petrelharp, line 3481: 2, not 1?

petrelharp: Oh! sorry, should have read the whole thing!

docs/demography.md
jeromekelleher: These are diploids though

tests/test_demography.py
```diff
@@ -1408,28 +1408,28 @@ def test_mean_one_pop(self):
         #  + 400 * exp(-100/200) * (1 - exp(-(300-100)/400))
         #  + 200 * exp(-100/200 - (300-100)/400)
         ddb = self.one_pop_example()
-        m = ddb.mean_coalescence_time(num_samples=[2])
+        m = ddb.mean_coalescence_time(samples=4)
```
petrelharp, line 1411: this is harmless, but so we don't wonder "why 4" maybe this should be put back to `1`.

tests/test_demography.py
```diff
@@ -1494,9 +1494,15 @@ def test_two_pop(self):
...
-        rates2, P2 = ddb.coalescence_rate_trajectory(steps=steps, num_samples=[1, 1])
+        rates, P = ddb.coalescence_rate_trajectory(steps=steps, samples={0: 1})
+        rates1, P1 = ddb.coalescence_rate_trajectory(steps=steps, samples={1: 1})
+        rates2, P2 = ddb.coalescence_rate_trajectory(
+            steps=steps,
+            samples=[
+                msprime.SampleSet(1, population=0, ploidy=1),
+                msprime.SampleSet(1, population=1, ploidy=1),
+            ],
+        )
```
petrelharp, line 1505: this is also fine, but `samples={0:1, 1:1}` is equivalent.

tests/test_demography.py
```diff
@@ -1572,9 +1578,15 @@ def test_mean_two_pop(self):
...
-        mB = ddb.mean_coalescence_time(num_samples=[0, 2])
-        mC = ddb.mean_coalescence_time(num_samples=[1, 1])
+        mA = ddb.mean_coalescence_time(samples={0: 1})
+        mB = ddb.mean_coalescence_time(samples={1: 1})
+        mC = ddb.mean_coalescence_time(
+            samples=[
+                msprime.SampleSet(1, population=0, ploidy=1),
+                msprime.SampleSet(1, population=1, ploidy=1),
+            ],
+        )
```
petrelharp, line 1588: same as previous comment

tests/test_demography.py
```diff
@@ -1628,9 +1640,15 @@ def test_convergence(self):
...
-        rates2, PP = ddb.coalescence_rate_trajectory(steps=steps, num_samples=[1, 1])
+        rates, PP = ddb.coalescence_rate_trajectory(steps=steps, samples={"A": 1})
+        rates1, PP = ddb.coalescence_rate_trajectory(steps=steps, samples={"B": 1})
+        rates2, PP = ddb.coalescence_rate_trajectory(
+            steps=steps,
+            samples=[
+                msprime.SampleSet(1, population="A", ploidy=1),
+                msprime.SampleSet(1, population="B", ploidy=1),
+            ],
+        )
```
petrelharp, line 1651: again

petrelharp:
Is the confusion because all these functions only depend on the *relative* number of samples? So, `samples={"A":1, "B":1}` is the same as `samples={"A":100, "B":100}`?  We could add a note about this if you think it would help, although we do explain that e.g. what's computed for [coalescence_rate_trajectory](https://tskit.dev/msprime/docs/stable/api.html#msprime.DemographyDebugger) is "the probability that the lineages of a randomly chosen pair of samples has not yet coalesced " - thus only depending on the relative distribution of lineages, not the total number.

tests/test_demography.py
jeromekelleher: Will do - I stuck this in wondering if the number actually changes anything.

jeromekelleher:
Right, I see - it's the *relative* distribution of lineages. OK, will tidy up. Thanks.

tests/test_demography.py
jeromekelleher: Doesn't appear to be, I get:
```
            if time >= 0 and time < 100:
                self.assertAlmostEqual(rA, 1 / (2 * 100))
                self.assertAlmostEqual(rB, 1 / (2 * 1000))
>               self.assertAlmostEqual(rC, 0)
E               AssertionError: 0.0009166666666666666 != 0 within 7 places (0.0009166666666666666 difference)
```

tests/test_demography.py
jeromekelleher: same here

jeromekelleher:
Pushed an update fixing the mistakes @petrelharp. We still need the SampleSet stuff though for the examples, though.

tests/test_demography.py
petrelharp: Oh, of course not. Hmph.

docs/demography.md
```diff
@@ -1822,9 +1822,9 @@ change as we make the migration rate less asymmetric:
 def get_mean_coaltimes(m):
     demography.set_migration_rate(source="A", dest="B", rate=m)
     debug = demography.debug()
-    TAA = debug.mean_coalescence_time([2, 0])
-    TBB = debug.mean_coalescence_time([0, 2])
-    TAB = debug.mean_coalescence_time([1, 1])
+    TAA = debug.mean_coalescence_time({"A": 2})
+    TBB = debug.mean_coalescence_time({"B": 2})
+    TAB = debug.mean_coalescence_time({"A": 1, "B": 1})
```
petrelharp, line 1827: I guess we need to do the `ploidy=1` thing here, too.

docs/demography.md
```diff
@@ -1795,16 +1795,16 @@ demography.set_migration_rate(source="B", dest="A", rate=0.02)
 
 debug = demography.debug()
 T = np.linspace(0, 500, 51)
-RAA, _ = debug.coalescence_rate_trajectory(T, [2, 0])
-RBB, _ = debug.coalescence_rate_trajectory(T, [0, 2])
-RAB, _ = debug.coalescence_rate_trajectory(T, [1, 1])
+RAA, _ = debug.coalescence_rate_trajectory(T, {"A": 2})
+RBB, _ = debug.coalescence_rate_trajectory(T, {"B": 2})
+RAB, _ = debug.coalescence_rate_trajectory(T, {"A": 1, "B": 1})
```
petrelharp, line 1800: need ploidy=1

petrelharp:
And, whoops - it's the relative distribution of *pairs* of lineages. So, what I said above is not true.  Perhaps we need a note in the documentation:
```
... note::
This describes statistics of coalescence rates averaged over pairs of lineages,
so for instance a single diploid sample in each of two populations computes the
average of the values for each of the four pairs of haploid chromosomes - two
between-population pairs and one pair within each population.
```

msprime/demography.py
```diff
@@ -3645,11 +3663,12 @@ def coalescence_rate_trajectory(
...
-        num_pops = self.num_populations
-        if not len(num_samples) == num_pops:
-            raise ValueError(
-                "`num_samples` must have the same length as the number of populations"
-            )
+        sample_sets = ancestry._parse_samples(samples, self.demography)
+        num_samples = np.zeros(self.num_populations, dtype=int)
+        for sample_set in sample_sets:
+            pop_id = self.demography[sample_set.population].id
+            ploidy = 2 if sample_set.ploidy is None else sample_set.ploidy
```
petrelharp, line 3670: we could default to ploidy=1 here, no?

jeromekelleher:
So setting default ploidy=1 seems confusing to me, since we're computing the rates according to the *diploid* coalescent. This would take some explaining, and also shut down the possibility of computing under the k-ploid coalescent later on.

Do we need this full generality? It seems to me that we could do everything we're doing in the examples with a ``populations`` argument, that takes a list of populations. We then assume a single lineage in each population. If we need to full generality later we can add a ``samples`` keyword argument that's mutually exclusive to ``populations``.

docs/demography.md
jeromekelleher: It's pretty ugly and confusing then though, if we have specify the SampleSet thing every time.

petrelharp:
Good idea.... but the general case would be useful *sometimes*, so I'm not sure?

> So setting default ploidy=1 seems confusing to me, since we're computing the rates according to the diploid coalescent. This would take some explaining, and also shut down the possibility of computing under the k-ploid coalescent later on.

Would it? Ploidy of the model, if it's adjustable, should be an attribute of the demographic model, not the samples, so it should be unaffected by what we do here. I'm imagining saying that dict-style samples are haploid, and if someone wants to do diploid samples then they should use the SampleSet( ) interface.  If we were to allow for changing the ploidy of the coalescent model, we'd add a `ploidy=2` argument which doesn't affect the samples.  Ah - and, how about we change the name from `samples` to `lineages`?

jeromekelleher:
> Good idea.... but the general case would be useful sometimes, so I'm not sure?

Sure - there's nothing stopping us from adding the ``lineages`` argument later, and making it mutually exclusive to the ``populations`` argument. But, for now, we could add the ``populations`` argument which would be simple to do and explain. I don't want to have to get into stuff like "the lineages argument is a bit like the ``samples`` argument for sim_ancestry, except the ploidy is by default one, except that isn't actually the ploidy of the model it's the ploidy of the samples, but you can set the sample ploidy explicitly if you want using the SampleSet form". 

I would much rather write "Draw a single haploid lineage from each of the specified populations", and be happy that this was a non-confusing interface that we could support and build on long term.

petrelharp:
Actualy, I think a `lineages` argument would be easier to explain ("this says how many lineages are drawn from each population"), since a `populations` argument would have to say "... unless there is only one population, in which case we draw two lineages from it".

I'm happy to do this, btw, but can't do it until my afternoon.

jeromekelleher:
Yeah, that's a good point. I'll add a commit with the lineages arg.

jeromekelleher:
I've added on a commit to change the name to "lineages" @petrelharp. If you're OK with how it looks, more or less, I'd suggest merging this and following up with a PR with any changes you think would help. (I'm happy to add support for the numpy array for of the lineages argument back in at some point in addition to the dict form, but I can't be bothered with it right now.)

Apologies for the churn, but I hope it is all a bit less confusing now!

petrelharp (approved):
Ah, much better!

petrelharp:
Looks good. I have no edits to make!

PR #1656, opened 2021-04-23
Initial demes support.

jeromekelleher:
WIP

cc @grahamgower, @apragsdale, @molpopgen

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -954,6 +954,8 @@ class Demography(collections.abc.Mapping):
         """
         if populations is None:
             populations = range(self.num_populations)
+        if len(populations) != self.num_populations:
+            raise ValueError("populations must have Demography.num_populations entries")
         copy_demog = Demography()
         for identifier in populations:
             pop = self[identifier]
@@ -977,11 +979,13 @@ class Demography(collections.abc.Mapping):
             ]
 
         def remap(identifier):
+            if identifier == -1:
+                return -1
             return self[identifier].name
 
         for event in self.events:
             copy_event = copy.deepcopy(event)
-            if isinstance(event, MassMigration):
+            if isinstance(event, (MassMigration, MigrationRateChange)):
                 copy_event.source = remap(event.source)
                 copy_event.dest = remap(event.dest)
             elif isinstance(event, PopulationSplit):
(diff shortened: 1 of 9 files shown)


msprime/demography.py
```diff
@@ -1754,6 +1813,82 @@ def from_old_style(
...
+                initial_size = last_epoch.end_size
+                growth_rate = get_growth_rate(last_epoch)
+            demography.add_population(
+                name=deme.name,
+                description=deme.description,
+                growth_rate=growth_rate,
+                initial_size=initial_size,
+                default_sampling_time=deme.end_time,
+            )
+            for epoch in deme.epochs:
```
grahamgower, line 1846: Shouldn't this be `for epoch in reversed(deme.epochs[:-1]):`? The ZigZag model is a good many-epochs test case.

grahamgower (commented):
Looks like a good start, although I think the epoch iteration ought to be reversed.

msprime/demography.py
jeromekelleher: I don't think it matters - we're sorting the events at the end anyway.

jeromekelleher: Ah, I see what you mean now. Hmm, not sure how the zigzag models works at the moment then.

msprime/demography.py
```diff
@@ -1754,6 +1813,82 @@ def from_old_style(
                 population_map,
             )
 
+    @staticmethod
+    def from_demes(graph):
+        def get_growth_rate(epoch):
+            ret = 0
+            if epoch.size_function not in ["constant", "exponential"]:
+                raise ValueError(
+                    "msprime only supports constant exponentially changing "
```
apragsdale, line 1831: constant _or_ exponentially changing

msprime/demography.py
```diff
@@ -1754,6 +1813,82 @@ def from_old_style(
...
+            )
+            for epoch in deme.epochs:
+                new_initial_size = None
+                if initial_size != epoch.end_size:
+                    new_initial_size = epoch.end_size
+                new_growth_rate = None
+                alpha = get_growth_rate(epoch)
+                if growth_rate != alpha:
+                    new_growth_rate = alpha
+                if new_growth_rate is not None or new_initial_size is not None:
```
apragsdale, line 1863: Do you also need to check that the `epoch.end_time` is not 0, since for those demes that exist at time zero, you've already added that population's initial size and growth rate above? Maybe it's just redundant, and that's fine?

jeromekelleher:
I'm going to have to leave this for a bit now, unfortunately. If anyone was super motivated and picked it up (there's a good pattern for the tests and examples now, I think) I wouldn't be the least bit upset!

jeromekelleher:
Closing in favour of #1662

PR #1678, opened 2021-05-05
Add hypothesis testing strategy for demes graphs.

grahamgower:


grahamgower:
The hypothesis-generated dems graphs are possibly too pathological!

jeromekelleher:
Hmm, maybe not - the first one ( "derived should be active") might be a real problem.

jeromekelleher:
I've pushed a commit here that fixes the problems that were arising from the hypothesis generated graphs. The bad news is that this breaks some existing tests.

The problem hinges around the ``initially_active`` flag - the way that it was being set was conflicting with the internal logic around handling population splits. I think we need to rethink the conversion a little bit.

@apragsdale, can you comment here?

apragsdale:
I had wanted to make populations active only for the time between that deme's start and end time, because we shouldn't be allowed to sample from a population outside of that time interval. I'm not sure exactly what type of examples the hypothesis tests cook up that make it fail, but my suspicion is it's when deme end times are greater than zero, and it's a population that just 'fizzles' instead of admixing, or something like that.

Maybe a solution is to allow the demography API to have an activate/deactivate population function? Something like `demography.activate_population(population=name, time=end_time` and `demography.deactivate_population(population=name, time=start_time)` for cases that aren't internally handled by demographic events like splits and admixtures?

jeromekelleher:
Thanks @apragsdale  - I think it's a good goal that populations shouldn't be ``active`` outside of the Deme start and end_time and ideally add this as a verification test. I'll have another look - hypothesis tests are doing their job I think!

jeromekelleher:
I've identified one issue which I think is easily enough to resolve, but I still seem to be hitting some state machine errors, so I'm not sure how to proceed.

The easy issue is that the hypothesis generator is producing graphs that have an end_time that is not zero. The simplest case is this:
```yaml
        time_units: generations
        demes:
          - name: A
            epochs:
              - start_size: 2000
                end_time: 100
```
This model doesn't make sense in msprime because populations can only become active if they are involved in events. Msprime's demography model is really assuming that time starts at zero, so the simplest thing from my perspective is to just throw an error.  Have we discussed whether the standard should require that time starts (ok, ends) at zero?

More generally, there's no way that we can represent "lone" populations like this that have a intervals not equal to (0, inf). Either we relax the state machine requirements (and just ignore the start/end times) or we'll have to add manual state machine management events (which I'm fairly reluctant to do - it'll be a lot of work).

I guess the simplest way forward for now would be to reliably detect the class of graph that Msprime can't represent and refuse to simulate those, and maybe update the hypothesis generator to just generate graphs that msprime works with. I don't think it'll be much of a limitation in practise, these types of model seem pretty pathological to me.

Thoughts?

grahamgower:
> I guess the simplest way forward for now would be to reliably detect the class of graph that Msprime can't represent and refuse to simulate those, and maybe update the hypothesis generator to just generate graphs that msprime works with. I don't think it'll be much of a limitation in practise, these types of model seem pretty pathological to me.

How about we start by wrapping the demes version of the hypothesis strategy, to exclude cases msprime doesn't care about (using [hypothesis.assume()](https://hypothesis.readthedocs.io/en/latest/details.html?highlight=assume#making-assumptions))? This ought to be a useful way forward in the event we want to keep more pathological cases in the upstream hypothesis strategy (which we can consider separately for each case).

```python
import hypothesis as hyp
import hypothesis.strategies as st

@st.composite
def permissible_graphs(draw):
    """
    Wrapper around the hypothesis strategy for generating demes graphs,
    which excludes pathological cases we don't care about.
    """
    graph = draw(demes_delete_me.graphs())
    graph_end_time = min(deme.end_time for deme in graph.demes)
    hyp.assume(graph_end_time == 0)
    return graph
```

jeromekelleher:
Sure, sounds good @grahamgower. Can we also get rid of any isolated demes that don't exist for all time?

grahamgower:
> Have we discussed whether the standard should require that time starts (ok, ends) at zero?

I don't recall having this discussion, no. I suppose this could be dealt with in the same way as https://github.com/tskit-dev/msprime/issues/983? But maybe we should disallow this in the demes spec anyhow? To me it doesn't seem as crazy as allowing populations sizes between 0 and 1.

> Sure, sounds good @grahamgower. Can we also get rid of any isolated demes that don't exist for all time?

Um, I guess. Maybe msprime wants to assume a *connected* graph?

jeromekelleher:
> I suppose this could be dealt with in the same way as #983? 

It's not so much that we can't simulate it (that's no big deal), it's that the Demography model doesn't support it right now, and, personally, I don't think it's worth spending the effort to make it support it. 

More generally, I think msprime wants a graph that 

a) All root demes have a start_time of infinity (I think we require that already?)
b) All leaf demes have an end_time of zero
c) The start_time of each deme is associated with a population split or admixture event

Does that make sense? I would be great if we can nail down a subset of demes models that map 1-1 to the msprime Demography model and the population state machine.

grahamgower:
I think (a) and (c) are taken care of by the demes data model: all demes either have an ancestor or an infinite start time. I've added a the wrapper I suggested, and modified leaf demes to have an end time of zero. There's a new failure for you to check out now :)

apragsdale:
For (b), do we want all leaf demes to have an end time of zero, or just at least one leaf deme has an end time of zero?

I'd imagine the second, as we could want to model a ghost population that doesn't persist until present time but also doesn't contribute to an admixture event, just continuous migration with that ghost population.

apragsdale:
> I'd imagine the second, as we could want to model a ghost population that doesn't persist until present time but also doesn't contribute to an admixture event, just continuous migration with that ghost population.

Actually, I'm not sure if that's currently allowed in the demography setup - would require being able to manually activate a population at some time? What are your thoughts about allowing that? Or is it adding too much complication..

grahamgower:
Hmm, yeah that makes sense @apragsdale. @jeromekelleher, can you provide a more concrete definition of what you meant by 'leaf' in (b)?

jeromekelleher:
You're right @apragsdale, we do want to have ghost populations, like we have in [this example](https://github.com/tskit-dev/msprime/blob/main/msprime/demography.py#L2262).

So it seems like we'll have to let go of the idea of keeping the state-machine fully consistent with the demes model for now. It can be done easily enough by adding manual ``activate`` and ``deactivate`` demographic events, but I just can't put the time in for at least a month. The state-machine is really only there to help prevent errors though, so I think that's OK.

So, is it sufficient that we set the default sampling time to the ``end_time`` of each Deme? The model will still be correct, it just won't enforce people taking samples from populations that shouldn't exist at a particular time.

This has been a really useful process I think.

grahamgower:
The [preliminary demes->msprime converter](https://github.com/popsim-consortium/demes-python/blob/main/demes/convert/msprime_.py#L13) doesn't use the new split/admixture demographic events, just restricting itself to mass migration events. It uses postive/zero population sizes to indicate active/inactive. This completely bypasses the state machine stuff, so maybe its a simpler way forward?

jeromekelleher:
Maybe - but it makes the debug output much less useful for common cases.

grahamgower:
Well how about this as a compromise? It keeps the split/admixture events, but demes that aren't activated by such an event are never inactive (and thus persist until time=0).
```diff
diff --git a/msprime/demography.py b/msprime/demography.py
index f16dfe29..40db36bd 100644
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -1839,28 +1839,29 @@ class Demography(collections.abc.Mapping):
             return ret
 
         graph = graph.in_generations()
+        events = graph.discrete_demographic_events()
+        initially_active = collections.defaultdict(lambda: True)
+        for split in events["splits"]:
+            initially_active[split.parent] = False
+        for merge in events["mergers"]:
+            for parent in merge.parents:
+                initially_active[parent] = False
 
         demography = Demography()
-        min_end_time = min(deme.end_time for deme in graph.demes)
-        if min_end_time != 0:
-            raise ValueError("Must have at least one Deme with end_time=0")
         for deme in graph.demes:
             initial_size = 0
             growth_rate = 0
             last_epoch = deme.epochs[-1]
-            if last_epoch.end_time == min_end_time:
+            if last_epoch.end_time == 0:
                 initial_size = last_epoch.end_size
                 growth_rate = get_growth_rate(last_epoch)
-                initially_active = True
-            else:
-                initially_active = False
             demography.add_population(
                 name=deme.name,
                 description=deme.description,
                 growth_rate=growth_rate,
                 initial_size=initial_size,
                 default_sampling_time=deme.end_time,
-                initially_active=initially_active,
+                initially_active=initially_active[deme.name],
             )
             for epoch in reversed(deme.epochs):
                 new_initial_size = None
@@ -1880,7 +1881,6 @@ class Demography(collections.abc.Mapping):
                     initial_size = new_initial_size
                     growth_rate = new_growth_rate
 
-        events = dict(graph.discrete_demographic_events())
         for split in events.pop("splits"):
             demography.add_population_split(
                 time=split.time, ancestral=split.parent, derived=split.children

```

And here's a diamond-shaped demographic model, where the youngest deme goes exinct at a time > 0:
```python
# diamond.py
import demes
import msprime

yaml1 = """\
time_units: generations
demes:
- name: A
  epochs:
  - start_size: 2000
    end_time: 333
- name: B
  ancestors: [A]
  epochs:
  - start_size: 2000
    end_time: 222
- name: C
  ancestors: [A]
  epochs:
  - start_size: 2000
    end_time: 222
- name: D
  ancestors: [B, C]
  proportions: [0.5, 0.5]
  start_time: 222
  epochs:
  - start_size: 2000
    end_time: 111
"""

graph = demes.loads(yaml1)
demography = msprime.Demography.from_demes(graph)
ddb = demography.debug()
ddb.print_history()
```
```
$ python diamond.py
DemographyDebugger
╠════════════════════════════════╗
║ Epoch[0]: [0, 111) generations ║
╠════════════════════════════════╝
╟    Populations (total=4 active=1)
║    ┌────────────────────────────────┐
║    │   │  start│   end│growth_rate  │
║    ├────────────────────────────────┤
║    │  D│    0.0│   0.0│ 0           │
║    └────────────────────────────────┘
╟    Events @ generation 111
║    ┌───────────────────────────────────────────────────────────────────────────────┐
║    │  time│type        │parameters         │effect                                 │
║    ├───────────────────────────────────────────────────────────────────────────────┤
║    │   111│Population  │population=D,      │initial_size → 2e+03 for population D  │
║    │      │parameter   │initial_size=2000  │                                       │
║    │      │change      │                   │                                       │
║    └───────────────────────────────────────────────────────────────────────────────┘
╠══════════════════════════════════╗
║ Epoch[1]: [111, 222) generations ║
╠══════════════════════════════════╝
╟    Populations (total=4 active=1)
║    ┌─────────────────────────────────────┐
║    │   │    start│      end│growth_rate  │
║    ├─────────────────────────────────────┤
║    │  D│   2000.0│   2000.0│ 0           │
║    └─────────────────────────────────────┘
╟    Events @ generation 222
║    ┌─────────────────────────────────────────────────────────────────────────────────┐
║    │  time│type        │parameters          │effect                                  │
║    ├─────────────────────────────────────────────────────────────────────────────────┤
║    │   222│Population  │population=B,       │initial_size → 2e+03 for population B   │
║    │      │parameter   │initial_size=2000   │                                        │
║    │      │change      │                    │                                        │
║    │┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈│
║    │   222│Population  │population=C,       │initial_size → 2e+03 for population C   │
║    │      │parameter   │initial_size=2000   │                                        │
║    │      │change      │                    │                                        │
║    │┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈│
║    │   222│Admixture   │derived=D           │Moves all lineages from admixed         │
║    │      │            │ancestral=[B, C]    │population 'D' to ancestral             │
║    │      │            │proportions=[0.50,  │populations. Lineages move to 'B' with  │
║    │      │            │0.50]               │proba 0.5; 'C' with proba 0.5. Set 'D'  │
║    │      │            │                    │to inactive, and all migration rates    │
║    │      │            │                    │to and from 'D' to zero.                │
║    └─────────────────────────────────────────────────────────────────────────────────┘
╠══════════════════════════════════╗
║ Epoch[2]: [222, 333) generations ║
╠══════════════════════════════════╝
╟    Populations (total=4 active=2)
║    ┌─────────────────────────────────────────────┐
║    │   │    start│      end│growth_rate  │ B │ C │
║    ├─────────────────────────────────────────────┤
║    │  B│   2000.0│   2000.0│ 0           │ 0 │ 0 │
║    │  C│   2000.0│   2000.0│ 0           │ 0 │ 0 │
║    └─────────────────────────────────────────────┘
╟    Events @ generation 333
║    ┌────────────────────────────────────────────────────────────────────────────────┐
║    │  time│type        │parameters         │effect                                  │
║    ├────────────────────────────────────────────────────────────────────────────────┤
║    │   333│Population  │population=A,      │initial_size → 2e+03 for population A   │
║    │      │parameter   │initial_size=2000  │                                        │
║    │      │change      │                   │                                        │
║    │┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈│
║    │   333│Population  │derived=[C, B],    │Moves all lineages from derived         │
║    │      │Split       │ancestral=A        │populations 'C' and 'B' to the          │
║    │      │            │                   │ancestral 'A' population. Also set the  │
║    │      │            │                   │derived populations to inactive, and    │
║    │      │            │                   │all migration rates to and from the     │
║    │      │            │                   │derived populations to zero.            │
║    └────────────────────────────────────────────────────────────────────────────────┘
╠══════════════════════════════════╗
║ Epoch[3]: [333, inf) generations ║
╠══════════════════════════════════╝
╟    Populations (total=4 active=1)
║    ┌─────────────────────────────────────┐
║    │   │    start│      end│growth_rate  │
║    ├─────────────────────────────────────┤
║    │  A│   2000.0│   2000.0│ 0           │
║    └─────────────────────────────────────┘
```

apragsdale:
> So it seems like we'll have to let go of the idea of keeping the state-machine fully consistent with the demes model for now. It can be done easily enough by adding manual `activate` and `deactivate` demographic events, but I just can't put the time in for at least a month. The state-machine is really only there to help prevent errors though, so I think that's OK.

To me this also seems like the most natural and most satisfying solution, but I understand the lack of time to devote to it. I can have a go at adding that in a separate PR, perhaps, which would make this all fairly straightforward.

> So, is it sufficient that we set the default sampling time to the `end_time` of each Deme? The model will still be correct, it just won't enforce people taking samples from populations that shouldn't exist at a particular time.

And I think that's fine for a temporary solution as well.



> Well how about this as a compromise? It keeps the split/admixture events, but demes that aren't activated by such an event are never inactive (and thus persist until time=0).

This looks really reasonable to me.

jeromekelleher:
Your solution looks good to me @grahamgower - can you push the changes here please?

I think the manual activate/deactivate events combined with MassMigrations will be the way to go for the next version, but for now let's get something that simulates the correct model and mostly matches the state-machine semantics we'd like.

grahamgower:
Ok, I've just pushed this @jeromekelleher. There's still failures for randomly generated graphs. I presume that events with very small event times are being rounded somewhere, so we end up with events at time 0.

jeromekelleher:
This should fix the problems @grahamgower. I've added a test to check that the sampling time is the right value but hit some annoying floating point problems associated with ``in_generations`` (which only happen with the ridiculously large 10^300 end_time range). Feel free to put in a better solution.

grahamgower:
Looks good to me. I think we should fix that in `demes`, before the graph-generating strategy is exported. We'll need to strike a balance between generating graphs that test obscure corner cases and graphs that have technically allowable values but just don't make sense.

grahamgower:
Feel free to squash to something you're happy with here. It turned out to be a bit of a mess...

jeromekelleher:
I'll tag a release when I get a chance.

PR #1683, opened 2021-05-12
clarified mutation map length error

### LINKED ISSUE #1682: Error with MutationMap
Hi,

I am getting this error when using sim_mutations with a RateMap instead of a constant rate:
`
msprime._msprime.InputError: Input error in mutation rate map: Mutation map is not compatible with specified tables.
`
I tried to see where the error msg came from but could not figure it out. Any hint to what could cause this error? My recombination map is a priori ok (and has no missing data). The input tree sequence was generated by SLiM, where only deleterious mutations in coding regions were simulated, with a constant rate. I am now trying to add neutral mutations with a variable rate (piecewise constant rate) along the genome.

Many thanks in advance for your insights,

Julien.

petrelharp:
I think this will close #1682, although @jydu should confirm.

### DIFF
--- a/lib/mutgen.c
+++ b/lib/mutgen.c
@@ -811,7 +811,7 @@ mutgen_set_rate_map(mutgen_t *self, size_t size, double *position, double *rate)
         goto out;
     }
     if (rate_map_get_sequence_length(&self->rate_map) != self->tables->sequence_length) {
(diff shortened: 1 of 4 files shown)


jydu:
Thanks @petrelharp ! That is indeed clearer.

jeromekelleher (approved):
LGTM

PR #1708, opened 2021-06-04
Implement population activation

apragsdale:
Work in progress.. this is my first attempt at trying to dive into the guts of the C side of things to implement more flexible population activation and deactivation. I've tried to get the simulator to recognize those types of events, but am a bit stuck at the point of actually having the simulator apply the population activation. My hope was to be able to reverse engineer how to implement it based on other demographic event types, but I worry I've just ended up just getting lost and messing things up, so it's possible this should just be tossed and started over.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -5875,17 +5875,12 @@ msp_activate_population(msp_t *self, int population_id)
     int ret = 0;
     population_t *pop = &self->populations[population_id];
 
-    /* It's not currently possible to do this because we're only calling
-     * msp_activate_population from population split, where we only activate
-     * if it's not already active. But, leaving this here as a reminder
-     * in case we ever do call activate from somewhere else.
-     */
-    /* if (pop->state == MSP_POP_STATE_ACTIVE) { */
-    /*     ret = MSP_ERR_POPULATION_CURRENTLY_ACTIVE; */
-    /*     goto out; */
-    /* } */
     if (pop->state == MSP_POP_STATE_PREVIOUSLY_ACTIVE) {
         ret = MSP_ERR_POPULATION_PREVIOUSLY_ACTIVE;
+        goto out;
+    }
+    if (pop->state == MSP_POP_STATE_ACTIVE) {
+        ret = MSP_ERR_POPULATION_CURRENTLY_ACTIVE;
         goto out;
     }
     pop->state = MSP_POP_STATE_ACTIVE;
@@ -6482,6 +6477,46 @@ msp_add_census_event(msp_t *self, double time)
     de->print_state = msp_print_census_event;
     ret = 0;
 
+out:
+    return ret;
+}
+
+/* Add an activate_population_event at a given time */
+
+static int
+msp_activate_population_event(msp_t *self, demographic_event_t *event)
+{
+    population_id_t population = event->params.activate_population.population;
+
+    return msp_activate_population(self, population);
+}
+
+static void
+msp_print_activate_population_event(
+    msp_t *MSP_UNUSED(self), demographic_event_t *event, FILE *out)
+{
+    fprintf(out, "%f\tactivate_population_event:\n", event->time);
+}
+
+int MSP_WARN_UNUSED
+msp_add_activate_population_event(msp_t *self, double time, int population_id)
+{
+    int ret = 0;
+    demographic_event_t *de;
+    int N = (int) self->num_populations;
+
+    if (population_id < 0 || population_id >= N) {
+        ret = MSP_ERR_POPULATION_OUT_OF_BOUNDS;
+        goto out;
+    }
+    ret = msp_add_demographic_event(self, time, &de);
+    if (ret != 0) {
+        goto out;
+    }
+    de->params.activate_population.population = population_id;
+    de->change_state = msp_activate_population_event;
+    de->print_state = msp_print_activate_population_event;
+    ret = 0;
 out:
     return ret;
 }
--- a/lib/tests/test_ancestry.c
+++ b/lib/tests/test_ancestry.c
@@ -2532,6 +2532,96 @@ test_census_event(void)
 }
 
 static void
+test_activate_population_event(void)
+{
+    int ret;
+    msp_t msp;
+    gsl_rng *rng = safe_rng_alloc();
+    sample_t samples[] = {
+        { .time = 2, .population = 0 },
+        { .time = 2, .population = 0 },
+        { .time = 0, .population = 1 },
+        { .time = 0, .population = 1 },
+    };
+    tsk_table_collection_t tables;
+
+    ret = build_sim(&msp, &tables, rng, 1, 3, samples, 4);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    ret = msp_set_population_configuration(&msp, 0, 1, 0, false);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    ret = msp_add_activate_population_event(&msp, 1, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    int derived[] = { 0, 1 };
+    ret = msp_add_population_split(&msp, 4, 2, derived, 2);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    ret = msp_initialise(&msp);
+    CU_ASSERT_EQUAL(ret, 0);
+    CU_ASSERT_EQUAL(msp.populations[0].state, MSP_POP_STATE_INACTIVE);
+    CU_ASSERT_EQUAL(msp.populations[1].state, MSP_POP_STATE_ACTIVE);
+
+    ret = msp_run(&msp, DBL_MAX, UINT32_MAX);
+    CU_ASSERT_EQUAL(ret, 0);
+    msp_verify(&msp, 0);
+    msp_print_state(&msp, _devnull);
+
+    CU_ASSERT_EQUAL(msp.populations[0].state, MSP_POP_STATE_PREVIOUSLY_ACTIVE);
+    CU_ASSERT_EQUAL(msp.populations[1].state, MSP_POP_STATE_PREVIOUSLY_ACTIVE);
+    CU_ASSERT_EQUAL(msp.populations[2].state, MSP_POP_STATE_ACTIVE);
+
+    ret = msp_free(&msp);
+    CU_ASSERT_EQUAL(ret, 0);
+    gsl_rng_free(rng);
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_activate_population_event_errors(void)
+{
+    int ret;
+    uint32_t n = 2;
+    msp_t msp;
+    gsl_rng *rng = safe_rng_alloc();
+    tsk_table_collection_t tables;
+
+    /* Activate already active population. */
+    ret = build_sim(&msp, &tables, rng, 1, 1, NULL, n);
+    CU_ASSERT_EQUAL(ret, 0);
+    ret = msp_add_activate_population_event(&msp, 1, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    ret = msp_initialise(&msp);
+    CU_ASSERT_EQUAL(ret, 0);
+    ret = msp_run(&msp, DBL_MAX, ULONG_MAX);
+    CU_ASSERT_EQUAL(ret, MSP_ERR_POPULATION_CURRENTLY_ACTIVE);
+
+    msp_free(&msp);
+    tsk_table_collection_free(&tables);
+
+    /* Activate nonexistent populations. */
+    ret = build_sim(&msp, &tables, rng, 1, 1, NULL, n);
+    CU_ASSERT_EQUAL(ret, 0);
+    ret = msp_add_activate_population_event(&msp, 1, -1);
+    CU_ASSERT_EQUAL(ret, MSP_ERR_POPULATION_OUT_OF_BOUNDS);
+
+    msp_free(&msp);
+    tsk_table_collection_free(&tables);
+
+    ret = build_sim(&msp, &tables, rng, 1, 1, NULL, n);
+    CU_ASSERT_EQUAL(ret, 0);
+    ret = msp_add_activate_population_event(&msp, 1, 1);
+    CU_ASSERT_EQUAL(ret, MSP_ERR_POPULATION_OUT_OF_BOUNDS);
+
+    msp_free(&msp);
+    tsk_table_collection_free(&tables);
+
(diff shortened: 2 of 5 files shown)


msprime/_msprimemodule.c
```diff
@@ -1242,6 +1242,29 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
...
+    PyObject *value;
+    int err, population_id;
+
+    value = get_dict_value(py_event, "population_id");
+    if (value == NULL) {
+        goto out;
+    }
+    population_id = (int) PyLong_AsLong(value);
+
+    err = msp_activate_population(self->sim, time, population_id);
```
grahamgower, line 1258: I think this should be `msp_add_activate_population()`

lib/msprime.c
```diff
@@ -5893,6 +5893,24 @@ msp_activate_population(msp_t *self, int population_id)
     return ret;
 }
 
+/* Adds a population activation event.
+ * Note: we use the int32_t type here so we can be sure we're passing in
+ * arrays of the correct type from numpy where we have to specify the size.
+ */
+int MSP_WARN_UNUSED
+msp_add_activate_population(
```
grahamgower, line 5901: This function needs to create a demographic event for the population activation (e.g. see the `msp_add_population_split()` function). The demographic event's `change_state` field should be a pointer to the function that does the activation (you'll need to write this function, which I guess will be a thin wrapper around `msp_activate_population`). The "activate population" demographic event's `change_state()` function will then be triggered at the appropriate time from the simulation's main loop (see `msp_run_coalescent()`, which calls `msp_apply_fixed_events()`, etc.). You'll need to add a struct in msprime.h for the demographic event (eg. see the one for the struct typedef'ed to `population_split_t`).

grahamgower (commented):
It's looking good @apragsdale. I think you should read through `msp_run_coalescent()` and look at how the demographic events get triggered via the `msp_get_next_fixed_event_time()` and `msp_apply_fixed_events()` functions. You'll need to add a new data structure to msprime.h for the population activate demographic event, and then your `msp_add_activate_population()` function should add the demographic event to the events list like the other `msp_add_xxx()` functions.

lib/msprime.c
jeromekelleher, line 5901: I'd change the name of this to ``msp_add_activate_population_event`` just for clarity.

msprime/_msprimemodule.c
```diff
@@ -1445,7 +1468,8 @@ Simulator_parse_demographic_events(Simulator *self, PyObject *py_events)
     int err, population_id, source, destination;
     int is_population_parameter_change, is_migration_rate_change, is_mass_migration,
         is_population_split, is_admixture, is_symmetric_migration_rate_change,
-        is_simple_bottleneck, is_instantaneous_bottleneck, is_census_event;
+        is_simple_bottleneck, is_instantaneous_bottleneck, is_census_event,
+        is_population_activation;
```
jeromekelleher, line 1472: Let's standardise on ``activate_population_event``, and use this systematically. So, ``is_activate_population_event`` here.

msprime/demography.py
```diff
@@ -705,6 +705,42 @@ def add_census(self, time: float) -> CensusEvent:
         """
         return self.add_event(CensusEvent(time))
 
+    def activate_population(self, time: float, population: Union[int, str]):
```
jeromekelleher, line 708: We're not going to be adding this event to the public API, so I don't think we need the method here. We can just use the classes manually  within the code (or, make this a private method)

jeromekelleher (commented):
Looks good, thanks @apragsdale. Good advice also from @grahamgower, thanks for the review.

I would start this at the opposite end - start at the bottom C level and only fill in the high-level Python stuff when that's all working. I would use the "census" event as the template. Start at ``msp_add_census_event`` and  work your way up from there. Add a test case like ``test_census_event`` and use that to exercise the code. Don't even look at the Python layer until the C stuff is working.

Thanks!

apragsdale:
Thanks both @grahamgower and @jeromekelleher for the reviews and suggestions. I might end up starting from scratch here, but I'll give it another shot and hopefully can make progress!

jeromekelleher:
No need to start again @apragsdale, what you have is good. Just focus on the C side first, the rest of the plumbing is routine then.

apragsdale:
I'm pretty lost here.. taking Jerome's advice, and running `ninja -C build test` in the lib directory, I can't get it to compile, and I think I'm just fundamentally misunderstanding something with how to set up and use the activate population data structure.

sorry for the hassle - this is all very new to me

jeromekelleher:
Looks OK to me so far. I think you're probably hitting up against the default semantics and maybe need to set up a more complex test case, in which some populations are initially inactive.

Would you like me to sort out the low-level stuff here?

apragsdale:
The error I'm getting, when running `ninja -C build test` in the `lib` directory, is and incompatible pointer types in the `msp_add_activate_population_event` function, when trying `de->change_state = msp_activate_population;`. I'm having a hard time tracking down figuring out the exact issue there. @jeromekelleher If you're able to have a look at that, that could help get me back on track. Thanks!

benjeffery:
Haven't got my head deep into this, but it seems currently you're assigning `msp_activate_population` to `de->change_state` on `msprime.c:5918`.

That function has signature `static int msp_activate_population(msp_t *, int)` but it needs to be `(msp_t *, struct demographic_event_t_t *)`  so I _think_ you need to assign a function that extracts the pop id out of the demographic event and then calls `msp_activate_population`.  

I'll tag @jeromekelleher here in case I'm way off!

jeromekelleher:
I've added in a commit that'll fix up the plumbing @apragsdale. Still plenty to do here, but at least it'll compile.

apragsdale:
Thanks a lot Ben and Jerome. That makes a lot of sense in hindsight, as well as the changes. I'll carry on working on this. Thanks again!

apragsdale:
I've got another simple (I think) question, about why the test I've written is failing. Essentially, I've set up two populations, one activated at time zero, another activated at time in the past, which then further in the past are joined in a population split event. I suspect that what I'm doing (`msp.populations[0].state = MSP_POP_STATE_INACTIVE;` on line 2732 in `test_ancestry.c`) is breaking the flow of inactive -> active -> previously active, but I'm not entirely sure. Any ideas if that is the issue, or if it's something else? Thank you!

grahamgower:
I think your population is getting activated during the test's call to `msp_initialise()`. Try calling `msp_set_population_configuration(&msp, 0, 1, 0, false);` instead of setting the population state manually.
```
diff --git a/lib/tests/test_ancestry.c b/lib/tests/test_ancestry.c
index 5dd0dca1..ad37118c 100644
--- a/lib/tests/test_ancestry.c
+++ b/lib/tests/test_ancestry.c
@@ -2729,8 +2729,8 @@ test_population_activation_event(void)
 
     ret = build_sim(&msp, &tables, rng, 1, 3, samples, 4);
     CU_ASSERT_EQUAL(ret, 0);
-    msp.populations[0].state = MSP_POP_STATE_INACTIVE;
-    CU_ASSERT_EQUAL(msp.populations[0].state, MSP_POP_STATE_INACTIVE);
+    ret = msp_set_population_configuration(&msp, 0, 1, 0, false);
+    CU_ASSERT_EQUAL(ret, 0);
 
     ret = msp_add_activate_population_event(&msp, 1, 0);
     CU_ASSERT_EQUAL(ret, 0);
@@ -2741,13 +2741,15 @@ test_population_activation_event(void)
 
     ret = msp_initialise(&msp);
     CU_ASSERT_EQUAL(ret, 0);
+    CU_ASSERT_EQUAL(msp.populations[0].state, MSP_POP_STATE_INACTIVE);
 
     ret = msp_run(&msp, DBL_MAX, UINT32_MAX);
     CU_ASSERT_EQUAL(ret, 0);
     msp_verify(&msp, 0);
     msp_print_state(&msp, _devnull);
 
-    CU_ASSERT_EQUAL(msp.populations[0].state, MSP_POP_STATE_ACTIVE);
+    CU_ASSERT_EQUAL(msp.populations[0].state, MSP_POP_STATE_PREVIOUSLY_ACTIVE);
 
     ret = msp_free(&msp);
     CU_ASSERT_EQUAL(ret, 0);

```

grahamgower:
Also, it might be worth running the simulation until t=1.5 (`msp_run(&msp, 1.5, UNIT32_MAX)`), stopping the simulation after the activation but before the split event, to check that your population is ACTIVE.

apragsdale:
I think I've made some progress here -- in the `from_demes()` method, we need to activate populations when they are "ghosts" (meaning, they have end time > 0 and they peter out instead of being directly involved in a split or admixture). For such populations, I check that they have end time > 0 and all their successors' start times precede the population's end time. In that case, we need to add an activate population event. I think more tests would be very helpful here, but wanted to see what others think of this so far.

jeromekelleher:
Sounds good @apragsdale - I wonder if it would be better to get the low-level stuff wrapped up first, though, and then tackle the from_demes stuff? This will be a very big if we're going to do it all at once.

apragsdale:
That sounds fine to me @jeromekelleher  - I can separate the C-level implementation from the demography/from_demes methods into two PRs if that makes it more manageable. What more do you reckon needs attention on the low-level side of things?

lib/util.c
```diff
@@ -281,6 +281,9 @@ msp_strerror_internal(int err)
         case MSP_ERR_POPULATION_PREVIOUSLY_ACTIVE:
             ret = "Attempt to set a previously active population to active";
             break;
+        case MSP_ERR_POPULATION_STATE_ACTIVE:
```
jeromekelleher, line 284: CURRENTLY_ACTIVE would be follow the existing patter a bit better.

lib/msprime.c
```diff
@@ -5875,11 +5875,7 @@ msp_activate_population(msp_t *self, int population_id)
     int ret = 0;
     population_t *pop = &self->populations[population_id];
 
-    /* It's not currently possible to do this because we're only calling
-     * msp_activate_population from population split, where we only activate
-     * if it's not already active. But, leaving this here as a reminder
-     * in case we ever do call activate from somewhere else.
-     */
+    /* TODO implement this now that we have the activate_population event */
```
jeromekelleher, line 5878: Delete this commented out block (you've implemented it below, right?)

lib/msprime.c
```diff
@@ -5888,6 +5884,10 @@ msp_activate_population(msp_t *self, int population_id)
         ret = MSP_ERR_POPULATION_PREVIOUSLY_ACTIVE;
         goto out;
     }
+    if (pop->state == MSP_POP_STATE_ACTIVE) {
```
jeromekelleher, line 5887: This isn't being covered by tests. Need to try to activate an already active population.

lib/msprime.c
```diff
@@ -6486,6 +6486,46 @@ msp_add_census_event(msp_t *self, double time)
...
+
+int MSP_WARN_UNUSED
+msp_add_activate_population_event(msp_t *self, double time, int population_id)
+{
+    int ret = 0;
+    demographic_event_t *de;
+    int N = (int) self->num_populations;
+
+    if (population_id < 0 || population_id >= N) {
+        ret = MSP_ERR_POPULATION_OUT_OF_BOUNDS;
```
jeromekelleher, line 6514: This isn't being covered by tests.

lib/tests/test_ancestry.c
```diff
@@ -2713,6 +2713,54 @@ test_census_event(void)
     tsk_table_collection_free(&tables);
 }
 
+static void
+test_population_activation_event(void)
```
jeromekelleher, line 2717: Rename to ``test_activate_population_event`` (this will help when you're search-replacing to implement deactivate population). Also, we want another test case called ``test_activate_population_event_errors`` which provokes the error cases identified above.

msprime/_msprimemodule.c
```diff
@@ -1507,6 +1532,10 @@ Simulator_parse_demographic_events(Simulator *self, PyObject *py_events)
     if (growth_rate_s == NULL) {
         goto out;
     }
+    activate_population_event_s = Py_BuildValue("s", "activate_population_event");
```
jeromekelleher, line 1535: Need to XDECREF this value in the out block, otherwise we have a memory leak.

msprime/_msprimemodule.c
```diff
@@ -1677,6 +1715,13 @@ Simulator_parse_demographic_events(Simulator *self, PyObject *py_events)
                     strength);
         } else if (is_census_event) {
             err = msp_add_census_event(self->sim, time);
+        } else if (is_activate_population_event) {
```
jeromekelleher, line 1718: This is a duplicate of the above, isn't it?

msprime/_msprimemodule.c
```diff
@@ -1242,6 +1242,29 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
     return ret;
 }
 
+static int
+Simulator_parse_activate_population_event(Simulator *self, double time, PyObject *py_event)
+{
+    int ret = -1;
```
jeromekelleher, line 1248: Need to get some tests into the test_lowlevel file that'll exercise these code paths. See how the other events are done.

jeromekelleher (commented):
There's a few things need to be done to close up the C side of things @apragsdale. I suggest you  take a copy of the branch, and then drop all the changes to the high-level Python. That way we can see exactly what the low-level changes are and be sure we're not missing any important test coverage.

Probably the best thing to do then is to follow up with a similar PR to do the deactivate event, which should be pretty routine.

Finally, we can then add in the high-level Python stuff to actually use the events.

apragsdale:
I'm not really sure what's going on with the lint test - running `pre-commit run --all-files` locally shows that the clang-format check passes. Although I notice running `clang-format -i lib/tests/* lib/!(avl).{c,h}` locally did change some code that was outside what I added or modified with this PR.

Otherwise, addressed reviews here, including dropping the changes to the Python side of things, and adding some tests for activation errors.

apragsdale:
cc @benjeffery - the only test failing is the lint test, it looks like.

jeromekelleher:
@apragsdale - can't you try ``pip install clang-format=6.0.1`` please and try to linting again? (**SUPER** annoying that clang-format doesn't do cross version compatibility!!!)

benjeffery:
I've pushed a commit to fix the lint. Note that this PR needed a squash anyway.

apragsdale:
Ok, rebased and squashed!

jeromekelleher (approved):
Looks great,thanks @apragsdale!

PR #1716, opened 2021-06-08
Initial proposal for demes CLI

jeromekelleher:
Here's a proposal for simulating demographic models from the command line, allowing us to do stuff like
```
$ msp ancestry YRI:10 CHB:10 -d ooa.yaml -o tmp.trees
```
ie., sample 10 individuals from the YRI population and 10 from the CHB population, directly corresponding to the mapping form of the ``samples`` argument (and kind-of sharing notation, by using ":" as the separator).

It pains me to have substructure that we have to parse out of the command line args, but I think this form is about as good as it's going to get. 

Thoughts? @apragsdale @grahamgower @molpopgen @petrelharp

(If we like this CLI notation, then I'd suggest we update stdpopsim to use the same approach)

### DIFF
--- a/msprime/cli.py
+++ b/msprime/cli.py
@@ -21,12 +21,16 @@ Command line interfaces to the msprime library.
 """
 import argparse
 import hashlib
+import logging
 import os
 import random
 import signal
 import sys
 import warnings
+from typing import Dict
+from typing import List
 
+import demes
 import tskit
 
 import msprime
@@ -34,10 +38,24 @@ from . import ancestry
 from . import mutations
 
 
+logger = logging.getLogger(__name__)
+
+
 def set_sigpipe_handler():
     if os.name == "posix":
         # Set signal handler for SIGPIPE to quietly kill the program.
         signal.signal(signal.SIGPIPE, signal.SIG_DFL)
+
+
+def setup_logging(args):
+    log_level = "WARN"
+    if args.verbosity > 0:
+        log_level = "INFO"
+    if args.verbosity > 1:
+        log_level = "DEBUG"
+    logging.basicConfig(
+        level=log_level, format="%(asctime)s %(name)s %(levelname)s %(message)s"
+    )
 
 
 #######################################################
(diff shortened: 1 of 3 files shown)


jeromekelleher:
I think using ":" as the separator is pretty kosher from a shell perspective. Although colon is a [special character](https://tldp.org/LDP/abs/html/special-chars.html) in bash, it's can be used, e.g., as part of a filename without issues and indeed is used as a separate in things like PATH.

grahamgower:
Yes! That will make simulation very easy, particularly for folks without Python experience. I assume you considered `=` as a separator? I don't have a strong opinion, but that's the first thing I'd have thought of doing.
```
$ msp ancestry YRI=10 CHB=10 -d ooa.yaml -o tmp.trees
```

I assume also that you don't want to support a sampling time? I think it's worth avoiding a specific CLI syntax for this and just telling people to use the Python API.

jeromekelleher:
> I assume also that you don't want to support a sampling time? I think it's worth avoiding a specific CLI syntax for this and just telling people to use the Python API.

Yes, this is exactly what I was thinking. If you want to do something more complicated than using the default sampling time for the deme (== its start time time, backwards in time) then you should use the Python API.

jeromekelleher:
> I assume you considered = as a separator? I don't have a strong opinion, but that's the first thing I'd have thought of doing.

No, it didn't occur to me actually. I'm a bit wary of it though, since ``=`` is the assignment operator in bash. So, we can do stuff like 
```
$ a=b  #works fine, assigns b to variable a
```
whereas 
```
$ a:b

Command 'a:b' not found, did you mean:
```
bash doesn't tokenize around ":" even in this case.

I worry that there could be some unintended consequences when scripting, so that one would have to put things in single quotes if we used the ``a=b`` form, rather than ``a:b``? Shell programming is very tricky, I'd rather not take chances here.

molpopgen:
I like this idea.  Any idea how `:` vs `=` interact with other CLI tools that one may use?  I'm thinking `xargs` and `parallel` here.

petrelharp:
This is a *great* idea. (I was recommending someone use msprime the other day, only to find out that they don't know python, so was using fastsimcoal's command-line interface.)

apragsdale:
This would be really awesome! I like it, and don't see any sticking points with the CLI format. And I don't think #1708 needs to be finished to get the default sampling to be at the start time of the deme (backwards in time).

msprime/cli.py
```diff
@@ -1027,8 +1028,26 @@ def run_mutations(args):
...
+        graph = demes.load(args.demography)
+        demography = msprime.Demography.from_demes(graph)
+        for pop_spec in args.samples:
+            # We only accept ID:num_samples values
+            tokens = pop_spec.split(":")
+            if len(tokens) != 2:
+                raise ValueError(
+                    "sample specifications must be in form population_id:num_samples"
+                )
+            samples[tokens[0]] = int(tokens[1])
```
grahamgower, line 1042: Maybe `int(float(tokens[1]))` here so that things like 1e5 can be parsed?

jeromekelleher:
Thanks for the input all, ready for review.

jeromekelleher:
Ack, that's not the commit I wanted to push! Dammit.

jeromekelleher:
OK, that's better. Thank goodness for git reflog!

grahamgower (approved):
LGTM!

PR #1723, opened 2021-06-11
Add logo

hyanwong:
I wasn't sure where to put it, so I simply dumped it at the top level! I think it's better to call it `msprime_logo.svg` rather than something generic like `logo.svg` as then it's easier to find if you download a copy.


jeromekelleher (approved):
LGTM!

This is the standard die format, right?

hyanwong:
> This is the standard die format, right?

Yeah, AFAIK the only constraint is the the opposite sides need to add up to 7.

PR #1724, opened 2021-06-11
Add Demography.to_demes().

### LINKED ISSUE #1721: add `Demography.to_demes()`
See #1710

grahamgower:
Closes #1721.

---

WIP, but seems to be behaving itself. At least, conversion of the `Demography._*_model()` models look correct when plotting them with demesdraw.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -30,6 +30,7 @@ import itertools
 import logging
 import math
 import numbers
+import operator
 import re
 import sys
 import textwrap
@@ -38,8 +39,11 @@ from typing import Any
 from typing import ClassVar
 from typing import Dict
 from typing import List
+from typing import MutableMapping
+from typing import Tuple
 from typing import Union
 
+import demes
 import numpy as np
 import tskit
 
@@ -1826,7 +1830,23 @@ class Demography(collections.abc.Mapping):
             )
 
     @staticmethod
-    def from_demes(graph):
+    def from_demes(graph: demes.Graph) -> Demography:
+        """
+        Creates a :class:`.Demography` object from the specified
+        :ref:`demes graph <demes:sec_introduction>`.
+
+        .. code::
+
+            import demes
+
+            graph = demes.load("model.yaml")
+            demography = msprime.Demography.from_demes(graph)
+
+        :param demes.Graph graph: A demes graph.
+        :return: A :class:`.Demography` instance corresponding to the demes model.
+        :rtype: .Demography
+        """
+
         def get_growth_rate(epoch):
             ret = 0
             if epoch.size_function not in ["constant", "exponential"]:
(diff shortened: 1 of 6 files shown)


msprime/demography.py
```diff
@@ -2431,6 +2459,301 @@ def _american_admixture_model():
...
+        """
+        resolved = self.validate()
+        b = demes.Builder()
+        for pop in resolved.populations:
+            initial_epoch = dict(
+                end_size=pop.initial_size, growth_rate=pop.growth_rate, end_time=0
+            )
+            b.add_deme(
+                pop.name,
+                description=pop.description if pop.description else None,
```
jeromekelleher, line 2501: Not sure the ``if`` is doing much here? Might as well pass through the empty string value

(I'm not sure why I made the default the empty string rather than None, maybe I did this when demes required a ``description`` field? I think it would probably be better to have the description = None if not specified in Demography.add_population()``

msprime/demography.py
```diff
@@ -2431,6 +2459,301 @@ def _american_admixture_model():
...
+                start_time=math.inf,
+                epochs=[initial_epoch],
+            )
+
+        deme_map = dict()
+        for pop in resolved.populations:
+            deme_map[pop.name] = b.data["demes"][pop.id]
+            deme_map[pop.id] = b.data["demes"][pop.id]
```
jeromekelleher, line 2509: Why store both? Probably clearer to do something like ``resolved[population].name`` to be certain that we're converting to names.

msprime/demography.py
```diff
@@ -2431,6 +2459,301 @@ def _american_admixture_model():
...
+            """
+            migration_matrix = mm_list[0]
+            if time > mm_end_times[0]:
+                # We need a new migration matrix.
+                migration_matrix = copy.deepcopy(migration_matrix)
+                mm_list.insert(0, migration_matrix)
+                mm_end_times.insert(0, time)
+            return migration_matrix
+
+        for time, events_group in itertools.groupby(
```
jeromekelleher, line 2560: There's a lot of logic in this loop... I've found it a lot easier in things like ``assert_equivalent`` to let the debugger do some of the heavy lifting. It saves a lot of reasoning about the effects of events, because you can just detect the change in state of the model. I think this makes things like changing migration rates a lot easier, and definitely detecting the start/end times of Demes (you can just look at the state of the Population in each epoch).

Any thoughts?

jeromekelleher (commented):
Looks good - is it worth letting the DemographyDebugger do some of the heavy lifting though? It's a good way of avoiding some of the multiple implementations of the event logic.

msprime/demography.py
grahamgower: Demes doesn't allow empty strings (but allows None).

msprime/demography.py
grahamgower: Ah, yes. The migration matrices from the debugger should be a drop in replacement. I'll take a look.

grahamgower:
I updated to use the demography debugger for the migration matrices, and to get the deme start/end times with respect to the population's ACTIVE state. It's also desirable to check start/end times of demes based on mass migrations with proportion 1, to support legacy models, so I've left that in. Also, the big loop still iterates the events rather than the demography debugger epochs --- there's some quirk with the granularity of times returned by the demography debugger epochs such that the round tripping gives problems. I don't think it's worth the effort to try and figure that out just to shuffle a couple of lines of code.

grahamgower:
Oh, and I updated the docstring to mention the bottleneck events can't be converted, and that legacy models could be misspecified and in general may not be convertible.

msprime/demography.py
```diff
@@ -2431,6 +2459,275 @@ def _american_admixture_model():
...
+
+            Demographic models using bottlenecks added via the
+            :meth:`add_simple_bottleneck` or :meth:`add_instantaneous_bottleneck`
+            methods are not able to be converted into a demes graph.
+
+        .. note::
+
+            Demes is stricter than msprime with regard to how a demographic
+            model is structured, so models that can be simulated with msprime
+            are not gauranteed to be convertable to a demes graph. In particular,
```
jeromekelleher, line 2476: typo: gauranteed

jeromekelleher (approved):
Looks great, thanks @grahamgower. One tiny typo spotted.

I'm generally a bit uneasy about how much logic is going in here, but I guess it can't be avoided unless we somehow modified the low-level C code to emit a demes model through the debug interface.

grahamgower:
Typo fixed, rebased, and lightly squashed.

jeromekelleher:
Ah, sorry - can you update the CHANGELOG with this please? This is a big feature, don't want to forget about it on release.

grahamgower:
Changelog updated.

PR #1730, opened 2021-06-14
Explicitly define "source" and "dest" with targets

hyanwong:
Casting another glance over the docs, I worry somewhat that the “reversed terminology” of source and dest in msprime is confusing (and will only become more so as we incorporate demes models). I think it might help if, the first time in a tut the words are mentioned, we could link the words ("source" and "dest") themselves to actual definitions. This is my attempt to do that. Any improvements to the wording very welcome, as this is a particularly hard thing to phrase clearly.

### DIFF
--- a/docs/demography.md
+++ b/docs/demography.md
@@ -172,10 +172,28 @@ great way to deepen your understanding of the coalescent in general.
 Mostly it doesn't matter, and it's just a point of clarification
 for reading this documentation and understanding why the models
 we define are usually expressed backwards in time. However,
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
LGTM

@grahamgower, do you think this un-confuses things a big?

grahamgower (approved):
Yep, LGTM. This is a good place to put this, I think.

petrelharp (approved):
LGTM! Thanks.

PR #1738, opened 2021-06-17
added simple pedigree simulator

nikbaya:
Created new pedigree simulator with accompanying tests.

### DIFF
--- a/tests/test_pedigree.py
+++ b/tests/test_pedigree.py
@@ -5,6 +5,7 @@ import unittest
 
 import numpy as np
 import pytest
+import tskit
 
 import msprime
 from msprime import pedigrees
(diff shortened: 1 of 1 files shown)


jeromekelleher:
@nikbaya - it would be good to update the description here so that others know what's going on (for others, Nik and I went through this today and we didn't bother updating the description at the time)

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
 
         with pytest.raises(ValueError):
             ped.get_times(individual=ped.individual)
+
+
+def simulate_pedigree(
+    n_founders=8,
+    n_children_prob=None,
+    n_generations=3,
+    random_sex=True,
```
jeromekelleher, line 184: I'm not sure we need the non-random sex option, do we? Probably isn't going to matter for the simulations we'll be doing (we're not going to take sex into account yet) so I think it's just extraneous complexity.

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+) -> tskit.IndividualTable:
```
jeromekelleher, line 185: We need to control randomness here, so we want to have a ``random_seed`` argument. Since this is just a testing method we can make it default to some value (``random_seed=42``, say)

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+    def choose_sex(random=True, sex=None):
+        # male = 1, female = 2
+        if random:
+            sex = np.random.binomial(n=1, p=0.5) + 1
```
jeromekelleher, line 189: By using ``np.random`` methods we're affecting the global state in the numpy module, which we don't want to do for test cases (test cases should always be self contained and deterministic). So, we should do something like 
```
rng = np.random.RandomState(random_seed)
...
sex = rng.binomial(n=1, p=0.5) + 1
```

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
 
         with pytest.raises(ValueError):
             ped.get_times(individual=ped.individual)
+
+
+def simulate_pedigree(
+    n_founders=8,
```
jeromekelleher, line 181: Generally the convention within msprime and tskit is to be a bit more verbose than this, we'd have ``num_founders`` rather than ``n_founders``. Probably no need to give a default here I'd say, easy enough to specify a number of founders.

tests/test_pedigree.py
jeromekelleher, line 185: Also need some documentation here describing what the function does. It doesn't need to  be the full sphinx approach we'd use for a public function that's part of the API, but developers should be able to read a quick summary to understand what the function does and what the parameters mean.

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+    if n_children_prob is None:
+        n_children_prob = [0, 0, 1]
+
+    if random_sex:
+        sex = None
+    else:
+        sex = 1  # male=1, female=2
+
+    # first generation
+    n_males = int(n_founders / 2)
```
jeromekelleher, line 221: Use the integer division operaror: ``num_males = num_founders // 2``

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+) -> tskit.IndividualTable:
+    def choose_sex(random=True, sex=None):
+        # male = 1, female = 2
+        if random:
+            sex = np.random.binomial(n=1, p=0.5) + 1
+        else:  # non-random: alternate male female to ensure equal balance
+            sex = (sex % 2) + 1
+        return sex
+
+    def choose_n_children(n_children_prob, n_families=1):
```
jeromekelleher, line 194: Hardly worth making a function out of this I think, it's a one-liner.

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+            "properties": {
+                "sex": {"type": "integer"},
+            },
+            "required": ["sex"],
+            "additionalProperties": True,
+        }
+    )
+
+    if n_children_prob is None:
+        n_children_prob = [0, 0, 1]
```
jeromekelleher, line 213: Not obvious what this means, can you explain the function documentation?

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+        next_gen = [[], []]
+        avail_pat = np.random.permutation(curr_gen[0])
+        avail_mat = np.random.permutation(curr_gen[1])
+        n_pairs = min(len(curr_gen[0]), len(curr_gen[1]))
+        if n_pairs == 0 and n_children_prob[0] != 1:
+            raise Exception(
+                f"Not enough people to make children in generation {gen_idx}"
+            )
+        pairs = zip(avail_pat[:n_pairs], avail_mat[:n_pairs])
+        n_children_per_pair = choose_n_children(
```
jeromekelleher, line 237: Not sure why you wouldn't do this in the loop? We're not worried about efficiency here, something like
```python
for pair in zip(avail_pat[:n_pairs], avail_mat[:n_pairs]):
   num_children = ...
```

tests/test_pedigree.py
```diff
@@ -174,3 +175,153 @@ def test_pedigree_sanity_checks(self):
...
+    ):
+        if n_children_prob is None:
+            n_children_prob = [0, 0, 1]
+        tb = simulate_pedigree(
+            n_founders=n_founders,
+            n_children_prob=n_children_prob,
+            n_generations=n_generations,
+            random_sex=random_sex,
+        )
+        return tb
```
jeromekelleher, line 266: We should add a test to make sure that the pedigree we return is considered valid by tskit, i.e., create a TableCollection and see if we can call ``tables.tree_sequence()`` successfully.

jeromekelleher (commented):
Very nice, thanks @nikbaya. Just a few style points and simplifications and I think we're good to go.

tests/test_pedigree.py
nikbaya: I needed it for `test_grandparents`, so that there are enough males and females in the second generation to make the third generation. Should I simplify the code even further and not pair mates by sex?

jeromekelleher: It might be worth it - it's probably a good thing if we're producing super weird tangled up pedigrees. I guess the first pass at this could be the simplest possible pedigree simulator, which makes no attempt at all at realism.

nikbaya: ok, I'll change it to ignore sex when pairing up mates

jeromekelleher:
Can you ping me when this is ready for final review please @nikbaya?

nikbaya:
@jeromekelleher it should be ready now, I think I've made all the requested changes

nikbaya:
Just realised I need to uncomment and edit one of the tests, should be ready soon

nikbaya:
@jeromekelleher ok should be ready now, sorry about that

jeromekelleher (approved):
Very nice! Great work @nikbaya, thanks.

Issue #1485, opened 2021-02-15
"Sandbox violation" with meson 0.57


jeromekelleher:
Looks like our C library build is broken by meson 0.57. The recommended approach for now is to make sure you have an older version installed:
```
python3 -m pip install meson==0.56
```

```
$ meson build
The Meson build system
Version: 0.57.0
Source dir: /home/jk/work/github/msprime/lib
Build dir: /home/jk/work/github/msprime/lib/build
Build type: native build
Project name: msprime
Project version: undefined
C compiler for the host machine: cc (gcc 9.3.0 "cc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0")
C linker for the host machine: cc ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64

|Executing subproject kastore method meson 
|
|Project name: kastore
|Project version: undefined
|C compiler for the host machine: cc (gcc 9.3.0 "cc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0")
|C linker for the host machine: cc ld.bfd 2.34
|C++ compiler for the host machine: c++ (gcc 9.3.0 "c++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0")
|C++ linker for the host machine: c++ ld.bfd 2.34

subprojects/kastore/meson.build:14:0: ERROR: Sandbox violation: Tried to grab file kastore.c outside current (sub)project.

A full log can be found at /home/jk/work/github/msprime/lib/build/meson-logs/meson-log.txt
jk@tempest$ 
```

grahamgower:
It doesn't seem like meson is in the `requirements/development.txt` file. Perhaps it should be, pinned to the working version? I just dumbly installed meson, it failed of course, but then it took me a while to remember that I'd seen this issue in my inbox some months ago.

jeromekelleher:
Hmm yes, probably should be. I think it's a fairly harmless dependency right, it doesn't pull in loads of other stuff? (That's probably why I left it out originally)

grahamgower:
Zero dependencies. One must install ninja manually too!
```
$ pip show ninja meson
Name: ninja
Version: 1.10.0.post2
Summary: Ninja is a small build system with a focus on speed
Home-page: http://ninja-build.org/
Author: Jean-Christophe Fillion-Robin
Author-email: scikit-build@googlegroups.com
License: Apache 2.0
[...]

PR #1741, opened 2021-06-18
Add meson and ninja to developer requirements.

grahamgower:
See #1485.

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -8,8 +8,11 @@ demes>=0.1.2
 flake8
 hypothesis
 matplotlib>=3.4.0
+# More recent meson is too strict. See #1485.
+meson<=0.56
 mock
 mypy
+ninja
 numpy
 pre-commit
 pytest


jeromekelleher:
Didn't realise there was a pip distribution of ninja, that's pretty handy. I don't think there's any harm in putting these in the dev requirements - @benjeffery, what do you think?

benjeffery:
> Didn't realise there was a pip distribution of ninja, that's pretty handy. I don't think there's any harm in putting these in the dev requirements - @benjeffery, what do you think?

Meson doesn't have wheels and is built on `pip install` - does this cause issues for OSX/WIN?

jeromekelleher:
It's pure python AFAIK, so doesn't matter?

benjeffery (approved):
Yep, looks like it is. All good!

jeromekelleher:
Should probably add these to tskit dev requirements too so

PR #1742, opened 2021-06-22
Workflow updates

benjeffery:
Switch to using new dependabot, which has worked well over at tskit.

Also update pre-commit, which has some opinions on typing specifications, and caught two missing `asserts`.


benjeffery:
Seems the docs don't like the new style, will revert.

grahamgower:
Pipes for union types don't work until Python 3.10, right? https://www.python.org/dev/peps/pep-0604/

benjeffery:
> Pipes for union types don't work until Python 3.10, right? https://www.python.org/dev/peps/pep-0604/

Yes, I'm not sure why pyupgrade is putting them in considering we specify 3.7.

benjeffery:
@grahamgower msprime does `from __future__ import annotations` which enables this syntax.

benjeffery:
All fixed up, letting `jupyter-book` choose its own `sphinx` version fixed this.

jeromekelleher (approved):
Looks great, thanks @benjeffery! The pipe notation for typing is really nice, excellent we can use that.

PR #1743, opened 2021-06-22
Update Python version of simulator to use tables as input

### LINKED ISSUE #1739: Update algorithms.py to use tskit individual table
The Python [algorithms.py](https://github.com/tskit-dev/msprime/blob/main/algorithms.py) contains Python versions of the main algorithms implemented in msprime, and is intended to mirror the low-level implementations so that we can use it to prototype new features. The pedigree simulation is out of date now, and needs to be updated to use the tskit Individual table as the means of providing input.

Iinitially, let's assume that the ``pedigree_file`` is a tskit Tree sequence that contains a complete pedigree, with two nodes associated with each individual (and therefore complete time information), and trim out any extra functionality that's currently in there.

jeromekelleher:
These are minimal changes needed to get the pedigree simulator working with the new specification of pedigrees, i.e., via the information in a TableCollection.

To simulate a simple pedigree, we can now do:
```
python3 tests/test_pedigree.py 10 10 > pedigree.trees
```
This writes out a tree sequence containing pedigree information (10 founders for 10 generations). I've udpated this simulator to output node information also, so that the times and locations etc of individuals are stored in the node properties, ``time``, ``population``, ``flags`` etc.

Then, we can provide this pedigree as input to algorithms.py via 
```
python3 algorithms.py 0 -r 0.1 --model=wf_ped -F pedigree.trees tmp.trees -v 
```

The advantage here is that we can completely specify the required state of the simulation with existing data structures, and there's a clear path towards making things more complex, e.g., via multiple populations etc.

There's still a fair bit of refactoring to do, but at least this gets the code working.

Closes #1739

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -14,7 +14,6 @@ import numpy as np
 import tskit
 
 import msprime
-from msprime import pedigrees
 
 
 logger = daiquiri.getLogger()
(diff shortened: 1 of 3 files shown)


jeromekelleher:
@DomNelson - I've pinged you for a review her as you're the original author. I realise you don't have much time for this stuff, but it'd be great if you could take a quick look and see if you can spot any mistakes I've made, or point out any further opportunities for simplification.

PR #1747, opened 2021-06-22
mention missing flanks, closes #1746

### LINKED ISSUE #1746: behavior of sim_ancestry with nan values in rate map not obviously explained
@jradrion recently was surprised to find that if you read in a HapMap file whose first row has a start position that is above zero and a 0.0 map position, then the rate value on the first chunk of chromosome is `nan`, and there are no trees simulated there. This makes sense (and we discussed it) - effectively, the hapmap file is saying the the sequencable part of the chromosome doesn't start at coordinates 0. And, [the docs say](https://tskit.dev/msprime/docs/stable/api.html?highlight=read_hapmap#msprime.RateMap.read_hapmap) "If the first genetic position is 0 the interval from position 0 to the first physical position in the file is marked as unknown.", but I think we ought to say explicitly that you'll get missing data up until the first position.

petrelharp:
A better fix would be to also fill out the stub in [ancestry simulations: missing data](https://tskit.dev/msprime/docs/stable/ancestry.html?highlight=missing%20data#missing-data) but this is a start.

(also, the docs aren't building locally for me for some reason, grrr...)

### DIFF
--- a/msprime/intervals.py
+++ b/msprime/intervals.py
@@ -463,7 +463,7 @@ class RateMap(collections.abc.Mapping):
         # so we explicitly disable reformatting here.
         # fmt: off
         """
-        Parses the specified file in HapMap format and returns a RateMap.
+        Parses the specified file in HapMap format and returns a :class:`.RateMap`.
         HapMap files must white-space-delimited, and by default are assumed to
         contain a single header line (which is ignored). Each subsequent line
         then contains a physical position (in base pairs) and either a genetic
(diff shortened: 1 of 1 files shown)


petrelharp:
@jradrion want to take a look at this?

jeromekelleher:
> (also, the docs aren't building locally for me for some reason, grrr...)

Try updating Jupyterbook, there was a major version bump recently

jeromekelleher (approved):
LGTM

hyanwong (approved):
LGTM

PR #1751, opened 2021-07-01
Fixed pedigree simulation error where unnecessary nodes are added

nikbaya:


algorithms.py
```diff
@@ -784,19 +789,22 @@ def free_segment(self, u):
...
+            time=self.t, flags=flags, population=population
+        )
 
     def flush_edges(self):
         """
         Flushes the edges in the edge buffer to the table, squashing any adjacent edges.
         """
         if len(self.edge_buffer) > 0:
-            parent = len(self.tables.nodes) - 1
+            # print(self.edge_buffer)
```
jeromekelleher, line 801: Some stray prints here, good to clean up this up. Also in a few other places.

tests/test_algorithms.py
```diff
@@ -169,13 +169,45 @@ def test_from_ts(self):
...
         for tree in ts.trees():
             assert tree.num_roots > 1
-        assert ts.num_individuals == len(tables.individuals)
-        assert ts.num_samples == 20
+        for node in ts.nodes():
+            assert node.individual != tskit.NULL
+        assert (
+            len(tables.individuals)
+            == ts.num_individuals
+            == len(ts.dump_tables().individuals)
```
jeromekelleher, line 189: No need to check this last one, that's a basic tskit thing.

tests/test_algorithms.py
```diff
@@ -169,13 +169,45 @@ def test_from_ts(self):
...
+        )
+        assert len(ts.dump_tables().nodes) == len(tables.nodes)
```
jeromekelleher, line 191: Likewise here, tskit guarantees that the number of nodes in the tables is the same as the ts

tests/test_algorithms.py
```diff
@@ -169,13 +169,45 @@ def test_from_ts(self):
...
+        # founder nodes have correct time
+        founder_ids = {
+            idx
+            for idx, ind in enumerate(ts.dump_tables().individuals)
+            if np.array_equal(ind.parents, [-1, -1])
+        }
+        assert all(
+            [
+                (node.individual in founder_ids)
+                for node in ts.dump_tables().nodes
```
jeromekelleher, line 203: You can also do ``for node in ts.nodes()``

tests/test_algorithms.py
```diff
@@ -169,13 +169,45 @@ def test_from_ts(self):
...
+                if node.time == (num_generations - 1)
+            ]
+        )
+
```
jeromekelleher, line 207: Good to make this a separate tests case, ``test_one_gen_pedigree`` or something.

jeromekelleher, line 207: So we're not really testing the trees here. What about something like
```python
for tree in ts.trees():
    for node_id in tree.nodes():
         node = ts.node(node_id)
         individual = ts.individual(node.individual)
         if tree.parent(node_id) != tskit.NULL:
              parent_node = ts.node(tree.parent(node_id))
              assert parent_node.individual in individual.parents
         else:
              assert node.individual in founder_ids
```
right? This checks that the trees agree with the pedigree.

jeromekelleher (commented):
Looks great, thanks @nikbaya! Minor updates, and we're good to merge I think.

tests/test_algorithms.py
nikbaya: This is throwing an error because I'm getting trees that look like this: 
```
             36      35         21                 31
             ┏┻━┓   ┏━┻━━┓    ┏━━┻━━┓            ┏━━┻━┓
     55      ┃  ┃  48    ┃   46     ┃      52    ┃    ┃
    ┏━┻━┓    ┃  ┃  ┏┻━┓  ┃  ┏━┻┓    ┃      ┏┻━┓  ┃    ┃
    ┃  69    ┃  ┃  ┃  ┃  ┃  ┃  ┃   76      ┃  ┃  ┃   79
    ┃  ┏┻━┓  ┃  ┃  ┃  ┃  ┃  ┃  ┃  ┏━┻┓     ┃  ┃  ┃  ┏━┻┓
80 81 92 94 82 83 84 90 93 85 91 96 98 86 87 89 88 97 99 95
```
Would we expect the nodes of founders (0-19) to be missing?

jeromekelleher: Hmm, so it looks like we're missing edges joining the nodes where coalescences occured to the founders. To have a full record, we'd need to have "unary" edges joining (say) 80 back to its founder.

I'll take a look, this is an important detail.

jeromekelleher:
I did a bit of updating here @nikbaya, as I realised these tests I had suggested we're actually correct. I also added an edge back to the founder nodes, so it should all be working pretty well now.

I think I'll merge this much, and then we can apply the same logic changes to the C code.

jeromekelleher (approved):

PR #1756, opened 2021-07-06
Implement deactivate population

apragsdale:
Related to #1708, this fills in the deactivate population event at the low level.


jeromekelleher (approved):
Perfect!

jeromekelleher:
I'd suggest the next step would be to fill in the stuff in ``_msprimemodule.c`` and to add tests for it in ``test_lowlevel.py``.

PR #1758, opened 2021-07-08
Activate populations in from_demes

apragsdale:
Probably needs a few more tests, but this is a start to ensure that the interval that a population is active matches the deme lifespan from the demes model input.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -392,6 +392,19 @@ class Demography(collections.abc.Mapping):
             # the KeyError message
             self[pop_ref]
 
+    def _add_activate_population_event(
+        self, time: float, *, population: str | int
+    ) -> ActivatePopulationEvent:
+        """
+        Activates a population at the specified time. The population is expected to be
+        initially inactive.
+
+        :param float time: The time at which this event occurs in generations.
+        :param str, int population: The population to activate.
+        """
+        self._check_population_references([population])
+        return self.add_event(ActivatePopulationEvent(time=time, population=population))
+
     def add_population_split(
         self, time: float, *, derived: list[str | int], ancestral: str | int
     ) -> PopulationSplit:
@@ -1856,29 +1869,40 @@ class Demography(collections.abc.Mapping):
 
         graph = graph.in_generations()
         events = graph.discrete_demographic_events()
-        initially_active = collections.defaultdict(lambda: True)
-        for split in events["splits"]:
-            initially_active[split.parent] = False
-        for merge in events["mergers"]:
-            for parent in merge.parents:
-                initially_active[parent] = False
 
         demography = Demography()
         for deme in graph.demes:
             initial_size = 0
             growth_rate = 0
             last_epoch = deme.epochs[-1]
+            initially_active = False
+            to_activate = False
             if last_epoch.end_time == 0:
                 initial_size = last_epoch.end_size
                 growth_rate = get_growth_rate(last_epoch)
+                initially_active = True
+            else:
+                # add activation if needed
+                # needed if isn't the ancestor of any split or merger
+                to_activate = True
+                for split in events["splits"]:
+                    if deme.name == split.parent:
+                        to_activate = False
+                for merger in events["mergers"]:
+                    if deme.name in merger.parents:
+                        to_activate = False
             demography.add_population(
                 name=deme.name,
                 description=deme.description,
                 growth_rate=growth_rate,
                 initial_size=initial_size,
                 default_sampling_time=deme.end_time,
-                initially_active=initially_active[deme.name],
+                initially_active=initially_active,
             )
+            if to_activate:
+                demography._add_activate_population_event(
+                    time=deme.end_time, population=deme.name
+                )
             for epoch in reversed(deme.epochs):
                 new_initial_size = None
                 if initial_size != epoch.end_size:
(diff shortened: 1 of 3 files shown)


apragsdale:
The random demes Graph testing using hypothesis looks to be creating models with absurdly large event times (order 1e+27), which looks to be causing floating point errors or something like that. That could be an issue on the import demes hypothesis strategies? cc: @grahamgower

grahamgower:
> The random demes Graph testing using hypothesis looks to be creating models with absurdly large event times (order 1e+27), which looks to be causing floating point errors or something like that. That could be an issue on the import demes hypothesis strategies? cc: @grahamgower

Ughhh... :( I spent quite some time trying to avoid those problems, but I must have missed something. Maybe it's more trouble than it's worth. Probably best to just skip the hypothesis tests for now.

msprime/demography.py
```diff
@@ -392,6 +392,24 @@ def _check_population_references(self, populations: list[str | int]):
...
+        """
+        Activates a population at the specified time. The population is expected to be
+        initially inactive.
+
+        :param float time: The time at which this event occurs in generations.
+        :param str, int population: The population to activate.
+        """
+        self._check_population_references([population])
+        pop = self[population]
+        if pop.initially_active is None:
```
jeromekelleher, line 407: It's not clear to me that we want to put in this complexity - we'll only really be using this functionality within Demes models, where we'll be controlling the ``initially_active`` state explicitly anyway, won't we?

msprime/demography.py
```diff
@@ -3150,6 +3183,35 @@ def _as_lineage_movements(self):
         ]
 
 
+@dataclasses.dataclass
+class ActivatePopulationEvent(DemographicEvent):
+    """
+    :param int population: The ID of the population to activate.
```
jeromekelleher, line 3189: The attribute docstring gets associated with the attribute itself, see the Population class, e.g.

jeromekelleher (approved):
LGTM - nice to simplify the semantics if we can, though.

msprime/demography.py
apragsdale: For now, I imagine it will really only be used in the Demes import interface. But I wasn't sure how much to put into it, in case someone does want to use it outside of Demes in the future.

msprime/demography.py
apragsdale: Ok, I'll try to clean this all up as well

msprime/demography.py
jeromekelleher: I'd put in the minimum possible semantics for what we need right now - we're not putting it into the public interface, so we don't have to worry about forwards compatibility.

apragsdale:
Ok, minimal makes sense to me. Just so I understand, would you suggest completely removing that ActivatePopulationEvent class and the _add_activate_population_event function, and then just handling those actions entirely within the from_demes function?

jeromekelleher:
> Ok, minimal makes sense to me. Just so I understand, would you suggest completely removing that ActivatePopulationEvent class and the _add_activate_population_event function, and then just handling those actions entirely within the from_demes function?

No, it's good to have these because they'll do things like helping with Debugging. I'm just talking about simplifying the semantics around ``initially_active``.

msprime/demography.py
apragsdale: I'm not sure if how I changed it is what you had in mind. Not really familiar with this way of adding doc to the attribute itself, so let me know if I'm off here.

msprime/demography.py
```diff
@@ -3186,10 +3181,14 @@ def _as_lineage_movements(self):
...
 class ActivatePopulationEvent(DemographicEvent):
     """
-    :param int population: The ID of the population to activate.
+    A population activation event, which changes the state of the population
+    from inactive to active at the given time.
     """
 
     population: int | str
+    """
+    :param int population: The ID of the population to activate.
```
jeromekelleher: We don't want the :param: notation - see the example [here](https://github.com/tskit-dev/msprime/blob/975710471cd5efd26efb00eda4dd1c66e9811a71/msprime/demography.py#L88) for the recommended way to document dataclass attributes.

apragsdale: The docstring from the latest commit has just "The ID of the population to activate.", and I removed the :param: notation. Not sure why it wasn't updated here.

jeromekelleher (approved):
LGTM, thanks @apragsdale!

Pinging @grahamgower for a review before merging

grahamgower (approved):
LGTM.

jeromekelleher:
Great - shall we disable the hypothesis tests for now @grahamgower or keep this PR on ice until there's an update?

grahamgower:
> Great - shall we disable the hypothesis tests for now @grahamgower or keep this PR on ice until there's an update?

Best to disable the hypothesis tests I think.

jeromekelleher:
OK - would you mind adding a commit here to disable the hypothesis tests please @apragsdale? Good to merge then I think.

apragsdale:
I commented out the hypothesis testing instead of deleting, so it's more or less there if we come back to it, but could also delete if that's preferable.

apragsdale:
Ok - squashed!

jeromekelleher (approved):

jeromekelleher:
@mergifyio update

PR #1759, opened 2021-07-09
Updated pedigree sim and tests in C

nikbaya:
Updated WF pedigree model to match the Python implementation.

Modified existing tests for the pedigree simulations.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -838,6 +838,7 @@ msp_free_pedigree(msp_t *self)
         for (i = 0; i < self->pedigree->num_inds; i++) {
             msp_safe_free(ind->parents);
             msp_safe_free(ind->segments);
+            msp_safe_free(ind->nodes);
             ind++;
         }
     }
@@ -1471,6 +1472,16 @@ msp_print_individual(msp_t *self, individual_t *ind, FILE *out)
             fprintf(out, " None");
         }
     }
+
+    fprintf(out, " ], Nodes: [");
+
+    for (j = 0; j < self->ploidy; j++) {
+        if (ind->nodes[j] != TSK_NULL) {
+            fprintf(out, " %d", ind->nodes[j]);
+        } else {
+            fprintf(out, " None");
+        }
+    }
     fprintf(out, " ]\n");
 }
 
@@ -1786,8 +1797,7 @@ msp_store_edge(msp_t *self, double left, double right, tsk_id_t parent, tsk_id_t
     int ret = 0;
     tsk_edge_t *edge;
     const double *node_time = self->tables->nodes.time;
-
-    tsk_bug_assert(parent > child);
+    tsk_bug_assert(parent < child);
     tsk_bug_assert(parent < (tsk_id_t) self->tables->nodes.num_rows);
     if (self->num_buffered_edges == self->max_buffered_edges - 1) {
         /* Grow the array */
@@ -2116,7 +2126,8 @@ alloc_individual(individual_t *ind, size_t ploidy)
     // Better to allocate these as a block?
     ind->parents = malloc(ploidy * sizeof(individual_t *));
     ind->segments = malloc(ploidy * sizeof(avl_tree_t));
-    if (ind->parents == NULL || ind->segments == NULL) {
+    ind->nodes = malloc(ploidy * sizeof(avl_node_t *));
+    if (ind->parents == NULL || ind->segments == NULL || ind->nodes == NULL) {
         ret = MSP_ERR_NO_MEMORY;
         goto out;
     }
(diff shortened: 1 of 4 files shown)


lib/msprime.c
```diff
@@ -2116,7 +2126,8 @@ alloc_individual(individual_t *ind, size_t ploidy)
     // Better to allocate these as a block?
     ind->parents = malloc(ploidy * sizeof(individual_t *));
     ind->segments = malloc(ploidy * sizeof(avl_tree_t));
-    if (ind->parents == NULL || ind->segments == NULL) {
+    ind->nodes = malloc(ploidy * sizeof(avl_node_t *));
```
jeromekelleher, line 2129: This is the wrong size - should be ``sizeof(tsk_id_t)`` or (better) ``sizeof(*ind->nodes)``, so you don't have to write down what the type is twice.

lib/msprime.c
```diff
@@ -3169,15 +3140,14 @@ msp_merge_ancestors(msp_t *self, avl_tree_t *Q, population_id_t population_id,
                 }
             }
         } else {
-            if (!coalescence) {
+            if (new_node_id==TSK_NULL) {
                 coalescence = true;
                 l_min = l;
-                ret = msp_store_node(self, 0, self->time, population_id, individual);
+                new_node_id = (tsk_id_t) msp_store_node(self, 0, self->time, population_id, individual);
```
jeromekelleher, line 3146: Need to check the error here:
```c
if (new_node_id < 0) {
    ret = (int) new_node_id;
    goto out;
}
```

jeromekelleher: I think you also need to modify msp_store_node to return the node ID.

lib/msprime.c
```diff
@@ -4366,55 +4340,72 @@ msp_pedigree_climb(msp_t *self)
...
                 }
-                continue;
-            }
-
-            /* Recombine and climb to segments to the parents */
-            if (rate_map_get_total_mass(&self->recomb_map) > 0) {
-                ret = msp_dtwf_recombine(self, merged_segment, &u[0], &u[1]);
-                if (ret != 0) {
-                    goto out;
+                x = merged_head;
```
jeromekelleher, line 4365: We can do this a bit more concisely in C:
```c
for (x = merged_head; x != NULL; x = x->next) {
    ...
}
```

lib/msprime.c
```diff
@@ -4366,55 +4340,72 @@ msp_pedigree_climb(msp_t *self)
...
+            ret = msp_merge_ancestors(self, segments, 0, 0, &merged_head, parent_id, node);
             if (ret != 0) {
                 goto out;
             }
-            if (merged_segment == NULL) {
+            if (merged_head == NULL) {
                 // This lineage has coalesced
                 continue;
             }
+            ret = msp_flush_edges(self);
```
jeromekelleher, line 4351: I think we should only have the flush edges at the end of the loop - we're doing it twice here, right?

lib/msprime.c
```diff
@@ -7103,6 +7104,7 @@ msp_set_simulation_model_wf_ped(msp_t *self)
     }
 
     parents = malloc(num_individuals * self->ploidy * sizeof(*parents));
+    nodes = malloc(num_individuals * self->ploidy * sizeof(*nodes));
```
jeromekelleher, line 7107: We don't need the parents, nodes, etc arrays at all now, right?

jeromekelleher (commented):
Looks good!

It might be a good idea to make the change msp_merge_ancestors first separately though, and to make sure this all works by running it through CI. Always good to do things bit-by-bit and try to break as few things as possible at one time.

Could be simplest to just do it as a separate PR.

lib/msprime.c
nikbaya: I think tsk_node_table_add_row already returns the node ID to msp_store_node (see the return of tsk_node_table_add_row_internal)

nikbaya:
> It might be a good idea to make the change msp_merge_ancestors first separately though, and to make sure this all works by running it through CI. Always good to do things bit-by-bit and try to break as few things as possible at one time.

@jeromekelleher Argh yeah, I totally forgot that changing msp_merge_ancestors would break things outside of the pedigree sims because I was just running the pedigree test suite. Starting a new branch now!

jeromekelleher:
> @jeromekelleher Argh yeah, I totally forgot that changing msp_merge_ancestors would break things outside of the pedigree sims because I was just running the pedigree test suite. Starting a new branch now!

Haha, compile and run tests early and often!

jeromekelleher:
I think we can close this @nikbaya?

PR #1768, opened 2021-07-13
Workaround GitHub's broken package parsers

jeromekelleher:
So that we're visible in the Dependency Graph.


jeromekelleher:
Tests pass, so hopefully this is harmless. Going to merge to see if it works and msprime's Dependent packages show up.

PR #1772, opened 2021-07-16
clarify units of `initial_size`

petrelharp:
There's a number of places where we just say "population size" without reminding the reader it's in units of individuals (i.e., multiplied by ploidy). Maybe this makes things more clear?

### DIFF
--- a/docs/demography.md
+++ b/docs/demography.md
@@ -303,8 +303,11 @@ and makes it easier to avoid
 
 #### Initial size
 
-The `initial_size` of a population is its size at the start of a simulation
+The `initial_size` of a population is its size, as a number of individuals,
+at the start of a simulation
 (looking {ref}`backwards in time<sec_demography_direction_of_time>`).
+In other words, an initial size of ``n`` means that the population had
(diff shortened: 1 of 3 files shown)


mufernando (approved):
LGTM! I think this is specially necessary because in earlier versions of msprime and in other simulators population size refers to the number of chromosomes.

petrelharp:
Do you think it's confusing that it says "initial number of individuals" when it's really a haploid simulator under the hood? Alternatively we could say "initial size (which will be multiplied by ploidy to get the number of chromosomes)" or something.

jeromekelleher (approved):
Good call, thanks @petrelharp !

jeromekelleher:
> Do you think it's confusing that it says "initial number of individuals" when it's really a haploid simulator under the hood?

I don't think it's less confusing if we try to insulate people from the details of what's happening under the hood. Strictly speaking it's model dependent anyway - DTWF deals with individuals (although we don't explicitly track them) whereas Hudson approximates by assuming individuals don't matter.

PR #1779, opened 2021-07-21
Added new_node_id arg to msp_merge_ancestors

nikbaya:


jeromekelleher:
This isn't quite the same logic as we implemented in the Python version @nikbaya - the value of ``new_node_id`` that's passed in is never checked and so won't be used.

Also looks like you need to pass the code through clang-format, see [here](https://tskit.dev/tskit/docs/latest/development.html#sec-development-c-code-style) for info

PR #1783, opened 2021-07-23
Pin demes < 0.2.

grahamgower:
We anticipate making a breaking change with 0.2.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,7 @@ install_requires =
     numpy
     newick>=1.3.0
     tskit>=0.3.5
-    demes
+    demes<0.2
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Are we doing this in case we release a version before Demes 0.2 is released?

grahamgower:
I think having an msprime point release before demes 0.2 would be ideal. Right now we're in the unfortunate situation that msprime 1.0.1 depends on demes (any version), which will then become broken when demes 0.2 is released.

jeromekelleher:
OK, merging. I'll see what's involved in a point release.

Issue #1775, opened 2021-07-19
mean_coalescence_time(): convergence problems?


grahamgower:
`DemographyDebugger.mean_coalescence_time()` seems to have difficulty with more complex population relationships. For the example below, the mean value does not appear to be converging (at least, not quickly enough). Increasing the `max_iter` parameter doesn't seem to be a viable option, because the number of time steps is doubled in each iteration (so the cpu-time doubles each iteration). The `rtol` parameter is also not useful in this case --- the logged `mean_diff` value is smallest in the first iteration, so either we take the mean coalescent time estimated in that first iteration, or we run forever. Perhaps there's some fine tuning that could be done here, or the time steps could be chosen differently?

```python
import demes
import msprime
import daiquiri

test_case = """\
time_units: generations
defaults:
  epoch: {start_size: 1000}
demes:
- name: A
- name: B
  ancestors: [A]
  start_time: 3000
- name: C
  ancestors: [B]
  start_time: 2000
- name: D
  ancestors: [C]
  start_time: 1000
migrations:
- demes: [A, D]
  rate: 1e-5
"""

graph = demes.loads(test_case)
dbg = msprime.Demography.from_demes(graph).debug()
print(dbg)
daiquiri.setup(level="DEBUG")
t = dbg.mean_coalescence_time({"A": 1, "C": 1}, max_iter=20)
```

Output before I killed the run.
```
DemographyDebugger
╠══════════════════════════════════╗
║ Epoch[0]: [0, 1e+03) generations ║
╠══════════════════════════════════╝
╟    Populations (total=4 active=4)
║    ┌─────────────────────────────────────────────────────────────┐
║    │   │    start│      end│growth_rate  │   A   │ B │ C │   D   │
║    ├─────────────────────────────────────────────────────────────┤
║    │  A│   1000.0│   1000.0│ 0           │   0   │ 0 │ 0 │ 1e-05 │
║    │  B│   1000.0│   1000.0│ 0           │   0   │ 0 │ 0 │   0   │
║    │  C│   1000.0│   1000.0│ 0           │   0   │ 0 │ 0 │   0   │
║    │  D│   1000.0│   1000.0│ 0           │ 1e-05 │ 0 │ 0 │   0   │
║    └─────────────────────────────────────────────────────────────┘
╟    Events @ generation 1e+03
║    ┌──────────────────────────────────────────────────────────────────────────────────┐
║    │  time│type            │parameters         │effect                                │
[...]

PR #1787, opened 2021-07-25
Change matrix exponential algorithm

petrelharp:
I've implemented an alternative matrix exponential method as discussed in #1775 - the tests for matrix exponenetials don't pass because they're not all stochastic matrices; but something else isn't working and I haven't had a chance to look into it.

### DIFF
--- a/tests/test_demography.py
+++ b/tests/test_demography.py
@@ -1873,44 +1873,43 @@ class TestMatrixExponential:
     Test cases for the matrix exponential function.
     """
 
-    def verify(self, A):
+    def verify(self, A, E=None):
+        print(A)
+        assert np.max(np.diag(A)) <= 0
+        assert np.min(A - np.diag(np.diag(A))) >= 0
+        assert np.allclose(np.sum(A, 1), np.zeros((np.shape(A)[0],)))
         E1 = scipy.linalg.expm(A)
         E2 = msprime.demography._matrix_exponential(A)
+        assert np.min(E2) >= 0
+        assert np.allclose(np.sum(E2, 1), np.ones((np.shape(A)[0],)))
         assert E1.shape == E2.shape
         assert np.allclose(E1, E2)
-
-    def test_singleton(self):
-        for j in range(10):
-            A = np.array([[j]])
-            self.verify(A)
+        if E is not None:
+            assert E.shape == E2.shape
+            assert np.allclose(E, E2)
 
     def test_zeros(self):
         for j in range(1, 10):
             A = np.zeros((j, j))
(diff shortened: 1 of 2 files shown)


petrelharp:
Ok - I think I've got a better expm algorithm in here, and the demography tests pass (at least locally?). Want to put in a test that exercises #1775, @grahamgower? I think you can push to this PR? We could merge this in, since I think this algorithm is superior, except that I've left the old expm still here for easy comparison (e.g., we'll need it to check that the new tests actually do the bad thing).

tests/test_demography.py
```diff
@@ -1873,44 +1873,42 @@ class TestMatrixExponential:
...
     def test_identity_exp(self):
         # (-1) * np.eye(k), compared to exp(-1) * np.eye(k)
         for k in range(2, 5):
-            A = msprime.demography._matrix_exponential((-1) * np.eye(k))
-            B = np.exp(-1) * np.eye(k)
-            assert np.allclose(A, B)
+            A = (-1) * np.eye(k)
+            for i in range(k):
+                A[i, (i + 1) % k] = 1.0
+            print(A)
```
jeromekelleher, line 1910: print

jeromekelleher (commented):
LGTM. To avoid having untested code in the library (pulling down the test coverage stats) we should either (a) delete it, (b) move it into the test module; (c) run some tests on it.

I think it's probably worth making a small ``test_matrix_exponential.py`` test module to isolate this stuff, since it's easy for it to get lost in the (very large) test_demography.py file.

@grahamgower, are you OK picking this one up? If you can't  push to the branch, then you can make your own PR, preserving Peter's commits?

grahamgower:
Thanks @petrelharp, this is excellent! It works for my example from #1775, and there's a nice speed improvement too.

@jeromekelleher, yes, I can pick this up.

jeromekelleher:
Great, thanks @grahamgower - I've pushed this into the 1.0.2 release, since you probably want a fix for this out ASAP also?

grahamgower:
> Great, thanks @grahamgower - I've pushed this into the 1.0.2 release, since you probably want a fix for this out ASAP also?

Sounds good!

petrelharp:
> LGTM. To avoid having untested code in the library (pulling down the test coverage stats) we should either (a) delete the old version (b) move it into the test module; (c) run some tests on it.

I think we should delete it, unless we want to move it to the test code (eg we could use it instead of scipy to compare against in validation). But I'm assuming @grahamgower has the edit token now, so I'll leave it alone.

PR #1788, opened 2021-07-26
Change matrix exponential algorithm

grahamgower:
Successor to #1787. (Only maintainers can push to pr branches, so I needed to open a new pr)

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -3480,25 +3480,28 @@ def _matrix_exponential_poisson(A, n=5):
     """
     dA = (-1) * np.diag(A)
     dmax = np.max(dA)
-    if dmax == 0:
-        expA = np.eye(A.shape[0])
-    else:
+    expA = np.eye(A.shape[0])
+    if dmax > 0:
         P = A / dmax
         np.fill_diagonal(P, 1 - dA / dmax)
+        # nscales is the number of scaling-and-squaring steps:
+        # we compute exp(tA) and then square the result nscales times
         nscales = int(max(0, np.ceil(np.log2(dmax) - np.log2(0.2))))
         t = dmax / (2 ** nscales)
-        Pk = np.exp(-t) * t * P
-        expA = Pk
-        np.fill_diagonal(expA, np.exp(-t) + np.diag(expA))
-        for k in range(2, n + 1):
+        # Now, exp(tA) = exp(-t) * (I + tP + (tP)^2 / 2 + ...)
+        # and the k-th term in this sum is Pk=(tP/k times the previous)
+        Pk = np.eye(A.shape[0])
+        for k in range(1, n + 1):
             Pk = (t / k) * np.matmul(P, Pk)
             expA = expA + Pk
+        expA *= np.exp(-t)
         for _ in range(nscales):
             expA = np.matmul(expA, expA)
     return expA
 
 
-_matrix_exponential = _matrix_exponential_poisson
+def _matrix_exponential(*args, **kwargs):
+    return _matrix_exponential_poisson(*args, **kwargs)
 
 
 class PopulationStateMachine(enum.IntEnum):
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -3459,7 +3459,7 @@ def _proportions_to_sequential(P):
     return C
 
 
-def _matrix_exponential(A):
+def _matrix_exponential_eigen(A):
     """
     Returns the matrix exponential of A.
     https://en.wikipedia.org/wiki/Matrix_exponential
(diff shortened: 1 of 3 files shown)


msprime/demography.py
```diff
@@ -3459,18 +3459,36 @@ def _proportions_to_sequential(P):
...
-def _matrix_exponential(A):
+def _matrix_exponential(A, n=5):
     """
     Returns the matrix exponential of A.
-    https://en.wikipedia.org/wiki/Matrix_exponential
-    Note: this is not a general purpose method and is only intended for use within
-    msprime.
+    Only works if the offdiagonals of A are nonnegative
+    and the row sums of A are less than or equal to zero.
+    Melloy & Bennett (1993), https://doi.org/10.1016/0377-0427(93)90036-B
```
grahamgower, line 3467: @petrelharp, I wasn't sure if you'd looked at this ref, or come up with this indepentently. It helped me understand what you were doing anyhow.

grahamgower:
Hmm, looks like numpy's linalg backend behaves differently on different platforms...

msprime/demography.py
petrelharp: Oh, great! That not only does just what we do, but proves that it's optimal in some sense! (I worked it out myself, but it's very natural - we're "computing the transition probability of a continuous-time Markov process by Poissonization"; and the reason I did it that way was that Simon Tavaré told me once that that's the best way to do it.)

Note that the reference wants row sums are zero, when in fact it works if they are less than or equal to zero.

msprime/demography.py
```diff
@@ -3459,28 +3459,16 @@ def _proportions_to_sequential(P):
...
-    """
     Only works if the offdiagonals of A are nonnegative
     and the row sums of A are less than or equal to zero.
+    Melloy & Bennett (1993), https://doi.org/10.1016/0377-0427(93)90036-B
     """
-    # TODO: check if these affect the timing
     assert np.max(np.diag(A)) <= 0
-    assert np.min(A - np.diag(np.diag(A))) >= 0
+    assert np.min(np.tril(A, k=-1)) == 0
+    assert np.min(np.triu(A, k=1)) == 0
```
petrelharp, line 3471: oh good! I was hoping there was a better way to do that.

msprime/demography.py
```diff
@@ -3499,15 +3487,10 @@ def _matrix_exponential_poisson(A, n=5):
             Pk = (t / k) * np.matmul(P, Pk)
             expA = expA + Pk
         expA *= np.exp(-t)
-        for _ in range(nscales):
-            expA = np.matmul(expA, expA)
+        expA = np.linalg.matrix_power(expA, 2 ** nscales)
```
petrelharp, line 3490: I'm a bit nervous about this, not knowing about the numerical stability and efficiency of `np.linalg.matrix_power`? Oh, never mind! From [the docs](https://numpy.org/doc/stable/reference/generated/numpy.linalg.matrix_power.html) I see that
```
For positive integers n, the power is computed by repeated matrix squarings and matrix multiplications
```

msprime/demography.py
```diff
@@ -4169,6 +4153,8 @@ def mean_time(steps, P):
                 len(steps),
                 max(steps),
             )
+            P1 = P2
+            m1 = m2
```
petrelharp, line 4157: good call!

msprime/demography.py
```diff
@@ -4200,7 +4186,7 @@ def coalescence_rate_trajectory(
         :math:`z`, and let :math:`N(z,t)` be the diploid effective population
         size of population :math:`z` at time :math:`t`. Then the mean
         coalescence rate at time :math:`t` is :math:`r(t) = (\\sum_z p(z,t) /
-        (2 * N(z,t)) / p(t)`.
+        (2 * N(z,t))) / p(t)`.
```
petrelharp, line 4189: good catch

tests/test_matrix_exponential.py
```diff
@@ -0,0 +1,136 @@
...
+    Test cases for the matrix exponential function.
+    """
+
+    def verify(self, A, E=None):
+        assert np.max(np.diag(A)) <= 0
+        assert np.min(A - np.diag(np.diag(A))) >= 0
+        assert np.max(np.sum(A, 1)) <= 0
+        E1 = scipy.linalg.expm(A)
+        E2 = msprime.demography._matrix_exponential(A)
+        E3 = _matrix_exponential_eigen(A)
```
petrelharp, line 34: since we're comparing to the eigendecomposition, shall we remove the scipy version?

tests/test_matrix_exponential.py
```diff
@@ -0,0 +1,136 @@
...
+            E = np.exp(-j) * np.eye(j) + (1 - np.exp(-j)) * np.ones((j, j)) / j
+            self.verify(A, E)
+
+    def test_identity_exp(self):
+        # (-1) * np.eye(k), compared to exp(-1) * np.eye(k)
+        for k in range(2, 5):
+            A = (-1) * np.eye(k)
+            for i in range(k):
+                A[i, (i + 1) % k] = 1.0
+            self.verify(A)
```
petrelharp, line 65: we could use maybe one more example, like maybe:
```
for k in range(2,5):
   A = np.arange(k**2).reshape(k, k)
   np.fill_diagonal(-1 * np.sum(A, 1) + np.diag(A))
   self.verify(A)
```

tests/test_matrix_exponential.py
```diff
@@ -0,0 +1,137 @@
...
+        T = np.linspace(0, last_epoch + 12 * last_N, 101)
+        rates1, P1 = ddb.coalescence_rate_trajectory(
+            T, {"A": 1, "C": 1}, double_step_validation=False
+        )
+        assert np.all(rates1 >= 0)
+        assert np.all(P1 >= 0)
+        assert np.all(P1 <= 1)
+        assert np.all(np.diff(P1) <= 0)
+
+        # scipy.linalg.expm() is unstable for this example.
```
petrelharp, line 106: Do we even want to keep in everything from here down, now that we know this is a good test? It's a cautionary note to future us, I guess, but we could just put a comment in the code saying that "this example is unstable with scipy.linalg.expm and _matrix_exponential_eigan"?

petrelharp (commented):
Looks good! I vote to insert one more test of expm (as I put in) and remove all the scipy stuff, but happy to hear otherwise?

msprime/demography.py
grahamgower: I did some informal timing tests, and was surprised that I could measure a difference with/without the asserts. This change made the time differences harder to measure (closer to the noise).

msprime/demography.py
grahamgower: This is potentially not as obvious (one has to read the docs). But it's measurably faster when calling `mean_coalescence_time()`.

petrelharp (approved):
Great, thanks! LGTM!

jeromekelleher:
Happy to merge @grahamgower?

grahamgower:
> Happy to merge @grahamgower?

Yep, unless you want me to squash?

jeromekelleher:
No, looks about right - merging..

PR #1797, opened 2021-07-30
Update ancestry.md

TPPSellinger:
I added a small section to describe the practical consequences of having branch lengths proportional to N^2 in the dirac coalescent. @JereKoskela  can you check if what I wrote is correct and clear ?

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -912,7 +912,7 @@ ts = msprime.sim_ancestry(
 
 In the following visualisation, we trace the migration history of two samples backwards
 in time to their most recent common ancestor. Each one of the ten populations in the
-model is assigned a 
+model is assigned a
 colour and a position on the x axis. The time of migration events is recorded on the y
 axis.
 
@@ -2070,6 +2070,38 @@ tree = ts.first()
 print(tree.tmrca(0,1))
 
 ```
+
+Because branch lengths are proportional to {math}`N^2` generations, the number
+of events simulated along branches (such as mutations or recombinations) will
+also be proportional to {math}`N^2`. For example, if a tree sequence generated
+using a given census size {math}`N` is used as input for simulating segregating
+sites (see {ref}`Mutation simulations<sec_mutations>` for details on how to do
+this), the expected number of segregating sites obtained will be proportional
+to {math}`N^2`.
+
+```{code-cell}
+ts = msprime.sim_ancestry(
+    samples=1, ploidy=2, random_seed=1, population_size=10000,recombination_rate=1e-8,
+    sequence_length=10**4,model=msprime.DiracCoalescent(psi=0.1, c=0))
+mts = msprime.sim_mutations(ts, rate=1e-8)
+print(str(mts.get_num_sites()))
+```
+
+If the desired amount of diversity (as measured by the number of segregating
+sites) is proportional to {math}`N`, then {math}`N^{1/2}` should be used for
+the population size under the Dirac-coalescent.
+
+
+```{code-cell}
+ts = msprime.sim_ancestry(
+    samples=1, ploidy=2, random_seed=1, population_size=100,recombination_rate=1e-8,
+    sequence_length=10**4,model=msprime.DiracCoalescent(psi=0.1, c=0))
+mts = msprime.sim_mutations(ts, rate=1e-8)
+print(str(mts.get_num_sites()))
+```
+
+Therefore the population size must be carefully chosen (and
+potentially rescaled) to obtain the desired simulated data in these models.
 
 (sec_ancestry_models_selective_sweeps)=
 


docs/ancestry.md
```diff
@@ -2071,6 +2071,27 @@ print(tree.tmrca(0,1))
 
 ```
 
+Because branch lengths are proportional to {math}`N^2` generations, other events (such as mutations or recombinations) will also be proportional to {math}`N^2`. For example, if c is set to 0, the dirac coalescent will behave as if under a Moran model. However if believed census size is inputted (e.g. {math}`N=10000`), a number of segregating sites proportional to {math}`N^2` will be obtained. Leading to unrealistic simulated data.
```
JereKoskela, line 2074: I don't think it's a good idea to introduce the concept of a Moran model here. We don't want to make this any worse than it has to be for people who are not familiar with obtaining coalescent processes as scaling limits from forward-time models. I also don't think that we should say any particular outcome is more or less realistic - that's highly application-specific. Instead, we should help practitioners calibrate what number they should plug in for `N` to get the results they expect. I've suggested a slight redraft below.

Because branch lengths are proportional to {math}`N^2` generations, the number of events typically simulated along branches (such as mutations or recombinations) will also be proportional to {math}`N^2`.
For example, if a tree sequence generated using a given census size {math}`N` is used as input for simulating segregating sites (see {ref}`Mutation simulations<sec_mutations>` for details on how to do this), the expected number of segregating sites obtained will be proportional to {math}`N^2`.

docs/ancestry.md
```diff
@@ -2071,6 +2071,27 @@ print(tree.tmrca(0,1))
...
+
+```{code-cell}
+ts = msprime.sim_ancestry(
+    samples=1, ploidy=2, random_seed=1, population_size=10000,recombination_rate=1e-8,
+    sequence_length=10**4,model=msprime.DiracCoalescent(psi=0.1, c=0))
+mts = msprime.sim_mutations(ts, rate=1e-8)
+print(str(mts.get_num_sites()))
+```
+However if the square root of the population size is inputted, the expected (and desired) ammount of diversity is obtained.
```
JereKoskela, line 2083: Continuing from above, how about:

If the desired amount of diversity (as measured by the number of segregating sites) is proportional to {math}`N`, then {math}`N^{1/2}` should be used for the population size under the Dirac-coalescent.

JereKoskela:
Thanks @TPPSellinger! I've suggested a few tweaks above, but I think overall this is very clear.

@jeromekelleher, it would be good to get your thoughts on this. It was born out of a suggestion by @TPPSellinger that the unusual scaling of the multiple merger models would be clearer if it was more explicitly connected to an observable quantity like the number of segregating sites, which is typically what people care about calibrating. Invariably that creates something of a link between the model used to simulate the tree sequence and a subsequent `sim_mutations` call, but I think something like this PR is the least intrusive way to introduce that if it is to be done.

TPPSellinger:
I agree with what you said. The point is to not lose the reader, my mistake for introducing unecessary terms. I will wait for Jerome's comments and do the modifications.

jeromekelleher:
Seems like a useful clarification to me - I don't have anything to add to @JereKoskela's points.

docs/ancestry.md
petrelharp: Sorry, I'm really very late here - but, is the reason for branch lengths being `N^2` the same as for the Moran model? For the Moran model, it takes order of `N^2` *units of time* to coalesce, but it also takes `N` units of time for a typical individual to reproduce, so I'd argue that the time to coalescence in the Moran is still `N` generations (well, `2N`, I guess). So, if I were to implement the Moran model in msprime, I'd have chosen that time scaling, so that things like mutation rates match what they mean in other models (i.e., mutations per generation).

TPPSellinger:
I changed the text according to your comments @JereKoskela . Let me know if it's better now.

docs/ancestry.md
```diff
@@ -2071,7 +2071,7 @@ print(tree.tmrca(0,1))
 
 ```
 
-Because branch lengths are proportional to {math}`N^2` generations, other events (such as mutations or recombinations) will also be proportional to {math}`N^2`. For example, if c is set to 0, the dirac coalescent will behave as if under a Moran model. However if believed census size is inputted (e.g. {math}`N=10000`), a number of segregating sites proportional to {math}`N^2` will be obtained. Leading to unrealistic simulated data.
+Because branch lengths are proportional to {math}N^2 generations, the number of events simulated along branches (such as mutations or recombinations) will also be proportional to {math}N^2. For example, if a tree sequence generated using a given census size {math}N is used as input for simulating segregating sites (see {ref}Mutation simulations<sec_mutations> for details on how to do this), the expected number of segregating sites obtained will be proportional to {math}N^2.
```
jeromekelleher, line 2074: You're missing backticks here for the math expressions, ``{math}`N^2```

TPPSellinger: My mistake. I will fix this.

docs/ancestry.md
JereKoskela: @petrelharp, the reason for the `N^2` scaling for the Dirac-coalescent is exactly the same as it is for the Moran model, and you're totally right about the scaling. On the other hand, the reason for the `N^{a-1}` scaling for the Beta-coalescent is different, and can't be addressed by a similar "time units to generations" mapping. From memory, given that we were going to have to explain the latter anyway, we thought it would be less obscure to use `N^2` for the former as well (matching the time scaling in the source material we cite for the model), and avoid any mention of time units distinct from generations.

jeromekelleher:
@JereKoskela, are you happy?

JereKoskela:
> @JereKoskela, are you happy?

Looks good to me @jeromekelleher. Thanks @TPPSellinger!

jeromekelleher (approved):

docs/ancestry.md
petrelharp: ok, great - I figured you were thinking about this, but wanted to make sure.

PR #1801, opened 2021-08-03
Full ARG for gene conversion

### LINKED ISSUE #1773: Add gene conversion to record_full_arg option
We should add support for gene conversion nodes to the record full ARG option.

Should be fairly similar to recombination nodes.

If someone is interested in this functionality (and ideally helping to implement/test it), please comment here.

JereKoskela:
Adds support for recording gene conversion nodes when `store_full_arg = True`, as well as some basic tests.

Edit: Closes #1773

### DIFF
--- a/lib/tests/test_ancestry.c
+++ b/lib/tests/test_ancestry.c
@@ -1299,7 +1299,7 @@ test_mixed_hudson_smc(void)
 
 static void
 run_gc_simulation(double sequence_length, double gc_rate, double tract_length,
-    double recombination_rate, bool discrete_genome)
+    double recombination_rate, bool discrete_genome, bool full_arg)
 {
     int ret;
     uint32_t n = 10;
@@ -1321,6 +1321,7 @@ run_gc_simulation(double sequence_length, double gc_rate, double tract_length,
     CU_ASSERT_EQUAL_FATAL(msp_set_recombination_rate(&msp, recombination_rate), 0);
     CU_ASSERT_EQUAL_FATAL(msp_set_gene_conversion_rate(&msp, gc_rate), 0);
     CU_ASSERT_EQUAL_FATAL(msp_set_gene_conversion_tract_length(&msp, tract_length), 0);
+    CU_ASSERT_EQUAL_FATAL(msp_set_store_full_arg(&msp, full_arg), 0);
     /* Set a very small block size to force lots of fenwick tree rebuilds */
     CU_ASSERT_EQUAL_FATAL(msp_set_segment_block_size(&msp, 16), 0);
     ret = msp_initialise(&msp);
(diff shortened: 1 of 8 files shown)


fbaumdicker (changes requested):
Nice @JereKoskela ! I think you missed the second process that handles gene conversions that start from the left of the first segment. If there is no particular reason to not record those events you should add the full-arg-recording also to the `msp_gene_conversion_left_event` function.

JereKoskela:
Thanks @fbaumdicker, I had completely missed that there was a second function! I'll add recording to that as well and push another commit.

JereKoskela:
Ok, now covering both `msp_gene_conversion_event` and `msp_gene_conversion_left_event`.

lib/tests/test_ancestry.c
```diff
@@ -1378,25 +1381,33 @@ test_gc_tract_lengths(void)
...
 }
 
 static void
 test_gc_rates(void)
 {
-    run_gc_simulation(10, 0.1, 1.5, 1.0, false);
-    run_gc_simulation(5, 5.0, 2.5, 0.01, false);
-    run_gc_simulation(10, 1, 1, 1.0, false);
-    run_gc_simulation(30, 1.0, 6.5, 1.0, true);
+    run_gc_simulation(10, 0.1, 1.5, 1.0, false, false);
```
jeromekelleher, line 1403: Might be more readable to do a loop here and above
```c
for (full_arg = 0; full_arg < 2; full_arg++) {
    run_gc_simulation(5, 5.0, 2.5, 0.01, false, full_arg);
    // etc
}

File: tests/test_ancestry.py [hunk 2]
```diff
@@ -178,6 +178,46 @@ def test_multimerger(self):
         )
         self.verify(sim, multiple_mergers=True)
 
+    def test_gene_conversion_n25(self):
```
jeromekelleher, line 181: We're not actually checking for  properties of the GC ARG here. We can either update the ``verify`` method above to also check for GC nodes, or make a separate test class where the verify is just checking for properties of the ARG with GC. I don't mind which.

jeromekelleher (commented):
This is great, thanks @JereKoskela! The only issue is that the tests aren't actually checking any semantics - the function ``msp_store_arg_gene_conversion`` could be a no-op and these tests would still pass (always a good metric for judging your tests!).

lib/tests/test_ancestry.c
JereKoskela: Good idea. Done.

tests/test_ancestry.py
JereKoskela: Well spotted. I've extended the `verify` method to check GC nodes. The alternative would have resulted in a lot of duplicated code.

jeromekelleher:
Excellent! Good to merge after a squash.

jeromekelleher:
Ah - actually, can you update the CHANGELOG to document the addition of this feature please? You might need to make a new release block (put in 2021-XX-XX as the date).

JereKoskela:
> Ah - actually, can you update the CHANGELOG to document the addition of this feature please? You might need to make a new release block (put in 2021-XX-XX as the date).

Looks doable at a glance. I gather the block should be ## [1.0.3] 0 2021-XX-XX? I'll include that in the squash update and you can tell me if I've done anything dumb @jeromekelleher.

JereKoskela:
Squashed and Changelog updated.

jeromekelleher (approved):

Issue #1780, opened 2021-07-21
Class design for "selective sweep" concept.


molpopgen:
This issue is motivated by #1762 

The current class for modeling the [structured coalescent](https://pubmed.ncbi.nlm.nih.gov/3066685/) is `msprime.SweepGenicSelection`.  This class handles the following scenario:

* Given a start/end frequency (e.g., moving backwards in time so that end < start), we generate a trajectory using the conditional diffusion methods described in [Coop and Griffiths (2004)](https://pubmed.ncbi.nlm.nih.gov/15465123/).  The trajectory of a mutation with fitnesses 1, 1+s/2 and 1+s for the 3 diploid genotypes, conditional on it moving from frequency `start` to `end`.
* That trajectory is associated with a genomic position.
* Given that trajectory, we do the operations described [here](https://pubmed.ncbi.nlm.nih.gov/7498754/) to generate the history of the two parts of the population--lineages linked to the site and those that have recombined away from it (or weren't linked in the first place if start frequency is low enough).

The `init` arguments that matter most are start/end frequencies and `s`.

Moving forwards, there are many more modeling possibilities:

* Handling [arbitrary dominance](https://pubmed.ncbi.nlm.nih.gov/16219788/), which amounts to a new trajectory-making function.
* Adding an option that guarantees that the model end at a frequency 1/(2N).  (Not guaranteeing this is the cause of #1762.)  We want this to be optional so that one could chain together multiple sweep models to do "stranger" things like stitch together trajectories where `s` and/or dominance vary over time.  By default, we probably want this feature to be **on**, reflecting the most common anticipated use case.
* Allowing for population size changes, and possibly migration of the beneficial allele, thus getting closer to feature parity with [discoal](https://pubmed.ncbi.nlm.nih.gov/27559153/).

So, at the very least, we want the `init` method to account for:

* Dominance, via an `h` `kwarg`
* Guaranteeing that the trajectory goes down to a very small frequency, a.k.a "1/2N".

We also need a name, bearing in mind the assumptions as I understand them:

1. The Hudson model (as opposed to DWTF or others).
2. The trajectory applies to a single site and a mutation that arose in a **single copy** at some point in the past. (In other words, we are not allowing for the other flavor of sweeps from standing variation by which the beneficial allele was present in multiple copies of different evolutionary origin at the onset of selection.  Nor are we allowing recurrent mutation to the beneficial allele during the "selective phase" of the simulation.)

I've likely missed a few things.

[...]

Issue #1762, opened 2021-07-12
Sweeps from standing variation give incorrect results.


molpopgen:
The genic selection model is not giving the correct output when the starting frequency is high.

The code to reproduce the plot below is [here](https://github.com/molpopgen/msprime_discoal_comparison), and the README discusses the dependencies.

The following plots show the ECDF for the number of segregating sites for:

1. dicsoal (blue) with selection
2. msprime (black) with selection
3. msprime (red) without selection.

The upper left panel is a hard sweep, and the two simulations with selection agree.  As the initial frequency increases, though, msprime "gets neutral" too quickly.

What is not shown here is that forward simulations agree with discoal for all 4 panels.  I can provide figures if desired, but it looks like there's a bug in msprime.

![compare](https://user-images.githubusercontent.com/6416159/125341688-c0dec300-e308-11eb-8be4-12925a6933f2.png)

andrewkern:
huh interesting... i wonder what i messed up here. i'll dig a bit.

molpopgen:
For the record, I caught this b/c using msprime instead of discoal made the accuracy of diploSHIC "quite poor" and I tracked it down to this diff in the number of segregating sites.

andrewkern:
so browsed this quickly-- i think the difference is that discoal is doing a full soft sweep (i.e. conditioning on a neutral trajectory from freq f_0 -> 1/2N) where as we don't yet have that for msprime, so msprime is going straight from f_0 back to a standard coalescent sim. does that make sense you reckon Kevin?

molpopgen:
Msprime must be doing something, as initial frequency has a predictable effect in the expected direction.  But it does seem like msprime is exiting the sweep phase too early.

andrewkern:
yeah i think that's it. to get msprime to do something equivalent to the `discoal -f` option we need to add a neutral trajectory model or maybe just a soft sweep model. it's on my radar.

molpopgen:
So we need to be a bit careful here, and will want @jeromekelleher to weigh in on this.  We have a 1.0 API that has a `start_frequency` in there, so fiddling with that could be API-breaking.  Seems that we also don't want to add more classes here?  It could be confusing to have a separate SoftSweep class as a model in addition to GenicSelection--what if SelectionWithDominance is added later?  Would we want a separate SoftSweepWithDominance for consistency? It can kinda blow up on you.

[...]

PR #1803, opened 2021-08-06
Drift trajectory implementation

andrewkern:
this PR is dealing with issues #1762 and #1780. this implements a NeutralFixation model, which deals with a structured coalescent model akin to a selective sweep but for the case where the allele that we condition upon is neutral rather than beneficial.

This will need a rebase, etc, but wanted to get some feedback on what i've done thus far

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -7077,9 +7077,13 @@ genic_selection_generate_trajectory(sweep_t *self, msp_t *simulator,
         }
         pop_size = get_population_size(&simulator->populations[0], sim_time);
         alpha = 2 * pop_size * trajectory.s;
-        x = 1.0
-            - genic_selection_stochastic_forwards(
-                  trajectory.dt, 1.0 - x, alpha, gsl_rng_uniform(rng));
+        if (alpha > 0) {
+            x = 1.0
+                - genic_selection_stochastic_forwards(
+                      trajectory.dt, 1.0 - x, alpha, gsl_rng_uniform(rng));
+        } else {
+            x = neutral_stochastic_backwards(trajectory.dt, x, gsl_rng_uniform(rng));
+        }
         /* need our recored traj to stay in bounds */
         t += trajectory.dt;
         sim_time += trajectory.dt * pop_size * simulator->ploidy;
@@ -7335,7 +7339,7 @@ msp_set_simulation_model_sweep_genic_selection(msp_t *self, double position,
         ret = MSP_ERR_BAD_TIME_DELTA;
         goto out;
     }
-    if (s <= 0) {
+    if (s < 0) {
         ret = MSP_ERR_BAD_SWEEP_GENIC_SELECTION_S;
         goto out;
     }
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -7004,7 +7004,7 @@ out:
 }
 
 /**************************************************************
- * Allele frequency trajectory simulation for genic selection
+ * Allele frequency trajectory simulation for sweep routines
  *
  **************************************************************/
 
@@ -7014,6 +7014,16 @@ genic_selection_stochastic_forwards(double dt, double freq, double alpha, double
     /* this is scaled following Ewens chapter 5 e.g.,
      * w_11=1+s; w_12=1+s/2; w_22=1; that is h=0.5 */
     double ux = ((alpha / 2.0) * freq * (1 - freq)) / tanh((alpha / 2.0) * freq);
+    int sign = u < 0.5 ? 1 : -1;
+    return freq + (ux * dt) + sign * sqrt(freq * (1.0 - freq) * dt);
+}
+
+static double
+neutral_stochastic_backwards(double dt, double freq, double u)
+{
+    /* returns allele freq of neutral allele drifting conditional
+    on loss (looking backwards in time) following Ewens */
+    double ux = -1.0 * freq;
     int sign = u < 0.5 ? 1 : -1;
     return freq + (ux * dt) + sign * sqrt(freq * (1.0 - freq) * dt);
 }
--- a/msprime/_msprimemodule.c
+++ b/msprime/_msprimemodule.c
@@ -1077,9 +1077,10 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
     PyObject *dirac_s = NULL;
     PyObject *beta_s = NULL;
     PyObject *sweep_genic_selection_s = NULL;
+    PyObject *neutral_fixation_s = NULL;
     PyObject *value;
     int is_hudson, is_dtwf, is_smc, is_smc_prime, is_dirac, is_beta, is_sweep_genic_selection;
-    int is_wf_ped;
+    int is_wf_ped, is_neutral_fixation;
     double psi, c, alpha, truncation_point;
 
     hudson_s = Py_BuildValue("s", "hudson");
@@ -1112,6 +1113,10 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
     }
     sweep_genic_selection_s = Py_BuildValue("s", "sweep_genic_selection");
     if (sweep_genic_selection_s == NULL) {
+        goto out;
+    }
+    neutral_fixation_s = Py_BuildValue("s", "neutral_fixation");
+    if (neutral_fixation_s == NULL) {
         goto out;
     }
 
@@ -1219,9 +1224,23 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
             goto out;
         }
     }
+    
+    is_neutral_fixation = PyObject_RichCompareBool(py_name,
+            neutral_fixation_s, Py_EQ);
+    if (is_neutral_fixation == -1) {
+        goto out;
+    }
+    if (is_neutral_fixation) {
+        /* currently using all the machinery from the genic model */
+        ret = Simulator_parse_sweep_genic_selection_model(self, py_model);
+        if (ret != 0) {
+            goto out;
+        }
+    }
 
     if (! (is_hudson || is_dtwf || is_smc || is_smc_prime || is_dirac
-                || is_beta || is_sweep_genic_selection || is_wf_ped)) {
+                || is_beta || is_sweep_genic_selection || is_neutral_fixation
+                || is_wf_ped)) {
         PyErr_SetString(PyExc_ValueError, "Unknown simulation model");
         goto out;
     }
@@ -1239,6 +1258,7 @@ out:
     Py_XDECREF(beta_s);
     Py_XDECREF(dirac_s);
     Py_XDECREF(sweep_genic_selection_s);
+    Py_XDECREF(neutral_fixation_s);
     return ret;
 }
 
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1314,7 +1314,9 @@ class Simulator(_msprime.Simulator):
         """
         num_labels = 1
         for model in models:
-            if isinstance(model, SweepGenicSelection):
+            if isinstance(model, SweepGenicSelection) or isinstance(
+                model, NeutralFixation
+            ):
                 num_labels = 2
         return num_labels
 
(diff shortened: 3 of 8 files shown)


molpopgen:
Will take a look next week--probably Tuesday.

andrewkern:
kevin note this is just the neutral fixation case, an incremental step towards getting the soft sweep model in

andrewkern:
also i should probably add verification against SLiM for patterns of linked polymorphism to this PR. Right now in `verification.py` i've only got tests of the sojourn time of the neutral variants against the analytic expectation (which checks out) and i'm relying on the rest of the structured coalescent machinery to just work. it should though.

molpopgen:
Yes, we don't want anything slipping through the cracks, so more
verification is a good idea.

lib/msprime.c
```diff
@@ -7018,6 +7018,16 @@ genic_selection_stochastic_forwards(double dt, double freq, double alpha, double
...
 }
 
+static double
+neutral_stochastic_backwards(double dt, double freq, double u)
+{
+    /* returns allele freq of neutral allele drifting conditional
+    on loss (looking backwards in time) following Ewens */
+    double ux = -1.0 * freq;
+    int sign = u < 0.5 ? 1 : -1;
+    return freq + (ux * dt) + sign * sqrt(freq * (1.0 - freq) * dt);
```
molpopgen, line 7028: Can we think of reasonable ways to handle the possibility of numeric errors here?  It seems that there are no restrictions placed on the input values, so one can get negative frequencies out:

```python
In [1]: dt = 1e-8

In [2]: freq = 1e-8

In [3]: ux = -freq

In [4]: sign = -1

In [5]: import math

In [7]: freq + (ux * dt) + sign * math.sqrt(freq * (1.0 - freq) * dt)
Out[7]: -4.999999918875795e-17
```

Perhaps this should return `max(value, 0.0)` ?

msprime/_msprimemodule.c
```diff
@@ -1219,9 +1224,23 @@ Simulator_parse_simulation_model(Simulator *self, PyObject *py_model)
             goto out;
         }
     }
+
```
molpopgen, line 1227: I can't comment on the CPython bits -- need @jeromekelleher for that.

msprime/ancestry.py
```diff
@@ -1922,3 +1924,63 @@ def __init__(
...
+@dataclasses.dataclass
+class NeutralFixation(ParametricAncestryModel):
+    """
+    A fixation of a neutral mutation has occured in the history of the sample.
+    This will lead to a weak reduction in polymorphism  near the fixed site.
+
+    The model is one of a structured coalescent where selective backgrounds are
+    defined as in
+    `Braverman et al. (1995) <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1206652/>`_
+    The implementation details here follow closely those in discoal
```
molpopgen, line 1938: Is the scaling identical to Schrider and Kern?  The sweep model wasn't, so any differences here also need to be documented.

tests/test_models.py
```diff
@@ -1084,3 +1137,206 @@ def test_lots_of_sweeps(self):
...
+
+class TestNeutralFixation:
+    """
+    Tests for the neutral fixation model.
+    """
+
+    def test_model_end_broken(self):
+        # Checking that we're correctly detecting the fact that
+        # sweeps are non renentrant.
+        model = msprime.NeutralFixation(
```
molpopgen, line 1150: Totally opaque to me what this test is doing?

tests/test_models.py
```diff
@@ -1084,3 +1137,206 @@ def test_lots_of_sweeps(self):
...
+        )
+        with pytest.raises(RuntimeError, match="does not support interruption"):
+            msprime.sim_ancestry(10, model=model, end_time=0.0001)
+
+    def test_incorrect_num_labels(self):
+        model = msprime.NeutralFixation(
+            position=0.5, start_frequency=0.1, end_frequency=0.9, dt=0.01
+        )
+        for num_labels in [1, 3, 10]:
+            # Not the best error, but this shouldn't be exposed to the user anyway.
```
molpopgen, line 1161: This test is also unclear--`num_labels` is an undocumented part of the public API.  What's going on here?

tests/test_models.py
```diff
@@ -1084,3 +1137,206 @@ def test_lots_of_sweeps(self):
...
+        # make any difference.
+        model = msprime.NeutralFixation(
+            position=0.5, start_frequency=0.1, end_frequency=0.9, dt=0.01
+        )
+        ts = msprime.sim_ancestry(
+            10, sequence_length=1, recombination_rate=20, model=model
+        )
+        assert ts.num_trees == 1
+
+    def test_neutral_fixation_no_recomb(self):
```
molpopgen, line 1182: is this test helpful?  If it failed, that wouldn't imply an error in the neutral fixations, but somewhere else in the library, right?

tests/test_models.py
```diff
@@ -1084,3 +1137,206 @@ def test_lots_of_sweeps(self):
...
+            start_frequency=1.0 / (2 * N),
+            end_frequency=1.0 - (1.0 / (2 * N)),
+            dt=1e-6,
+        )
+        ts = msprime.sim_ancestry(10, model=model, population_size=1000, random_seed=2)
+        assert ts.num_trees == 1
+        for tree in ts.trees():
+            assert tree.num_roots == 1
+
+    def test_neutral_fixation_recomb(self):
```
molpopgen, line 1195: Tests like this are a bit strange--number of trees is only > 1 some times, depending on rec rate, seed, etc., and could fail at a later time if some other part of the library is touched.

tests/test_models.py
```diff
@@ -1084,3 +1137,206 @@ def test_lots_of_sweeps(self):
...
+            random_seed=2,
+        )
+        assert all(tree.num_roots == 1 for tree in ts.trees())
+
+    def test_many_neutral_fixations_regular_times_model_change(self):
+        models = []
+        for j in range(0, 10):
+            models.extend(
+                [
+                    # Start each sweep after 0.01 generations of Hudson
```
molpopgen, line 1302: is this really generations, or "time units"

verification.py
```diff
@@ -1641,6 +1641,60 @@ def test_sojourn_time2(self):
         pyplot.savefig(f, dpi=72)
         pyplot.close("all")
 
+    def test_sojourn_time3(self):
+        """
+        testing against expected sojourn time of a
```
molpopgen, line 1646: Comments need more detail for future readers: what is the expectation, etc.?

verification.py
```diff
@@ -1641,6 +1641,60 @@ def test_sojourn_time2(self):
...
+            for x in replicates:
+                tree_times = np.zeros(x.num_trees)
+                j = 0
+                for tree in x.trees():
+                    tree_times[j] = np.max([tree.time(root) for root in tree.roots])
+                    j += 1
+                reptimes[i] = np.max(tree_times)
+                i += 1
+            data["alpha_means"].append(np.mean(reptimes))
+            data["exp_means"].append(2 * n)
```
molpopgen, line 1684: This is why comments are useful: many will think that the expected time to fixation is 4N, but here it is written as 2N--why?

tests/test_models.py
jeromekelleher: I think these are copy-pastes of the Sweep selection model tests, and testing a bunch of internal things. This is checking for a known limitation.

jeromekelleher (commented):
I haven't gone through the details @andrewkern, but it mostly seems good. My high-level input would be that there seems to be a lot of copy-paste boilerplate which is basically identical to the SweepGenicSelection model. Perhaps we could think about generalising that, and make the SweepGenicSelection class as a subclass of this for compatibility purposes?

andrewkern:
Yeah I agree- this is too much of the same code repeated. One thing that I was having a hard time wrapping my head around is how to generalize this sort of model and deal with the `_to_lowlevel()` stuff. in particularly i'm not sure the best way to deal with the parameters that are associated, e.g., this ugly thing that i've done here:
https://github.com/andrewkern/msprime/blob/c7de424a686d4bba11a93e010d345b0c5cd57cfc/msprime/ancestry.py#L1969

any advice would be appreciated here Jerome.

tests/test_models.py
andrewkern: yep-- testing all this stuff in the `s=0` case. thought these would be good to include, though not necessary.

tests/test_models.py
andrewkern: right-- `num_labels` refers to the number of labels we use internally in the structured coalescent. we generalized the bookkeeping of nodes to include a `label` a ways back. here we are making sure that the model returns the correct number of labels, in this case 2

tests/test_models.py
andrewkern: fair. can probably be removed.

tests/test_models.py
andrewkern: should be generations-- internal time in `msprime`

jeromekelleher:
It looks like the only thing that's actually different is the trajectory generator, so what I'd like to do is to decouple the allele frequency trajectory generator from the ``Sweep`` model. Would it make sense to have something like
```python

class TrajectoryGenerator:
    start_frequency:float
    end_frequency: float
    dt: float

class SweepModel:
     position: float
     trajectory: TrajectoryGenerator
```

We could then stuff like 
```python
model = SweepModel(position=0.5, trajectory=GenicSelectionTrajectory(0.1, 0.9, 1e-6))
# or 
model = SweepModel(position=0.5, trajectory=NuetralFixationTrajectory(0.1, 0.9, 1e-6))
```

We could keep support for the existing SweepGenicSelection class easily enough by using these building blocks.

These TrajectoryGenerator classes would front-end a bunch of stuff from the CPython, and would eventually end up referring to the corresponding struct + function pointer in C (a bit like ancestry models or demographic events). This would mean that we could also test the trajectory generation code independently of the actual simulations (a big issue at the moment), and also do things like allow people define trajectory generators in Python which would be super slow, but handy for prototyping (or maybe not - we're just generating numpy arrays, some models might be possible to implement efficiently using numpy).

Would that work? Are there more types of sweep that would fit into this framework, i.e., where we just define the trajectory generator and don't touch the actual ancestry simulation code at all? 

If so, then maybe the best thing to do would be to back out of all the Python level changes and just commit in the nuetral trajectory in C (and maybe any others you have lying around handy) and then I'll jump in and do the plumbing?

andrewkern:
so i think it's a bit easier than this? all of the trajectories that we currently want (hard sweeps, soft sweeps, these neutral fixations) can be generated by the current C function `genic_selection_generate_trajectory()` (which itself should be renamed to something more general). 

currently it is doing neutral fixations by asking about the selection coefficient, so what i did was tack on the python `NeutralFixation` model to basically do everything that the current `SweepGenicSelection` model was doing, but fix the selection coefficient at `s=0`. if we added one more parameter, call it f_0, which told the C trajectory function to switch between neutrality and selection, then we could add a `SoftSweep` model. 

So maybe at the python level we have something like this?

```python
class SweepModel:
     position: float
     start_frequency:float
     end_frequency: float
     dt: float

class HardSweepModel(SweepModel):
     s: float

class SoftSweepModel(SweepModel):
     s: float
     f_0: float
```
etc.?

andrewkern:
other use cases that I can think of that we would want to cover would be:
1. hard partial sweeps (hard sweep that ends at freq < 1)
2. soft partial sweeps 
3. neutral fixations (already implemented)
4. long term balanced polymorphism (no trajectory there, just constant deme sizes among backgrounds)

jeromekelleher:
Thinking ahead though @andrewkern, can you see reasons why we would want to specify different trajectories? I would rather put a bit of time in now and get the framework right. If this is what the sweep model really *is* (a trajectory supplied to the existing sweeps simulatation), then I'd rather specify it this way rather than trying to hide this detail.

@molpopgen, any thoughts here?

andrewkern:
so the biggest issue i see in separating the trajectory generation from the model entirely is that eventually (a few PRs down the road) we will want to get the trajectory generation connected to changing population size e.g. https://github.com/tskit-dev/msprime/blob/c7de424a686d4bba11a93e010d345b0c5cd57cfc/lib/msprime.c#L7049

i was thinking we'd do this like `pop_size = get_population_size(sim->populations[0], time);` so that would would need the simulation object during trajectory generation.

molpopgen:
> Thinking ahead though @andrewkern, can you see reasons why we would want to specify different trajectories? I would rather put a bit of time in now and get the framework right. If this is what the sweep model really _is_ (a trajectory supplied to the existing sweeps simulatation), then I'd rather specify it this way rather than trying to hide this detail.
> 
> @molpopgen, any thoughts here?

For the case of a single trajectory, restricted to no migration between demes, this is the way to go.

molpopgen:
> Would that work?

This is something that I've been wanting to see for a long time--there should be no need to generate entire trajectories and store them in RAM.  I don't see any issue with this, but it'd take a good validation test to make sure.

andrewkern:
> So, we try to abstract out the details of everything _except_ the core function for transitioning an allele frequency randomly. All the parameters that we need are encapsulated in the particular type of trajectory.
> 
> Would that work?

Yeah I think that can work. I'd like to keep these allele freq transition functions on the C side if that's okay by you?

@molpopgen we are storing these trajectories in RAM currently, and will continue to need to for the rejection sampling that we will eventually add to deal with population size change

jeromekelleher:
> Yeah I think that can work. I'd like to keep these allele freq transition functions on the C side if that's okay by you?

Sure yeah - I just want to make it *possible* to specify one of these in Python.

> For the case of a single trajectory, restricted to no migration between demes, this is the way to go.

So looking ahead, is there a more complex scenario we'd want to model? A trajectory going through multiple demes that would need multiple population sizes? Or would that be better modelled with a specific class?

andrewkern:
i've never figured out the details of how to simulate alleles sweeping through multiple populations simultaneously in the coalescent. the weird bit is specifying the how / when of the beneficial allele reaching a 2nd (or 3rd etc) population in the context of migration between demes. 

 we can however do everything else in a multipop setting e.g., migration of neutral alleles, etc

andrewkern:
resurrected this cause of recent attention on slack, #1955, #1962

should we try to pull this functionality in finally?

molpopgen:
The thing that slowed this down, I think, is #1780.  There's some idea that we need to get the right API sorted out "at once".  Since we're now at 1.x, and fiddling we do risks breaking the expectations of semantic versioning.  So it behooves us to get this right the first time.

jeromekelleher:
After spending a fair bit of time on this a few months ago trying to see what the "right" API would look like, I've come to the conclusion that this is a hard problem and realistically it's something that we'd need someone working on full time for a couple of months to really nail down. None of us have this kind of time on our hands, so I think we need to decide whether we want to wait and get it right, or patch things up as best we can for now while keeping in mind that we'll probably deprecate the current API some day, replacing with something more general (when hopefully we'll have someone to work on it).

I'd vote for patching things up as best we can for now, without adding in *too* much extra code complexity or duplication.

So, I guess the first thing to do is to rebase this branch and bring it up to date, and maybe resolve any issues that have been sorted out? Or perhaps it would be simpler to make a fresh PR so that we can do clean reviews?

PR #1807, opened 2021-08-12
Pass node to merge ancestors in pedigree climb

nikbaya:
This PR is the next step in updating simulation through pedigrees. In `msp_pedigree_climb` we pass an individual's node to `msp_merge_ancestors`.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1782,7 +1782,6 @@ msp_store_node(msp_t *self, uint32_t flags, double time, population_id_t populat
     if (ret < 0) {
         goto out;
     }
-    ret = 0;
 out:
     return ret;
 }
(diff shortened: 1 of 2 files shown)


nikbaya:
@jeromekelleher Does the coverage failure mean I have to add more tests?

jeromekelleher:
Not in this case @nikbaya because you happen to be touching code that's involved in difficult-to-test error conditions.

nikbaya:
@jeromekelleher then I think if there's nothing to do, this is ready to be merged

jeromekelleher:
We should be able to test some properties of the simulations through the pedigree now in C though, shouldn't we? Like in a really simple example, make sure we're actually using the input nodes?

nikbaya:
Yeah that's true, I'll do that

nikbaya:
@jeromekelleher sorry, I've realised I don't quite follow. What do you mean by checking if we're "using the input nodes"? Would that mean checking in test_pedigrees.c if the node table looks correct? Or would it mean adding some `tsk_bug_assert` calls in msprime.c?

jeromekelleher:
I think we could add some simple checks to the C tests, about [here](https://github.com/tskit-dev/msprime/blob/7a49bc6ff3dc44171300ed8d7b2f15fa89fd1c8b/lib/tests/test_pedigrees.c#L95) and maybe elsewhere in that file. I guess checking that the number of nodes before and after running the simulation through the pedigree would be a minimal check (that would fail before making this change, right?)

nikbaya:
@jeromekelleher I think this is ready with the changes to test_pedigrees.c that we discussed.

jeromekelleher (approved):
LGTM

PR #1808, opened 2021-08-12
Initial outline for Trajectory refactoring

jeromekelleher:
Here's a rough outline for what I'm thinking the low-level C will look like. I think it's worth getting things right here in the Python version before looking at the C --- it's much faster to try different things out here. 

As discussed in #1803 

Does this look like roughly the right design @andrewkern, @molpopgen?

Looks like we're missing quite a bit of stuff from the Python version though @andrewkern?

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -2,6 +2,7 @@
 Python version of the simulation algorithm.
 """
 import argparse
+import dataclasses
 import heapq
 import logging
 import math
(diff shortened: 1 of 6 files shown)


jeromekelleher:
We don't have ploidy in the Python version @molpopgen, but we probably should. But yes, I think it would be best if we did things in the general case with arbitrary ploidy.

molpopgen:
> But yes, I think it would be best if we did things in the general case with arbitrary ploidy.

Yeah, seems prudent to iron out any wrinkles ASAP

PR #1814, opened 2021-08-18
remove skipif windows on ooa test, closes #1629

### LINKED ISSUE #1629: numerical issues with coalescence rate trajectories on windows
As observed in #1620, trying to calculate coalescence rate trajectories for the OOA model produces NAN's in Windows (only). I'm not inclined to fix this unless it actually bites someone. We could probably be smarter about solving the relevant differential equations.

(Note: maybe there's a way to do the solution analytically? But, with different populations growing at different exponential rates, we're solving a time-inhomogeneous Feynman-Kac problem; and while `u_t = a u - exp(g t) u` has a solution if a is a real number, it's not obvious how to get a solution with `a` a matrix and `g` a diagonal matrix.)

petrelharp:
Let's see!

### DIFF
--- a/tests/test_demography.py
+++ b/tests/test_demography.py
@@ -1853,7 +1853,6 @@ class TestDemographyTrajectories(unittest.TestCase):
         assert abs(coaltime - coaltime2) < 2
 
     @pytest.mark.slow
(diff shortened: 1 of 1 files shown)


petrelharp:
Yes! It works!

jeromekelleher (approved):
Windows tests are definitely valuable!

PR #1842, opened 2021-09-18
Experimental yaml input format

jeromekelleher:
This is an experiment to see what a yaml/json input format (building on demes) would look like. It mostly works I think, except for the basic confusion about the direction of time. We can easily imagine adding to this to allow for things like recombination maps.

Here's an example input file:

```yaml
demography:
  # This is an **embedded** Demes yaml model.
  time_units: generations
  demes:
    - name: X
      epochs: [{end_time: 1000, start_size: 2000}]
    - name: A
      ancestors: [X]
      epochs: [{start_size: 2000}]
    - name: B
      ancestors: [X]
      epochs: [{start_size: 2000}]

# Note: We are **referring** to the Demes model here.
samples: {A: 100, B: 100}
sequence_length: 100000
recombination_rate: 1e-8
ploidy: 1
model: hudson
```

The idea is that we *embed* the Demes yaml description within the larger simulation configuration context. When we're parsing the input yaml, we just hand-off the parsing of the ``demography`` object to ``demes-python`` which will do all the hard work for us. 

I'm not suggesting this as a general specification for popgen simulations, I just want to illustrate the power that we get from keeping making Demes simple and self-contained.  To me, the ability to make a simple configuration file for a specific simulator like this is a powerful argument for not over-specifying the standard.

Any thoughts @molpopgen @grahamgower @apragsdale? I've been talking about simulation configurations being able to "refer" to elements of the Demes model for a while, and this is an attempt to make things concrete. (I guess we shouldn't get into detailed discussions about Demes itself here though: if someone wants to follow up, maybe create an issue on the spec repo to discuss?)

### DIFF
--- a/msprime/json_input.py
+++ b/msprime/json_input.py
@@ -1,0 +1,51 @@
+#
+# Copyright (C) 2021 University of Oxford
+#
+# This file is part of msprime.
+#
+# msprime is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# msprime is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with msprime.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Define formats used for simulation input as JSON and related formats.
+"""
+from __future__ import annotations
+
+import copy
+import json
+
+import demes
+from ruamel.yaml import YAML
+
+import msprime
+
+
+def parse_ancestry_json(data):
+    data = copy.deepcopy(data)
+    if "start_time" in data or "end_time" in data:
+        raise ValueError(
+            "specifying time values not currently supported as too confusing"
+        )
+    if "demography" in data:
+        demes_dict = data["demography"]
+        # TODO nasty going back to JSON here - can we make a demes.fromdict()
+        # function to do this directly?
+        demes_model = demes.loads(json.dumps(demes_dict), format="json")
+        data["demography"] = msprime.Demography.from_demes(demes_model)
+    return data
+
+
+def parse_ancestry_yaml(text):
+
+    yaml = YAML(typ="safe")
+    return parse_ancestry_json(yaml.load(text))
--- a/msprime/cli.py
+++ b/msprime/cli.py
@@ -35,6 +35,7 @@ import tskit
 
 import msprime
 from . import ancestry
+from . import json_input
 from . import mutations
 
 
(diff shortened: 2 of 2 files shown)


petrelharp:
Two thoughts:
- it seems like this could be a nice bridge for the folks who aren't comfortable in python? It might be worth finding some of those people to test it out on.
- Perhaps this all should be within an `ancestry: ` block, to be followed by a `mutations:` block (and then maybe an `output:` block?) for a more complete specificatoin?

msprime/json_input.py
```diff
@@ -0,0 +1,51 @@
...
+    data = copy.deepcopy(data)
+    if "start_time" in data or "end_time" in data:
+        raise ValueError(
+            "specifying time values not currently supported as too confusing"
+        )
+    if "demography" in data:
+        demes_dict = data["demography"]
+        # TODO nasty going back to JSON here - can we make a demes.fromdict()
+        # function to do this directly?
+        demes_model = demes.loads(json.dumps(demes_dict), format="json")
```
grahamgower, line 43: [`demes.Graph.fromdict()`](https://popsim-consortium.github.io/demes-docs/main/api.html#demes.Graph.fromdict)?

grahamgower:
I agree with @petrelharp that it maybe needs to have separate `ancestry:` and `mutations:` blocks. But then it doesn't neatly align with the current CLI `msp ancestry` subcommand. Also, maybe the demography could be either inline or refer to a file path?

msprime/json_input.py
jeromekelleher: Aha! Thanks @grahamgower.

jeromekelleher:
Thanks, great points @petrelharp and @grahamgower ! I think a combined ``ancestry`` and ``mutation`` format is the right approach, and yes, this would be a good bridge for people who aren't comfortable with Python.

WRT to the CLI, I've already created an ``msp ancestry-yaml`` as a quick way of getting something working without having to worry about the semantics of ``msp ancestry``. So, we just need a command to run a simulation from a yaml config. Unfortunately ``msp simulate`` is already used as the legacy interface. We could do ``msp yaml``?

jeromekelleher:
Update: I've added the proposed mutations/ancestry sections and the config looks like this now:
```yaml
ancestry:
  sequence_length: 100000
  recombination_rate: 1e-8
  samples: {A: 100, B: 100}
  ploidy: 1
  model: hudson
  demography:
    time_units: generations
    demes:
      - name: X
        epochs: [{end_time: 1000, start_size: 2000}]
      - name: A
        ancestors: [X]
        epochs: [{start_size: 2000}]
      - name: B
        ancestors: [X]
        epochs: [{start_size: 2000}]

mutations:
  rate: 1e-8
  model: blosum62
```

To make this fully general we'd need to 

1. Add support for reading RateMaps from dictionaries (easy)
2. Support parsing Ancestry and Mutation models from dictionaries (should be pretty easy, this is basically what we turn the classes into anyway). Since the ancestry models use a duration, we actually sidestep the awkward time business
3. Think properly about time and implement start_time and end_time accordingly (but, these are pretty niche options, so could just be dropped)

apragsdale:
This looks really nice to me. Agree that ancestry/mutations/output blocks makes a lot of sense, and those updates look clean. If I'm reading the changes correctly, you can place any valid argument to sim_ancestry and sim_mutations into this yaml? So specify seeds, or more complicated models (e.g. dtfw then switch to hudson), etc. For an "output" block, it might be nice to be able to specify "trees" vs "vcf", plus all the bells and whistles that go with those. Not sure how general you intend this input approach to be.

Overall, I think this would be a nice middle ground between avoiding both python scripting and the cli (which can sometimes be confusing for some). Looking forward to discussing more today in a bit.

molpopgen:
I like the approach overall.  I think embedding the `demes` bits is quite elegant.

PR #1846, opened 2021-09-20
Refactor and simplify simulation through pedigrees

jeromekelleher:


jeromekelleher:
Making progress here - but there's a memory crash on OSX, so need to improve the C tests.

jeromekelleher:
Notes from meeting with @sgravel, and others:

- It would make more sense to call the variable that we accumulate ancestral material on an individual ``descendents``, since that is what they are (the descendants that happen to to be carrying ancestral material)
- The way to deal with stuff popping out of the pedigree is to run two simulations: first the "wf_ped" simulation, and then the more complex population model, using the output tree sequence as it's ``initial_state``. Once we set the population IDs correctly on the input pedigree tree sequence, it should all work itself out.
- It would be simpler to think about this in terms of the male and female parents. We should probably add a ``sex`` field to the individuals, and be concrete about this.

PR #1904, opened 2021-11-11
Do not use inspect to generate provenance.

### LINKED ISSUE #1899: memory leak in sim_ancestry()
I'm seeing a fairly rapid increase in memory use over time with the following test case.
```python
import msprime

print(msprime.__version__)
for seed in range(1, 10_001):
    msprime.sim_ancestry(samples=64, random_seed=seed)
```

I ran it under [memory-profiler](https://pypi.org/project/memory-profiler/):
```
$ time mprof run sim-sim-sim.py 
mprof: Sampling memory every 0.1s
running new process
running as a Python program...
1.0.2

real    2m0.464s
user    1m53.911s
sys     0m5.716s
```

```
$ mprof plot -o leaky.png mprofile_20211110210243.dat
```
And see this
leaky

jeromekelleher:
Closes #1899

Turns out to be much easier to just remove the calls to inspect and put in some simple dictionary code. I changed around some tests for sim_mutations, because there didn't seem to be much reason to make the semantics different to sim_ancestry (which stores almost all parameters as passed in).

@benjeffery, does this look OK to you?

Confirming this fixes the memory leak:

fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [1.0.3] - 2021-XX-XX
+## [1.0.3] - 2021-11-12
 
 **New features**:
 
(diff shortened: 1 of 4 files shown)


benjeffery:
Looks good to me, sorry this slipped through last week

Issue #1857, opened 2021-10-06
Add parse_ped function 


jeromekelleher:
This should take a pedigree file as input and return a tskit TableCollection ready for simulating with. This will be the primary interface for people setting up the input to running a pedigree simulation, so we need to take some care about how we think about populations, etc, and whether we do things like supporting inserting additional samples.

This can be a more specialised version of the code that we added to tskit (probably we can just move the function over as a starting point).

jeromekelleher:
Closed in #1866 

Not clear that there's enough information in the FAM file format for us to actually build a simulation though.

PR #1913, opened 2021-11-14
PedigreeBuilder API

jeromekelleher:
One of the major missing pieces for the fixed pedigree simulations is to decide on the user interface: how does the user input their pedigree, along with the information required for msprime to run the simulations? For the first iteration of the fixed pedigree simulations we need to know (a) the time of each individual; (b) its population (if this is ultimately a multi-population simulation); (c) whether this individual is a sample.

One option (considered in #1857) is to use an existing pedigree file format like the plink FAM file. However, FAM files are quite limited (fixed columns of FID, IID, PAT, MAT, SEX, PHENOTYPE). We could add some extra columns to the FAM file format, but then it becomes confusing: what do we do with the phenotype column? How do we validate the "POPULATION" column? 

Another option is to create our own text file format, say something like this:
```
# is_sample parent0 parent1 time    population  metadata
0   -1  -1  1.0 0   {"a": "b"}
0   -1  -1  1.0 0   {"a": "b"}
1   0   1   0.0 0   {"a": "b"}
```
This is a direct mapping into the tskit/msprime model of the pedigree (zero based IDs derived from the row number), with all the information we need and arbitrary JSON metadata associated with each individual. The problem with this approach is that users will have to write some code to parse the pedigree information they have in whatever format it's in, and produce the tskit model - this will require detailed documentation and explanation of what's required *and* it's a non-trivial piece of work to write the parser for this format (having the JSON metadata breaks np.loadtxt, e.g.). Also, the format is quite rigid, and would make it difficult to add a more columns in the future (e.g., ``sex``).

So, I suggest a middle ground - provide an API that will help users build an msprime pedigree model, and provide a framework for doing things the filling in missing times etc later. This is a first draft of what that API might look like.

Here's some example usage:
```python
demography = msprime.Demography.isolated_model([100])
builder = msprime.PedigreeBuilder(demography)
p0 = builder.add_individual(time=1, metadata={"whaever":"you want to store"})
p1 = builder.add_individual(time=1)
child = builder.add_individual(time=0,  parents=[p0, p1])
pedigree = builder.finalise(sequence_length=10)
```
First, we *require* a demography to be specified. This allows us to validate population ID/names (in a multi-population model) but also makes it clear that you *must* specify the demography. This isn't so much for the fixed pedigree simulation, but to define what happens *outside* the fixed pedigree. Requiring the demography to be specified up front is less obviously necessary here, but we'll see an example in a minute where it's useful. Then, we run it like this:

```python
ped_ts = msprime.sim_ancestry(
    initial_state=pedigree, demography=demography, random_seed=1, model="wf_ped"
)       
print(ped_ts.draw_text())    
recap_ts = msprime.sim_ancestry(  
    initial_state=ped_ts, demography=demography, random_seed=1
)
print(recap_ts.draw_text())
```
giving
```
1.00┊ 0 3 ┊                                                                                    
    ┊ ┃ ┃ ┊                                                                                    
0.00┊ 4 5 ┊                                                                                    
    0    10                                                                                    
                                                                                               
108.92┊  6  ┊                                  
      ┊ ┏┻┓ ┊                                  
1.00  ┊ 0 3 ┊                                  
      ┊ ┃ ┃ ┊                                  
0.00  ┊ 4 5 ┊                                  
      0    10                                  
```

Because we're specifying the ``initial_state`` in both cases, the ``demography`` argument is required. So, it's a good idea to make it a requirement through the whole process. (Recapitating like this might not be the best idea, but that's another question.)

## Population structure

Suppose we want to specify a more complex demography outside the pedigree, and state that various individuals come from different populations. We do this like:
```python
demography = msprime.Demography()
demography.add_population(name="A", initial_size=100)
demography.add_population(name="B", initial_size=100)
demography.add_population(name="C", initial_size=100)
demography.add_population_split(time=10, derived=["A", "B"], ancestral="C")

builder = msprime.PedigreeBuilder(demography)
p0 = builder.add_individual(time=1, population="A")
p1 = builder.add_individual(time=1, population="B")
child = builder.add_individual(time=0, is_sample=True, parents=[p0, p1], population="A")
pedigree = builder.finalise(sequence_length=10)
```
Note that we're using the population names here to make things less error prone, and that we can validate the populations at ``add_individual`` time so that users get better error messages. 

## Conclusion

I'm hoping that this API will help bridge the gap between the pedigree data that people have (which will be in a range of different formats) and the data model needed by msprime. To me, it's less work for users to learn and use this API that it would be to generate the custom text format that would be required, and it lets us provide error messages early.

Any thoughts @sgravel, @hyanwong @apragsdale @LukeAndersonTrocme ?

### DIFF
--- a/msprime/__init__.py
+++ b/msprime/__init__.py
@@ -56,6 +56,7 @@ from msprime.demography import (
     Population,
 )
 
+from msprime.pedigrees import PedigreeBuilder
 from msprime.intervals import RateMap
 from msprime.likelihood import log_arg_likelihood, log_mutation_likelihood
 
@@ -130,6 +131,7 @@ __all__ = [
     "NODE_IS_GC_EVENT",
     "NUCLEOTIDES",
     "PAM",
+    "PedigreeBuilder",
     "Population",
     "PopulationConfiguration",
     "PopulationParametersChange",
--- a/msprime/pedigrees.py
+++ b/msprime/pedigrees.py
@@ -26,6 +26,64 @@ import os
 import numpy as np
 import tskit
 
+from msprime import demography as demog_mod
+
+
+class PedigreeBuilder:
+    def __init__(self, demography=None):
+        if demography is None:
+            demography = demog_mod.Demography.isolated_model([1])
+        self.demography = demography
+        self.tables = tskit.TableCollection(0)
+        demography.insert_populations(self.tables)
+
+        assert len(self.tables.individuals) == 0
+        self.tables.individuals.metadata_schema = tskit.MetadataSchema(
+            {
+                "codec": "json",
+                "type": "object",
+                "properties": {},
+                "additionalProperties": True,
+            }
+        )
+
+    def add_individual(
+        self, *, time, is_sample=None, parents=None, population=None, metadata=None
+    ):
+        if is_sample is None:
+            # By default, individuals at time 0 are marked as samples
+            is_sample = time == 0
+
+        if parents is None:
+            parents = [-1, -1]
+        if len(parents) != 2:
+            raise ValueError("Must have exactly two parents")
+
+        ind_id = self.tables.individuals.add_row(parents=parents, metadata=metadata)
+        if population is None:
+            if self.demography.num_populations == 1:
+                population = 0
+            else:
+                raise ValueError(
+                    "Must specify the population for each individual in a "
+                    "multi-population demography"
+                )
+        population = self.demography[population].id
+        flags = tskit.NODE_IS_SAMPLE if is_sample else 0
+        for _ in range(2):
+            self.tables.nodes.add_row(
+                flags=flags, time=time, individual=ind_id, population=population
+            )
+        return ind_id
+
+    def finalise(self, sequence_length):
+        copy = self.tables.copy()
+        copy.sequence_length = sequence_length
+        # Not strictly necessary, but it's useful when the tables are written
+        # to file and are loaded as as a tree sequence.
+        copy.build_index()
+        return copy
+
 
 def sim_pedigree(
     *,
@@ -39,22 +97,10 @@ def sim_pedigree(
     # part of the public API and subject to arbitrary changes/removal
     # in the future.
 
-    tables = tskit.TableCollection(
-        sequence_length=-1 if sequence_length is None else sequence_length
-    )
-    # TODO we should be using the demography class here to set up the
-    # population metadata
-    permissive_json = tskit.MetadataSchema(
-        {
-            "codec": "json",
-            "type": "object",
-            "properties": {},
-            "additionalProperties": True,
-        }
-    )
-    tables.populations.metadata_schema = permissive_json
-    # tables.individuals.metadata_schema = permissive_json
-    tables.populations.add_row()
+    # We use the builder to get a standard set of tables, but bypass
+    # it so that we can use numpy methods add to it more efficiently.
+    tables = PedigreeBuilder().tables
+    tables.sequence_length = -1 if sequence_length is None else sequence_length
 
     population = np.array([tskit.NULL], dtype=np.int32)
     rng = np.random.RandomState(random_seed)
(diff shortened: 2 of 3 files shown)


LukeAndersonTrocme:
Just putting my two cents. 

As a novice python user I can confirm that formatting the json metadata was not very straight forward when loading data into tskit. However it was not impossibly difficult, with a bit of googling I was able to figure it out. Though I agree with Jerome's comments about documentation, and also agree that this API middle ground makes sense. From what is described above this API seems pretty straight forward to learn how to use. 

Just a thought about standardized input files, I think most people working with genealogical data would be able to generate a flat text file (i.e. no JSON metadata) with specified column names (leaving the parsing up to the user). I would imagine that some users would find it easier to generate such a file than to learn to use this API. (although this doesn't address the issue of validating populations). 

I guess the question I would ask is: would it be useful/practical to have a "quick start" simple model to get users going with the software before they have to learn the details of more complex out-of-pedigree demography models?

From my experience with the French-Canadian genealogy (which may not be the most representative user case), it seems like there's less of a need to include complex models beyond the pedigree because of how extensive this pedigree is. In this case, I would imagine that a vanilla model would do just fine as a starting point. And if this is the case, then a ped file with a few extra columns specifying generation time, and is_sample would be enough to get a novice user to get started with "out of the box" settings.

apragsdale:
I think this is a great middle ground, and agree with @LukeAndersonTrocme that it is a bit daunting at first to figure out the metadata formatting. This does help quite a bit with that.

I also agree that there are too many corner cases to reasonably expect to come up with a new file format that covers everything someone might want (plus it's a lot of added work, as mentioned). I like the approach of this API - can be used directly to build a pedigree on the fly, or wrapped pretty readily into a parser for an existing text file format. The PedigreeBuilder looks good to me.

jeromekelleher:
Thanks for the input @apragsdale and @LukeAndersonTrocme this is really helpful. I think having the PedigreeBuilder is a good building block, so I'll wrap this up and commit.

Regarding text files:

> Edit: maybe a function to parse the standard FAM file would be useful to include. Covers a lot of cases, and would give an example of how to write a parser using the API?

The trouble with the FAM file is that it doesn't have the information we need. We have to have a time field, as a minimum. OK, so big deal, we add a column to the text file right? It's not that simple because standard FAM files don't have a header, so it's really not going to be obvious what this last column means, and it'll be really easy to mix it up with existing phenotype values.

> And if this is the case, then a ped file with a few extra columns specifying generation time, and is_sample would be enough to get a novice user to get started with "out of the box" settings.

@LukeAndersonTrocme, can you point to some examples of ped files that you're thinking of? I've made a new issue #1915 to discuss this, so let's move discussion of text files over there. I agree that a text file input would be handy for many users (and for making examples etc) so good to get your thoughts here.

hyanwong:
> As a novice python user I can confirm that formatting the json metadata was not very straight forward when loading data into tskit. 

Just on this point (but slightly off topic for this thread) I'm in the middle of revising the "editing a tree sequence" part of the tutorials and will give examples of editing the metadata in a tree sequence, which is a lot easier in tskit 0.4.0. I'll ping @LukeAndersonTrocme for feedback when I PR it.

hyanwong:
> I suggest a middle ground - provide an API that will help users build an msprime pedigree model, and provide a framework for doing things the filling in missing times etc later. 

This SGTM. We can always provide wrapper code for reading in other formats, e.g. in the tsconvert repo. I note that [tsconvert](https://github.com/tskit-dev/tsconvert) mentions the PLINK pedigree format too.

I think that what would help a lot is writing the documentation, plus a tutorial, along with some basic visualisations (e.g. https://github.com/tskit-dev/tskit/issues/621 and I thought that @jeromekelleher posted a nice networkx example in ?slack recently). We also want to make sure that we make clear the intent of the fields etc (see https://github.com/tskit-dev/tskit/discussions/1582).

apragsdale:
> The trouble with the FAM file is that it doesn't have the information we need.

Right of course, I think I was confusing my file formats.

This looks good to me! Thanks Jerome

jeromekelleher:
OK, going to merge this much. I've added an issue to track documenting the interface (which I want to work with a bit before documenting)

Issue #1865, opened 2021-10-07
Recapitating from an incomplete pedigree


jeromekelleher:
Consider the following code:
```python
import msprime
import msprime.pedigrees

N = 5
tables = msprime.pedigrees.sim_pedigree(
    population_size=N, random_seed=1234, end_time=10
)

tables.sequence_length = 100
ped_ts = msprime.sim_ancestry(
    initial_state=tables,
    model="wf_ped",
    random_seed=12345,
)
final_ts = msprime.sim_ancestry(
    initial_state=ped_ts, population_size=N, model="dtwf", random_seed=123456
)

print(final_ts.draw_text())
```
The trees we get are:

```
41.00┊                              110        ┊   
     ┊                        ┏━━━━━━┻━━━━━━┓  ┊   
10.00┊                        8             9  ┊   
     ┊                        ┃             ┃  ┊   
9.00 ┊                        ┃             ┃  ┊   
     ┊                        ┃             ┃  ┊   
8.00 ┊                        ┃             ┃  ┊   
     ┊                        ┃             ┃  ┊   
7.00 ┊                       37             ┃  ┊   
     ┊                ┏━━━━━━━┻━━━━━━━┓     ┃  ┊   
6.00 ┊                ┃               ┃     ┃  ┊   
     ┊                ┃               ┃     ┃  ┊   
5.00 ┊                ┃              50     ┃  ┊   
     ┊                ┃             ┏━┻━┓   ┃  ┊   
4.00 ┊                ┃             ┃   ┃   ┃  ┊   
     ┊                ┃             ┃   ┃   ┃  ┊   
3.00 ┊                ┃             ┃   ┃   ┃  ┊   
     ┊                ┃             ┃   ┃   ┃  ┊   
2.00 ┊               83             ┃   ┃   ┃  ┊   
     ┊      ┏━━━━━━━┳━┻━━━┳━━━━━┓   ┃   ┃   ┃  ┊   
1.00 ┊     93       ┃    98     ┃   ┃   ┃   ┃  ┊   
     ┊  ┏━━━╋━━━┓   ┃   ┏━┻━┓   ┃   ┃   ┃   ┃  ┊   
0.00 ┊ 100 107 108 101 102 104 109 105 106 103 ┊   
   0.00                                     100.00 
```

[...]

Issue #1863, opened 2021-10-07
Change the model name for "wf_ped"


jeromekelleher:
After taking some folks on a tour of the simulate-through-pedigree API there was universal agreement that the name "wf_ped" is unhelpful and confusing. The "Wright-Fisher" bit got in there because under the hood the details of how recombination happens is handled by the same code as the DTWF model, but actually this doesn't really have anything to do with a Wright-Fisher process.

So, any suggestions? Here are the [existing models](https://tskit.dev/msprime/docs/latest/ancestry.html#sec-ancestry-quickref) for reference.

The simplest and most obvious is to call the ancestry model class ``Pedigree()`` with a shorthand string "pedigree", but I wonder if this will confuse people into thinking that we are simulating a pedigree rather than simulating *within* a pedigree? Perhaps this confusion would be alleviated if we documented the ``sim_pedigree()`` function as part of the public API?

hyanwong:
`FixedPedigree` perhaps?

jeromekelleher:
I like it!

jeromekelleher:
``FixedPedigree`` sounds right for the class name, what should we call the short-cut string? "fixed_pedigree" is the obvious choice I guess?

hyanwong:
> `FixedPedigree` sounds right for the class name, what should we call the short-cut string? "fixed_pedigree" is the obvious choice I guess?

Yes, make it as obvious as possible I reckon. We could also use "KnownPedigee", but I think "Fixed" is better.

PR #1914, opened 2021-11-14
Implement multiple populations in ped sim

jeromekelleher:
Adds basic support for simulating through a fixed pedigree with multiple populations.

Each individual is assigned a population, and then we "migrate" the chunks of ancestry that come from different populations to this parent's population, if necessary.

@apragsdale, any chance you could take a look, or maybe try this out and see if the results make sense to you?

cc @sgravel @LukeAndersonTrocme

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1470,7 +1470,8 @@ msp_print_individual(msp_t *self, individual_t *ind, FILE *out)
 {
     size_t j;
 
-    fprintf(out, "\tID: %d - Time: %f, Parents: [", ind->id, ind->time);
+    fprintf(out, "\tID: %d; time: %f, population: %d, Parents: [", ind->id, ind->time,
+        ind->population);
     for (j = 0; j < self->ploidy; j++) {
         fprintf(out, " %d", ind->parents[j]);
     }
@@ -2203,6 +2204,7 @@ msp_pedigree_initialise(msp_t *self)
     segment_t *segment;
     avl_node_t *a;
     label_id_t label = 0;
+    tsk_size_t j;
 
     if (self->next_demographic_event != NULL || self->store_full_arg) {
         ret = MSP_ERR_UNSUPPORTED_OPERATION;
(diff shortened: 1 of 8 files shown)


apragsdale:
I'll have a close look and will give it a spin, hopefully today or tomorrow.

apragsdale:
I ran into an issue when there are demographic events in the demography, such as a population split. I opened a PR to your branch, which I hope worked.

Adding a `demography.add_population_split(time=10, derived=[0, 1], ancestral=2)`, I get `msprime._msprime.LibraryError: Current simulation configuration is not supported.` Haven't had a chance to look into cause of it yet.

jeromekelleher:
I've made a quick change to make the pedigree simulation work with a demographic model. However, having done it I wonder if it's such a good idea. Here's the example:
```python
        demography = msprime.Demography.isolated_model([10, 10, 10])
        demography.add_population_split(time=10, derived=[0, 1], ancestral=2)
        ts = msprime.sim_ancestry(
            initial_state=tables, demography=demography, model="wf_ped", random_seed=1
        )       
        recap_ts = msprime.sim_ancestry(initial_state=ts, demography=demography)
```

We provide the demographic model to ``sim_ancestry`` when running the fixed pedigree simulation and we implement those demographic events as they happen. However, they don't fundamentally *do* anything - we can change population sizes, move lineages around etc etc, but it doesn't affect the simulation within the fixed pedigree in any way. So, it's a bit misleading to require it. 

Requiring the input demography makes sense in two cases:
1. We continue the simulation under some other model after the fixed pedigree has been fully traversed (but I think this is a bad idea because it handles things that leave the pedigree early poorly - they are basically just lineages hanging around in stasis until we reach the top of the pedigree). We currently don't allow this.
2. We somehow synchronously ran another model in parallel with the fixed pedigree, and reasoned about e.g. how many lineages  were currently in the pedigree and adapted the population sizes accordingly (which would be hard, but ultimately the only correct way of doing things I think).  

Otherwise, if the entire simulation is determined by the fixed pedigree, then there's really no point in providing a demography at all, and really it's just misleading and confusing to do so. This is just a case of turning off some input parameter checks, which is easy to do. I guess what we have to worry about then is what happens if we allow for either 1 or 2 above in the future. I don't think 1 will ever be a good idea unless we do 2 as well, and once 2 is a distinguishable case from that we're currently doing, then that's OK. If we *did* run a second model in parallel with the fixed pedigree (we're renaming the model, see #1863), we'd have to specify what that model is and so we'd have a way of detecting the condition. We could have a notation like

```python
ts = msprime.sim_ancestry(intial_state=pedigree, model=FixedPedigree() + DiscreteTimeWrightFisher())
```
to indicate this compound model. *Then* we'd require a demography to define the behaviour of the parallel model. We'd still want to be able to simulate through *just* the fixed pedigree sometimes, and in that case the demography would be meaningless, as now.

OK, I think I've convinced myself that the right thing to do is required that ``demography`` isn't specified when we run a ``FixedPedigree`` simulation. Any recapitation simulations will require a demography.

What do you think @apragsdale?

Thanks @apragsdale, this was super helpful and helped to clarify an important point!

sgravel:
I agree with what you said Jerome. Restating in my own words:  I think we should not specify a demography when simulating in the genealogy for now, but only make sure that the genetic material of each each founder genomic material is tagged with a time and population, with the idea that these will be handled by recapitation later on. 

As I understand things, this is similar to Jerome's case 1, but the difference is that dangling lineages are handled after recapitation rather than during the simulation of genealogical transmissions.

sgravel:
After reading issue #1913, I guess what remains to be determined is how we ensure that the population IDs are correctly validated if we don't specify a demography. Perhaps a middle ground would be to have a compulsory "source_populations" tuple argument when using model="wf_ped". It would make the expected behaviour of the simulation much more explicit, would make it possible to check on the fly whether provided founder populations make sense, and it would also allow for checking in a more streamlined way when performing recapitation with a demography.

apragsdale:
The demography seems most useful to make sure that the individuals are labeled with a valid population. If you intend to recapitate or expect lineages to conform to a given demography, even though the demography is not used for the pedigree simulation, it might be important to be able to raise an error if a node is somewhere that conflicts with the demography?

Perhaps the demography can be optional, and then if the demography is not given, it is assumed that all nodes are placed in population 0 and anything else raises an error?

jeromekelleher:
Thanks for the input @sgravel and @apragsdale, I think we're all on the same page. I'll update this with my proposal ASAP.

jeromekelleher:
I've made some updates here which basically shuts down the options that you can supply with the fixed pedigree simulation, and I'm feeling pretty positive that we're on the right track now. I'm going to merge and push through a few more updates, and I think we're ready to start documenting this and considering the higher-level issues like #1865 (which I think we will just need to document as being an approximation).

Issue #1930, opened 2021-11-23
Update `from_demes()` once demes-0.1.3 is released


apragsdale:
Demes 0.1.3 will introduce some breaking changes (e.g. pulses will now specify sources and proportions as lists, instead of source and proportions as single values). We're quite close there to hitting all the milestones to get 0.1.3 out the door (https://github.com/popsim-consortium/demes-python/milestone/3), so this issue is just to make sure we are on top of it. (The requirements also has "demes>=0.1.2", so demes-0.1.3 will end up not working with the current msprime version if the user has demes-0.1.3 installed.)

The changes in msprime should only require updating `Demography.from_demes()`, specifically a few simple changes in lines here: https://github.com/tskit-dev/msprime/blob/7fbaab3b031f7208cd82b79763e924260f412723/msprime/demography.py#L1984

jeromekelleher:
Sounds good @apragsdale. I'm currently hoping to get msprime 1.1 out in a few weeks (conditioned on tskit 0.4.0 and getting the pedigree stuff wrapped up). I'll put this into the 1.1 milestone for now, so we don't forget about it (in case the demes release is out in time).

apragsdale:
Great! Thanks @jeromekelleher. I think it's quite doable, as it's just two "small" issues over in Demes to resolve. I'll ping @grahamgower and see if we can make it in time for msprime-1.1.

grahamgower:
Thanks @apragsdale for opening this. I think we should make the next demes-python release version 0.2.0, because of this breaking change. msprime 1.0.x already requires demes<0.2 in the setup.cfg, so the next demes release shouldn't cause big problems (fingers crossed).

jeromekelleher:
Does this need doing now that version 0.2.0 of demes is out @apragsdale? We can push out a quick point release here with any updates?

Looks like we should just be able to wait though?

apragsdale:
I have already started on those updates to msprime for demes 0.2 and been using them locally, so it will be quick to get a PR together for you two to have a look at. If 1.1 is close, I don't know if it's so important to rush out a point release, but it could be nice to have if 1.1 is a bit further off

jeromekelleher:
Cool, let's see where we are after the PR goes in. Probably simplest to pop out a quick point release, it's no big deal.

jeromekelleher:
Closed in #1936

PR #1936, opened 2021-12-01
Handle pulses in demes 0.2.0

apragsdale:
See #1930.

Here, pulses with multiple sources are parsed into individual mass migration events. Here, I haven't worried about trying to collapse simultaneous pulses back into a single event with multiple sources when calling `to_demes()`, since demes also maintains the convention of apply pulses that occur at the same time in the order that they are specified.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -1982,12 +1982,14 @@ class Demography(collections.abc.Mapping):
         # Add pulses in reverse order, so that pulses with the same time
         # correspond to the correct backwards-time mass migration ordering.
         for pulse in reversed(events.pop("pulses")):
-            demography.add_mass_migration(
-                time=pulse.time,
-                source=pulse.dest,
-                dest=pulse.source,
-                proportion=pulse.proportion,
-            )
+            sequential_props = _proportions_to_sequential(pulse.proportions)
+            for prop, source in zip(sequential_props, pulse.sources):
+                demography.add_mass_migration(
+                    time=pulse.time,
+                    source=pulse.dest,
+                    dest=source,
+                    proportion=prop,
+                )
         for merger in events.pop("mergers"):
             demography.add_admixture(
                 time=merger.time,
@@ -2710,9 +2712,9 @@ class Demography(collections.abc.Mapping):
             for lm in lineage_movements:
                 if (lm.source, lm.dest, lm.proportion) in pulses:
                     b.add_pulse(
-                        source=resolved[lm.dest].name,
+                        sources=[resolved[lm.dest].name],
                         dest=resolved[lm.source].name,
-                        proportion=lm.proportion,
+                        proportions=[lm.proportion],
                         time=time,
                     )
 
--- a/tests/test_demes.py
+++ b/tests/test_demes.py
@@ -286,14 +286,14 @@ class TestFromYamlExamples:
             epochs:
               - start_size: 2000
         pulses:
-          - source: X
+          - sources: [X]
             dest: A
             time: 500
-            proportion: 0.1
-          - source: A
+            proportions: [0.1]
+          - sources: [A]
             dest: X
             time: 100
-            proportion: 0.2
+            proportions: [0.2]
         """
         d = self.from_yaml(yaml)
         assert d.num_populations == 2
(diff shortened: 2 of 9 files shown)


grahamgower:
> I haven't worried about trying to collapse simultaneous pulses back into a single event with multiple sources when calling `to_demes()`

Do we have code in demes to "simplify" this case? (if not, we should open a ticket)

apragsdale:
> > I haven't worried about trying to collapse simultaneous pulses back into a single event with multiple sources when calling `to_demes()`
> 
> Do we have code in demes to "simplify" this case? (if not, we should open a ticket)

We don't, I don't think. But demes would be the place to implement that instead of here, I imagine. It could find use beyond just the msprime conversion.

grahamgower:
> > > I haven't worried about trying to collapse simultaneous pulses back into a single event with multiple sources when calling `to_demes()`
> > 
> > 
> > Do we have code in demes to "simplify" this case? (if not, we should open a ticket)
> 
> We don't, I don't think. But demes would be the place to implement that instead of here, I imagine. It could find use beyond just the msprime conversion.

Yes, that's what I meant sorry.

msprime/demography.py
```diff
@@ -1982,12 +1982,16 @@ def get_growth_rate(epoch):
...
+            total_ancestry_moved = 0
+            for prop, source in zip(pulse.proportions, pulse.sources):
+                demography.add_mass_migration(
+                    time=pulse.time,
+                    source=pulse.dest,
+                    dest=source,
+                    proportion=prop / (1 - total_ancestry_moved),
+                )
+                total_ancestry_moved += prop
+            assert total_ancestry_moved == sum(pulse.proportions)
```
jeromekelleher, line 1994: We need to either delete the asertion or change to ``math.isclose(total_ancestry_moved, sum(pulse.proportions))``

It seems like something we should just do some unit tests for though. Should be easy to make a few examples and test that the output has the required properties?

jeromekelleher: If we're happy that the unit tests are covering this, we don't need the assertion. And, we should be happy that the unit tests are covering this :smile:

jeromekelleher (approved):
LGTM, one small comment.

Can you update the CHANGELOG also please?

msprime/demography.py
grahamgower: I think you wrote a couple of utility functions for this pattern @jeromekelleher. `_sequential_to_proportions()` and `_proportions_to_sequential()`.

jeromekelleher: Ah yes, better to use those functions then.

apragsdale: Yes, that assert was leftover from testing and I forgot to remove it! I'll also update to use those functions as well - I didn't realize they were already in there

tests/test_demes.py
```diff
@@ -367,6 +367,24 @@ def test_pulses_ordering(self):
...
+    def test_multipulse(self):
+        b = demes.Builder(defaults=dict(epoch=dict(start_size=1)))
+        b.add_deme("a")
+        b.add_deme("b")
+        b.add_deme("c")
+        b.add_pulse(sources=["a", "b"], proportions=[0.1, 0.2], dest="c", time=1)
+        g = b.resolve()
+        d = msprime.Demography.from_demes(g)
+        assert len(d.events) == 2
+        assert d.events[0].__class__ == msprime.MassMigration
```
jeromekelleher, line 379:
```suggestion
        assert isinstance(d.events[0], msprime.MassMigration)
```

jeromekelleher: isinstance is the idiomatic way of doing this, comparing to ``__class__`` can be brittle

tests/test_demes.py
```diff
@@ -1018,6 +1036,44 @@ def test_census_event_is_ignored(self):
         demog.add_census(100)
         demog.to_demes()
 
+    @pytest.mark.filterwarnings(
+        # demes warns about multiple pulses with the same time
+        "ignore:Multiple pulses:UserWarning"
+    )
+    def test_multipulse_roundtrip(self):
```
jeromekelleher, line 1043: This is really two tests, isn't it? Can just call them ``test_multipulse_roundtrip_ex1``, ``test_multipulse_roundtrip_ex2``

tests/test_demes.py
```diff
@@ -367,6 +367,24 @@ def test_pulses_ordering(self):
...
+        b = demes.Builder(defaults=dict(epoch=dict(start_size=1)))
+        b.add_deme("a")
+        b.add_deme("b")
+        b.add_deme("c")
+        b.add_pulse(sources=["a", "b"], proportions=[0.1, 0.2], dest="c", time=1)
+        g = b.resolve()
+        d = msprime.Demography.from_demes(g)
+        assert len(d.events) == 2
+        assert d.events[0].__class__ == msprime.MassMigration
+        assert d.events[0].__class__ == msprime.MassMigration
```
jeromekelleher, line 380: Probably meant ``d.events[1]`` here?

jeromekelleher (approved):
LGTM, couple of nits.

apragsdale:
Thanks for the comments, Jerome - I made each of those changes you suggested and squashed.

jeromekelleher (approved):

PR #1941, opened 2021-12-07
Remove demes/hypothesis code.

grahamgower:
This was more trouble than it was worth and is being removed from demes.


jeromekelleher (approved):
LGTM

PR #1944, opened 2021-12-08
Fix tests for tskit 0.4.0

benjeffery:
Apart from the dict encoding changes, these are the only broken tests be 0.4.0! :partying_face: 

Obviously don't merge now, but thought I'd PR them as I was checking compatibility.


jeromekelleher:
Well, isn't that nice! These were just testing on the broken behaviour anyway, so all is good.

jeromekelleher:
superseded by #1945

PR #1945, opened 2021-12-09
Tskit C 0.99.15/python 0.4.0 and code reorg

jeromekelleher:
Updates tskit to the latest version, and simplifies the repo structure to use the embedded kastore.

We now have two versions of the C API (one through the submodule and the other through meson) so we need to document this somewhere.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
           # we can invalidate the cache when we need to. Note that we need
           # to update the key in the ``save_cache`` block below also when
           # doing that.
-          key: msprime-{{ .Branch }}-v3
+          key: msprime-{{ .Branch }}-v4
       - run:
           name: Checkout submodules
           command: |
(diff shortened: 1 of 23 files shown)


jeromekelleher:
All works fine locally now :tada:

We just have to wait for the conda packages for tskit 0.4.0 to come out.

The tskit tarballs worked great @benjeffery, the meson setup is *way* simpler here now.

benjeffery (approved):
Looks great!

jeromekelleher:
Not ready for merging - we should see whether we want to pop out a quick point release for #1947 before making the jump to tskit 0.4.0.

petrelharp:
Hm - turns out that since the upcoming SLiM release isn't putting individual parents before children, recapitation won't work until this goes in. Right now, recapitate fails with:
```
E       msprime._msprime.InputError: Input error in initialise: Individuals must be provided in an order where children are after their parent individuals
```

I'm happy to have a look at this, but I'm trying to get everything sorted with pyslim first.

PR #1953, opened 2021-12-13
Time units and refseq

jeromekelleher:
Adds support for the new top-level info to msprime. Basics:

- sim_mutations doesn't alter the value. 
- The output of sim_ancestry has time_units = "generations", except when recapitating.

Errors/warnings:
- sim_mutations and sim_ancestry will error if inputs have time_units="uncalibrated". (This seems fine, you have to specifically go out of your way to set the time_units to uncalibrated, and it's just intended for things like tsinfer where the time units really are pretty weird)
- Currently, sim_ancestry issues a warning if time_units != "generations". Maybe this is premature though?

Pinging @petrelharp @molpopgen @hyanwong @grahamgower @bhaller for opinions here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
 # Changelog
 
-**New features**:
+## [1.1.0] - 2021-12-14
 
+**New features**
+
+- Add support for tree sequence ``time_units`` field. The ``time_units`` will
+  be set to "generations" for the output of ``sim_ancestry`` (and ``simulate``),
+  unless the ``initial_state`` argument is used. In this case, the
+  ``time_units`` value will be inherited from the input.
+  ({pr}`1953`, {issue}`1951`, {issue}`1877`, {issue}`1948`, {user}`jeromekelleher`).
 
 **Bug fixes**:
 
@@ -17,6 +24,10 @@
 - Read the population name from PopulationConfiguration ``metadata`` in
   ``Demography.from_old_style`` ({issue}`1950`, {pr}`1954`,
   {user}`jeromekelleher`)
+
+**Maintenance **:
+
+- Update tskit to Python 0.4.0 and C 0.99.15.
 
 ## [1.0.4] - 2021-12-01
 
--- a/msprime/__init__.py
+++ b/msprime/__init__.py
@@ -45,6 +45,7 @@ from msprime.ancestry import (
     StandardCoalescent,
     SweepGenicSelection,
     FixedPedigree,
+    TimeUnitsMismatchWarning,
 )
 
 from msprime.core import __version__
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -30,6 +30,7 @@ import logging
 import math
 import sys
 import tempfile
+import warnings
 from typing import Any
 from typing import ClassVar
 
@@ -44,6 +45,15 @@ from . import provenance
 from msprime import _msprime
 
 logger: logging.Logger = logging.getLogger(__name__)
+
+TIME_UNITS_GENERATIONS = "generations"
+
+
+class TimeUnitsMismatchWarning(UserWarning):
+    """
+    Warning raise when the time units specified in different parts of a
+    simulation do not match.
+    """
 
 
 def _model_factory(model: None | str | AncestryModel) -> AncestryModel:
@@ -195,6 +205,7 @@ def _demography_factory(
 def _build_initial_tables(*, sequence_length, samples, ploidy, demography):
     # NOTE: this is only used in the simulate() codepath.
     tables = tskit.TableCollection(sequence_length)
+    tables.time_units = TIME_UNITS_GENERATIONS
 
     for index, (population, time) in enumerate(samples):
         tables.nodes.add_row(
@@ -804,6 +815,20 @@ def _parse_sim_ancestry(
             raise TypeError(
                 "initial_state must either be a TreeSequence or TableCollection instance"
             )
+        if initial_state.time_units == tskit.TIME_UNITS_UNCALIBRATED:
+            raise ValueError(
+                "Cannot use a tree sequence with uncalibrated time_units as "
+                "the initial state"
+            )
+        if initial_state.time_units != TIME_UNITS_GENERATIONS:
+            message = (
+                f"The initial_state has time_units={initial_state.time_units} but "
+                "time is measured in generations in msprime. This may lead to "
+                "significant discrepancies between the timescales. "
+                "If you wish to suppress this warning, you can use, e.g., "
+                "warnings.simplefilter('ignore', msprime.TimeUnitsMismatchWarning)"
+            )
+            warnings.warn(message, TimeUnitsMismatchWarning)
 
     if sequence_length is None:
         # These are all the cases in which we derive the sequence_length
@@ -940,6 +965,7 @@ def _parse_sim_ancestry(
                 "Either the samples or initial_state arguments must be provided"
             )
         initial_state = tskit.TableCollection(sequence_length)
+        initial_state.time_units = TIME_UNITS_GENERATIONS
         demography.insert_populations(initial_state)
         if not init_for_debugger:
             sample_sets = _parse_samples(samples, demography)
--- a/msprime/mutations.py
+++ b/msprime/mutations.py
@@ -1356,6 +1356,11 @@ def sim_mutations(
     if not isinstance(rate_map, intervals.RateMap):
         raise TypeError("rate must be a float or a RateMap")
 
+    if tables.time_units == tskit.TIME_UNITS_UNCALIBRATED:
+        raise ValueError(
+            "Simulating mutations doesn't make sense when time is uncalibrated"
+        )
+
     start_time = -sys.float_info.max if start_time is None else float(start_time)
     end_time = sys.float_info.max if end_time is None else float(end_time)
     if start_time > end_time:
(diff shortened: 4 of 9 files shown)


grahamgower:
> * Currently, sim_ancestry issues a warning if time_units != "generations". Maybe this is premature though?

SLiM is outputting time_units = "ticks", so this warning will be a bit annoying. But I'm not sure what it means to recapitate if time units aren't generations?

bhaller:
Note that SLiM presently sets `time_units` to `"ticks"`.  At the moment this is just a nod towards models with overlapping generations, where one unit of time really doesn't correspond to anything that one would call a biological "generation".  It also aligns with future plans for supporting multiple species, at which point the time unit in SLiM itself will really be called a "tick" across the board, and each species being modeled will run its "generations" (which are still not really biological generations, but that ship sailed long ago) on a different tick timescale.  I also plan to add the ability for the user to set the time unit's name, so that if one tick is a day or a month or a year, they can state that explicitly; for overlapping generations, this seems like it is what really makes sense.  Setting it to "`generations`" seems like it would be wrong for SLiM, at least for nonWF models.

So, I think it ought to be OK for the user to set whatever time unit they want to set; I think they should "own" this field.  And I'm leery of the idea of "sim_ancestry issues a warning if time_units != "generations"".  Isn't the timescale of recapitation whatever the user says it is, and dependent upon the scale of the recombination/mutation parameters?  Or am I missing the point?  But erroring out on `"uncalibrated"` seems good, probably.

jeromekelleher:
>  Isn't the timescale of recapitation whatever the user says it is, and dependent upon the scale of the recombination/mutation parameters? Or am I missing the point? 

If they are using the population_size/Demography parameters as intended, then yes, the simulation is in units of "generations". They can also set things up in "coalescent" time if they want by doing the scalings themselves (or just working directly in that space for theoretical work) but nearly always the time units should be interpreted as "generations".

jeromekelleher:
We should probably back out of the warning then, since this will create a lot of confusion. @petrelharp, any thoughts here?

bhaller:
> > Isn't the timescale of recapitation whatever the user says it is, and dependent upon the scale of the recombination/mutation parameters? Or am I missing the point?
> 
> If they are using the population_size/Demography parameters as intended, then yes, the simulation is in units of "generations". They can also set things up in "coalescent" time if they want by doing the scalings themselves (or just working directly in that space for theoretical work) but nearly always the time units should be interpreted as "generations".

Hmm.  Well, that seems like a huge mismatch with SLiM, then, which I've never heard anybody discuss before!  If you do a SLiM simulation with overlapping generations, each time unit is not going to represent a biological "generation", and it sounds like maybe that would mean that recapitating such a tree sequence naively would in fact produce incorrect results.  @petrelharp is this on your radar?  Is it discussed anywhere in the SLiM or pyslim documentation?  I'm pretty sure we didn't mention it as an issue in our tree-seq recording paper – happily, the recapitation example presented there used a WF model with non-overlapping generations, so I guess it is not wrong (but it will produce the warning if run under SLiM 3.7 and the proposed msprime, I guess).  The recapitation example shown in the SLiM manual also uses a WF model, and thus also dodges this bullet; but I guess it needs a big red warning label in its discussion, about using recapitation with overlapping generations, which was not on my radar as a problem at all.  Yikes.

If this is really a problem – and it sounds like it is – then actually, having msprime warn because the time unit is not `"generations"` seems appropriate and useful.  I guess it would warn the user that they might, in fact, be doing something wrong and they need to stop and think.  So now I'm leaning towards issuing that warning – and I'm thinking that adding the ability to set the time unit in SLiM, so that users can state that their model ticks are, in fact, "generations" to avoid this warning – is higher priority than I realized.  I think I will do that today.

bhaller:
Maybe the warning can provide some helpful advice, like "If the time unit for this tree sequence really is generations, you can set that by executing '...'; if not, take care that you have rescaled the recapitation parameters as appropriate for your rescaled time unit."  I love helpful warnings like that.

jeromekelleher:
Yep, happy to add more informative warnings @bhaller. 

Let's see what @petrelharp has to say. I'm certain @molpopgen has thought about these issues as well.

bhaller:
OK, the time unit can now be set in `initializeTreeSeq()`.  If not set, it defaults to `"generations"` for WF models, `"ticks"` for nonWF models; this allows WF models to get recapitated without a warning, while nonWF models, as proposed here, ought to warn (but the user can get rid of that warning by passing `timeUnit="generations"` in their nonWF model, if that is in fact the case).  So, with this tweak done, I believe I am in support of this proposed warning, but let's see what others say.

molpopgen:
> I'm certain @molpopgen has thought about these issues as well.

Actually, I haven't.  I'm still in a generation world, and support for overlapping generations in fwdpy11 is only in the embryonic stages.

petrelharp:
>  Hmm. Well, that seems like a huge mismatch with SLiM, then, which I've never heard anybody discuss before!

We had to deal with this (and have discussion of it) in ["Space is the Place"](https://www.genetics.org/content/215/1/193). It's been at the back of my mind to suggest that SLiM make it easy somehow to compute, empirically, generation times (as well as Ne) to make "correct" recapitation easier.

Although throwing a warning would be more the correct thing to do, I think this is a situation where (a) it will horribly confuse everyone; (b) everything is wrong anyways so this small bit of wrongness isn't so terrible, and (c) we don't have a simple solution available yet. Putting a warning in the future when we've got something worked out for how to do it right (e.g., a way to make the warning not happen) would be nice, but I don't think we should put one in at the moment.

petrelharp:
Otherwise, I like this propose.

jeromekelleher:
OK, so is the consensus to put this warning on ice until we have something constructive to tell people (or at least some documentation to point to explaining what they need to do?)

bhaller:
Hmm.  I don't think I agree, actually.  I like the warning.  Recapitating on the assumption that one time unit equals one generation, when in fact one time unit equals, say, one 25th of a generation or something (e.g., for humans), is a pretty big error, right?  And the user can do something about it, if they are aware that the problem even exists at all, by rescaling their recapitation parameters, right?  The warning serves to make the user aware of the problem and guide them to the solution.  Yes, it may confuse people (although I think a helpful warning message could go a long way towards mitigating that), but I don't think just sweeping this problem under the rug is a good idea; it could lead to pretty bad screwups without the user having any idea that anything went wrong.  Until this was brought to my attention by this discussion thread, it would not have occurred to me to worry about the timescale matchup for recapitation; I imagine I am not entirely alone in that level of cluelessness.  :->

bhaller:
> @bhaller what if SLiM put "generations" in for time units in WF models, but "ticks" in for nonWF models? Once we got the warnings out of the way for WF models, I'd be all for it?

Indeed, that's what I did!  :->  See https://github.com/MesserLab/SLiM/issues/260.

bhaller:
> Although: I think that as a matter of principle if we throw a warning there should be some way, programatically, to avoid the warning; and we don't really have one currently, so I guess I'm still in favor of not warning.

Languages generally have a way of suppressing warnings; doesn't Python?  Apart from the use of such a facility, it's commonplace to warn about dangerous situations that might pose a problem, with no other way (apart from the use of that language facility) to suppress the warning, no?  R certainly does this is all kinds of situations.  I think it's very helpful in cases like this where, without a warning, it is very easy for a naive user to go completely off-piste without ever realizing it.

hyanwong:
FWIW I agree with Ben here. Better to warn the user if they are very likely to be doing something wrong (recapitating a non WF model). The other option is to actually error out, but have a parameter that says "I know what I'm doing, don't throw an error", but that's a bit of an annoying thing to have to maintain, right?

petrelharp:
If recapitating WF doesn't throw the warning, I'm ok.

But, if we are on the topic of throwing warnings, how about:
```
UserWarning: You are simulating data from a large population that mates completely at random (Wright-Fisher). Although some results are widely generalizable, it is unlikely that any real species, anywhere in the world, behaves like this. Please take care to assess whether your results are sensitive to this choice of demographic model.
```
(j/k)

jeromekelleher:
OK, let's keep the warning for now so. If it proves to be annoying we can remove it. I'll do my best to put in a helpful warning message.

jeromekelleher:
OK, this is what the message looks like:
```
  /home/jk/work/github/msprime/msprime/ancestry.py:831: TimeUnitsMismatchWarning: The initial_state has time_units=ticks but time is measured in generations in msprime. This may lead to significant discrepancies between the timescales. If you wish to suppress this warning, you can use, e.g., warnings.simplefilter('ignore', msprime.TimeUnitsMismatchWarning)
```

If we're happy with this I'll merge and get the ball rolling on the release.

petrelharp:
Perhaps it should say "time is measured by default in generations"? This is because: what do we want people to do? Right now, the only options are:

1. rescale all the time units in the SLiM simulation so that they're in generations, or
2. rescale all the Ne's (and other units) in msprime so that they're in ticks.

The second thing is what I'd recommend to people in practice; the unfortunate thing about it is that Ne is no longer in units of individuals (it's in units of ticks per coalescence). This *works*, and would mean that time units in the resulting msprime simulation were in ticks, not generations, so it's not quite right to say that time units are in generations. But - I would rather continue to just say that "time units are in generations" if people didn't have to do this sort of thing.

I think the solution to this (for the next msprime?) is to include a `generation_time` argument that scales all the times, and would make this warning go away. That would make the warning still accurate if it says "by default".

bhaller:
I'm fine with either the warning message you used, or what @petrelharp suggests.  I see the logic of his suggestion; but maybe it would make sense to change the warning to that message once his suggested `generation_time` argument is added?  Dunno, fine with either.

PR #1961, opened 2021-12-15
sim_ancestry not simulate

petrelharp:
There were some stray references to `simulate( )` that should have been `sim_ancestry( )`.


jeromekelleher (approved):
Good spot, thanks

PR #1955, opened 2021-12-13
Add warning that soft sweeps are currently not possible.

molpopgen:

PR #1962, opened 2021-12-15
changing this doc string

andrewkern:
editing the warning introduced in #1955 for clarity as to what the issue is

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1915,9 +1915,10 @@ class SweepGenicSelection(ParametricAncestryModel):
     Thus fitness of the heterozygote is intermediate to the
     two homozygotes.
 
-    .. warning:: Currently, models where `start_frequency` is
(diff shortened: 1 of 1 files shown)


petrelharp:
I'm confused - what *does* it mean if you run it back to a high initial frequency, if not a "sweep from standing genetic variation"?

andrewkern:
that you aren't conditioning on allele frequency dynamics back in time, past when the allele hit the final frequency

molpopgen:
> I'm confused - what _does_ it mean if you run it back to a high initial frequency, if not a "sweep from standing genetic variation"?

It means that you've taken a structured coalescent back to some point determined by "initial frequency" and then suddenly unstructure it (b/c there's no merged API that means "finish it w/a neutral trajectory").

jeromekelleher (approved):
LGTM

PR #1978, opened 2022-01-13
Quick hack to add single RE and breakpoints

jeromekelleher:
This adds single recombination nodes and storing of breakpoints to metadata in the algorithms.py script. Completely untested, just an easy way to get some input for other discussions (but I think we should do something like then when updating the real code).

@hyanwong, when you run something like this:
```
python3 algorithms.py 5 -L 100 -d -r 0.01 tmp.trees --full-arg
```
we get 
```
$ tskit nodes tmp.trees
id      is_sample       time    population      individual      metadata
0       1       0.000000        0       -1      {}
1       1       0.000000        0       -1      {}
2       1       0.000000        0       -1      {}
3       1       0.000000        0       -1      {}
4       1       0.000000        0       -1      {}
5       0       0.029150        0       -1      {'breakpoint': 82}
6       0       0.051952        0       -1      {}
7       0       0.129681        0       -1      {}
8       0       0.139396        0       -1      {'breakpoint': 13}
9       0       0.139502        0       -1      {}
10      0       0.161148        0       -1      {}
11      0       0.165444        0       -1      {}
12      0       0.405434        0       -1      {}
```
(annoyingly we put in ``is_sample`` here rather than ``flags`` - but note the ``breakpoint`` in the metadata)
and 
```
$ tskit edges tmp.trees
left    right   parent  child   metadata
0.000000        100.000000      5       1
0.000000        100.000000      6       3
0.000000        100.000000      6       4
0.000000        100.000000      7       0
0.000000        100.000000      7       6
0.000000        100.000000      8       2
82.000000       100.000000      9       5
0.000000        100.000000      9       7
0.000000        82.000000       10      5
13.000000       100.000000      10      8
0.000000        13.000000       11      8
0.000000        100.000000      11      10
0.000000        100.000000      12      9
0.000000        100.000000      12      11
```

This will make it easier to go back and forth between Griffiths ARGs and our edge annotated version.

@JereKoskela does the basic change of just dropping one "add_node" look right to you?

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -741,10 +741,14 @@ class Simulator:
             self.gc_mass_index[u.label].set_value(u.index, 0)
         self.segment_stack.append(u)
 
-    def store_node(self, population, flags=0):
+    def store_node(self, population, flags=0, metadata=None):
+        metadata = {} if metadata is None else metadata
         self.flush_edges()
         return self.tables.nodes.add_row(
-            time=self.t, flags=flags, population=population
+            time=self.t,
+            flags=flags,
+            population=population,
+            metadata=metadata,
         )
 
     def flush_edges(self):
@@ -797,7 +801,10 @@ class Simulator:
                 if node == tskit.NULL:
                     # Add a new node for the current ancestor
                     node = self.tables.nodes.add_row(
-                        flags=0, time=current_time, population=population.id
+                        flags=0,
+                        time=current_time,
+                        population=population.id,
+                        metadata={},
                     )
                 # Add in edges pointing to this ancestor
                 seg = ancestor
@@ -1381,9 +1388,13 @@ class Simulator:
         self.set_segment_mass(alpha)
         self.P[alpha.population].add(alpha, label)
         if self.full_arg:
-            self.store_node(lhs_tail.population, flags=msprime.NODE_IS_RE_EVENT)
+            self.store_node(
+                lhs_tail.population,
+                flags=msprime.NODE_IS_RE_EVENT,
+                metadata={"breakpoint": bp},
+            )
             self.store_arg_edges(lhs_tail)
-            self.store_node(alpha.population, flags=msprime.NODE_IS_RE_EVENT)
+            # self.store_node(alpha.population, flags=msprime.NODE_IS_RE_EVENT)
             self.store_arg_edges(alpha)
         ret = None
         if return_heads:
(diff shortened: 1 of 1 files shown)


JereKoskela:
Dropping one "add_node" is exactly what I've done in the C code in my local repo. I'm hoping to PR it some time next week - the likelihood function get a lot fiddlier as a result and needs some more tiem to get working again.

jeromekelleher:
> the likelihood function get a lot fiddlier as a result and needs some more tiem to get working again.

Dammit - there was always going to be a drawback!

hyanwong:
> Dropping one "add_node" is exactly what I've done in the C code in my local repo. I'm hoping to PR it some time next week - the likelihood function get a lot fiddlier as a result and needs some more tiem to get working again.

Oh, that's a pain! Sorry to hear that. I'd be interested to know why, over coffee some time. Thanks for doing it, though. I think this should make it a lot easier for others to work with our ARGs

JereKoskela:
> > Dropping one "add_node" is exactly what I've done in the C code in my local repo. I'm hoping to PR it some time next week - the likelihood function get a lot fiddlier as a result and needs some more tiem to get working again.
> 
> Oh, that's a pain! Sorry to hear that. I'd be interested to know why, over coffee some time. Thanks for doing it, though. I think this should make it a lot easier for others to work with our ARGs

I'd be happy to elaborate over coffee whenever the chance arises, but essentially it's because finding all edges with a given node as either a child or a parent isn't straightforward. 

Suppose you're looking at a node whose parent is a recombination node. In order to work out the length of the non-ancestral material in which the recombination happens (which may be zero), you have to find the edges which go from the parent to the two grandparents and check their ancestral segments. That results in a fair bit more scrolling back and forth through the edge table than when everything was readable from a node and its two recombinant parents.

The extra complication isn't totally unexpected, and I think a single recombinant node is still the better way to do things. Hopefully there will e.g. be a non-trivial memory saving in full-args.

jeromekelleher:
Could we add some metadata to make this easier @JereKoskela? I'm currently looking at adding the breakpoint anyway, would saving more information about the event help?

JereKoskela:
> Could we add some metadata to make this easier @JereKoskela? I'm currently looking at adding the breakpoint anyway, would saving more information about the event help?

Storing the closest points of ancestral material to the left and to the right of the breakpoint would make things simpler again. That said, I'm pretty much done with this, so probably not worth doing unless there's a more general motivation for it.

JereKoskela:
Trying to get the likelihood tests to pass has reminded me of what exactly the issue was with single-node recombination storing.

Suppose you have a lineage with a segment of trapped non-ancestral material, and that it forms a diamond. It will be impossible to recover the breakpoint from the nodes and edges, and therefore impossible to determine whether or not the trapped segment should be regarded as "extant" links for the purpose of working out a rate of effective recombinations in the likelihood calculation.

Fortunately, #1979 is exactly what we need to fix this, but it means the metadata update and fixing the likelihood need to be done concurrently.

@jeromekelleher, what do you think is the best way forward? Should I PR the update for the data structure (with a broken likelihood function) so that you can fix the metadata, or is there a better way?

PR #1976, opened 2022-01-13
Pin meson

molpopgen:

PR #1981, opened 2022-01-14
Pin meson

### LINKED ISSUE #1975: Pin meson
Meson should be pinned in development requirements to some recent version that works.

benjeffery:
Fixes #1975

Pinned to greater than or equal to the first version that works.


jeromekelleher:
We already have #1976 - but I guess this is more specific?

benjeffery:
> We already have #1976 - but I guess this is more specific?

Doh, didn't see that!

Issue #1979, opened 2022-01-13
Store ARG breakpoints in node metadata


jeromekelleher:
As prototyped in #1978, it would be very helpful if we stored the actual breakpoint when running ``full_arg`` simulations. We can infer *some* breakpoints from the resulting edges, but not all (they might fall in the middle of non ancestral material). The major update will be adding the metadata schema to the NodeTable, and ensuring that we set valid JSON on every node.

We can do this after #1942

hyanwong:
It would then be trivial to output a Griffiths ARG from an msprime full arg: just take these breakpoints form the metadata and drop the right and left from the edges, right? Is it worth defining an msprime helper function to do this export? I suspect not, but thought I would mention it (note that this isn't a tskit function, because a TS can be much more general than a Griffiths ARG)

jeromekelleher:
Yes, all quite tricky. I think the first step is to do the simplest form of this so we get things working again, and then to take a step back to see if we can refactor things a bit with an eye to generalisation. Until we have some infrastructure for converting back and forth from the Griffiths representation (tskit-dev/tskit#2098) we're going to be talking too much in the abstract and stand a good chance of forgetting about important details.

hyanwong:
> Until we have some infrastructure for converting back and forth from the Griffiths representation ([tskit-dev/tskit#2098](https://github.com/tskit-dev/tskit/issues/2098)) we're going to be talking too much in the abstract and stand a good chance of forgetting about important details.

Absolutely. I think this is a really good way forward, so thanks for taking that on! We need to remember, however, that the Griffiths representation is only one example of a (node annotated) ARG, and I don't think we want to paint ourselves into a corner. Having said which, the idea of using the metadata means that this is kept pretty flexible for future changes.

jeromekelleher:
I've made a start on this in  #1990 and in principle it's straightforward, but in practise a whole bunch of painful compatibility issues start cropping up when we set a metadata schema on the NodeTable. Things that were equal before are no long equal because one has a schema and the other doesn't, even though we're not actually recording any metadata. There's a real risk that we'll break a lot of code and I'm not sure the benefits outweigh the costs (a significant one being, the time it takes to figure out how significant the problems are).

[...]

Issue #1942, opened 2021-12-07
Should the msprime ARG representation have 2 recombination nodes or one?


hyanwong:
In https://github.com/tskit-dev/msprime/pull/665#issuecomment-458903503 @JereKoskela discussed the rationale for having each recombination event represented by 2 nodes in msprime. However, it seems like that argument is moot, as it should be possible to use a single node. This might be clearer (and tie in with the standard ARG visualisations, e.g. in the msprime paper, see below). We discussed this a bit on slack: some pertinent comments extracted below.

>>  [Yan] why do we create 2 separate nodes in [the case of recombination nodes). The docs say that of the 2 nodes “one identif(ies) the individual providing the genetic material to the left of the breakpoint and the other identif(ies) the individuals providing the genetic material to the right”...Wouldn’t it be possible (even preferable) to have one node (the individual in whom the nodes recombine), just like in fig 5?
>
>  [Jere] I don't think there is a reason as such. It is arbitrary whether you record the "child" node in which the lineages remain together, or the "parents" in which the material first splits up. Either way can be made to work by recording the right set of edges, and under the ARG, all three individuals live simultaneously. From a memory cost point of view, storing one individual would presumably be slightly better, and I think the only reason for two separate nodes is that it occurred to me first. I'm not sure whether it would break things to change now, how much work it would be, or whether it is worth doing.
> ... I'm not opposed to the single node approach. That's almost certainly what I'd do if I was setting this up from scratch.

This would, however, risk being a breaking change, possibly for no good reason, so it's worth discussing here.

hyanwong:
Jere also says:

> To return briefly to the discussion about 1 vs 2 recombination nodes, do you have a view about a potential change @Jerome? As far as I can tell, the benefits of switching to one node are
An improved constant in the memory / parsing cost of output.
Easier visualisation (or other similar tasks) when the node table has been reordered.
A cleaner match with tskit if there were to be a "canonical" storage format with just one node. The canonical format could also be two nodes, though there seems little reason to go in that direction except compatibility with existing msprime code.
The downside is that this would break things if people are storing full ARGs and manipulating them in pipelines.

hyanwong:
The msprime paper figure is like this:

![image](https://user-images.githubusercontent.com/4699014/145016302-31de022b-b766-4da7-8608-0afe68d555e6.png)

jeromekelleher:
[...]

PR #1982, opened 2022-01-17
Initial pass at single node recombination

### LINKED ISSUE #1974: Rename or change behaviour of NODE_IS_CA_EVENT
At the moment, when we set `record_full_arg=True`, we mark those nodes that are "Common Ancestor" events but *not* coalescence events with `NODE_IS_CA_EVENT`. In fact, the CA class should also include ("normal") coalescent nodes too, as all nodes should be either CA or RE nodes (or both), so we should either change the flag name or the behaviour.

This can probably be done as part of #1942

JereKoskela:
This changes the infrastructure for recombination nodes under the `full-arg` from two nodes per recombination to one. I've introduced a new flag for the new recombination node, but commented out the old one rather than introduced error handling if old `MSP_NODE_IS_RE_EVENT`s are encountered. Is that something you can add easily along with the breakpoint metadata @jeromekelleher? Background for this change is in #1942.

As discussed, the likelihood function will be broken until I've had a chance to fix it, which I'll be able to do one #1979 has been resolved.

I also changed the flag for `MSP_NODE_IS_CA_EVENT` to the more descriptive `MSP_NODE_IS_NONGENETIC_CA`. I'm happy with further tweaks if anyone has better ideas.

To my surprise, this change causes `test_multimerger` in `test_ancestry.py` to fail. Apparently, a full-arg has 172 edges while its simplified tree sequence has 173. The test involves an instantaneous bottleneck, which is a part of the code I haven't changed, and with which I'm not too familiar. @hyanwong, based on your visualisation plots, do you think there can be some edge case in which simplifying increases the number of edges?

Closes #1974

### DIFF
--- a/lib/likelihood.c
+++ b/lib/likelihood.c
@@ -152,7 +152,7 @@ msp_log_likelihood_arg(tsk_treeseq_t *ts, double r, double Ne, double *r_lik)
         parent = edges->parent[edge];
         lik -= rate * (nodes->time[parent] - sim_time);
         sim_time = nodes->time[parent];
-        if (nodes->flags[parent] & MSP_NODE_IS_RE_EVENT) {
+        if (nodes->flags[parent] & MSP_NODE_IS_RECOMBINANT) {
             if (r <= 0) {
                 *r_lik = -DBL_MAX;
                 ret = 0;
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -2328,20 +2328,14 @@ msp_store_arg_recombination(msp_t *self, segment_t *lhs_tail, segment_t *rhs)
 {
     int ret = 0;
 
-    /* Store the edges for the LHS */
+    /* Store a node for the recombinant child */
     ret = msp_store_node(
-        self, MSP_NODE_IS_RE_EVENT, self->time, lhs_tail->population, TSK_NULL);
+        self, MSP_NODE_IS_RECOMBINANT, self->time, lhs_tail->population, TSK_NULL);
     if (ret < 0) {
         goto out;
     }
     ret = msp_store_arg_edges(self, lhs_tail);
     if (ret != 0) {
-        goto out;
-    }
-    /* Store the edges for the RHS */
-    ret = msp_store_node(
-        self, MSP_NODE_IS_RE_EVENT, self->time, rhs->population, TSK_NULL);
-    if (ret < 0) {
         goto out;
     }
     ret = msp_store_arg_edges(self, rhs);
@@ -2929,7 +2923,7 @@ msp_merge_two_ancestors(msp_t *self, population_id_t population_id, label_id_t l
     if (self->store_full_arg) {
         if (!coalescence) {
             ret = msp_store_node(
-                self, MSP_NODE_IS_CA_EVENT, self->time, population_id, TSK_NULL);
+                self, MSP_NODE_IS_NONGENETIC_CA, self->time, population_id, TSK_NULL);
             if (ret < 0) {
                 goto out;
             }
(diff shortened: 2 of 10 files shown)


hyanwong:
> To my surprise, this change causes `test_multimerger` in `test_ancestry.py` to fail. Apparently, a full-arg has 172 edges while its simplified tree sequence has 173. The test involves an instantaneous bottleneck, which is a part of the code I haven't changed, and with which I'm not too familiar. @hyanwong, based on your visualisation plots, do you think there can be some edge case in which simplifying increases the number of edges?

Yes, definitely. This can be quite common. It's exactly the same issue we have been debating over in https://github.com/tskit-dev/what-is-an-arg-paper/issues/17#issuecomment-1009410925

JereKoskela:
> > @hyanwong, based on your visualisation plots, do you think there can be some edge case in which simplifying increases the number of edges?
> 
> Yes, definitely. This can be quite common. It's exactly the same issue we have been debating over in [tskit-dev/what-is-an-arg-paper#17 (comment)](https://github.com/tskit-dev/what-is-an-arg-paper/issues/17#issuecomment-1009410925)

Excellent, thanks. In that case I'll go ahead and just delete that assert from the test.

lib/likelihood.c
```diff
@@ -121,12 +121,13 @@ msp_log_likelihood_arg(tsk_treeseq_t *ts, double r, double Ne, double *r_lik)
     double material = lineages * tsk_treeseq_get_sequence_length(ts);
     double material_in_children, material_in_parent, rate, gap;
     double lik = 0;
+    double left_end, right_end;
```
jeromekelleher, line 124: Did you intend to include these changes in the likelihood calculation? I thought the idea was to leave this to a follow-up?

lib/msprime.c
```diff
@@ -5090,9 +5084,9 @@ msp_run(msp_t *self, double max_time, unsigned long max_events)
     }
     if (self->store_full_arg
         && !(self->model.type == MSP_MODEL_HUDSON || self->model.type == MSP_MODEL_SMC
-               || self->model.type == MSP_MODEL_SMC_PRIME
```
jeromekelleher, line 5093: Looks like you're on the wrong version of clang-format. See [here](https://tskit.dev/tskit/docs/stable/development.html#sec-development-c-requirements) for info on getting the right one.

Yes, clang-format is awful for changing the format between versions in a non-compatible way.

jeromekelleher (commented):
Basic changes LGTM.

I'm not sure what the simplest approach to dealing with the likelihood code is?

lib/msprime.c
JereKoskela: Thanks! I think that's fixed now.

lib/likelihood.c
JereKoskela: Yes. This works whenever there isn't an offending diamond in the ARG, and is as far as I got before realising the problem. I thought it would be quickest to include it here to have as a starting point for the fix once the metadata is in. Happy to do something different too if you think that's not good practice.

jeromekelleher: I think it probably is a good idea to back out of these changes on this branch, and try to keep them separate.

The easiest thing to do is to start a new branch based on your current one, which can be the starting point for likelihood updates. You can then reset the state of the likelihood files to what they are in upstream/main on this branch and it should all go smoothly.

JereKoskela: Sounds good to me. I'll push one more update to that effect and ping you.

JereKoskela:
Ok @jeromekelleher, I've reverted both the C and Python implementations of the likelihoods back to the current versions. The remaining changes in the tests files are either updating NODE_IS_RE_EVENT labels to NODE_IS_RECOMBINANT, or changing the manual construction of the example ARGs there to conform to the new data structure.

The test pipeline will obviously fail because the likelihood functions are wrong. I'm not sure what the best way around that is.

jeromekelleher:
I've added a commit to skip the tests temporarily @JereKoskela. The other failures are unrelated I think, I'll fix separately and update this PR afterwards.

Look OK to you?

jeromekelleher:
@mergifyio rebase

jeromekelleher (approved):
LGTM - let's merge an move on

PR #1991, opened 2022-01-23
RateMap docs

petrelharp:
This greatly confused me while trying to convert between SLiM maps and msprime maps, so I fleshed out this bit of the docs. The only thing I'm not sure about is how to deal with the temporary file `hotspot.txt`?


jeromekelleher (commented):
Looks good, but I'd get rid of the "reading from a text file" bit. It's tricky, and we'd end up having to write test cases to make sure that the supplied snippet worked properly across a range of inputs.

petrelharp:
So, we *do* want a `read_hapmap( )` example, right? Just not a "write_hapmap" one? So, perhaps I'll re-write that to download a hapmap file from somewhere? Or, just include one?

jeromekelleher:
Maybe - but perhaps we could separate it from the clarification changes that you have here?

Alternatively, just put in a fragment of the hapmap data we have in the docs for read_hapmap, like
```
data = io.StringIO("""
Chromosome  Position(bp)  Rate(cM/Mb)  Map(cM)
chr10       48232         0.1614       0.002664
chr10       48486         0.1589       0.002705
chr10       50009         0.159        0.002947
chr10       52147         0.1574       0.003287
""")
rate_map = RateMap.read_hapmap(data)
```

or something like this?

petrelharp:
The code you posted doesn't run exactly as written because of that initial newline, but how's this?

jeromekelleher (approved):
LGTM!

The alternative for the quoting is to do
```python
data = io.StringIO("""\
Chromosome  Position(bp)  Rate(cM/Mb)  Map(cM)
chr10       48232         0.1614       0.002664
chr10       48486         0.1589       0.002705
chr10       50009         0.159        0.002947
chr10       52147         0.1574       0.003287
""")
```
I tend to use this form, but I don't think one is clearly clearer than the other.

PR #1992, opened 2022-01-24
Revert "Initial pass at single node recombination"

jeromekelleher:
Reverts tskit-dev/msprime#1982

Given that it's quite likely that two-node recombination is actually the correct thing to do, I think it's wise to revert this PR before the state of the repo moves on and it becomes messier.

If we decide to go back to the single-node recombination I'll make a new PR with the original commits cherry picked and rebased, so we won't have to redo the work.

Does this sound OK @JereKoskela?

Apologies for the chopping and changing @JereKoskela - this is all very tricky!

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -1381,8 +1381,9 @@ class Simulator:
         self.set_segment_mass(alpha)
         self.P[alpha.population].add(alpha, label)
         if self.full_arg:
-            self.store_node(lhs_tail.population, flags=msprime.NODE_IS_RECOMBINANT)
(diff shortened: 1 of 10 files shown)


JereKoskela:
> Reverts #1982
> 
> Given that it's quite likely that two-node recombination is actually the correct thing to do, I think it's wise to revert this PR before the state of the repo moves on and it becomes messier.
> 
> If we decide to go back to the single-node recombination I'll make a new PR with the original commits cherry picked and rebased, so we won't have to redo the work.

Fine by me. I'm not sure I fully understand why the two node way is ultimately more helpful, but completely agree with keeping things as they have been until we figure it out.

Issue #1997, opened 2022-02-03
mutation generation bug


chriscrsmith:
I'm running into a bug using msprime.sim_mutations(). This pops up very infrequently so it’s not easy to reproduce, at least in the tree sequences I’m working with. But I was able to find a combo of seed (18112) and tree sequence that will generate the error. The tree sequence can be downloaded [here](http://sesame.uoregon.edu/~chriscs/ )(330Mb). 

MRE:
```
import tskit
import msprime
ts = tskit.load("/home/chriscs/theone.trees")
ts = msprime.sim_mutations(ts, rate=1e-8, random_seed=18112)

```

The error message is: 
“Bug detected in lib/mutgen.c at line 1137. Please report this issue on GitHub, ideally with a reproducible example. (https://github.com/tskit-dev/msprime/issues)
Aborted”

@andrewkern helped me add a printf near the bug assert
```
+++ b/lib/mutgen.c
@@ -1134,6 +1134,7 @@ mutgen_place_mutations(mutgen_t *self, bool discrete_sites)
                 } while (avl_node != NULL && !discrete_sites);

                 time = gsl_ran_flat(self->rng, branch_start, branch_end);
+                printf("DEBUG -- site_left: %lf position: %lf site_right: %lf\n", site_left, position, site_right);
                 tsk_bug_assert(site_left <= position && position < site_right);
                 tsk_bug_assert(branch_start <= time && time < branch_end);
                 if (avl_node != NULL) {
```

The following output shows the variable `position` from the gsl random number generator is hitting `site_right` 
DEBUG -- site_left: 66196102.000000 position: 66196121.000000 site_right: 66196127.000000
DEBUG -- site_left: 66195597.000000 position: 66195999.000000 site_right: 66196083.000000
DEBUG -- site_left: 28918378.000000 position: 28918563.000000 site_right: 28918696.000000
DEBUG -- site_left: 28918378.000000 position: 28918556.000000 site_right: 28918696.000000
DEBUG -- site_left: 28918378.000000 position: 28918623.000000 site_right: 28918696.000000
DEBUG -- site_left: 28918378.000000 position: 28918648.000000 site_right: 28918696.000000
DEBUG -- site_left: 72983639.000000 position: 72983787.000000 site_right: 72983828.000000
DEBUG -- site_left: 72983532.000000 position: 72983540.000000 site_right: 72983543.000000
DEBUG -- site_left: 72983639.000000 position: 72983811.000000 site_right: 72983828.000000
[...]

PR #2000, opened 2022-02-07
gsl_ran_flat() patch

chriscrsmith:
Pull request to solve gsl_ran_flat() bug #1997

### DIFF
--- a/lib/tests/test_core.c
+++ b/lib/tests/test_core.c
@@ -130,6 +130,39 @@ test_tskit_version(void)
     CU_ASSERT_EQUAL(TSK_VERSION_PATCH, 15);
 }
 
+static void
+test_gsl_ran_flat_patch(void)
+{
+    double left, right, x;
+    int i;
+    gsl_rng *rng = gsl_rng_alloc(gsl_rng_default);
+
+    right = 85190021.000000;
+    left = nexttoward(right, 0.0);
+    gsl_rng_set(rng, 12345);
+    x = gsl_ran_flat(rng, left, right);
+    CU_ASSERT(!(right > x));
+    gsl_rng_set(rng, 12345);
+    x = msp_gsl_ran_flat(rng, left, right);
+    CU_ASSERT(right > x);
+    CU_ASSERT(left <= x);
+
+    left = 0.0;
+    right = 1.0;
+    for (i = 0; i < 1000; i++) {
+        x = msp_gsl_ran_flat(rng, left, right);
+        CU_ASSERT(left <= x);
+        CU_ASSERT(right > x);
+    }
+
+    left = 2.0;
+    right = 2.0;
+    x = msp_gsl_ran_flat(rng, left, right);
+    CU_ASSERT(left == x && right == x);
+
+    gsl_rng_free(rng);
+}
+
 int
 main(int argc, char **argv)
 {
@@ -138,6 +171,7 @@ main(int argc, char **argv)
         { "test_strerror_tskit", test_strerror_tskit },
         { "test_probability_list_select", test_probability_list_select },
         { "test_tskit_version", test_tskit_version },
+        { "test_gsl_ran_flat_patch", test_gsl_ran_flat_patch },
         CU_TEST_INFO_NULL,
     };
 
--- a/lib/util.c
+++ b/lib/util.c
@@ -25,6 +25,7 @@
 #include <errno.h>
 
 #include <gsl/gsl_math.h>
+#include <gsl/gsl_randist.h>
 
 #include <tskit/core.h>
 #include "util.h"
@@ -712,3 +713,28 @@ fast_search_free(fast_search_t *self)
 extern inline size_t sub_idx_1st_strict_upper_bound(
     const double *base, size_t start, size_t stop, double query);
 extern inline size_t fast_search_idx_strict_upper(fast_search_t *self, double query);
+
+/*   The gsl_ran_flat() function is supposed to output lo<=x<hi, but
+ *   sometimes (with probability <1e-9) we find x=hi.
+ *   This function includes a while loop to ensure x<hi.
+ *   https://github.com/tskit-dev/msprime/issues/1997
+ */
+double
+msp_gsl_ran_flat(gsl_rng *rng, double lo, double hi)
+{
+    double x;
+
+    /* The logic here requires that lo <= hi, but */
+    /* gsl_ran_flat produces output in this case. Keep */
+    /* this assert to make sure that msprime's usage */
+    /* has the required properties.  */
+    tsk_bug_assert(lo <= hi);
+    if (lo == hi) {
+        x = lo;
+    } else {
+        do {
+            x = gsl_ran_flat(rng, lo, hi);
+        } while (x >= hi);
+    }
+    return x;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.1.1] - 2022-02-xx
+
+**Bug fixes**:
+- Fix (very) rare assertion trip caused by underlying GSL bug.
+  ({issue}`1997`, {pr}`2000`, {user}`chriscrsmith `, {user}`molpopgen`,
+  {user}`andrewkern`)
+
 ## [1.1.0] - 2021-12-14
 
 **New features**
--- a/lib/mutgen.c
+++ b/lib/mutgen.c
@@ -202,7 +202,7 @@ mutation_matrix_choose_root_state(mutation_model_t *self, gsl_rng *rng, site_t *
 {
     int ret = 0;
     mutation_matrix_t params = self->params.mutation_matrix;
-    double u = gsl_ran_flat(rng, 0.0, 1.0);
+    double u = msp_gsl_ran_flat(rng, 0.0, 1.0);
     size_t j = probability_list_select(u, params.num_alleles, params.root_distribution);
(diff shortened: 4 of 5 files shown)


lib/tests/test_core.c
```diff
@@ -130,6 +131,45 @@ test_tskit_version(void)
...
+    gsl_rng_set(rng, 18112);
+    FILE *rng_state;
+    rng_state = fopen("../tests/test_core_rngstate0.bin", "r");
+    gsl_rng_fread(rng_state, rng);
+    fclose(rng_state);
+    left = 85189995.000000;
+    right = 85190021.000000;
+    x = msp_gsl_ran_flat(rng, left, right);
+    CU_ASSERT(left <= x);
+    CU_ASSERT(right > x);
```
andrewkern, line 151: it might be good to run this to _create_ the error too. e.g.,
```suggestion
    x = gsl_ran_flat(rng, left, right);
    CU_ASSERT(left <= x);
    CU_ASSERT(!(right > x));
```

molpopgen (commented):
I think that the method used to trigger the bug is not safe. The binary versions of the random number generator state are not portable across platforms. So if they work in the CI testing, that is luck and not design.  I have an idea of how to trigger it. I hope to get a chance to test the idea tomorrow.

molpopgen:
Quoted from the GSL docs:

```
This function w[r](https://www.gnu.org/software/gsl/doc/html/rng.html#c.gsl_rng_fwrite)ites the random number state of the random number generator r to the [stream](https://www.gnu.org/software/gsl/doc/html/rng.html#c.gsl_rng_fwrite) stream in binary format. The return value is 0 for success and GSL_EFAILED if there was a problem writing to the file. Since the data is written in the native binary format it may not be portable between different architectures.
```

molpopgen:
Following my comments in #1997, this triggers the bug for any RNG seed.  Note the requirement for C99, which is okay since msprime is compiled with C99:

```c
// gcc  -std=c99 -o trigger trigger.c -lgsl -lgslcblas -lm  

#include <stdio.h>
#include <math.h>
#include <assert.h>
#include <gsl/gsl_rng.h>
#include <gsl/gsl_randist.h>

int main(int argc, char ** argv)
{
    const double hi = 85190021.000000;

    // Get the next double different from hi
    // in the direction towards 0.0
    const double next_to_zero = nexttoward(hi, 0.0);
    double rv;
    gsl_rng * r = gsl_rng_alloc(gsl_rng_default);
    gsl_rng_set(r, 12345);

    // This will  print 0.0 for the last value (on most systems at least)...
    fprintf(stdout,"%lf %lf %lf\n", hi, next_to_zero, hi - next_to_zero);

    // ... but it is not 0.0!!
    assert(next_to_zero < hi);
    assert(next_to_zero != hi);
    rv = gsl_ran_flat(r, next_to_zero, hi);

    assert(rv > next_to_zero);
    assert(rv == hi);
}
```

molpopgen:
It will be interesting to know if the rejection sampling approach leads to an infinite loop or not for the low end high values used in my example to trigger the bug.  I'll hazard a guess that it will, for at least some  floating point values with large integer components.

molpopgen:
> It will be interesting to know if the rejection sampling approach leads to an infinite loop or not for the low end high values used in my example to trigger the bug.  I'll hazard a guess that it will, for at least some  floating point values with large integer components.

Actually it won't be an infinite loop now that I think about it more. But it could be very long, requiring waiting until the random number generator returns zero exactly.

chriscrsmith:
@molpopgen is this how you envisioned the test? This loops just a handful of times (1-10)

`#include <stdio.h>
#include <math.h>
#include <assert.h>
#include <gsl/gsl_rng.h>
#include <gsl/gsl_randist.h>

int main(int argc, char ** argv)
{
  const double hi = 85190021.000000;
  const double next_to_zero = nexttoward(hi, 0.0);
  double rv;
  gsl_rng * r = gsl_rng_alloc(gsl_rng_default);
  gsl_rng_set(r, 12345);

  do {
    rv = gsl_ran_flat(r, next_to_zero, hi);
    fprintf(stdout,"%lf %lf %lf\n", next_to_zero, hi, rv);
  } while (rv >= hi && next_to_zero != hi);

  assert(rv >= next_to_zero);
  assert(rv < hi);
}`

lib/util.c
```diff
@@ -712,3 +713,24 @@ fast_search_free(fast_search_t *self)
...
+ *   This function includes a while loop to ensure x<hi.
+ *   https://github.com/tskit-dev/msprime/issues/1997
+ */
+double
+msp_gsl_ran_flat(gsl_rng *rng, double lo, double hi)
+{
+    double x;
+
+    if (lo > hi) {
+        double temp = lo;
```
jeromekelleher, line 728: Style nitpick: we do all the declarations at the start of the function, so here it'd be ``double x, temp;``.

No particularly good reason for it anymore tbh, but best to stick with the established style at this point.

lib/util.c
```diff
@@ -712,3 +713,24 @@ fast_search_free(fast_search_t *self)
...
+        lo = hi;
```
jeromekelleher, line 729: Is this the behaviour of gsl_ran_flag when lo > hi?

I'd be tempted to just put in a ``tsk_bug_assert(lo < hi);`` here. That's the assumption we're making in all the usages from msprime.

That might simplify things a bit too?

jeromekelleher (commented):
Looks great, thanks @chriscrsmith! Some minor comments.

Mainly: can we ``tsk_bug_assert`` out the awkward corner cases?

molpopgen:
> @molpopgen is this how you envisioned the test? This loops just a handful of times (1-10)
> 
> ```
> #include <stdio.h>
> #include <math.h>
> #include <assert.h>
> #include <gsl/gsl_rng.h>
> #include <gsl/gsl_randist.h>
> 
> int main(int argc, char ** argv)
> {
>   const double hi = 85190021.000000;
>   const double next_to_zero = nexttoward(hi, 0.0);
>   double rv;
>   gsl_rng * r = gsl_rng_alloc(gsl_rng_default);
>   gsl_rng_set(r, 12345);
> 
>   do {
>     rv = gsl_ran_flat(r, next_to_zero, hi);
>     fprintf(stdout,"%lf %lf %lf\n", next_to_zero, hi, rv);
>   } while (rv >= hi && next_to_zero != hi);
> 
>   assert(rv >= next_to_zero);
>   assert(rv < hi);
> }
> ```

Yeah this looks right. That's good that it doesn't loop many times. I think that makes sense--there must be many  small values close to zero or one that are smaller than the size of the range used here.  That's a good thing.

lib/util.c
chriscrsmith: I do think this is what's going on with gsl_ran_flat when lo>hi. I think the `tsk_bug_assert(lo < hi);` makes sense

chriscrsmith:
Hey @jeromekelleher and @molpopgen I incorporated your changes and squashed my commits.

molpopgen:
I'm a bit concerned about checking for `lo == hi`.  I think we may want something more like the function below?  @jeromekelleher?

```c
double safe_gsl_ran_flat(const gsl_rng * r, double lo, double hi) {
    double rv;
    if (hi < lo + nextafter(0.0,1.0)) {
        return lo;
    }
    rv = gsl_ran_flat(r,lo,hi);
    while(!(hi > rv)) {
        rv = gsl_ran_flat(r,lo,hi);
    }
    return rv;
}
```

lib/tests/test_core.c
```diff
@@ -130,6 +130,49 @@ test_tskit_version(void)
...
+        CU_ASSERT(left <= x);
+        CU_ASSERT(right > x);
+    }
+
+    left = 2.0;
+    right = 2.0;
+    x = msp_gsl_ran_flat(rng, left, right);
+    CU_ASSERT(left == x && right == x);
+
+    left = 3.0;
```
jeromekelleher, line 164: Ooh, clever! I think this might end up being a bit brittle though, so maybe not worth the complication? I wouldn't bother personally, we're not going to remove the assert from the msp_gsl_ran_flat function and that's all this is testing essentially.

lib/util.c
```diff
@@ -712,3 +713,24 @@ fast_search_free(fast_search_t *self)
...
+/*   The gsl_ran_flat() function is supposed to output lo<=x<hi, but
+ *   sometimes (with probability <1e-9) we find x=hi.
+ *   This function includes a while loop to ensure x<hi.
+ *   https://github.com/tskit-dev/msprime/issues/1997
+ */
+double
+msp_gsl_ran_flat(gsl_rng *rng, double lo, double hi)
+{
+    double x;
+
```
jeromekelleher, line 726: I guess a comment here saying "The logic here requires that lo <= hi, but gsl_ran_flat produces output in this case. Keep this assert to make sure that msprime's usage has the required properties. "

jeromekelleher (approved):
Looks great, thanks @chriscrsmith!

I don't think there's any issue with using "==" @molpopgen, I think this is good. 

In general, I don't see much point in doing things like ``!(hi > rv)`` instead of ``hi <= rv`` --- IEEE comparison semantics are well defined, aren't they? We are assuming IEEE in many places already.

molpopgen:
> Alright @jeromekelleher and @molpopgen, here is an updated version based on your suggestions.
> 
> 

I'm happy with this pull request. Unless Jerome has any comments, I'm happy with a squash and merge at this point.

jeromekelleher:
@mergifyio rebase

andrewkern:
congrats on your first PR @chriscrsmith 🎉

PR #2004, opened 2022-02-10
Disable multithreaded tests

jeromekelleher:


jeromekelleher:
That seems to have fixed the problems on Circle - let's not worry about why

PR #2011, opened 2022-02-24
Minor tweaks to Ancestry and Pedigree docs

gregorgorjanc:
@jeromekelleher this is just a minor polish of the docs as I have read through them. Great work. Happy to give it another read. 

One bit that's confusing is the `draw_pedigree()` function - its called before its mentioned where its defined. Should the function be part of msprime?

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -117,6 +117,7 @@ this.
 
 
 (sec_ancestry_samples)=
+
 ## Specifying samples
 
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
Thanks @gregorgorjanc, this is great! I fixed a small issue with the labels and squashed your commits. 

(Note that this changes the "main" branch on your fork - probably best to avoid using this for feature branches)

PR #2023, opened 2022-03-31
Update pre-commit

benjeffery:


jeromekelleher (approved):
LGTM

I suspect the test failure is due to an update in pytest. Probably simplest to pin to an older version for the moment and register an issue.

benjeffery:
> I suspect the test failure is due to an update in pytest. Probably simplest to pin to an older version for the moment and register an issue.

`pytest` is already pinned for CI on this repo, so I don't understand why these failures are happening. I can't recreate it locally.

benjeffery:
I've managed to re-create this, it only happens in the latest patch version of Python 3.9. Fix on the way.

jeromekelleher:
@mergifyio rebase

PR #2033, opened 2022-04-12
log2 docs fixup

### LINKED ISSUE #2026: Docs don't specify the units in which recombination rate is measured
In the old (0.x) msprime, my understanding was that the recombination_rate parameter gave the probability of recombination between sites that were 1 unit of distance apart, in the units used to specify sequence_length. This was different from Hudson's ms, in which recombination rate describes the probability of recombination between sites at either end of the sequence.

The new docs don't specify which convention is now being used.

Edit: [Elsewhere in the docs](https://tskit.dev/msprime/docs/stable/ancestry.html#sec-ancestry-discrete-genome) there is a discussion that does relate recombination_rate to sequence_length. However, I also find that discussion confusing. It says that 1-exp(-r) is the expected number of recombinations per unit time and per unit map length. Do you mean the number of cross-over events? If so, this should be rewritten in terms of Haldane's mapping formula, because the recombination rate is the probability that the number of crossovers is odd. If recombinations really are a thing you can count, wouldn't 1-exp(-r) be the probability of at least 1 event rather than the expected number of events?

petrelharp:
Closes #2026.

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -663,7 +663,7 @@ SVG(ts.draw_svg())
 assert 1 < ts.num_trees < 5
 ```
 
-We can also simulate a continous genome by setting
+Alternatively, we can simulate a continous genome by setting
 `discrete_genome=False`:
 
 ```{code-cell}
@@ -683,14 +683,6 @@ positions. Simulating a continuous genome sequence can be useful for
 theoretical work, but we recommend using discrete coordinates for most
 purposes.
 
-:::{note}
-With a discrete genome and a recombination rate of {math}`r`,
-the expected number of recombinations per unit time and per unit
-length is {math}`1-e^{-r}`. Therefore the total map length for a
-segment of length {math}`L` and recombination rate {math}`r`
-is {math}`L (1 - e^{-r})`
-and so {math}`Lr (1 - r/2) < L (1 - e^{-r}) < Lr`.
-:::
 
 (sec_ancestry_multiple_chromosomes)=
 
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
LGTM

petrelharp:
I finally figured out what it was talking about; hopefully this explains it better!

alanrogers:
> @alanrogers would you mind taking a quick look at these changes please?

I'd be happy to. But where can I find them?

petrelharp:
I agree, it is hard to make the terminology describing the coalescent process agree with the genetics terminology. I've tweaked it (and removed the "no recombination" phrase); how's that?

You're right, though, that "recombination rate" as we use it is "expected number of crossovers per generation", and so can be unbounded, not the other thing that is a probability.

alanrogers:
> I agree, it is hard to make the terminology describing the coalescent process agree with the genetics terminology. I've tweaked it (and removed the "no recombination" phrase); how's that?
> 
> You're right, though, that "recombination rate" as we use it is "expected number of crossovers per generation", and so can be unbounded, not the other thing that is a probability.

The phrase "co-inheritance" is still a bit unclear. It might mean that there were no crossovers or that the number of crossovers was even. Why not say 

> the
> probability that no crossovers intervene between two adjacent base pairs in {math}`t` generations is {math}`e^{-rt}`, where {math}`r` is probability of a crossover in a single generation.

petrelharp:
That would make sense; the problem is that "crossover" is not a term we use anywhere else in the documentation, instead referring to "recombinations". This is a bit defensible, because since msprime is a coalescent simulator, it doesn't actually perform any crossing over, it just splits an ancestral chunk of genome in two. The reason a chunk would have got split in two is the result of recombination in the ancestor, and the location of the split is the location of a crossover in that ancestor... but, you can see how the terminology is a bit tricky. I think we say "location of the recombination" as short for "location of the split induced by the recombination event" and really we should probably say "location of the crossover" instead.

But - in this example, I think 'co-inheritance' is in fact just right, since we're talking about adjacent base pairs - you can't have 3 crossovers between adjacent base pairs in a single event?

petrelharp:
Any other input here, @alanrogers? I'd like to get this merged in; I thought about how else to say things and didn't come up with anything concrete.

jeromekelleher:
Bumping this - @alanrogers any thoughts on Peter's proposed text?

jeromekelleher:
I'm going to merge this since we haven't had any further comments. Please to follow up with an issue if you think it can be further improved though @alanrogers

alanrogers:
Sorry to be slow in replying. End-of-semester blues.

Alan

PR #2043, opened 2022-04-21
Fix setuptools_scm issue in docker container

benjeffery:
Yikes, seems that running as the user in the docker container upsets git which setuptools_scm was using.
As part of fixing this I switch the build to use the now recommended `build` module.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,7 @@ jobs:
             python setup.py sdist
             python setup.py check
             python -m twine check dist/*.tar.gz --strict
+            rm dist/*
             python -m build
 
       - run:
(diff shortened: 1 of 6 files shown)


benjeffery:
Hmm, seems this isn't getting the version number right.

jeromekelleher (approved):
LGTM, thanks for updating this stuff.

This is a pretty major change to how we build though, and I don't understand how some things are happening now. I guess the most important thing is how we get the correct version of numpy at build time.

Maybe we could document the changes a bit more in the commit message in case we need to bisect back here at any point?

benjeffery:
I've split the commit in two, with an explanation in the commit message.

PR #2058, opened 2022-05-19
Fix linux wheel building and testing

benjeffery:
Turns out the test for the wheel was not a good test!

### DIFF
--- a/.github/workflows/docker/buildwheel.sh
+++ b/.github/workflows/docker/buildwheel.sh
@@ -21,12 +21,14 @@ git config --global --add safe.directory /project
 # Fetch the full history as we'll be missing tags otherwise.
 git fetch --unshallow
 for V in "${PYTHON_VERSIONS[@]}"; do
+    git reset --hard
+    git clean -fd
     PYBIN=/opt/python/$V/bin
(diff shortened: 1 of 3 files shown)


.github/workflows/docker/buildwheel.sh
```diff
@@ -21,12 +21,14 @@ git config --global --add safe.directory /project
...
 for V in "${PYTHON_VERSIONS[@]}"; do
+    git reset --hard
+    git clean -fd
     PYBIN=/opt/python/$V/bin
     rm -rf build/       # Avoid lib build by one Python is used by another
     $PYBIN/python -m venv env
     source env/bin/activate
-    python -m pip install --upgrade build
-    SETUPTOOLS_SCM_DEBUG=1 python -m build
+    $PYBIN/python -m pip install --upgrade build
```
jeromekelleher, line 30: That's weird - this looks like the wrong thing to do in a venv. I guess what's happened is that because we're running as root, that PATH works a little differently than usual and so ``python`` within the venv doesn't resolve to the right version?

benjeffery: Yes, I think that's what's happening, not ideal. Don't think I'll touch it though!

PR #2066, opened 2022-06-09
note about census nodes (closes #2065)

### LINKED ISSUE #2065: msprime census docs
Paraphrasing from tskit-slack.

The census docs state:
```
This can be useful when you wish to study haplotypes that are ancestral to your simulated sample, or when you wish to know which lineages were present in which populations at specified times.
```

However, census nodes are *not* sample nodes, making it currently impossible to study such haplotypes via the statistics framework.  Related: I cannot find docs stating that these nodes do not have `NODE_IS_SAMPLE` set.

To "study ancient haplotypes", shouldn't one prefer `SampleSet`?  The main difference is that census nodes are in the direct ancestry of samples at time 0, but it isn't really clear what one can do with that info?

petrelharp:
Does this help, @molpopgen?

I didn't say specifically "you can't compute statistics with these" because (a) you could mark these as samples and (b) we plan in the future to do missing data right with the stats, and when that happens I won't remember to go change this documentation.

### DIFF
--- a/msprime/demography.py
+++ b/msprime/demography.py
@@ -777,7 +777,9 @@ class Demography(collections.abc.Mapping):
         time, and to extract other information related to these haplotypes: for
         instance to trace the local ancestry of a sample back to a set of
         contemporaneous ancestors, or to assess whether a subset of samples has
-        coalesced more recently than the census time.
+        coalesced more recently than the census time. (However, these added
+        nodes will only represent the portions of ancestral genomes inherited
+        by the samples, rather than complete ancestral genomes.)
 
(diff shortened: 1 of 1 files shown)


molpopgen:
Yes, that does clarify things.  How would one mark census nodes as samples?  The census function takes a single argument, `time`.  Are you suggesting to copy the tables, reset the node flags column, then rebuild a tree sequence?

petrelharp:
That's right. We don't currently have an API for that, and I have not proposed anything concrete because I feel unclear about the use cases. (Probably what you would like is removing of the restriction that statistics be calculated with samples only?)

PR #2071, opened 2022-06-15
Microsatellite mutations models

### LINKED ISSUE #2013: adding microsatellite mutation models
i'm hoping to add some simple mutation models that would be appropriate for microsatellite data.
i think this will be quite useful to folks who are working on non-model organisms, ecological genetics, 
and maybe even forensics? 

The simplest, most classical model is the Stepwise Mutation Model of [Kimura and Ohta (1978)](https://dx.doi.org/10.1073%2Fpnas.75.6.2868). 
I reckon that's a good place to start, although [this nice paper](https://doi.org/10.1093/molbev/msg084) by Calabrese and Durrett 
lays out a number of other models that would be simple enough to implement. 

Here is how I would imagine the implementation to look for the stepwise microsat model. 
generally these models have a _lo_ and _hi_ parameter which governs the bounds on the
number of repeat units of the microsatelllite 

```python
class SMM_MicrosatMutationModel(MatrixMutationModel):
    """
    Stepwise mutation model (Kimura and Ohta, 1978), 
    for microsatellite repeats.

    This is a :class:`.MatrixMutationModel` with alleles ``[lo, .. , hi]``,
    root distribution is uniform ``(lo, hi)``, and a transition matrix that reflects
    single-step mutation.

    :param int lo: lower bound of the microsatellite repeat length
    :param int hi: upper bound of the microsatellite repeat length
        See :ref:`??` for more details.
    """

    def __init__(self, lo, hi):
        alleles = list(range(lo, hi+1))
        root_distribution = [1/len(alleles)] * len(alleles)
        transition_matrix = np.zeros((len(alleles), len(alleles)))
        for i in range(len(alleles)):
            if i == 0:
                transition_matrix[i, i+1] = 1
            elif i == len(alleles)-1:
                transition_matrix[i, i-1] = 1
            else:
                transition_matrix[i, i-1] = 0.5
                transition_matrix[i, i+1] = 0.5

```

what do others think?

andrewkern:
Here is a PR adding a three specific microsatellite mutation models along with a more general backend to add more. 

The generalized parameterization I've chosen here follow from [this nice paper](https://academic.oup.com/genetics/article/168/1/383/6059470) from Rasmus Nielsen's group. It offers plenty of flexibility and room for growth. I've currently implemented just 3 of the models in that paper, but it will be trivial to add more. 

I've implemented these as a subclass of `MatrixMutationModel`, and so I'm translating the CTMC representation from the original paper into the scaled, jump process transition matrix that is used in `MatrixMutationModel`. I hope I got the details correct here but may be off by a scalar?

Hoping to get some feedback as to what others would like to see here and what clean up to do. 

i should add-- this would close #2013


jeromekelleher:
@petrelharp - any thoughts on the above? What do you think of my ideas for the interface?

msprime/__init__.py
```diff
@@ -71,6 +71,9 @@
     JC69,
     MatrixMutationModel,
     MutationModel,
+    SMM_MicrosatMutationModel,
```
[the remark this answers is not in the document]
andrewkern: so `SMM` is probably well enough known on it's own. Many of the others that this generalized system I implemented might not be. E.g. EL2. I doubt it will be _that_ confusing to users however as there will be no name clash with the AA or nucleotide models

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+            raise ValueError("p must be between 0 and 1")
+        if not (0 < m < 1.0):
+            raise ValueError("m must be between 0 and 1")
+        alleles = [str(x) for x in range(lo, hi + 1)]
+        # SMM parameters for generalized microsat model
+        s = 0
+        u = 0.5
+        v = 0
+        transition_matrix = general_microsat_rate_matrix(s, u, v, p, m, lo, hi)
+        if root_distribution is None:
```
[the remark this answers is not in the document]
andrewkern: yep that's a better design. I'll make this edit.

docs/mutations.md
```diff
@@ -661,9 +668,33 @@ print(f"Genetic diversity: {theta}.")
...
+tracked, and its evolution over time subject to one of a number of potential
+biases.
+
+Consider the [Combined DNA Index System](https://en.wikipedia.org/wiki/Combined_DNA_Index_System) (CODIS), a US National database
+of genotypes maintained by the justice system for use in forensic DNA analysis.
+The CODIS system relies on 13 core microsatellite loci, that are unlinked across
+the human genome. Let's simulate a sample of 20 individuals at 13 unlinked loci,
+and then apply the {class}`.SMM_MicrosatMutationModel` model to the sample.
+
+Finally we will summarize variation across our loci
```
petrelharp, line 687: fragment sentence here

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+    :param int hi: upper bound
+    """
+    return 1 + ((i - lo) * s)
+
+
+def alpha_microsat(i, u, v, lo):
+    return max(0, min(1, u - (v * (i - lo))))
+
+
+def general_microsat_rate_matrix(s, u, v, p, m, lo, hi):
```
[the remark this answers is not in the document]
petrelharp: I've had a look at this, and I guess my vote is to leave the parameters here matching what's in the paper - there's enough of them and it's confusing enough that I like the short names that are easier to compare to the paper.

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
         super().__init__(alleles, root_distribution, transition_matrix)
 
 
+def gamma_microsat(m, i, j, lo, hi):
```
petrelharp, line 202: maybe these helper functions should be defined internally to the `general_microsat_rate_matrix` function?

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+    return rate_matrix
+
+
+class SMM_MicrosatMutationModel(MatrixMutationModel):
+    """
+    Stepwise mutation model (Kimura and Ohta, 1978),
+    for microsatellite repeats number.
+
+    This is a :class:`.MatrixMutationModel` with alleles ``[lo, .. , hi]``,
+    root distribution is uniform ``(lo, hi)``, and a transition matrix that reflects
```
petrelharp, line 315:
```suggestion
    a root distribution that is uniform ``(lo, hi)``, and a transition matrix that reflects
```

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+    single-step mutation.
```
petrelharp, line 316:
```suggestion
    This is a :class:`.MatrixMutationModel` with alleles ``[lo, .. , hi]``,
    a root distribution that is uniform ``(lo, hi)``, and a transition matrix that allows
    only mutations that change the number of repeats by +/- 1.
```

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+    rate_matrix /= alpha
+    row_sums = rate_matrix.sum(axis=1, dtype="float64")
+    np.fill_diagonal(rate_matrix, 1 - row_sums)
+    return rate_matrix
+
+
+class SMM_MicrosatMutationModel(MatrixMutationModel):
+    """
+    Stepwise mutation model (Kimura and Ohta, 1978),
+    for microsatellite repeats number.
```
petrelharp, line 312:
```suggestion
    for microsatellite repeat number.
```

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+
+    This is a :class:`.MatrixMutationModel` with alleles ``[lo, .. , hi]``,
+    root distribution is uniform ``(lo, hi)``, and a transition matrix that reflects
+    single-step mutation.
+
+    Note that the values of ``lo`` and ``hi`` are in units of repeat number,
+    and so can be arbitrarily large with respect to repeat length
+    (e.g. dinucleotide, tri-nucleotide, etc.).
```
petrelharp, line 320:
```suggestion
    Note that the values of ``lo`` and ``hi`` are in units of repeat number,
    and are unrelated to repeat length (e.g., the repeat itself can be dinucleotide, tri-nucleotide, etcetera).
```
Is this what you meant?

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+
+    :param int lo: lower bound of the microsatellite repeat copy number
+    :param int hi: upper bound of the microsatellite repeat copy number
+    :param list[float] root_distribution: An array of float values of length hi-lo+1
+        which should sum to 1. These values are used to determine the ancestral state of
+        each microsatellite. Defaults to the value of stationary distribution
+        of the model
```
petrelharp, line 327:
```suggestion
    :param list[float] root_distribution: An array of float values of length hi-lo+1
        which should sum to 1. These values are used to determine the ancestral state of
        each microsatellite. Defaults to the stationary distribution
        of the model.
```

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+class TPM_MicrosatMutationModel(MatrixMutationModel):
+    """
+    Two-phase mutation model of DiRienzo et al., (1994),
+    https://doi.org/10.1073/pnas.91.8.3166
+
+    This models evolution of microsatellite repeats number in a manner
+    that allows for multi-step mutations in copy number of microsatellite repeats.
+    This is parameterized by `p` the probability of a single step mutation,
+    and `m` the success probability of the truncated gamma distribution.
+    For this model both `p` and `m` need to be set to < 1.0.
```
petrelharp, line 363:
```suggestion
    and `m` the success probability of the truncated gamma distribution
    describing the distribution of longer steps.
    For this model both `p` and `m` need to be set to < 1.0.
```

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+    (e.g. dinucleotide, tri-nucleotide, etc.).
+
+    :param float p: probability of a single step mutation
+    :param float m: success prob of truncated gamma distribution
+    :param int lo: lower bound of the microsatellite repeat copy number
+    :param int hi: upper bound of the microsatellite repeat copy number
+    :param list[float] root_distribution: An array of float values of length hi-lo+1
+        which should sum to 1. These values are used to determine the ancestral state of
+        each microsatellite. Defaults to the value of stationary distribution
+        of the model
```
petrelharp, line 381: see comment above.

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+    and `m` the success probability of the truncated gamma distribution.
+    For this model both `p` and `m` need to be set to < 1.0.
+
+    In addition EL2 has a constant bias parameter `u`, where
+    :math:`0 < u < 1` and a linear bias parameter `v`, where
+    :math:`-\\infty < v < \\infty`.
+
+    This is a :class:`.MatrixMutationModel` with alleles ``[lo, .. , hi]``,
+    root distribution is uniform ``(lo, hi)``, and a transition matrix that reflects
+    single-step mutation.
```
petrelharp, line 430: fixup here again

msprime/mutations.py
```diff
@@ -199,6 +199,285 @@ def __init__(self, state_independent=False):
...
+
+    Note that the values of ``lo`` and ``hi`` are in units of repeat number,
+    and so can be arbitrarily large with respect to repeat length
+    (e.g. dinucleotide, tri-nucleotide, etc.).
```
petrelharp, line 434: ditto

msprime/mutations.py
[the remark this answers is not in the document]
jeromekelleher: If we're matching the param names in the paper then that's a good justification. Let's keep them this way so.

tests/test_mutations.py
```diff
@@ -339,6 +339,37 @@ def test_new_model(self):
...
+        model = msprime.SMM_MicrosatMutationModel(7, 9)
+        self.validate_model(model)
+        self.validate_stationary(model)
+
+    def test_TPM_MicrosatMutationModel(self):
+        model = msprime.TPM_MicrosatMutationModel(0.9, 0.9, 2, 9)
+        self.validate_model(model)
+        self.validate_stationary(model)
+
+    def test_EL2_MicrosatMutationModel(self):
```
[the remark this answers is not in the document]
andrewkern: okay i added a couple tests like this that hit all the bits of the transition matrix that are currently being used. we should add more tests for the `s` param stuff once/if we add those models

msprime/mutations.py
andrewkern: 👍

docs/mutations.md
```diff
@@ -661,9 +668,33 @@ print(f"Genetic diversity: {theta}.")
...
+
+```{code-cell} python
+
+ts = msprime.sim_ancestry(20, random_seed=2, sequence_length=13,    recombination_rate=0.5, population_size=100_000)
+model = msprime.SMM_MicrosatMutationModel(1, 10)
+mts = msprime.sim_mutations(ts, rate=1e-5, model=model, random_seed=1)
+for var in mts.variants():
+    print(var.site.position, var.alleles, var.genotypes, sep="\t")
+```
+
```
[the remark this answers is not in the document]
andrewkern: done. also added another example on.

andrewkern:
okay, I think this is all cleaned up and ready to go. I've added a bunch more tests to verify the transition matrices, cleaned up docs, refactored stuff according to the nice design suggestions (thanks!), and rebased and squashed.

jeromekelleher:
LGTM, thanks @andrewkern! Can you update the CHANGELOG please? We're good to merge then.

I'll open an issue to track statistical validation of these sims.

andrewkern:
done

jeromekelleher (commented):
Looks great, thanks!

It's nice to have defaults, but we need to document what they are and have a good justification for setting them to that value. 

Also hard-coding the defuaults into the signatures means that we have no flexibility later in terms of changing the defaults depending on what the other inputs are. The general pattern is to use ``param=None`` in the signature to avoid painting ourselves into corners (unless it's totally clear that there'll never be another other params). Also it avoids having to put the hard coded params into several functions.

However, we can make an issue to track this if you'd rather merge this now?

jeromekelleher:
On second thoughts, let's just merge and update. Opened #2074 to track.

tests/test_mutations.py
```diff
@@ -339,6 +339,136 @@ def test_new_model(self):
...
 
+    def test_SMM(self):
+        model = msprime.SMM(lo=2, hi=50)
+        # test for symmetry of SMM trans mat
+        for i in range(model.transition_matrix.shape[0]):
+            for j in range(i):
+                if i != j:
+                    assert (
+                        model.transition_matrix[i, j] == model.transition_matrix[j, i]
+                    )
```
petrelharp, line 350:
```suggestion
                    )
                if np.abs(i - j) > 1:
                    assert model.transition_matrix[i, j] == 0.0
```

petrelharp:
Lots of minor nitpicks (apologies). But: if this is going to be used for simulating CODIS loci I think we should have more exhaustive tests for correctness. Let's see:

- test that alleles are actually between `lo` and `hi` (inclusive)
- test that rates are 0 for non-adjacent ones in SMM
- test that stationary distribution is uniform when we say it is
- write tests for ratios of expressions for EL2 without copy-pasting the helper functions (I've suggested how to do this for TPM
- test that the sum of qij for j > i divided by the sum for j < i is equal to to `(1 - alpha)/alpha`
- test that the sum of `qij` for `i \neq j` divided by the same thing for `i = lo` is equal to `beta`

And, I think that tests everything?

jeromekelleher:
OK, removing the AUTOMERGE then since we should capture @petrelharp's comments in this PR.

FYI @andrewkern, I pushed to this branch to update the CHANGELOG date (the date of release was set to today, which might be confusing on the "current" docs build). This means you'll need to fetch from your origin and reset --hard. Or, just force push over the change I made and we can do it again, no biggie.

docs/mutations.md
```diff
@@ -661,9 +668,61 @@ print(f"Genetic diversity: {theta}.")
...
+We have implemented a number of mutational models that are commonly used
+to model the evolution of microsatellite repeats.
+The basic idea here is that the number of copies of a given repeat is
+tracked, and its evolution over time subject to one of a number of potential
+biases.
+
+Consider the [Combined DNA Index System](https://en.wikipedia.org/wiki/Combined_DNA_Index_System) (CODIS), a US National database
+of genotypes maintained by the justice system for use in forensic DNA analysis.
+The CODIS system relies on 13 core microsatellite loci, that are unlinked across
+the human genome. Let's simulate a sample of 20 individuals at 13 unlinked loci,
```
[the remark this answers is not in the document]
andrewkern: well crap i'm behind the times.

will change this in the simulations as well on my next pass

docs/mutations.md
```diff
@@ -661,9 +668,61 @@ print(f"Genetic diversity: {theta}.")
...
+    mts = msprime.sim_mutations(ts, rate=r, model=model, random_seed=1)
+    vars.append(mts.genotype_matrix().var())
+
+plt.plot(rates, vars)
+plt.xlabel('mutation rate')
+plt.ylabel('variance in repeat number')
+```
+
+So here we are seeing a clear relationship between the variance
+in repeat number and mutation rate 🔥🔥🔥
```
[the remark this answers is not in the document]
andrewkern: 🔥

msprime/mutations.py
```diff
@@ -199,6 +199,283 @@ def __init__(self, state_independent=False):
...
+    :param int lo: lower bound of the microsatellite repeat copy number
+    :param int hi: upper bound of the microsatellite repeat copy number
+    :param list[float] root_distribution: An array of float values of length hi-lo+1
+        which should sum to 1. These values are used to determine the ancestral state of
+        each microsatellite. Defaults to the stationary distribution
+        of the model
+
+    """
+
+    def __init__(self, p=0.9, m=0.4, u=0.7, v=0.4, lo=2, hi=50, root_distribution=None):
```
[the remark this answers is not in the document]
andrewkern: these were defaults i was using for testing, loosely inspired by the parameter estimates from Sainudiin et al. (2004).

i think a good way to go for defaults for each of these models would be the exact estimates from that paper

tests/test_mutations.py
```diff
@@ -339,6 +339,136 @@ def test_new_model(self):
...
+                denom = 1 - (1 - m) ** (hi - i)
+            elif lo <= j < i <= hi:
+                num = m * (1 - m) ** (np.abs(i - j) - 1)
+                denom = 1 - (1 - m) ** (i - lo)
+            else:
+                raise TypeError(
+                    "microsat model error. i and j must be between lo and hi"
+                )
+
+            return num / denom
```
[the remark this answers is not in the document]
andrewkern: ah yeah this is cleaner

tests/test_mutations.py
```diff
@@ -339,6 +339,136 @@ def test_new_model(self):
...
+                    p + (gamma_microsat(m, i, i - 1, lo, hi) * (1 - p))
+                )
+                exp = e1 / e2
+                obs = (
+                    model.transition_matrix[ii, ii + 1]
+                    / model.transition_matrix[ii, ii - 1],
+                )
+                # this test might seem a little loosey goosey
+                # but the scaling is create numerical noise i think
+                assert np.abs(exp - obs) < 1e-2
```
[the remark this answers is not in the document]
andrewkern: i will recheck this

andrewkern:
> Lots of minor nitpicks (apologies). But: if this is going to be used for simulating CODIS loci I think we should have more exhaustive tests for correctness. Let's see:
> 
> * test that alleles are actually between `lo` and `hi` (inclusive)
> * test that rates are 0 for non-adjacent ones in SMM
> * test that stationary distribution is uniform when we say it is
> * write tests for ratios of expressions for EL2 without copy-pasting the helper functions (I've suggested how to do this for TPM
> * test that the sum of qij for j > i divided by the sum for j < i is equal to to `(1 - alpha)/alpha`
> * test that the sum of `qij` for `i \neq j` divided by the same thing for `i = lo` is equal to `beta`
> 
> And, I think that tests everything?

Okay I think I've not got all these tests in, EXCEPT for the last one which I'm confused on-- I don't think that's true for all models where s = 0? might need coffee

tests/test_mutations.py
```diff
@@ -412,6 +412,39 @@ def test_EL2(self):
...
+        model = msprime.MicrosatMutationModel(s, u, v, p, m, lo, hi)
+        row_sums = model.transition_matrix.sum(axis=1)
+        d = row_sums - model.transition_matrix.diagonal()
+        betas = [1 + ((i - lo) * s) for i in range(lo, hi + 1)]
+        alphas = [u - (v * (i - lo)) for i in range(lo, hi + 1)]
+        for j in range(len(betas)):
+            assert (d[0] / d[j]) - (
+                betas[0] / betas[j] * alphas[0] / (1 - alphas[j])
+            ) < 1e-6
+        # 2nd parameterization, slight variation
```
petrelharp, line 432: Glad you're doing multiple parameterizations - perhaps make sure to set `p` not equal to zero in one of them? And it'd be nicer to do this with `@pytest.mark.parametrize`?

andrewkern: will do

msprime/mutations.py
```diff
@@ -199,6 +199,301 @@ def __init__(self, state_independent=False):
...
+        u=0.5,
+        v=0.0,
+        p=1.0,
+        m=1.0,
+        lo=2,
+        hi=50,
+        root_distribution=None,
+    ):
+        if np.isinf(lo):
+            raise ValueError("lo is infinite")
```
petrelharp, line 347: Should probably check that `lo >= 0` and that `lo < hi`?

msprime/mutations.py
```diff
@@ -199,6 +199,301 @@ def __init__(self, state_independent=False):
...
+    def beta_microsat(i, s, lo):
+        """
+        beta helper function from Sainudiin et al. (2004)
+        represents the mutation rate from allele i
+        :param float s: strength of length dependence on mutation rate
+        (i.e. s=0 uniform rate)
+        :param int lo: lower bound
+        :param int hi: upper bound
+        """
+        return 1 + ((i - lo) * s)
```
petrelharp, line 241: Bad things happen if `s < -1/(hi - lo)`; we should probably check for and error in this case? (gee, if `s = -1/(hi-lo)` then `hi` is an absorbing state, and the eigenvalue computation probably fails?)

petrelharp: Ah in Sainudiin they say that `s > - 1/(hi - lo + 1)`.

msprime/mutations.py
```diff
@@ -199,6 +199,301 @@ def __init__(self, state_independent=False):
...
+    Concretely, mutation rates under this model are computed as follows.
+    Let ``qij`` be the transition rate from equation (3) in Sainudiin et al,
+    and let ``M`` be the largest row sum of the matrix whose i,j-th entry
+    is ``qij`` and whose diagonals are zero. Then if the mutation rate
+    passed to :func:`.sim_mutations` is ``m``, the mutation rate from
+    allele ``i`` to allele `j`` is ``m * qij / M``.
+
+    :param float s: strength of length dependence on mutation rate
+        (i.e. s=0 uniform rate)
+    :param float u: constant bias parameter; must be between 0 and 1
```
petrelharp, line 322: We should probably raise an error if u is not between 0 and 1?

tests/test_mutations.py
```diff
@@ -339,6 +339,119 @@ def test_new_model(self):
...
+                )
+                a_ratio = u - (v * (i - lo))
+                a_ratio /= 1 - a_ratio
+                exp = g_ratio * a_ratio
+                obs = (
+                    model.transition_matrix[ii, ii + 1]
+                    / model.transition_matrix[ii, ii - 1]
+                )
+                if np.abs(exp - obs) > 1e-6:
+                    raise ValueError(f"failed test on row: {i} exp:{exp} obs:{obs}")
```
petrelharp, line 405: Whoops - left over from debugging?
```suggestion
                assert np.abs(exp - obs) < 1e-6
```

tests/test_mutations.py
```diff
@@ -339,6 +339,119 @@ def test_new_model(self):
...
+    @pytest.mark.parametrize(
+        "m, u, v, p, s",
+        [
+            (1.0, 0.62, 0.015, 0, 0.88),
+            (0.43, 0.68, 0.037, 0, 0.88),
+            (0.43, 0.68, 0.037, 0.5, 0.88),
+            (0.43, 0.68, 0.037, 1.0, 0.88),
+        ],
+    )
+    def test_MicrosatMutationModel(self, m, u, v, p, s):
```
petrelharp, line 424: love it - looks great.

tests/test_mutations.py
```diff
@@ -339,6 +339,119 @@ def test_new_model(self):
...
+        model = msprime.SMM(lo=2, hi=50)
+        # test for symmetry of SMM trans mat
+        for i in range(model.transition_matrix.shape[0]):
+            for j in range(i):
+                if i != j:
+                    assert (
+                        model.transition_matrix[i, j] == model.transition_matrix[j, i]
+                    )
+                if np.abs(i - j) > 1:
+                    assert model.transition_matrix[i, j] == 0.0
```
petrelharp, line 352:
```suggestion
                    assert model.transition_matrix[i, j] == 0.0
                elif np.abs(i - j) == 1:
                    assert model.transition_matrix[i, j] == model.transition_matrix[0, 1]
```

petrelharp (approved):
LGTM!! - just a couple of minor things (aren't checking for validity of a few parameters).

msprime/mutations.py
andrewkern: okay i'll add this as a specific test in `MicrosatMutationModel`

tests/test_mutations.py
andrewkern: yup

andrewkern:
okay this should be all set i think @petrelharp @jeromekelleher? 

note I haven't edited the CHANGELOG further. not sure if i need to?

andrewkern:
well darn now i need to hit these new features tests with tests... lol. brb

msprime/mutations.py
```diff
@@ -199,6 +199,283 @@ def __init__(self, state_independent=False):
...
+    and `m` the success probability of the truncated gamma distribution
+    describing the distribution of longer steps such that the
+    mean multi-step mutation is length `1/m`.
+    For this model both `p` and `m` need to be set to < 1.0.
+
+    In addition EL2 has a constant bias parameter `u`, where
+    :math:`0 < u < 1` and a linear bias parameter `v`, where
+    :math:`-\\infty < v < \\infty`.
+
+    This is a :class:`.MicrosatMutationModel` with alleles ``[lo, .. , hi]``
```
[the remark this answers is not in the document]
petrelharp: What about this suggestion?

msprime/mutations.py
```diff
@@ -199,6 +199,308 @@ def __init__(self, state_independent=False):
...
+        if np.isinf(hi):
+            raise ValueError("hi is infinite")
+        if lo > hi:
+            raise ValueError("lo > hi")
+        if not lo >= 0:
+            raise ValueError("lo must be >= 0")
+        if s <= -1 / (hi - lo + 1):
+            raise ValueError("s must be > -1/(hi-lo+1)")
+        if u < 0 or u > 1:
+            raise ValueError("u must be between 0 and 1 inclusive")
```
petrelharp, line 357: I think this should be *non*-inclusive. Otherwise, some wierd things can happen, e.g. the chain is not ergodic:
```
>>> mm = msprime.MicrosatMutationModel(u=0, lo=3, hi=6, v=-10)
>>> mm.transition_matrix
array([[1., 0., 0., 0.],
       [0., 0., 1., 0.],
       [0., 0., 0., 1.],
       [0., 0., 0., 1.]])
>>> mm.root_distribution
array([1., 0., 0., 0.])

```

petrelharp:
Terribly sorry, but I have one more change (I think it should be 0 < u < 1 not <=), and there's two pending suggestions?

msprime/mutations.py
andrewkern: we can change this but i think it conflicts with the original paper
image

andrewkern:
pending suggestions should be all set in my commits-- i didn't use the GH interface to incorporate them, but let me know if i missed something

msprime/mutations.py
petrelharp: Well, you get a valid Markov chain for the endpoints, but where one or both of the endpoints are absorbing, and so you'd get a situation where the root distribution draws with probability 1 from an absorbing state, and so you'd never get any non-silent mutations. That seems like not what anyone would actually want? And, the derived models have 0 < u < 1? There's no harm in leaving it 0 <= u <= 1, I guess, but I'd vote to restrict it.

msprime/mutations.py
```diff
@@ -199,6 +199,308 @@ def __init__(self, state_independent=False):
...
+    Note that the values of ``lo`` and ``hi`` are in units of repeat number,
+    and are unrelated to repeat length
+    (e.g., the repeat itself can be dinucleotide, tri-nucleotide, etcetera).
+
+    :param int lo: lower bound of the microsatellite repeat copy number
+    :param int hi: upper bound of the microsatellite repeat copy number
+    :param list[float] root_distribution: An array of float values of length hi-lo+1
+        which should sum to 1. These values are used to determine the ancestral state of
+        each microsatellite. Defaults to the value of stationary distribution
+        of the model
```
petrelharp, line 396: This is what I meant here:
```suggestion
        which should sum to 1, and give the probabilities used to choose
        the ancestral state of each microsatellite.
        Defaults to the value of stationary distribution
        of the model.
```

petrelharp:
Here's the two missing ones. I think they're important because they say what precisely what the mutation rates are.
Screenshot from 2022-06-23 21-58-11
Screenshot from 2022-06-23 21-57-47

petrelharp:
Oh! And another important thing - where did the defaults come from?

andrewkern:
okay got the suggestions. will post a separate commit for limiting u

andrewkern:
okay got a commit in that will hopefully close #2074?

msprime/mutations.py
```diff
@@ -471,8 +471,8 @@ class EL2(MicrosatMutationModel):
     :math:`-\\infty < v < \\infty`.
 
     This is a :class:`.MicrosatMutationModel` with alleles ``[lo, .. , hi]``
-    For a precise definition of the mutation rates, see :class:`.MicrosatMutationModel`
-    with ``s=0``.
```
petrelharp: (same as above)

msprime/mutations.py
```diff
@@ -422,8 +422,8 @@ class TPM(MicrosatMutationModel):
     For this model both `p` and `m` need to be set to < 1.0.
 
     This is a :class:`.MicrosatMutationModel` with alleles ``[lo, .. , hi]``.
-    For a precise definition of the mutation rates, see :class:`.MicrosatMutationModel`
-    with ``s=0``, ``u=0.5``, and ``v=0``.
```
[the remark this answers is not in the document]
andrewkern: Sorry. I thought it was more precise to say `Default parameterization as estimated in Sainudiin et al. (2004).` but I'll keep in this bit above as well as my addition

PR #2079, opened 2022-06-27
Review interface and documentation.

### LINKED ISSUE #2074: Review and document default params for microsat models
In #2071 we added default parameters for all the new microsat models. This is convenient, but may be a bad idea as users might assume that this is *the* microsat model.

What was the reasoning for the various defaults that you chose @andrewkern?

jeromekelleher:
I tried to reduce the duplication in the interface here and make things consistent. There are a few problems still, but I've opened more specific  issues to track.

I'm extremely nervous about putting in default values here --- there could be real world consequences of us getting some details wrong, or not thinking through how these default parameters might be misused by people who don't quite understand the model. I would rather have no defaults and force people to look up the references a bit if there's any doubt at all in our minds that these parameters really are good defaults in all possible situations. 

Closes #2074

### DIFF
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -681,19 +681,22 @@ biases.
 Consider the [Combined DNA Index System](https://en.wikipedia.org/wiki/Combined_DNA_Index_System) (CODIS), a US National database
 of genotypes maintained by the justice system for use in forensic DNA analysis.
 The CODIS system relies on 20 core microsatellite loci, that are unlinked across
-the human genome. Let's simulate a sample of 20 individuals at 20 unlinked loci,
+the human genome. Let's simulate a sample of 5 individuals at 20 unlinked loci,
 and then apply the {class}`.SMM` model to the sample.
+After running the simulation we then examine the variation across the loci, by
+iterating through the variants.
 
-Finally we will summarize variation across our loci, by
-iterating through the variants at each position / locus.
-We will print the position of the locus, then a list of allelic states, followed
-by haploid genotypes for each chromosome at that site.
+```{note}
+The alleles in {class}`.MicrosatMutationModel`s such as this represent the
+copy number of the repeat, not the actual sequences observed. See below for some basic
+techniques for working with this kind of data.
+```
 
 ```{code-cell} python
 
-ts = msprime.sim_ancestry(20, random_seed=2, sequence_length=20,
+ts = msprime.sim_ancestry(5, random_seed=2, sequence_length=20,
     recombination_rate=0.5, population_size=100_000)
-model = msprime.SMM(1, 10)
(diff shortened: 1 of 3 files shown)


docs/mutations.md
```diff
@@ -681,28 +681,59 @@ biases.
...
 and then apply the {class}`.SMM` model to the sample.
+After running the simulation we then examine the variation across the loci, by
+iterating through the variants.
 
-Finally we will summarize variation across our loci, by
-iterating through the variants at each position / locus.
-We will print the position of the locus, then a list of allelic states, followed
-by haploid genotypes for each chromosome at that site.
+```{note}
+The alleles in {class}`.MicrosatMutationModel`s like this represent the
```
andrewkern, line 690:
```suggestion
The alleles in {class}`.MicrosatMutationModel`s such as this represent the
```

docs/mutations.md
```diff
@@ -681,28 +681,59 @@ biases.
...
+copy number, not the actual sequences observed. See below for some basic
```
andrewkern, line 691:
```suggestion
copy number of the repeat, not the actual sequences observed. See below for some basic
```

docs/mutations.md
```diff
@@ -681,28 +681,59 @@ biases.
...
+we can work with easily. The variation data is returned using the
+tskit {meth}`tskit.TreeSequence.variants` method, which uses an array of
+alleles (the different states observed at a site) and an array of
+"genotypes", which are the observations at that site for all of the samples.
+Usually, we think of alleles as being nucleotides or amino acids which
+are represented as strings. In this case, however, the alleles
+represent copy-numbers which are still represented as strings. So,
+to get the observed copy numbers, we must convert the alleles
+to a numeric type, and obtain the actual observations by
+indexing into this array. This is done in the following function:
```
andrewkern, line 720: this is great!

msprime/mutations.py
```diff
@@ -307,58 +307,75 @@ class MicrosatMutationModel(MatrixMutationModel):
...
-    :param float p: probability of a single step mutation
-    :param float m: 1 / (mean multistep mutation size)
-    :param int lo: lower bound on repeat number
-    :param int hi: upper bound on repeat number
+    Note that the values of ``lo`` and ``hi`` are in units of repeat number,
+    and are unrelated to repeat length
+    (e.g., the repeat itself can be dinucleotide, tri-nucleotide, etcetera).
+
+    The default values for ``lo`` and ``hi`` are chosen based on
+    FIXME.
```
andrewkern, line 326: so these aren't based on anything in particular-- defaults for `lo` and `hi` are just for convenience

andrewkern (approved):
looks good to me. a few minor suggestions above.

Issue #2078, opened 2022-06-27
Justify default values for lo and hi


jeromekelleher:
Not obvious where these came from. In particular, why 2 and not 1 (or zero) for lo?

I've put a placeholder into the MicrosatMutationModel class.

jeromekelleher:
Sound reasonable to me - we just need to document these choices. People will assume that the defaults are the Right Choice.

andrewkern:
going to close this out now that #2080

jeromekelleher:
See #2086 for discussion on changing lo to 1.

Issue #2077, opened 2022-06-27
Documentation for EL2 contradicts signature


jeromekelleher:
The documentation for EL2 talks about the ``p`` parameter, but this is not included in the signature and fixed at 0. The documentation is muddled also and needs another clarification pass.

andrewkern:
closed with #2080

Issue #2075, opened 2022-06-27
More details on defaults for TPM


jeromekelleher:
We set p and m to some values estimated by Sainudiin et al, but it's not obvious to me where these values come from. Need more detail on this.

andrewkern:
We could say something like:

Sainudiin et al (2004) estimated model parameters using 644 dinucleotide repeat loci identified as orthologous in human and chimpanzee genomes using maximum likelihood. TPM parameter estimates of `p` and `m` are taken from those MLEs.

is that what you are looking for @jeromekelleher?

jeromekelleher:
Just a reference to which Table they come from would be enough. I was having trouble tracking the figures down quickly - we just want to make it easy for users to trace back to exactly what bit of the paper we're pulling these numbers from.

More fundamentally though - should we be putting in defaults for humans here? Do people use these models for nonhumans too? We don't put in defaults for particular organisms anywhere else in the API, so unless this model is *only* for humans, I wonder if we should just leave them unparametrised and instead give these values in some examples?

andrewkern:
closed with #2080

PR #2080, opened 2022-06-28
clean up of microsatellite documentation and defaults

andrewkern:
here is a PR to clean up a number of issues in documentation and the default values of the microsatellites.

what i've tried to do here is:
- strip out defaults from the microsatellite models so that people are forced to think about parameterization #2075  
- clean up the documentation on the TPM model #2075 
- clean up the conflict in documentation on the EL2 model #2077 
- explain the `lo` and `hi` default values #2078

### DIFF
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -758,6 +758,49 @@ plt.ylabel('Variance in repeat number');
 So here we are seeing a clear relationship between the variance
 in repeat number and mutation rate, as expected. 🔥🔥🔥
 
+#### A more complicated microsatellite model
+In the 1990s a lot of work went in to describing patterns of
+mutation at microsatellite loci, and a number of models were
+put forward described various biases in expansion vs contraction,
+mutistep mutations, and mutation rate biases. We have implemented 
+a general parameterization of microsatellite mutation models
+developed in [Sainudiin et al. (2004)](https://doi.org/10.1534/genetics.103.022665) that allows users fine grained control of microsatellite mutation.
+
+One such model is the equal rate, linear biased, two-phase
+mutation model of [Garza et al. (1995)](https://doi.org/10.1093/oxfordjournals.molbev.a040239)
+that we have implemented in {class}`.EL2`. Let's simulate large
+samples under this model with different strengths of linear bias, 
+and compare the distribution of allele sizes we get back.
+
+```{code-cell} python
+from matplotlib import pyplot as plt
+from scipy import stats
+
+biases = np.logspace(-7, 1, num=9)
+ts = msprime.sim_ancestry(1000, random_seed=2, sequence_length=1, population_size=100_000)
+for v in biases:
+    model = msprime.EL2(
+            # these are values of m and u from Sainudiin et al. (2004)
+            m=0.01,
+            u=0.68,
+            v=v,
+        )
+    mts = msprime.sim_mutations(ts, rate=1e-3, model=model)
+    C = copy_number_matrix(mts)
+    kde = stats.gaussian_kde(C.flatten())
+    x = np.linspace(2, 51)
+    plt.plot(x, kde(x), label=f'v={v}')
+plt.xlabel('copy number')
+plt.ylabel('frequency')
+plt.legend()
+plt.show();
+```
+From these simulations we can see that the linear bias parameter `v`
+has a strong effect on the distribution of allele frequencies we
+might expect from our model. Care should be taken when choosing
+parameters for your own simulations. Potentially appropriate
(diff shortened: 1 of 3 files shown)


andrewkern:
one thing that i have not done in this PR is provide the user with an example of a full parameterization of either the TPM or the EL2 model. i reckon that would be a good thing to add to the docs. I'll do that now and then add a commit.

docs/mutations.md
```diff
@@ -758,6 +758,49 @@ plt.ylabel('Variance in repeat number');
...
 
+#### A more complicated microsatellite model
+In the 1990s a lot of work went in to describing patterns of
+mutation at microsatellite loci, and a number of models were
+put forward described various biases in expansion vs contraction,
+mutistep mutations, and mutation rate biases. We have implemented 
+a general parameterization of microsatellite mutation models
+developed in [Sainudiin et al. (2004)](https://doi.org/10.1534/genetics.103.022665) that allows users fine grained control of microsatellite mutation.
+
+One such model is the qqual rate, linear biased, two-phase
```
jeromekelleher, line 769:
```suggestion
One such model is the equal rate, linear biased, two-phase 
```

msprime/mutations.py
```diff
@@ -446,8 +446,10 @@ class TPM(MicrosatMutationModel):
     """
 
     def __init__(self, *, p=None, m=None, lo=None, hi=None, root_distribution=None):
```
jeromekelleher, line 448: Simpler here to just take out the ``=None`` bit for these parameters. They become mandatory keyword arguments then.

msprime/mutations.py
```diff
@@ -466,24 +468,29 @@ class EL2(MicrosatMutationModel):
...
+    This is parameterized by `m` the success probability of
+    the truncated gamma distribution
     describing the distribution of longer steps, such that the
-    mean multi-step mutation is length `1/m`.
-    For this model both `p` and `m` need to be set to < 1.0.
+    mean multi-step mutation is length `1/m`. `u` the constant
+    bias parameter which determines bias (if any) in expansion or
+    contraction of the repeat number, and `v` the linear bias
+    parameter which determines mutation rate variation associated
+    with repeat number
```
jeromekelleher, line 478:
```suggestion
    with repeat number.
```

msprime/mutations.py
```diff
@@ -495,9 +502,13 @@ class EL2(MicrosatMutationModel):
     def __init__(
         self, *, m=None, u=None, v=None, lo=None, hi=None, root_distribution=None
```
jeromekelleher, line 503: Likewise here, just make m, u and v mandatory keyword params

jeromekelleher (commented):
LGTM! A few small changes, but totally agree with the direction.

msprime/mutations.py
andrewkern: will odo

andrewkern:
okay @jeromekelleher i think this is good to go!

jeromekelleher (approved):

msprime/mutations.py
```diff
@@ -322,8 +325,12 @@ class MicrosatMutationModel(MatrixMutationModel):
     and are unrelated to repeat length
     (e.g., the repeat itself can be dinucleotide, tri-nucleotide, etcetera).
 
-    The default values for ``lo`` and ``hi`` are chosen based on
-    FIXME.
+    .. warning::
+        The default values for ``lo`` and ``hi`` are chosen arbitrarily.
+        The default of ``lo=2`` is chosed because repeat number need be greater
+        than for it to be considered a repeat. Please ensure
```
petrelharp, line 331: This justification doesn't make sense to me - if it's polymorphic in the population, and *some* individuals have > 1 copy, then it's a repeat. For instance, it could make sense to parameterize things so that `lo=1` and `ancestral_state` is always `1`, and so it's only variable sites that start getting repeat numbers. Because: everything is a *potential* repeat. (Not saying we should do it that way.)

andrewkern: sure, absolutely. I was thinking more about these particular satellite repeats that are ascertained as STRs or whatever, but you are definitely correct.

andrewkern: and it would be easy enough to change the default to `lo=1`

andrewkern: btw you can also look through human microsats here-- because of the ascertainment, all of these have min allele lengths at 3, 4, 5 https://strbase.nist.gov/fbicore.htm

petrelharp: Right, but we're simulating the mutation process, not the ascertainment? I vote to change to `lo=1`.

PR #2083, opened 2022-07-01
Docs for simulating multiple chromosomes with pedigrees

### LINKED ISSUE #2082: Update docs on simulating multiple chromosomes
There was some discussion (#2081) about simulating multiple chromosomes when you have a known fixed pedigree. The [simulating multiple chromosomes](https://tskit.dev/msprime/docs/stable/ancestry.html#multiple-chromosomes) section of the docs were written before the fixed pedigree model was released, and the discussion in #2081 I think reveals some common confusions that users might have about the best approach for simulating whole genomes. A second issue was that it also wasn't clear how to handle NaN flanking regions in recombination map when you attempt to stitch chromosomes together.

Proposed doc updates

1. Split the "Example" in the multiple chromosomes section into two examples: a new section on simulating with a pedigree, and then the existing example of simulating when the pedigree is unavailable
2. In the existing example that stitches chromosome maps together, discuss how to handle flanking regions of NaN, and a caution to discard those regions afterward. I think the suggested approach should be to set those regions to 0 rate, and then use `delete_intervals()` to remove trees in those regions.

I'd be happy to open a PR with these proposed updates.

apragsdale:
This is very much a work in progress, but wanted to get something sketched out. 

Closes #2082

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -688,9 +688,58 @@ purposes.
 
 ### Multiple chromosomes
 
-Msprime does not directly support simulating multiple chromosomes
+#### Do I need to do this?
+
+The main purpose of simulating multiple chromosomes together (instead of
+running independent simulations for each chromosome) is to capture
+correlations between trees on distinct chromosomes. In many scenarios,
+**simulating multiple chromosomes may not be necessary**, and independent
+simulations may be preferable as they will be much more efficient.
+
+That being said, this section describes how to simulate data over
+multiple chromosomes, or more generally, over multiple regions
+with free recombination between them. 
+
+#### Using a fixed pedigree
+
+The most natural way to simulate multiple regions with free recombination for
+a set of sampled individuals is by simulating through a known pedigree. The
+{ref}`fixed pedigree <sec_ancestry_models_fixed_pedigree>` simulation model
+constrains simulated genomic regions by an input pedigree, which specifies
+parent-offspring relationships. Since autosomal genetic material in an
+offspring has equal chance of being inherited from either their mother or their
+father, independent simulations through the same pedigree ensures that those
+regions are inherited from different parents or from the same parent each with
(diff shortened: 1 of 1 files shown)


docs/ancestry.md
```diff
@@ -688,9 +688,58 @@ purposes.
...
+can be found in the {ref}`Fixed pedigree <sec_ancestry_models_fixed_pedigree>`
+section.
+
+Why might simulating multiple chromosomes using a pedigree be preferable to
+other simulation options? First, there is no need to concatenate recombination
+maps for multiple chromosomes, as each region is an independent simulation.
+This also results in simpler, reusable code which can be parallelized. Finally,
+this approach also improves efficiency over a single simulation of multiple
+concatenated chromosomes, as computational costs scale as the square of the
+sequence length (to check: right?).
```
jeromekelleher, line 733: Yep, that's right.

jeromekelleher (commented):
Very nice, thanks. Hadn't appreciated that using the pedigree captured the between chrom correlations :+1:

docs/ancestry.md
apragsdale: Thanks!

apragsdale:
> Very nice, thanks. Hadn't appreciated that using the pedigree captured the between chrom correlations +1

Yes! I think the fixed pedigree model really shines in this type of application. It's a really versatile tool.

I think this is ready for a more thorough review. Happy to make any changes to the larger structure of how I've laid out this section as well if you think there is a more natural order of presenting things.

jeromekelleher (approved):
LGTM!

I guess we should probably provide an example of how to do that NaN-bits extraction, but we can log that as a follow-up issue?

apragsdale:
Great! Let's open a follow-up issue for docs on handling the NaN bits, and I can work on that when I've got a bit of time in the coming weeks.

I just rebased and squashed this one.

PR #2085, opened 2022-07-01
Microsat verify

andrewkern:
working on the microsatellite statistical verification here. the fundamental results i can find center on the variance in repeat number either as a function of 4Nu or as a function of normalized distance between populations.

In this PR I confirm that the `SMM` model implementation fits our expectations as to the linear relationship between 4Nu and var(repeat number).


verification.py
```diff
@@ -5919,6 +5919,63 @@ def test_bottleneck_coalescence_time_T2(self):
...
+                population_size=N,
+                sequence_length=1,
+            )
+            variances = []
+            for ts in reps:
+                mts = msprime.sim_mutations(ts, rate=r, model=mod)
+                C = copy_number_matrix(mts)
+                variances.append(np.var(C.flatten(), ddof=1))
+            avg_variances.append(np.mean(variances))
+            expected.append(2 * N * r)
```
andrewkern, line 5969: factor of 2 off here?

verification.py
```diff
@@ -5919,6 +5919,63 @@ def test_bottleneck_coalescence_time_T2(self):
...
+        """
+        Under the strict SMM model we expect the
+        Variance in repeat number of a sample
+        to be equal to 4Nu. This result has
+        been derived a number of times but a good
+        citation is Goldstein et al (1995)
+        https://doi.org/10.1093%2Fgenetics%2F139.1.463
+        """
+
+        def copy_number_matrix(ts):
```
jeromekelleher, line 5937: No need to make this a private function, it might be useful for other tests. Just put it at the top level.

verification.py
jeromekelleher: Maybe easier to follow if we set ``ploidy=1`` in sim_ancestry and compared with ``4Nr``?

verification.py
```diff
@@ -5919,6 +5919,63 @@ def test_bottleneck_coalescence_time_T2(self):
...
+                C[var.site.id] = alleles[var.genotypes]
+            return C
+
+        rates = np.linspace(1e-5, 1e-2, num=10)
+        avg_variances = []
+        expected = []
+        N = 1000
+        nsamp = 100
+        nreps = 100
+        mod = msprime.SMM(lo=3, hi=500)
```
jeromekelleher, line 5954: 3 seems like an odd choice for ``lo`` here. Why not ``1``?

jeromekelleher (commented):
LGTM. Agree on getting this nice simple comparison in first.

andrewkern:
> Maybe easier to follow if we set ploidy=1 in sim_ancestry and compared with 4Nr?

So I'm not quite sure what's going on here. The scaling of the model seems to be such that if `ploidy=1` the variance in allele count scales as `Nu` where as I would have expected it to be `2Nu`.... is there any reason to believe that the `MatrixMutation` class scales this way?

verification.py
andrewkern: see below comment

andrewkern:
so @petrelharp and I talked about this yesterday a bit and I've gotten to the bottom of it. Nothing is wrong here actually! 

The factor of two was coming from using a uniform root distribution over the alleles and then taking the variance. @petrelharp suggested that I pin the root distribution, such that all sites start at the same root state, and then calculate the avg squared distance from _that_ state, things might look scale as `2N` instead of `N` as I was seeing and presto 🎉 

image

this verification test is all set now @jeromekelleher. perhaps we could merge this, and then i'll add more in a bit?

jeromekelleher (approved):
LGTM!

PR #2114, opened 2022-10-15
Check for valid input in ARG likelihood

### LINKED ISSUE #2107: segmentation fault when trying to run msprime.log_arg_likelihood()
```
import tskit
import msprime
import random

random.seed(6)
nreps = 20
ne = [10, 1e2, 1e3, 1e4, 3e4, 5e4, 7e4, 1e5]

for i in range(nreps):
        ifile = "slim_msp_rep_" + str(i+1) + ".trees"
        ts = tskit.load(ifile)
        for j in range(len(ne)):
                print(msprime.log_arg_likelihood(ts, recombination_rate = 0, Ne = ne[j]))
```
This code results in a Segmentation fault more often than not. The tree sequences are from a single-locus SLiM model without recombination, followed by recapitation of the deep past with msprime. I use a random seed of 6 with SLiM and msprime. 
I'm not sure how to reproduce the error as the code runs fine for the same tree sequence on occasion.

JereKoskela:
This outputs a readable error message when ARGs are requested for trees which aren't binary.

I haven't put in a check for unary nodes because those will legitimately appear on trees due to RE nodes.

Closes #2107


JereKoskela:
Done @jeromekelleher

jeromekelleher:
Needs a test or two, can be something quite simple. Probably easiest to hand construct a few tree sequences that we'd like to reject.

JereKoskela:
I've added a test to cover the `ValueError`. It's only one, but any further tests seemed redundant since they'd just be more elaborate (and less human-readable) versions of the same scenario. Let me know if there's something else in particular we should be covering @jeromekelleher.

jeromekelleher (approved):
LGTM

Issue #1993, opened 2022-01-24
Use ``ancestral_to`` field on segments instead of global lookup table.


jeromekelleher:
A fundamental part of the work in coalescent simulation is to keep track of the amount of ancestral material remaining for each interval along the genome. In msprime we currently do this using an AVL tree (S) (see [here](https://github.com/tskit-dev/msprime/blob/45fd7b88e8309e7c152ea8aadd14ac17fa4e438b/algorithms.py#L594)), and we update this structure during common ancestor events (e.g., [here](https://github.com/tskit-dev/msprime/blob/45fd7b88e8309e7c152ea8aadd14ac17fa4e438b/algorithms.py#L1755)). 

It turns out this is unnecessary and we can achieve the same thing by adding an ``ancestral_to`` field to each segment, which counts the number samples it is ancestral to. Then, when we merge segments at coalescence, we simply add the ``ancestral_to`` values.

See [this code](https://github.com/tskit-dev/what-is-an-arg-paper/blob/93e103925643ffe83d67cfad5c2df0b8d1106058/utils/args.py#L1) for how the simulation works when we track common ancestry by segment.

This should simplify the main interval overlap routines in msprime, and also result in a significant performance boost. I think a substantial fraction of the current simulation time (say, 20% for a long genome and large sample size) is taken up with AVL tree operations to maintain S.

jeromekelleher:
I think you demonstrated that this is actually detrimental to performance @GertjanBisschop - can you comment please?

GertjanBisschop:
Yep, that is true. The main reason was that a lot of segments cannot be squashed when keeping track of a single `ancestral_to` value per segment, as squashing can only happen for segments with identical `ancestral_to` values.

jeromekelleher:
Closing this as "this was a bad idea" then!

PR #2121, opened 2022-11-07
Python algorithm without S avl tree

GertjanBisschop:
Proposal for adapting `algorithms.py`  (see #1993) such that we can track the number of samples each segment is ancestral to without having to rely on AVL tree `S`.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -125,15 +125,16 @@ class Segment:
         self.population = None
         self.label = 0
         self.index = index
+        self.ancestral_to = -1
 
     def __repr__(self):
-        return repr((self.left, self.right, self.node))
+        return repr((self.left, self.right, self.node, self.ancestral_to))
 
     @staticmethod
     def show_chain(seg):
         s = ""
         while seg is not None:
-            s += f"[{seg.left}, {seg.right}: {seg.node}], "
+            s += f"[{seg.left}, {seg.right}: {seg.node}, {seg.ancestral_to}], "
             seg = seg.next
         return s[:-2]
 
@@ -475,7 +476,13 @@ class OverlapCounter:
             curr_interval = curr_interval.next
         raise ValueError("Bad overlap count chain")
 
-    def increment_interval(self, left, right):
+    def yield_overlap_check(self):
+        curr_interval = self.overlaps
+        while curr_interval is not None:
+            yield (curr_interval.node, curr_interval.ancestral_to)
+            curr_interval = curr_interval.next
+
+    def increment_interval(self, left, right, ancestral_to):
         """
         Increment the count that spans the interval
         [left, right), creating additional intervals in overlaps
@@ -486,14 +493,17 @@ class OverlapCounter:
             if curr_interval.left == left:
                 if curr_interval.right <= right:
                     curr_interval.node += 1
+                    curr_interval.ancestral_to += ancestral_to
                     left = curr_interval.right
                     curr_interval = curr_interval.next
                 else:
                     self._split(curr_interval, right)
                     curr_interval.node += 1
+                    curr_interval.ancestral_to += ancestral_to
                     break
             else:
-                if curr_interval.right < left:
+                # verify this changed < to <= !!!!
+                if curr_interval.right <= left:
                     curr_interval = curr_interval.next
                 else:
                     self._split(curr_interval, left)
@@ -505,7 +515,7 @@ class OverlapCounter:
         from breakpoint to seg.right. Set the original segment's
         right endpoint to breakpoint
         """
-        right = self._make_segment(bp, seg.right, seg.node)
+        right = self._make_segment(bp, seg.right, seg.node, seg.ancestral_to)
         if seg.next is not None:
             seg.next.prev = right
             right.next = seg.next
@@ -513,11 +523,12 @@ class OverlapCounter:
         seg.next = right
         seg.right = bp
 
-    def _make_segment(self, left, right, count):
+    def _make_segment(self, left, right, count, ancestral_to=0):
         seg = Segment(0)
         seg.left = left
         seg.right = right
         seg.node = count
+        seg.ancestral_to = ancestral_to
         return seg
 
 
@@ -591,7 +602,7 @@ class Simulator:
             self.gc_mass_index = [
                 FenwickTree(self.max_segments) for j in range(num_labels)
             ]
-        self.S = bintrees.AVLTree()
+
         for pop in self.P:
             pop.set_start_size(population_sizes[pop.id])
             pop.set_growth_rate(population_growth_rates[pop.id], 0)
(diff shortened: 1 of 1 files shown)


algorithms.py
```diff
@@ -486,14 +493,17 @@ def increment_interval(self, left, right):
...
                     left = curr_interval.right
                     curr_interval = curr_interval.next
                 else:
                     self._split(curr_interval, right)
                     curr_interval.node += 1
+                    curr_interval.ancestral_to += ancestral_to
                     break
             else:
-                if curr_interval.right < left:
+                # verify this changed < to <= !!!!
```
jeromekelleher, line 505: This doesn't seem correct - why did we change from < to <=?

jeromekelleher (commented):
This looks fantastic, a major simplification!

jeromekelleher:
@benjeffery any idea what's up with CircleCI here?

benjeffery:
> @benjeffery any idea what's up with CircleCI here?

Try pushing again, seen some discussion that this might be transitory.

GertjanBisschop:
> Can we remove the bintrees dependency or are we still using it somewhere?
> 
> I think the main issues here are with the verification code - how much have you run this? The tests run it a little bit but it's good to run the algorithm for a longish time to really shake out weird problems.

Don't think I have run it enough then. Will do that.

algorithms.py
GertjanBisschop: When using $<$ instead of $\leq$ the `OverlapCounter` makes segments of lengths 0 for breakpoints. For example, this is the overlap counter (see last line) that gets constructed for the following set of lineages:
`[0.0, 100.0: 0, 1]
[0.0, 100.0: 1, 1]
[0.0, 63.77461897661405: 2, 1]
[0.0, 100.0: 3, 1]
[0.0, 100.0: 4, 1]
[0.0, 100.0: 5, 1]
[0.0, 100.0: 6, 1]
[0.0, 100.0: 7, 1]
[0.0, 100.0: 8, 1]
[0.0, 100.0: 9, 1]
[63.77461897661405, 100.0: 2, 1]
overlap_counter:
[0, 63.77461897661405: 10, 10], [63.77461897661405, 63.77461897661405: 11, 11], [63.77461897661405, 100.0: 10, 10]
`

GertjanBisschop:
> Can we remove the bintrees dependency or are we still using it somewhere?
> 
> I think the main issues here are with the verification code - how much have you run this? The tests run it a little bit but it's good to run the algorithm for a longish time to really shake out weird problems.

bintrees is still used [`dtwf_generation()`](https://github.com/GertjanBisschop/msprime/blob/f8915244025c07563c135fe41511d0b791e76a3a/algorithms.py#L1124)

algorithms.py
jeromekelleher: aha, good spot!

jeromekelleher:
Can you push the changes to the C code please @GertjanBisschop - I'd like to have a look at the segment squashing logic

jeromekelleher:
Hmm, maybe we're not defragging often enough. Can you set it up so that we always defrag the segment chain after msp_merge_two_segments (probably just set ``defrag_required=true``) and look at those numbers of segment plots again?

GertjanBisschop:
This does not have any impact:
human_main_defrag

jeromekelleher:
This is a real head scratcher... The extra segments all seem to be late in the simulation, when they should be pretty widely separated within the lineage (I would have thought!)

Can you paste in the top of the perf report (from a full run) for the Drosophila example please?

GertjanBisschop:
Sorry @jeromekelleher, forgot to mention that the time (x-axis) is in generations, and not on the coalescent time scale. So this is the very beginning of the coalescent process.

GertjanBisschop:
Perf report for the Drosophila example:

   35.37%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] msp_merge_two_ancestors
     8.90%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] rate_map_position_to_mass
     7.67%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] avl_rebalance
     6.68%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] fenwick_set_value
     6.32%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] msp_defrag_segment_chain
     5.31%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] avl_at
     4.66%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] fenwick_find
     4.35%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] cmp_individual
     2.90%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] avl_search_closest
     2.57%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] msp_set_segment_mass
     2.06%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] fenwick_increment
     1.62%  python3  libm-2.31.so                                      [.] __log1p
     1.42%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] rate_map_mass_between
     0.82%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] msp_run
     0.77%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] msp_choose_uniform_breakpoint
     0.63%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] msp_get_segment.isra.0
     0.61%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] avl_unlink_node
     0.45%  python3  libgsl.so.23.1.0                                  [.] gsl_rng_uniform_int
     0.39%  python3  libc-2.31.so                                      [.] msort_with_tmp.part.0
     0.36%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] cmp_node_mapping
     0.28%  python3  _msprime.cpython-38-x86_64-linux-gnu.so           [.] fenwick_get_cumulative_sum
     0.25%  python3  libgsl.so.23.1.0                                  [.] gsl_ran_exponential

GertjanBisschop:
For every pair of overlapping segments with non-corresponding starts and ends I think you create three squashable segments in the new merged lineage. One section will correspond to a coalescence event (ancestral_to = sum of both) and the two other bordering segments have the ancestral_to value of one of the original segments.

jeromekelleher:
Right, I guess that's it.

At times I've wondered if actually keeping track of these values is worth it at all. In principle, you could just look through all segments periodically and when you see an interval that contains only one segment, throw it away. There would a tradeoff, of course, between the frequency of doing this vs generated unnecessary recombination events. Do you think this would be worth prototyping?

petrelharp:
> For every pair of overlapping segments with non-corresponding starts and ends I think you create three squashable segments in the new merged lineage. One section will correspond to a coalescence event (ancestral_to = sum of both) and the two other bordering segments have the ancestral_to value of one of the original segments.

Gee, this sounds related to the 'extend edges' thing again! (although maybe only tangentially)

jeromekelleher:
> Gee, this sounds related to the 'extend edges' thing again! (although maybe only tangentially)

Hah, yes! Really great point Peter.

This is definitely worth exploring, but we should look at it in the algorithms.py before diving into C. What we want to do is do ``merge_ancestors`` in two passes: first check if there's any coalescence, and if so we map both lineages to the new node at all positions along the genome whether they coalesce or not. This is something we'll want to support in msprime anyway, so good to explore it in this context in case it fixes these performance issues.

GertjanBisschop:
Cool. These issues are related indeed.
Just thinking. When mapping both lineages to the new node, wouldn't we still have to merge all segments of both original lineages? So we would still be iterating over the extra segments we create by using `ancestral_to`.

jeromekelleher:
> So we would still be iterating over the extra segments we create by using ancestral_to.

Yes, you're right. So maybe this idea of  "keeping more of the ARG" is orthogonal and we should investigate any performance benefits we might get separately. Let's talk about this offline.

A thought that just struck me then is that if we are keeping this extra ARG information then we could probably take the max of the ``ancestral_to`` values in adjacent squashable segments, and it might all work out quite nicely.

not-a-feature:
Additional reason to switch is that `bintrees` is deprecated and the development has stopped:
https://github.com/mozman/bintrees

PR #2124, opened 2022-11-16
Check for tree seqs without recombination nodes

### LINKED ISSUE #2123: Raise error when calculating ARG likelihood on simplified tree seqs
I think it isn't meaningful to run `log_arg_likelihood` on a simplified TS, but it is currently possible to do so. Maybe if there is >1 tree, but no RE nodes in the TS, we should error out when running `log_arg_likelihood`?

hyanwong:
Fixes #2123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
   are added `SMM`, `TPM`, and `EL2`.  
   ({issue}`2013`, {user}`andrewkern`).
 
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):
Nice, thanks.

``ts.nodes_flags`` was introduced in tskit 0.5.2 which we're already depending on, so all good there.

Issue #2128, opened 2022-12-01
Add an option to keep unary nodes


jeromekelleher:
For a number of downstream operations we need to have a way to recording the fact that a given segment passes through an individual (or more precisely a genome/ploid/node). I'm not sure what we'd want to call it eventually in the context of other options, but let's just call it ``keep_unary`` for now and review when we get closer to release.

I think the easiest way to get this working is to adapt the [store_arg_edges](https://github.com/tskit-dev/msprime/blob/main/algorithms.py#L1248) function such that we call this at the end of the merge ancestors functions, (e.g., [here](https://github.com/tskit-dev/msprime/blob/0cef0923d46f336ed46d6771944b44b1fd24759e/algorithms.py#L1804))

It *may* be as simple as something like this:

```python
        if self.full_arg:
            if not coalescence:
                self.store_node(pop_id, flags=msprime.NODE_IS_CA_EVENT)
            self.store_arg_edges(z)
        elif self.store_unary:
              if coalescence:
                    self.store_arg_edges(z)
```

@GertjanBisschop - would you mind seeing if something like this would work in the algorithms.py prototype?

GertjanBisschop:
Hah, indeed. I was just looking into this and about to ping you. I already made the changes in `algorithms.py`. I think it is as simple as this indeed. 
Currently thinking about the best way to test whether the changes are doing what we want it to do.

jeromekelleher:
Great! I thought this was basically the same thing we were talking about all right.

A good first step is to eyeball the text trees - maybe you could paste in a small example here for us to look at?

GertjanBisschop:
Small example: this is with `self.store_arg_edges(z, u)` where u is the node that was already instantiated when a coalescence event was registered.
```
2.07┊                     ┊            22       ┊            22       ┊              22     ┊                     ┊   
    ┊                     ┊       ┏━━━━━┻━━━━━┓ ┊       ┏━━━━━┻━━━━━┓ ┊           ┏━━━┻━━━┓ ┊                     ┊   
1.39┊                     ┊       ┃          21 ┊       ┃          21 ┊           ┃      21 ┊              21     ┊   
    ┊                     ┊       ┃           ┃ ┊       ┃           ┃ ┊           ┃       ┃ ┊           ┏━━━┻━━━┓ ┊   
0.95┊            20       ┊      20           ┃ ┊      20           ┃ ┊          20       ┃ ┊          20       ┃ ┊   
[...]

PR #2130, opened 2022-12-05
first pass store unary

GertjanBisschop:
First pass at adding an option to keep unary nodes (#2128). Unit tests are very basic at the moment. 
I went with the option to explicitly pass the `parent_node_id` to `store_arg_edges()`. This seems to make sense going forward with different options of `keep_unary` but might be wrong here. 
Also, I stuck with the `store_full_arg`, `keep_full_arg` naming conventions for  now.

### DIFF
--- a/tests/test_algorithms.py
+++ b/tests/test_algorithms.py
@@ -34,6 +34,26 @@ def has_discrete_genome(ts):
     return edges_left and edges_right and migrations_left and migrations_right and sites
 
 
+def verify_unary(ts):
+    min_children = np.zeros(ts.num_nodes, dtype=int)
+    max_children = np.zeros_like(min_children)
+
+    for tree in ts.trees():
+        for i in range(ts.num_nodes):
+            n = tree.num_children_array[i]
+            if n > 0:
+                if min_children[i] > 0:
+                    min_children[i] = min(min_children[i], n)
+                else:
+                    min_children[i] = n
+                max_children[i] = max(max_children[i], n)
+
+    assert np.any(min_children == 1)
+    for minc, maxc in zip(min_children, max_children):
+        if minc == 1:
+            assert maxc >= 2
+
+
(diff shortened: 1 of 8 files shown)


tests/test_algorithms.py
```diff
@@ -34,6 +34,26 @@ def has_discrete_genome(ts):
...
+        for i in range(ts.num_nodes):
+            n = tree.num_children_array[i]
+            if n > 0:
+                if min_children[i] > 0:
+                    min_children[i] = min(min_children[i], n)
+                else:
+                    min_children[i] = n
+                max_children[i] = max(max_children[i], n)
+
+    assert any(n == 1 for n in min_children)
```
jeromekelleher, line 51: ``assert np.any(min_children == 1)`` here is a bit more efficient

tests/test_ancestry.py
```diff
@@ -228,6 +228,36 @@ def test_gene_conversion_n100(self):
...
+            for i in range(ts.num_nodes):
+                n = tree.num_children_array[i]
+                if n > 0:
+                    if min_children[i] > 0:
+                        min_children[i] = min(min_children[i], n)
+                    else:
+                        min_children[i] = n
+                    max_children[i] = max(max_children[i], n)
+
+        assert any(n == 1 for n in min_children)
```
jeromekelleher, line 255: Ditto here

jeromekelleher (approved):
LGTM! Let's merge this and then figure out how to this intersects with things like simulating through the pedigree etc

jeromekelleher:
Can merge after #2131 is resolved

jeromekelleher:
@mergifyio rebase

petrelharp:
Just checking - is this storing *all* the unary nodes?

jeromekelleher:
This is "store the full extent of edges recording the paths of ancestry through nodes that are coalescent somewhere along their span". There's no recombination nodes - we need to think through the API before release to make something that's not too confusing (#2132)

PR #2139, opened 2022-12-06
Use state_independent in docs

hyanwong:
Switch a few mentions of parent independence to state independence, since that's now the parameter name


petrelharp (approved):
Good catch - thanks!

Issue #2132, opened 2022-12-05
Review API around store_unary, store_full_arg etc


jeromekelleher:
#2130 added the ``store_unary`` option as a quick way of getting the basic changes in to allow us track these extra nodes, but we need to carefully look at this before release to think about what we want the long-term API to be and whether we want to deprecate the ``store_full_arg`` option in favour of something more flexible.

jeromekelleher:
In discussion with @GertjanBisschop this morning we decided that one way to do it would be to decompose the functionality in terms of how we create new nodes and edges, and have some rules for doing this. Specifically we have two new arguments call (something like) ``nodes`` and ``edges``. These have the possible values:

**nodes**
- COALESCENT
- COMMON_ANCESTOR
- RECOMBINANT
- MIGRANT
- PASS_THROUGH

**edges**
- COALESCING_SEGMENTS
- EVERYWHERE

The current default would be ``nodes=COALESCENT`` and ``edges=COALESCING_SEGMENTS``, and the ``full_arg`` option would correspond to ``nodes=COMMON_ANCESTOR|RECOMBINANT|MIGRANT`` and ``edges=EVERYWHERE``. The functionality we're looking for in the pedigree simulation (#2129) would be ``nodes=PASS_THROUGH`` and ``edges=EVERYWHERE``.

This should be reasonably flexible and implementable across the different models. Any thoughts? Ping @sgravel @petrelharp @hyanwong

petrelharp:
This sounds like a very nice proposal, although it doesn't quite translate to the python api?

jeromekelleher:
Talking to @GertjanBisschop again about this we've decided that really we just have one interesting parameter, ``additional_nodes`` which defines when we create a new node associated with a particular ancestral haplotype, besides the ones we must create to record coalescences. The second parameter is a boolean (say) ``coalescing_segments_only`` which controls whether we keep the current (odd, in retrospect) behaviour of only recording relationships between segments that have coalesced. This can only be used *without* the ``additional_nodes`` argument, because it doesn't make sense to record (say) migrant nodes without keeping track of how they relate to other nodes.

We could make the ``additional_nodes`` argument take an [enum](https://docs.python.org/3.7/library/enum.html) which might look something like this:
```python
import enum

class NodeType(enum.Flag):
        COMMON_ANCESTOR = 1
        RECOMBINANT = 2
        MIGRANT = 4
        PASS_THROUGH = 8
```

[...]

PR #2143, opened 2022-12-08
More helpful record_full_arg errors

hyanwong:
There are a few cases (e.g. the DTWF) where full_arg recording is not possible. We should give slightly more informative error messages when this is attempted. This PR should make it easy to add extra error messages for different cases like this.


jeromekelleher (commented):
LGTM - although this may be changing in the near future 

Is it worth adding this now @GertjanBisschop?

GertjanBisschop:
This should change in the near future indeed (see #2132). If the next msprime release is 1.3.0, then there probably is not much of a point in merging this. If there is going to be an earlier release, then I am all for the (temporary) more meaningful error message.

GertjanBisschop:
Can be closed @hyanwong . Implemented in #2176.

PR #2144, opened 2022-12-08
Clarify "transition matrix" in docs

### LINKED ISSUE #2140: Documentation for "transition_matrix" of a mutation model
When looking at the transition_matrix discussion in the msprime docs that [discuss mutation](https://tskit.dev/msprime/docs/stable/mutations.html#matrix-mutation-models), I think it is reasonable to describe the transition matrix as describing the probabilities that, if a mutation occurs, that mutation will result in a change to each of the states. The rate then describes the probability of mutations occurring over time. Would it be helpful to emphasise that in the docs? I can make a minor wording PR. 

Note that it is (confusingly, I think) different from the P(t) “transition matrix” used e.g. in the[ wikipedia article on substitution models](https://en.wikipedia.org/wiki/Substitution_model#The_mathematics_of_substitution_models) 

> The entry in the ith column and the jth row, Pij(t), is the probability, after time t, that there is a base j at a given position, conditional on there being a base i in that position at time 0

P(t) is also described as a “transition probability matrix” e.g. in [the RevBayes docs](https://revbayes.github.io/tutorials/ctmc/), so I think clearing up the terminological differences here might be helpful. It's presumably too late to call the msprime `transition_matrix` by another name, like the `mutation_matrix`, I presume? Maybe this was discussed when setting up the mutation model API?

hyanwong:
And remove duplicated doc text. Fixes #2140

### DIFF
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -493,8 +493,8 @@ Here are the available models; they are documented in more detail below.
 
 These classes are defined by an alphabet of possible alleles (`alleles`); an array of
 probabilities that determines how likely each allele is to be the root or ancestral
-allele (`root_distribution`); and a `transition_matrix` specifying the probabilities,
-given a mutation has occurred, that each allele will mutate to every other allele.
+allele (`root_distribution`); and a `transition_matrix` specifying the probabilities
+with which a given mutation changes each allele into each other allele.
 Each class generates particular values for these parameters to create the specific model.
 For your own custom model these parameters can be set using
 {class}`msprime.MatrixMutationModel`. For more detail about how mutations are simulated
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -492,12 +492,13 @@ Here are the available models; they are documented in more detail below.
 ### Matrix Mutation Models
 
 These classes are defined by an alphabet of possible alleles (`alleles`); an array of
-probabilities that determines how likely each allele is to be the root, ancestral allele
-(`root_distribution`); and a transition matrix specifying the probability for each allele
-to mutate to every other allele. Each class has specific values of these parameters to
-create the specific model. For your own custom model these parameters can be set using
+probabilities that determines how likely each allele is to be the root or ancestral
+allele (`root_distribution`); and a `transition_matrix` specifying the probabilities,
+given a mutation has occurred, that each allele will mutate to every other allele.
+Each class generates particular values for these parameters to create the specific model.
+For your own custom model these parameters can be set using
 {class}`msprime.MatrixMutationModel`. For more detail about how mutations are simulated
-in these models see {ref}`sec_mutations_matrix_mutation_models_details`.
+in these models see {ref}`sec_mutations_matrix_mutation_theory`.
 
 (sec_mutations_matrix_mutation_models_details)=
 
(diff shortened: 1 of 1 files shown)


docs/mutations.md
```diff
@@ -492,12 +492,13 @@ Here are the available models; they are documented in more detail below.
 ### Matrix Mutation Models
 
 These classes are defined by an alphabet of possible alleles (`alleles`); an array of
-probabilities that determines how likely each allele is to be the root, ancestral allele
-(`root_distribution`); and a transition matrix specifying the probability for each allele
-to mutate to every other allele. Each class has specific values of these parameters to
-create the specific model. For your own custom model these parameters can be set using
+probabilities that determines how likely each allele is to be the root or ancestral
+allele (`root_distribution`); and a `transition_matrix` specifying the probabilities,
+given a mutation has occurred, that each allele will mutate to every other allele.
```
petrelharp, line 497:
```suggestion
allele (`root_distribution`); and a `transition_matrix` specifying the probabilities
with which a given mutation changes each allele into each other allele.
```

hyanwong: Nice wording.

docs/mutations.md
```diff
@@ -551,30 +553,16 @@ As a result, roughly 5/7th of the states will be 🎄, with the remainder
...
-a root distribution, and a transition matrix.
-These leave one free parameter: an overall mutation rate,
-specified by the mutation `rate` in the call to {func}`.sim_mutations`.
-Concisely,
-the underlying model of mutation is a continuous-time Markov chain on the alphabet,
-started by a draw from `root_distribution`, and
-with instantaneous transition rate from `i` to `j` that is equal to
-`rate` multiplied by `transition_matrix[i,j]`.
-The `root distribution` and every row in the `transition_matrix`
-must give *probabilities*, i.e., they must be nonnegative numbers summing to 1.
```
petrelharp, line 575: good catch!

docs/mutations.md
```diff
@@ -596,28 +584,43 @@ contains nonzero entries, then the mutated tree sequence will
...
+that caused an {math}`i \to j` change. If allele {math}`i` mutates to a different allele,
+the chance it mutates to allele {math}`j` is proportional to `transition_matrix[i,j]` but
+excluding the diagonal (no-change) entry, so is equal to
+`transition_matrix[i,j] / (1 - transition_matrix[i,i])`.
 Second, suppose that an ancestor carries allele {math}`i` at a given position.
-The probability that her descendant some time {math}`t` in the future carries allele {math}`j` at that position
-is given by a matrix exponential of
-the scaled [infinitesimal rate matrix](<https://en.wikipedia.org/wiki/Transition_rate_matrix>) of the Markov chain,
-which can be computed as follows:
+The probability {math}`P(t)` that her descendant some time {math}`t` in the future
```
petrelharp, line 594: We shouldn't introduce new notation (here, P(t)) if we don't actually use the notation anywhere. Perhaps add the equation (in math) that P(t) = e^{tQ}?

petrelharp: whoops that's P(t) = e^{t r Q}, where r is the rate

docs/mutations.md
```diff
@@ -596,28 +584,43 @@ contains nonzero entries, then the mutated tree sequence will
...
-Pt = scipy.linalg.expm(t * rate * Q)[i,j]
+Q = model.transition_matrix - np.eye(len(model.alleles))
+t = 0.1  # time
+scaled_Q = t * rate * Q
+Pt = scipy.linalg.expm(scaled_Q)
+
 
+print("P(t) = probs of allele change along a branch of length", t)
+print()
+print("    ----  End  state ----")
```
petrelharp, line 610: I'm not sure what "end state" means here?

petrelharp: oh wait got it, nm

petrelharp (approved):
Looks great. I made a minor suggestion to make the description even more precise.

hyanwong:
> Looks great. I made a minor suggestion to make the description even more precise.

Yeah, thanks. I committed that (good suggestion)

PR #2145, opened 2022-12-09
add keep_unary for pedigrees and dtwf

GertjanBisschop:
This pull request stores all nodes through which some ancestral material passes for pedigrees and dtwf simulations when the flag `record_unary=True` is set. This is a first pass to allow for testing.
This pull request requires defining a `MSP_NODE_IS_PASS_THROUGH_EVENT` (see here). Is this as simple as simply using the next available number in line (`#define MSP_NODE_IS_PT_EVENT (1u << 22)`)?
As with #2130 unit tests are still basic and require further work.


jeromekelleher:
Nice!

@sgravel I think this should more-or-less do what we want with the pedigree sims, so we could start thinking about trying it out and seeing what we get?

sgravel:
Perfect! Presumably this does not record the ancestors above the MRCA yet, right?

GertjanBisschop:
> Perfect! Presumably this does not record the ancestors above the MRCA yet, right?

Yes, that's correct @sgravel. This would require an additional flag indicating the stopping point of the simulation. We have been discussing this exact issue though while thinking about potential performance improvements (see #2121).

sgravel:
Ok, we'll do some testing!

jeromekelleher:
Let's update this and get it merged I think @GertjanBisschop - I guess we should try to register some issues for any remaining TODOs and corner cases so we don't forget them. We can update to the new API later.

GertjanBisschop:
Hmm something has gone wrong during rebase here.

jeromekelleher:
Let's have a look through this tomorrow @GertjanBisschop - it might be simpler to push through the final API rather than getting the C test coverage up.

GertjanBisschop:
Implemented in #2176.

PR #2162, opened 2023-03-28
new additional_nodes API

GertjanBisschop:
First pass at new `additional_nodes` API for Hudson, SMC, SMC' and multi-merger coalescents.
@benjeffery would you want to take a look at how I solved the `asdict` representation of NodeType as required for the provenance recording?
EDIT: ignore that Ben. I think I figured it out. Is this the way to do it?
```python
def asdict(self):
        return {"__class__": "msprime.NodeType", "value": self.value}
```

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -543,8 +543,8 @@ class Simulator:
         max_segments=100,
         num_labels=1,
         sweep_trajectory=None,
-        full_arg=False,
-        store_unary=False,
+        coalescing_segments_only=True,
+        additional_nodes=None,
         time_slice=None,
         gene_conversion_rate=0.0,
         gene_conversion_length=1,
@@ -571,8 +571,10 @@ class Simulator:
         self.num_labels = num_labels
         self.num_populations = N
         self.max_segments = max_segments
-        self.full_arg = full_arg
-        self.store_unary = store_unary
+        self.coalescing_segments_only = coalescing_segments_only
+        self.additional_nodes = msprime.NodeType(additional_nodes)
+        if self.additional_nodes.value > 0:
+            assert not self.coalescing_segments_only
         self.pedigree = None
         self.segment_stack = []
         self.segments = [None for j in range(self.max_segments + 1)]
@@ -1276,7 +1278,7 @@ class Simulator:
         index = random.randint(0, source.get_num_ancestors(label) - 1)
         x = source.remove(index, label)
         dest.add(x, label)
-        if self.full_arg:
+        if self.additional_nodes.value & msprime.NODE_IS_MIG_EVENT > 0:
             self.store_node(k, flags=msprime.NODE_IS_MIG_EVENT)
             self.store_arg_edges(x)
         # Set the population id for each segment also.
@@ -1383,7 +1385,7 @@ class Simulator:
             lhs_tail = x
         self.set_segment_mass(alpha)
         self.P[alpha.population].add(alpha, label)
-        if self.full_arg:
+        if self.additional_nodes.value & msprime.NODE_IS_RE_EVENT > 0:
             self.store_node(lhs_tail.population, flags=msprime.NODE_IS_RE_EVENT)
             self.store_arg_edges(lhs_tail)
             self.store_node(alpha.population, flags=msprime.NODE_IS_RE_EVENT)
@@ -1791,7 +1793,9 @@ class Simulator:
                     pop.add(alpha, label)
                     merged_head = alpha
                 else:
-                    if self.full_arg:
+                    if (coalescence and not self.coalescing_segments_only) or (
+                        self.additional_nodes.value & msprime.NODE_IS_CA_EVENT > 0
+                    ):
                         defrag_required |= z.right == alpha.left
                     else:
                         defrag_required |= (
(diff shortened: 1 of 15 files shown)


GertjanBisschop:
We probably need a more elegant solution for specifying the full_arg option using the `FlagEnum`. Currently I have specified `FULL_ARG` as one of the possible NodeTypes. It is probably better to use something like this:
```python
class NodeType(enum.Flag):
    RECOMBINANT = 1 << 17
    COMMON_ANCESTOR = 1 << 18
    MIGRANT = 1 << 19
    GENE_CONVERSION = 1 << 21
    PASS_THROUGH = 1 << 22
    
    @classmethod
    def full_arg(cls, model=None):
        if model is None:
            return cls.RECOMBINANT | cls.COMMON_ANCESTOR | cls.MIGRANT | cls.GENE_CONVERSION
        elif model.lower() =="hudson":
            return cls.full_arg()
        else:
            raise ValueError
```

benjeffery:
The `asdict` you have looks ok. Do you want me to do a full review here? One thing that jumped is that I think `store_unary` should be deprecated with a warning and not removed - otherwise isn't this a breaking change for users?

benjeffery:
> I don't think `store_unary` has been part of an actual msprime release. So should be ok on that front. 
Ahh great - I missed that point.

lib/msprime.h
```diff
@@ -430,7 +432,8 @@ int msp_set_simulation_model_sweep_genic_selection(msp_t *self, double position,
 int msp_set_start_time(msp_t *self, double start_time);
 int msp_set_store_migrations(msp_t *self, bool store_migrations);
 int msp_set_store_full_arg(msp_t *self, bool store_full_arg);
-int msp_set_store_unary(msp_t *self, bool store_unary);
+int msp_set_additional_nodes(msp_t *self, int additional_nodes);
```
jeromekelleher, line 435: Since the additional nodes argument is a flags value, we want it to be a fixed size and with an unsigned type (there can be confusion with the sign bit, otherwise). So, change the type here to ``uint32_t``.

msprime/_msprimemodule.c
```diff
@@ -1784,7 +1784,8 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
     Py_ssize_t num_populations = 1;
     int store_migrations = false;
     int store_full_arg = false;
-    int store_unary = false;
+    int additional_nodes = 0;
```
jeromekelleher, line 1787: Annoyingly, the Python C API doesn't support fixed width types like ``uint32_t``. So, the simplest thing to do then is to use a type that's guaranteed to be at least this size, and  to cast to uint32_t later. So,
```
unsigned long additional_nodes
```

We use ``k`` for unsigned long in the [Python C API](https://docs.python.org/3/c-api/arg.html)

msprime/_msprimemodule.c
```diff
@@ -1926,8 +1927,9 @@ Simulator_init(Simulator *self, PyObject *args, PyObject *kwds)
             goto out;
         }
     }
+    msp_set_additional_nodes(self->sim, additional_nodes);
```
jeromekelleher, line 1930: Here then we do
```c
msp_set_additional_nodes(self->sim, (uint32_t) additional_nodes);
```

jeromekelleher (commented):
Looks great @GertjanBisschop. Some quick comments on how to be type safe with fixed size types, which we need for the flags, because they go over 16 bits. Technically, ``int`` types are only guaranteed to be 16 bits long (but will usually be 32).

Just annoying legacy C stuff.

jeromekelleher:
LMK when this is ready for a final look @GertjanBisschop

algorithms.py
```diff
@@ -2275,16 +2287,16 @@ def add_simulator_arguments(parser):
         help="Parameters for the allele frequency trajectory simulation",
     )
     parser.add_argument(
-        "--full-arg",
+        "--all-segments",
         action="store_true",
         default=False,
-        help="Store the full ARG with all recombination and common ancestor nodes",
+        help="Only record edges along coalescing segments.",
```
GertjanBisschop, line 2293: This should be: Record edges for nodes that are coalescent nodes anywhere in the sequence.

msprime/_msprimemodule.c
```diff
@@ -3010,9 +3009,9 @@ static PyGetSetDef Simulator_getsetters[] = {
     {"record_migrations",
             (getter) Simulator_get_record_migrations, NULL,
             "True if the simulator should store migration records." },
-    {"record_full_arg",
-            (getter) Simulator_get_record_full_arg, NULL,
-            "True if the simulator should store the full ARG." },
+    {"additional_nodes",
+            (getter) Simulator_get_additional_nodes, NULL,
+            "The numeric value of the stored NodeTypes." },
```
GertjanBisschop, line 3014: should be `NodeType`

GertjanBisschop:
Ready for a final look @jeromekelleher. I hope the idling lint tests will run again after a squash and rebase.

benjeffery:
The lint failing needs a PR to fix. I'll get to it.

benjeffery:
@mergifyio rebase

msprime/ancestry.py
```diff
@@ -249,7 +249,8 @@ def _parse_simulate(
     start_time=None,
     end_time=None,
     record_full_arg=False,
```
jeromekelleher, line 251: Ah OK, let's not add any new features to ``_parse_simulate``. That's deprecated code now and we only add new features to ``sim_ancestry``.

jeromekelleher (approved):
LGTM

Just need to remove the new parameters (all of them) from ``simulate``, and update the CHANGELOG and we're done. 

Can you follow up with some issues to track what needs to be done next, and also scan to see if there's any we should close now as well after merging this PR?

PR #2164, opened 2023-03-31
clarify `coalescence_rate_trajectory`, closes #2163

### LINKED ISSUE #2163: Coalescence rate trajectory, more than two lineages
Hi, 

I have a question regarding the coalescence_rate_trajectory functionality. 

If I simulate a history with 2 lineages (ploidy=2, num_samples=1), then get the coalescence rate with `debug.coalescence_rate_trajectory(time_array,{"A": 2})`, then I am satisfied. If I simulate with more samples and do `debug.coalescence_rate_trajectory(time_array,{"A": 3})` then the coalescence rate msprime gives me is exactly the same as `debug.coalescence_rate_trajectory(time_array,{"A": 2})`. This is counter to my expectation. 

Quick bit of theory then an example. In a panmictic population of constant size, sampling 2 lineages, the density function for time till first coalecsence is f=exp(-t), thus the distribution is F=1-exp(-t), then the coalescence rate is CR = f/(1-F) = 1, and effective population size (Ne) is taken as Ne=1/CR = 1. If we sample 3 lineages, then the density function for time till first coalescence is f=3*exp(-3t), so F=1-exp(-3t), so CR = 3, so Ne=1/3. Thus it is suprising to me that `debug.coalescence_rate_trajectory(time_array,{"A": 2}) = debug.coalescence_rate_trajectory(time_array,{"A": 3})` . 

As a minimal example to demonstrate this, I do the following: a) simulate a constant population with N=2e+04, size with ploidy=2,samples=2, b) record the tree sequence, c) discretise time into windows, and round the continuous times from the tree sequence into these windows, d) choose three random pairs of lineages, e) calculate their empirical coalescence (note 1) rate from the discrete tree sequence. I get this, which looks good: 
 
image
(noise because of short sequence length)

If I do the same as above, but instead choose three arbirtrary lineages and calculate their empirical coalescence rate (note 2), then indeed the CR is 3 times higher so Ne is three times lower:
image

Where as from the simulation, choosing 2 or 3 lineages looks the same:
`R2, _ = debug_p2_s2.coalescence_rate_trajectory(T_gens,{"A": 2}) `
`R3, _ = debug_p2_s2.coalescence_rate_trajectory(T_gens,{"A": 3})` 

image

Am i misundertsanding the coalescence_rate_trajectory ? I appreciate that in each case it correctly gets N=2e+04, but if there are e.g. three lineages this is not what I expect. I also wonder if this is a related issue: https://github.com/tskit-dev/msprime/issues/1771

Thanks, 

(note 1) calculate empirical estimate of the density f, essentially with a histogram, from which F then CR then Ne follow simply.
(note 2) Same as above, but look for first coalescence event of three lineages

Trevor

petrelharp:
@trevorcousins pointed out the docs here were not totally clear that the coalescence rate calculated is the mean per-pair rate. Perhaps this is sufficient to clarify, given the equations are given below?


benjeffery:
@mergifyio rebase

jeromekelleher (approved):

PR #2170, opened 2023-04-17
Fix codecov with token

benjeffery:
Last fix didn't push coverage successfully.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 orbs:
   codecov: codecov/codecov@3.2.4
+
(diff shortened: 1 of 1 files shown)


benjeffery:
This is very odd - if I hard code the value of the codecov token this all works - put it in an environment variable and it all fails!

benjeffery:
Solution was a checkbox buried in the circle CI config to allow secret sharing with PRs on forks.

petrelharp:
which github emoji means facepalm?

petrelharp:
also, thanks for figuring that out!

PR #2173, opened 2023-04-19
clarification to migration documentation

petrelharp:
Looking at the docs for https://github.com/tskit-dev/msprime/discussions/2172 it seemed like this section should say concretely what asymmetric migration means. (Also, I *don't* think it's true that asymmetric migration is uncommon, so I edited out that part.)


benjeffery (approved):
LGTM, makes sense, although I'm not completely au fait with the details here.

jeromekelleher (approved):
Sgtm.

PR #2176, opened 2023-05-02
Additional_nodes DTWF/FIXED_PEDIGREE

GertjanBisschop:
`algorithms.py` outline for tracking `additional_nodes` for `DTWF` and `FIXED_PEDIGREE` models.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -771,6 +771,11 @@ class Simulator:
             self.tables.edges.add_row(left, right, parent, child)
             self.edge_buffer = []
 
+    def update_node_flag(self, node_id, flag):
+        node_obj = self.tables.nodes[node_id]
+        node_obj = node_obj.replace(flags=node_obj.flags | flag)
+        self.tables.nodes[node_id] = node_obj
+
     def store_edge(self, left, right, parent, child):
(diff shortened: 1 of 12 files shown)


algorithms.py
```diff
@@ -771,6 +771,12 @@ def flush_edges(self):
             self.tables.edges.add_row(left, right, parent, child)
             self.edge_buffer = []
 
+    def update_node_flag(self, node_id, flag):
+        self.flush_edges()
```
jeromekelleher, line 775: Do we need this flush_edges? Should probably be done in the other function we just added?

GertjanBisschop: Adding `flush_edges()` to `store_additional_nodes_edges()` would flush the edges twice in some cases, as it is also called by `store_node()`.

jeromekelleher: I see - should we call flush_edges to the non-store-node path in store_additional_nodes_edges then?

jeromekelleher: It's a surprising side-effect to have here in ``update_node_flag``

jeromekelleher (approved):
LGTM!

PR #2180, opened 2023-05-10
Test and build wheels on 3.11

### LINKED ISSUE #2177: Python 3.11 wheels
It would be great to have wheels for Python 3.11

benjeffery:
Fixes #2177

### DIFF
--- a/tests/test_ancestry.py
+++ b/tests/test_ancestry.py
@@ -870,9 +870,7 @@ class TestSimulator:
         assert len(caplog.records) == 6
         assert caplog.messages[0].startswith("model[0] {'name': 'hudson'}")
         assert caplog.messages[-1].startswith("Completed at time")
-        assert (
-            caplog.messages[2] == "time=0.312845 ancestors=2 ret=ExitReason.MAX_EVENTS"
-        )
+        assert caplog.messages[2].startswith("time=0.312845 ancestors=2 ret=")
 
     def test_debug_logging_dtwf(self, caplog):
         sim = ancestry._parse_simulate(3, Ne=10, model="dtwf")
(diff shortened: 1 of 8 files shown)


benjeffery:
Ah, got it. Should be all good now.

msprime/ancestry.py
```diff
@@ -1510,7 +1510,7 @@ def _run_until(self, end_time, event_chunk=None, debug_func=None):
                     "Please open an issue on GitHub"
                 )
             logger.debug(
-                "time=%g ancestors=%d ret=%s", self.time, self.num_ancestors, ret
+                f"time={self.time} ancestors={self.num_ancestors} ret={repr(ret)}"
```
jeromekelleher, line 1513: Why this change? Are these logging converters not supported after 3.11? They only generate the string as needed with log level, so are useful

jeromekelleher (commented):
Lgtm - logging changes seem like a regression though?

msprime/ancestry.py
benjeffery: In Py3.11 `IntEnum`'s  `__str__` method has reverted to that of a normal int. The other way to fix this is to change the checked for text to test for an int instead of the string. I'll change it to that.

tests/test_ancestry.py
```diff
@@ -764,7 +765,11 @@ def test_debug_logging_dtwf(self, caplog):
                 " until max time: inf"
             )
             assert caplog.messages[-1].startswith("Completed at time")
-            assert caplog.messages[2] == "time=1 ancestors=3 ret=ExitReason.MAX_EVENTS"
+            print("WOP", caplog.messages[2])
```
jeromekelleher, line 768: Stray print

tests/test_ancestry.py
```diff
@@ -764,7 +765,11 @@ def test_debug_logging_dtwf(self, caplog):
...
+            assert (
+                caplog.messages[2] == "time=1 ancestors=3 ret=ExitReason.MAX_EVENTS"
```
jeromekelleher, line 770: Just change to starts with the prefix that's shared

jeromekelleher (approved):
Looks good, ready to merge after cleanup

benjeffery:
@mergify rebase

Issue #2182, opened 2023-05-23
Autoflush edges in the simulator


jeromekelleher:
We're manually calling flush_edges in a few places, when the conditions can reliably be detected and it simplifies the code.

GertjanBisschop:
Solved in #2183.

PR #2183, opened 2023-06-07
auto_flush_edges

GertjanBisschop:
As discussed in #2181 and #2182.

Note the `self->num_buffered_edges = 0;` in `msp_reset()`. When resetting the simulation there might be leftover edges in the buffer if the simulation ended with an error. This was the case for `test_demographic_events`.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -1814,9 +1814,19 @@ msp_store_edge(msp_t *self, double left, double right, tsk_id_t parent, tsk_id_t
 {
     int ret = 0;
     tsk_edge_t *edge;
+    tsk_edge_t last_edge;
     const double *node_time = self->tables->nodes.time;
 
     tsk_bug_assert(parent < (tsk_id_t) self->tables->nodes.num_rows);
+    if (self->num_buffered_edges > 0) {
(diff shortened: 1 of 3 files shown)


lib/msprime.c
```diff
@@ -1813,10 +1813,19 @@ static int MSP_WARN_UNUSED
 msp_store_edge(msp_t *self, double left, double right, tsk_id_t parent, tsk_id_t child)
 {
     int ret = 0;
-    tsk_edge_t *edge;
+    tsk_edge_t *edge, *last_edge;
     const double *node_time = self->tables->nodes.time;
 
     tsk_bug_assert(parent < (tsk_id_t) self->tables->nodes.num_rows);
+    if (self->num_buffered_edges > 0) {
+        last_edge = self->buffered_edges + self->num_buffered_edges - 1;
```
jeromekelleher, line 1821:
```
tsk_edge_t *edge, last_edge;

last_edge = self->buffered_edges [self->num_buffered_edges - 1]
if (last_edge.parent != parent) {
```
is a bit more efficient (and obvious I think - I don't know why I was so fond of pointer addition tricks in the old days!)

jeromekelleher (approved):
Looks great! Minor tweak suggested

Issue #2135, opened 2022-12-05
Documentation for simulating with unary nodes


jeromekelleher:
Once we've thought through the proper API for simulating with extra unary nodes in the output (#2132) we should add a short section in the docs explaining what's going on with examples.

GertjanBisschop:
Also, review the current docstrings both for the `additional_nodes` flag as well as the `coalescing_segments_only` flag.

GertjanBisschop:
Solved in #2185.

PR #2185, opened 2023-06-19
additional_nodes docs

GertjanBisschop:
First go at documenting the `additional_nodes` and `coalescing_segments_only` flags (see #2162, #2176, #2135).

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -945,61 +945,183 @@ to represent the simulated genealogical history of the samples. Sometimes
 we are interested in more detailed information; this section gives details
 of options that allow us to do this.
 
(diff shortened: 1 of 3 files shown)


petrelharp:
Good start. Where is there going to be the list of things you can pass to additional_nodes, with definitions? (Also, we could come up with some figures with the relevant nodes in colors; I did something like this for simplify recently here: see tree*.svg [in this directory](https://github.com/petrelharp/tskit_intro/tree/main/figs/ts_ex) and the code to make it in `make_ex.py`.)

jeromekelleher (approved):
LGTM - let's squash and merge, and get some feedback (and maybe polish a bit) before release.

PR #2201, opened 2023-09-05
Clean up config and get version working

jeromekelleher:
The versioning for the msprime docs stopped working a while back, this should fix it.

I can't get the navbar working, so just sticking the version in the footer. It's better than nothing. The specific theming for modes didn't seem to do much, so I removed it. Highlighting seems to work well in light and dark mode for me.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -24,8 +24,10 @@ html:
   use_edit_page_button: true
   # Do not edit this - the version placeholder is replaced by the
   # current version during a distribution build in the Makefile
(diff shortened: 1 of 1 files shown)


benjeffery:
I checked out the docs, both fixed code and jupyter-book cells. It seems like they both look ok. I prefer the colours of the monokai style, but its a matter of taste so no biggie to remove it.

jeromekelleher:
Happy to change back to monokai, if it works on both modes.

PR #2212, opened 2023-10-29
Clarify the population size is diploid (by default)

hyanwong:
I think this is right. It is helpful to say "diploid population size" somewhere in the docs, as people might search for this phrase.

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -348,7 +348,8 @@ sized population in the default demography.
 
 For example, here we run the same simulation twice, one with
 the default population size and one with a population size
-of 100:
+of 100 (this is the diploid population size because the
+default {ref}`sec_ancestry_ploidy` is 2):
 
 ```{code-cell}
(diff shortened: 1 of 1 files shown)


petrelharp (commented):
LGTM!

petrelharp:
I think this only failed because of
```
WARNING: The default value for `navigation_with_keys` will change to `False` in the next release. If you wish to preserve the old behavior for your site, set `navigation_with_keys=True` in the `html_theme_options` dict in your `conf.py` file.Be aware that `navigation_with_keys = True` has negative accessibility implications:https://github.com/pydata/pydata-sphinx-theme/issues/1492
```
Do you mind using this as an opportunity to check whether setting that value explicitly fixes things, @hyanwong? (Sounds like probably we want to set it explicitly to `False` but I haven't looked into it...)

hyanwong:
Thanks Ben 🏅 ⭐

PR #2232, opened 2023-12-07
modify checks arg lik

GertjanBisschop:
Slight modification to the polytomy check when computing likelihoods. 
There are a couple of test cases in the current test suite that include examples with multiple roots (within a single local tree). Thus far these did not fail as they were always limited to two roots at most. 
These checks could probably be a bit more sophisticated as there are many cases where a decapitated arg could easily be associated with a well-defined likelihood value. @JereKoskela, input welcome.

### DIFF
--- a/msprime/likelihood.py
+++ b/msprime/likelihood.py
@@ -139,12 +139,21 @@ def log_arg_likelihood(ts, recombination_rate, Ne=1):
         `-DBL_MAX`.
     """
     for tree in ts.trees():
-        if np.any(tree.num_children_array > 2):
+        if np.any(tree.num_children_array[:-1] > 2):
             raise ValueError(
                 "ARG likelihood encountered a polytomy."
                 " Tree sequences must contain binary mergers only for"
                 " valid likelihood evaluation."
             )
+        if tree.num_children_array[-1] > 1:
+            if ts.num_edges > 1:
+                # num_edges check is here because to avoid breaking the expected
+                # result of the TestOddToplogies tests.
+                raise ValueError(
+                    "ARG likelihood encountered a tree with multiple roots."
+                    " All local trees must have a single mrca for"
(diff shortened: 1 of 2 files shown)


JereKoskela:
It has been two years since I've really looked at this, but I don't think multiple roots should cause any problems in the likelihood function implementation. Was the point of your update in #2107 that the existing test was incorrectly picking up multi-root trees as error cases @GertjanBisschop? If so, then I think the right fix is to just tweak the test to not do that, without introducing a new error. Or am I missing something?

GertjanBisschop:
I was mainly unsure about what happens when we add in nodes by decapitating for example. Then we have a bunch of floating lineages essentially not ending with a common ancestor event. What is the likelihood of such a lineage given that the likelihood computation, from my understanding, checks whether given a coalescence and recombination rate what the next out of those two events is, and what the likelihood is associated with that.

JereKoskela:
Sounds like a plan to me.

JereKoskela:
If I'm reading the new test right, then I think there is an issue we need to solve before merging. It looks like this likelihood will return a multiple roots error for any ARG with more than one root. That will include nearly all msprime realisations, even with `full_arg = True`.

Before merging, we should get this to a stage where multiple common ancestor roots don't throw an error. Else we're rendering the likelihood function effectively unusable.

GertjanBisschop:
Yep was about to say the same. I agree with @JereKoskela. Modifying the computation to say that we stopped before another event happened is the way to go. I'll create an issue.

GertjanBisschop:
This will return an error in the presence of any local tree with more than one root. This root does not need to be the same across local trees, so the graph does not need to have a single root. Or am I misunderstanding your comment?

JereKoskela:
Ok, that sounds better. I was misunderstanding. In that case let's merge and continue in the issue you just opened.

PR #2238, opened 2023-12-12
avoid negative probs, closes #2237

### LINKED ISSUE #2237: TPM microsats model transition matrix error
When specifying the following `MicrosatMutationModel` I get an "Each row of the transition matrix must be non-negative and sum to one" error.
```{python}
lo = 50
hi = 500
m = 0.25
p = 0.5
model = msprime.TPM(p=p, m=m, lo=lo, hi=hi)
```
I just followed the instructions from the docs to generate a random model. If this is not a valid parameter combination then this should have been caught I think. @andrewkern any idea what is going on here?

petrelharp:
See discussion at #2237 - maybe there's a better way to do this, but at least this works for this example?


GertjanBisschop (approved):
Cool! Thanks @petrelharp. Can't think of a better way to tackle this either.

PR #2239, opened 2023-12-12
pin stdpopsim

GertjanBisschop:
Pinning stdpopsim version until we can align the tested demographic models in msprime again with those in stdpopsim.


jeromekelleher:
I see - we are already pinning to stdpopsim 0.1.2 for tests

jeromekelleher (approved):

Issue #2158, opened 2023-02-21
Define global stopping condition


jeromekelleher:
In #2157 we decided to add a boolean flag ``stop_at_local_mrca`` which states whether we should keep simulating the ancestry of genome segments after an mrca has been reached.

We also want some way of specifying the global stopping condition.

Note that we already have an [end_time](https://tskit.dev/msprime/docs/stable/ancestry.html#stopping-simulations-early) parameter, which currently allows us to stop simulations early, and could be used as the global stopping condition if we turn off the current default of stopping when all intervals have a local mrca.

I guess there's two options. 

1. We could have a ``stop_condition`` parameter which takes a string, and could have things like ``grand_mrca``, ``all_local_mrcas``, ``time`` (used in conjunction with ``end_time`` above)
2. Have a single boolean parameter which turns off the current stop condition, like ``stop_at_all_local_mrcas`` or something?

Option (1) seems neater. We'd have something like
```
ts = msprime.sim_ancestry(10,
      stop_at_local_mrcas=False,
      stop_condition="time",
      end_time=100)
```
to record the full process back to time 100, regardless of whether coalescence occurs or not?

PR #2248, opened 2023-12-19
global stop condition

GertjanBisschop:
Initial draft for changes discussed in #2158 to enable simulating beyond local roots.


jeromekelleher:
@hossam26644 - we discussed doing something like this a few weeks ago. It's worth having a look back through the linked issues etc to see if this is something that we want to do for the 1.4.0 release.

PR #2267, opened 2024-02-19
Disallow unstable beta coalescents

### LINKED ISSUE #2256: Discrepancy between expected and observed coalescence time in beta coalescent for large values of alpha
Hey!

There appears to be an upward bias in the simulated coalescence times of the beta coalescent. This is relative to my expectation based on the time scaling provided in the documentation. The bias starts to be noticeable from `alpha=1.8` and increases substantially for values closer to 2.

Relative error between observed and expected coalescent times plotted against alpha:
image

Code to generate figure:
```python
import multiprocessing as mp

import matplotlib.pyplot as plt
import msprime
import numpy as np
import scipy


def compute_beta_timescale(pop_size, alpha, ploidy):
    """
    Compute the generation time for the beta coalescent exactly as done in msprime when testing.
    https://github.com/tskit-dev/msprime/blob/804e0361c4f8b5f5051a9fbf411054ee8be3426a/verification.py#L3447
    """

    if ploidy > 1:
        N = pop_size / 2
        m = 2 + np.exp(
            alpha * np.log(2) + (1 - alpha) * np.log(3) - np.log(alpha - 1)
        )
    else:
        N = pop_size
        m = 1 + np.exp((1 - alpha) * np.log(2) - np.log(alpha - 1))
    ret = np.exp(
        alpha * np.log(m)
        + (alpha - 1) * np.log(N)
        - np.log(alpha)
        - scipy.special.betaln(2 - alpha, alpha)
    )
    return ret


def simulate_tree_height(alpha):
    """
    Simulate genealogies under the beta coalescent and compute the average tree height.
    Since we only have two lineages, the coalescence time should coincide with `compute_beta_timescale ()`.
    """

    # simulate genealogies under the beta coalescent
    g = msprime.sim_ancestry(
        samples=2,
        num_replicates=100000,
        model=msprime.BetaCoalescent(alpha=alpha),
        ploidy=1
    )

    return np.mean([ts.first().total_branch_length / 2 for ts in g])


if __name__ == "__main__":
    alphas = np.linspace(1.99, 1.999, 30)

    # simulate average tree heights in parallel
    with mp.Pool() as pool:
        heights_observed = np.array(pool.map(simulate_tree_height, alphas))

    # compute theoretical tree heights
    heights_theoretical = np.array([compute_beta_timescale(1, alpha, 1) for alpha in alphas])

    # compute relative difference between observed and theoretical
    diff_rel = (heights_observed - heights_theoretical) / heights_theoretical

    # plot relative difference against alpha
    plt.plot(alphas, diff_rel)
    plt.margins(0)
    plt.xlabel('alpha')
    plt.ylabel('diff_rel')
    plt.show()
```

Am I perhaps missing something?

JereKoskela:
Closes #2256 (again) by restricting the valid region of the Beta-coalescent alpha-parameter to `1 < alpha <= 1.991`.

I can't make the timescale near `alpha = 1` break even for values around `1 + 1e-15`. So I've left that boundary case as it is, restricted only by `alpha > 1`. I'm amazed that it's so robust!

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -2381,6 +2381,13 @@ tree = ts.first()
 print(tree.tmrca(0,1))
 
 ```
+:::{important}
+Both the haploid and polyploid timescales collapse to zero as {math}`\alpha`
+approaches 2. For reasons of numerical stability, values above
+{math}`\alpha = 1.991` have been prohibited. The
+{ref}`Hudson coalescent<sec_ancestry_models_hudson>` provides a very good
+approximation for {math}`1.991 <= \alpha <= 2`.
+:::
 
 The Dirac-coalescent is simulated similarly in both the diploid case:
 
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -7416,7 +7416,7 @@ msp_set_simulation_model_beta(msp_t *self, double alpha, double truncation_point
 {
     int ret = 0;
 
-    if (alpha <= 1.0 || alpha >= 2.0) {
+    if (alpha <= 1.0 || alpha > 1.991) {
         ret = MSP_ERR_BAD_BETA_MODEL_ALPHA;
         goto out;
     }
--- a/lib/tests/test_ancestry.c
+++ b/lib/tests/test_ancestry.c
@@ -1590,7 +1590,7 @@ test_beta_coalescent_bad_parameters(void)
     int ret;
     msp_t msp;
     unsigned int n = 10;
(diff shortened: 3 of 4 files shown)


JereKoskela:
This time I remembered the changelog.

JereKoskela:
Ping @jeromekelleher & @Sendrowski. I think this is ready to be merged unless you have comments.

Sendrowski:
It looks good to me!

jeromekelleher (approved):
LGTM

jeromekelleher:
I guess we should pop out another point release?

PR #2270, opened 2024-03-12
Hulls

GertjanBisschop:
First pass at a general version of the SMC (SMC(k)) not based on rejection sampling.


jeromekelleher:
I've had a quick look over, and it looks great. I'll need to spend a bit more time to check the C code, but I don't expect to find much. Basically I think we can merge this pretty quickly, and then should run a bunch of simulations to try and shake out any remaining subtleties.

algorithms.py
```diff
@@ -521,6 +684,113 @@ def _make_segment(self, left, right, count):
         return seg
 
 
+class Hull:
```
jeromekelleher, line 687: A few comments would help here on these classes. Briefly, what are they and what are they for?

algorithms.py
```diff
@@ -521,6 +684,113 @@ def _make_segment(self, left, right, count):
...
+        self.insertion_order = math.inf
+
+    def __lt__(self, other):
+        return (self.x, self.insertion_order) < (other.x, other.insertion_order)
+
+    def __repr__(self):
+        return f"x:{self.x}, io:{self.insertion_order}"
+
+
+class OrderStatisticsTree:
```
jeromekelleher, line 714: Explain here that bintrees doesn't quite do what we want and we're emulating that inefficiently.

algorithms.py
```diff
@@ -679,10 +961,50 @@ def initialise(self, ts):
         for node in range(ts.num_nodes):
             seg = root_segments_head[node]
             if seg is not None:
+                left_end = seg.left
```
jeromekelleher, line 964: Would it be easier to do this stuff as a separate loop through the nodes below? So, all the stuff we do during init is under the one "if model == "smc_k"?

algorithms.py
```diff
@@ -700,6 +1022,18 @@ def change_population_growth_rate(self, pop_id, rate, time):
     def change_migration_matrix_element(self, pop_i, pop_j, rate):
         self.migration_matrix[pop_i][pop_j] = rate
 
+    def alloc_hull(self, left, right, ancestor_node):
```
jeromekelleher, line 1025: I don't get why this is called ``ancestor_node``. This must be a segment or something, if we can ``prev`` it?

jeromekelleher: I think we need a different name here all right, this is confusing me quite a bit

algorithms.py
```diff
@@ -2195,6 +2688,32 @@ def verify(self):
                         self.gc_map,
                         self.get_gc_left_bound,
                     )
+        if self.model == "smc_k":
+            self.verify_hulls()
+
+
+def intersect_hulls(a, b):
```
jeromekelleher, line 2695: What about adding an ``intersects_with`` method to Hull. so would be ``a.intersects_with(b)``?

jeromekelleher (commented):
Had a quick look through this and spotted some minor things in the python version. Still need to go through the C in more detail.

lib/msprime.c
```diff
@@ -726,7 +882,71 @@ static segment_t *MSP_WARN_UNUSED
...
+out:
+    return hull;
+}
+
+static hullend_t *MSP_WARN_UNUSED
+msp_alloc_hullend(msp_t *self, double position, label_id_t label)
+{
+    hullend_t *hullend = NULL;
+
+    if (object_heap_empty(&self->hullend_heap[label])) {
```
jeromekelleher, line 936: Same here, we're not exercising the expand logic here in the C tests (will probably happen automatically, though?)

lib/msprime.c
```diff
@@ -3407,6 +4071,24 @@ msp_reset_memory_state(msp_t *self)
                 avl_unlink_node(&pop->ancestors[label], node);
                 msp_free_avl_node(self, node);
             }
+            if (pop->hulls_left != NULL) {
```
jeromekelleher, line 4074: This case isn't being hit in tests - I think you need to run some replicates of the smc_k?

lib/msprime.c
```diff
@@ -3462,13 +4145,32 @@ msp_insert_root_segments(msp_t *self, const segment_t *head, segment_t **new_hea
             if (ret != 0) {
                 goto out;
             }
-
+            if (self->model.type == MSP_MODEL_SMC_K) {
+                if (self->state == MSP_STATE_NEW) {
```
jeromekelleher, line 4149: This case isn't tested - what is it we're testing against?

I also wonder about having these type of "model specific" checks deep inside a loop. I think it's better to just do another loop later, just over the individuals (maybe)

lib/msprime.c
```diff
@@ -3746,6 +4448,64 @@ msp_apply_fixed_events(msp_t *self, double time)
...
+        hullend_target.position = left;
+        hullend_target.insertion_order = UINT64_MAX;
+        avl_search_closest(hulls_right, &hullend_target, &search);
+        if (search != NULL) {
+            floor_ptr = (hullend_t *) search->item;
+            if (floor_ptr->position > left) {
+                search = search->prev;
+            }
+            if (search != NULL) {
+                num_ending_before_hull = (uint64_t) avl_index(search) + 1;
```
jeromekelleher, line 4484: This isn't being hit in C tests

lib/msprime.c
```diff
@@ -6752,6 +7515,146 @@ msp_std_common_ancestor_event(
...
+    /* In very large simulations, the fenwick tree used as an indexing
+     * structure for genomic segments will experience some numerical
+     * drift, where the indexed values diverge from the true values
+     * associated with segments. We ensure that this drift does not
+     * become too large by rebuilding the indexing structure every
+     * now and again. */
+    mass_index = &self->populations[pop_id].coal_mass_index[label];
+    tsk_bug_assert(mass_index != NULL);
+    if (fenwick_rebuild_required(mass_index)) {
+        fenwick_rebuild(mass_index);
```
jeromekelleher, line 7589: This isn't hit in tests - how did we make sure it is in the other cases?

lib/msprime.c
```diff
@@ -7248,6 +8151,18 @@ msp_set_simulation_model(msp_t *self, int model)
             goto out;
         }
     }
+    /* SMC_K logic */
+    ret = msp_setup_smc_k(self);
```
jeromekelleher, line 8155: Why are we calling setup_smc_k outside of the check for whether it's the SMC_K?

lib/msprime.c
```diff
@@ -7270,6 +8185,28 @@ msp_set_simulation_model_smc_prime(msp_t *self)
     return msp_set_simulation_model(self, MSP_MODEL_SMC_PRIME);
 }
 
+int
+msp_set_simulation_model_smc_k(msp_t *self, double hull_offset)
+{
+    int ret = 0;
+
+    if (hull_offset < 0) {
+        ret = MSP_ERR_BAD_HULL_OFFSET;
```
jeromekelleher, line 8194: Not covered in C tests

jeromekelleher (commented):
Generally looks great @GertjanBisschop. I think we just need to beef up the C testing a bit - we're aiming for all reachable code branches to be covered by tests here. Otherwise, we can't be sure that valgrind is seeing those cases and we're not leaking memory or whatever.

So, everything except an unprovokable error (like MSP_ERR_NO_MEMORY) should be covered.

lib/msprime.c
GertjanBisschop: Good point. I'll restructure the model specific checks.
What I am testing against is the fact that one could instantiate a simulator using the lowlevel or C API with the model directly set to 'smc_k'. However, this is not the logic that is followed when running `sim_ancestry()` for example. Here we always initialise the simulator as if the model was 'hudson' first and only set the model once we start simulating. I have therefore not implemented any smc_k initialisation steps for the non-standard case. The second if-statement catches this cases.

lib/msprime.c
```diff
@@ -972,12 +1242,178 @@ msp_free_segment(msp_t *self, segment_t *seg)
...
+    /* step 2: num ending before hull->left */
+    hulls_right = &self->populations[u->population].hulls_right[u->label];
+    query.position = hull->left;
+    query.insertion_order = UINT64_MAX;
+    if (hulls_right->head == NULL) {
+        num_ending_before_left = 0;
+    } else {
+        c = avl_search_closest(hulls_right, &query, &query_node);
+        /* query < node->item ==> c = -1 */
+        num_ending_before_left = (uint64_t) avl_index(query_node) + (uint64_t)(c != -1);
```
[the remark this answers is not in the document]
GertjanBisschop: good point. I hadn't realised this while coding this up. This seriously limits the number of samples we can simulate.

lib/msprime.c
GertjanBisschop: The `msp_insert_root_segments` logic only gets called for the SMC_K when we add in historical samples. This approach will only work though if the inserted root segments are sorted based on left coordinate (or if they all have the same left coordinate).

lib/msprime.c
GertjanBisschop: This is due to the fact that when we reset the state we reset to 'hudson' (= initial_model) first before setting the model to smc_k. This is a remainder of the code from when before I realised that when going through the python API we always reset to 'hudson' first.

lib/msprime.c
GertjanBisschop: note to self: this requires setting up an initial state where the samples do not all span entire sequence length.

algorithms.py
```diff
@@ -2166,6 +2652,42 @@ def verify_mass_index(self, label, mass_index, rate_map, compute_left_bound):
...
+                    assert max(pop.hulls_right[label].rank.values()) == num_lineages - 1
+                # verify counts in avl tree
+                count = 0
+                for a, b in itertools.combinations(pop._ancestors[label], 2):
+                    # make_hulls:
+                    a_hull = make_hull(a, self.L, self.hull_offset)
+                    b_hull = make_hull(b, self.L, self.hull_offset)
+                    count += a_hull.intersects_with(b_hull)
+                avl_pairs = avl_count_pairs(pop.hulls_left[label])
+                print(avl_pairs, count)
```
jeromekelleher, line 2672: print

jeromekelleher (commented):
LGTM - I think we should merge and start working with it!

jeromekelleher (approved):

PR #2271, opened 2024-03-25
Clarify the meaning of the mutation rate when a matrix is used

hyanwong:
In particular, that the exact transition is determined once a mutation has been laid down (so that the rate is the sum of all the possible mutation rates at that site).

This is right, isn't it @GertjanBisschop ?


GertjanBisschop:
The [next section](https://tskit.dev/msprime/docs/stable/mutations.html#parameterisation-of-matrix-mutation-models) explains this quite well I think @hyanwong . Have a look to see if you still think we need to be more concrete in the section you modified as well.

hyanwong:
Ah, good catch, I didn't properly read that.

I think it could be useful to say that the rate should be the (poisson) sum of all the separate sorts of mutation that could occur?

PR #2273, opened 2024-03-30
add record_provenance argument to sim_mutations; closes #2272

### LINKED ISSUE #2272: no `record_provenance` argument to sim_mutations
The following code has per-bp resolution on the RateMap:
```
L = 1e5
ts = msprime.sim_ancestry(1000, sequence_length=L, recombination_rate=1e-8, population_size=1e4)

rm = msprime.RateMap(
    position=np.linspace(0, L, int(L+1)),
    rate=rng.exponential(size=int(L))*1e-8
)
mts = msprime.sim_mutations(ts, rate=rm)
```
and produces the warning:
```
The provenance information for the resulting tree sequence is 3.08MB. This is nothing to worry about as provenance is a good thing to have, but if you want to save this memory/storage space you can disable provenance recording by setting record_provenance=False
```
However, this isn't possible, as [sim_mutations](https://github.com/tskit-dev/msprime/blob/main/msprime/mutations.py#L1555) doesn't have that option. This might be an issue in practice, for someone wanting to do chromosome-scale per-bp resolution (which should only take ~1hr with human-like parameters, so is not crazy!).

petrelharp:
I just followed what's done in [ancestry.py](https://github.com/tskit-dev/msprime/blob/main/msprime/ancestry.py#L573) here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
           # we can invalidate the cache when we need to. Note that we need
           # to update the key in the ``save_cache`` block below also when
           # doing that.
-          key: msprime-{{ .Branch }}-v4
+          key: msprime-{{ .Branch }}-v5
       - run:
           name: Checkout submodules
           command: |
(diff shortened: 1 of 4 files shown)


jeromekelleher (approved):
Lgtm. Should we update the changelog?

Issue #2280, opened 2024-04-29
Enable gene conversion for SMC_K model


GertjanBisschop:
The general idea of how this should be done is already incorporated in `algorithms.py`, what's left is implementing this in C and adding a fair amount of tests.

GertjanBisschop:
Solved in #2282.

PR #2277, opened 2024-04-25
Fix OSX tests

benjeffery:

PR #2282, opened 2024-05-09
Smc(k) gene conversion

GertjanBisschop:
Implementation of gene conversion using SMC K logic. 
`alloc_hull` function was adapted and no longer automatically applies the `hull_offset`. This means there are now multiple `min(self.sequence_length, hull_right + self.hull_offset)` statements but it makes reasoning about the right endpoint of the hull easier in all other cases.

Will check if we need some additional statistical tests.
Solves #2280.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -918,8 +918,8 @@ msp_alloc_hull(msp_t *self, double left, double right, segment_t *lineage)
     tsk_bug_assert(hull->id < UINT_MAX * self->num_populations);
 
     hull->left = left;
-    hull->right = GSL_MIN(
-        right + self->model.params.smc_k_coalescent.hull_offset, self->sequence_length);
+    tsk_bug_assert(right <= self->sequence_length);
+    hull->right = right;
     hull->lineage = lineage;
     hull->count = 0;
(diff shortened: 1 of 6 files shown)


GertjanBisschop:
@benjeffery does this require merging #2277? The PR seems to not only affect the conda setup for MacOS.

benjeffery:
I'll try and get that one merged, not sure what is wrong with it.

lib/msprime.c
```diff
@@ -3466,6 +3492,18 @@ msp_gene_conversion_event(msp_t *self, label_id_t label)
         new_individual_head = head;
     }
     if (new_individual_head != NULL) {
+        if (self->model.type == MSP_MODEL_SMC_K) {
+            tsk_bug_assert(tract_hull_left < tract_hull_right);
+            tract_hull_right = GSL_MIN(
+                tract_hull_right + self->model.params.smc_k_coalescent.hull_offset,
+                self->sequence_length);
+            hull = msp_alloc_hull(
```
jeromekelleher, line 3500: Probably need to check ``hull == NULL`` here for MSP_ERR_NO_MEMROY?

jeromekelleher (approved):

PR #2286, opened 2024-06-20
Support numpy >= 2, drop Python 3.8

jeromekelleher:
Also add support for Python 3.12


jeromekelleher:
OK, looks like this mostly just works. We're successfully compiling against numpy 2.0, and running tests against numpy 1.x. We can't really test against numpy 2.0 at runtime until there's a tskit package.

Note the last commit isn't doing what it should here because we're pulling in multiple versions of numpy, and a 1.x version is coming from pip for some reason. I think we should try to reduce the number of packages coming from pip here, hopefully most of what we need is on conda-forge now.

Going to park this until a tskit package is available.

Issue #2294, opened 2024-07-17
Add Lineage/lineage_t concept


jeromekelleher:
Currently we store the head of the segment chain to represent a lineage in both the Python and C versions. This is getting increasingly unwieldly. Replace with an object representing a lineage, and move the lineage specific things like ``population`` and ``label`` into this structure.

This will be a noisy change, but a worthwhile one.

FYI I'm on this one @GertjanBisschop

jeromekelleher:
This is partially done, with #2296 creating the lineage_t struct, and #2297 moving the label, population and hull attributes to this struct in the minimal working way. Follow ups will start to take advantage of this cleaner structure to refactor.

jeromekelleher:
This is progressing as-of #2303 but still needs some work to complete. Some todos:

- [ ] There are redundant calls to lineage_reset_segments, probably mainly in the GC code path. If we fix this, we should be able to avoid passing ``new_right`` to msp_reset_hull_right. This also needs to be done in the Python implementation (see Lineage.reset_segments())
- [ ] The situation with lineages/segments is pretty confusing for DTWF and pedigree simulations. We should add some comments discussing this at a minimum.

jeromekelleher:
Notes on the checklist above - the first task has been shelved for issue #2386.

We should do some documentation on the second part all right.

PR #2296, opened 2024-07-23
Simpler lineage struct

jeromekelleher:
Work towards #2294

Work on adding a lineage struct that we store in the algorithm state in preference to using just segments. This does the initial bit where we just define this structure minimally in C and Python. Actually using it to some advantage can be done in follow-ups.

### DIFF
--- a/algorithms.py
+++ b/algorithms.py
@@ -2,6 +2,7 @@
 Python version of the simulation algorithm.
 """
 import argparse
+import dataclasses
 import heapq
 import itertools
 import logging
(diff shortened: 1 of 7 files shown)


jeromekelleher:
I think this is the basic changes needed to start with. Would you mind having a quick look over to see if I've made any obvious mistakes (like keeping an msp_verify() left over from debugging etc) please @GertjanBisschop?

jeromekelleher:
I'm going to go ahead and merge this, there's no point in keeping a big diff like this lying around. Sorry to anyone who's code is clobbered!

PR #2300, opened 2024-08-01
Insert merged lineage

jeromekelleher:
More work on #2294 

Factor out a bunch of repetitive and complicated code from the key merge_ancestors functions.


jeromekelleher:
Low coverage in diff can't be helped here, a large fraction of the update is unreachable

PR #2303, opened 2024-08-02
Factor out distributed SMC logic

jeromekelleher:
Bring SMC k logic together in one place as part of #2294


jeromekelleher:
I'm going to merge this as it should be significant improvement in terms of code readability, and I think it also fixes a few memory leaks and performance issues. I'll follow up with issues for what remains in order to release the SMC(k) code.

PR #2309, opened 2024-08-06
Multiple merger polyploid scaling

JereKoskela:
Ping @jeromekelleher. Adding tests for polyploid scaling for multiple merger coalescents turned out to be easy, but also sent me down a bit of a rabbit hole. Summary of changes:

- The extra spaces on various lines in `msprime.c` which are unrelated to anything I've done (1396, 1899...) were put in by clang. Let me know if you want me to do something about them.
- For the Dirac coalescent, population growth at rate `g` should yield a coalescent growth rate of `2g`,essentially because the coalescence probability in the Moran model is `1/N^2` rather than `1/N`. We had overlooked that, but it is now fixed. 
- Ploidy wasn't being handled quite right in the Dirac coalescent either, but is now fixed. In the process, I got rid of some `if (ploidy == 1) {...} else {...}` statements which became redundant.
- For both the Dirac and Beta coalescents, I moved the allocation of the ancestor buckets `Q` later into the rejection sampler. It is now only done once a multiple merger event is definitely taking place.
- The Dirac coalescent test in `test_ancestry.py` started failing after the ploidy fix because the rate of multiple mergers wasn't high enough to guarantee one anymore. I increased it.
- I also noticed we were disallowing `psi = 1` in the Dirac coalescent for no reason. The documented (and sensible) range is `0 < psi <= 1`. I tweaked the boundary values and associated tests to make the code consistent with the documentation.

### DIFF
--- a/msprime/simulations.py
+++ b/msprime/simulations.py
@@ -2034,8 +2034,8 @@ class BetaCoalescent(ParametricSimulationModel):
 
     name = "beta"
 
-    def __init__(self, reference_size=1, alpha=None, truncation_point=1):
-        self.reference_size = reference_size
+    def __init__(self, reference_size=None, alpha=None, truncation_point=1):
+        super().__init__(reference_size=reference_size)
         self.alpha = alpha
         self.truncation_point = truncation_point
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Sorry, making a mess of git stuff here, I'll fix it up

PR #2325, opened 2024-10-24
Test_pedigree with internal samples

abureau:
I have modified TestSimulateThroughPedigree and the simulate_pedigree function to specify generations of subjects as internal samples. Setting internal_sample_gen=[True, False] on line 562 specifies that founders i.e. parents of a two-generation pedigree are samples, but not their children (probands). This results in Fatal Python error: Aborted with core dump in my setup.

PR #2326, opened 2024-10-24
Test_pedigree.py with internal samples

abureau:
I have modified TestSimulateThroughPedigree and the simulate_pedigree function to specify generations of subjects as internal samples. Setting internal_sample_gen=[True, False] on line 562 specifies that founders i.e. parents of a two-generation pedigree are samples, but not their children (probands). This results in Fatal Python error: Aborted with core dump in my setup.

### DIFF
--- a/tests/test_pedigree.py
+++ b/tests/test_pedigree.py
@@ -48,6 +48,7 @@ def simulate_pedigree(
     num_generations=3,
     sequence_length=1,
     random_seed=42,
+    internal_sample_gen=(None,None,None),
 ) -> tskit.TableCollection:
     """
     Simulates pedigree.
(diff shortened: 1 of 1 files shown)


abureau:
I continue here the discussion of [the closed pull request](https://github.com/tskit-dev/msprime/pull/2325). I inserted the command
tables.dump("test_shallow_internal.ts") 
at line 564 of test_pedigree.py. Then I called
python3 algorithms.py --from-ts test_shallow_internal.ts 4 test.ts
and got no error message. The tree object in the output test.ts seems fine, but I am not sure what to inspect.

No rush to reply, I just report now for later.

PR #2341, opened 2025-03-05
fix wrong citation (78 instead of 73)

hossam26644:
that's the cited paper: https://doi.org/10.1073/pnas.75.6.2868
It is cited correctly in the docstring of the python method


benjeffery:
Congrats your your first msprime PR!

PR #2359, opened 2025-05-02
skip msp verify hulls for testing

hossam26644:


jeromekelleher:
I think the better approach here is to comment out msp_verify where ever it's being called in this file. This will kill performance on any of the models.

hossam26644:
@jeromekelleher what about I just add a return statement in the first line of msp verify?
msp verify is called many times in the test files; they could be left to run for testing, I can comment all the calls in the test files(i do not like this), or just cancel msp verify by the return statement I suggested.

jeromekelleher:
This was how it's designed - it should never be called in production code, it's meant to be used during development and called explicitly from the C test suite. So, just comment out/delete whereever it's currently getting called in the production code. It's probably just a few lines where debugging stuff has been left in place.

jeromekelleher:
Oh, I see - were you running all the evaluations through the dev CLI? Let's not do this,  the dev-cli isn't for profiling it's strictly for development (and almost never used now anyway).

So - I don't think we need this change, you just need to stop using the dev CLI and run through Python instead.

PR #2370, opened 2025-05-23
Allow Demography init from provenance with preset pop IDs

hyanwong:
The last step to allow demographies to be instantiated from provenance entries. We exceptionally allow a `demography.Population` object to be created with an existing ID, but only if generated when decoding provenance JSON. This ensures that the IDs in the events etc match with what we expect. It's a slight hack in that we check for the error message. The alternative is to pick a different error (?RuntimeError) or make our own bespoke error that can be allowed through in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Support for simulations after local mrca,
   ({issue}`2157`, {pr}`2396`, {user}`jeromekelleher`, {user}`hossam26644`).
 - Add wheels on Windows ({pr}`2414`, {issue}`2200`,{user}`benjeffery`)
+- Demography objects can now be created from provenance entries ({pr}`{2369}`, {user}`hyanwong`)
 
  **Performance improvements**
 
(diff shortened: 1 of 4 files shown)


msprime/provenance.py
```diff
@@ -188,10 +189,21 @@ def hook(obj):
...
-                return getattr(module, cls)(**obj)
+                cls = getattr(module, cls)
+                if cls == demography.Demography:
+                    # the Demography class normally generates its own population IDs, but
+                    # allow fixed IDs here to ensure they match e.g. IDs in event objects
+                    demography_object = object.__new__(cls)
+                    try:
+                        demography_object.__init__(**obj)
+                    except ValueError as e:
+                        if str(e).startswith("Population ID should not be set"):
```
jeromekelleher, line 202: This control flow by exception logic is fragile and distributes the logic of creating Demography objects around different modules. We should add a ``Demography.fromdict`` method that does this, and can be tested independently;.

jeromekelleher (commented):
I think we need a different way of instantiating Demography objects

msprime/provenance.py
hyanwong: I'm returning to some of my open PRs, and realise that this one has an easy fix, because I just found out that you can pass an InitVar parameter to a dataclass, in this case to say not to worry about making new population IDs in this specific case. I think this is probably the minimally intrusive thing to do.

PR #2376, opened 2025-07-21
specify units for recombination rate; closes #2375

### LINKED ISSUE #2375: Unit used for `recombination_rate` when specifying a single value
Hi, thanks for msprime! The `recombination_rate` argument of  `sim_ancestry` is a bit ambiguous when specifying a single value. [The docstring states:](https://github.com/tskit-dev/msprime/blob/main/msprime/ancestry.py#L1165)

> The rate of recombination along the sequence; can be either a single value (specifying a single rate over the entire sequence)...

I assume that the intention is to calculate the rate as Morgans / base positions?

petrelharp:
I'm not clear this improves things, since what exactly the units are depends? (like, probably "per generation per bp" but only if that's what time and sequence length are measured in)

Hm, I should have done this on my branch (I did this just by clicking "edit"; probably we should disable that, since the default resulting option was "commit this directly to main".)

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1162,7 +1162,8 @@ def sim_ancestry(
         case multiple events at precisely the same genome location are very
         unlikely (but technically possible).
         See the :ref:`sec_ancestry_discrete_genome` section for usage examples.
-    :param recombination_rate: The rate of recombination along the sequence;
+    :param recombination_rate: The rate of recombination along the sequence,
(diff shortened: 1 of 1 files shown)


timothymillar:
Thanks @petrelharp and @jeromekelleher. I think a note in the ancestry docs would do it. I found the ancestry docs very quickly when looking for more info.

jeromekelleher:
Closed in favour of #2391

Issue #2383, opened 2025-07-23
Merge the hullend_t struct into hull_t


jeromekelleher:
We're currently keeping a separate struct for hull right values, but we could just use one struct with all the information needed. This would simplify the memory management a bit.

May not be worth the work, but worth looking at.

PR #2384, opened 2025-07-23
Remove hullend struct

jeromekelleher:
Initial work on #2383 

As I work through it, I wonder if we can save some time on AVL tree maintenance by storing pointers to the left and right AVL tree nodes in the hull. It's not much space and would simplify things quite a bit.

Should probably work it out on the Python implementation first though.


jeromekelleher:
This also embeds the AVL nodes needed by the hull in the left and right AVL trees, which significantly reduces the need for searching around and simplifies the code.

PR #2385, opened 2025-07-24
Embed avl into lineage struct

jeromekelleher:
Following the ideas in #2383, we can also embed an avl_node_t in the lineage_t struct and save some memory management and AVL lookups.


jeromekelleher:
This gives another significant perf bump for the SMC(k) and may lead to modest improvements in some other cases (unlikely to be significant, though)

PR #2396, opened 2025-08-05
Feat:/simulate after local mrca

hossam26644:
Here you go: @jeromekelleher 
Still missing some tests, and I want to change the flag name of the python algorithm to match the msprime implementation.
As a reminder, this implementation consumes more memory than the when stopping at the youngest root. It could be the node table getting much larger, and could be a memory leak.

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -329,6 +329,13 @@ out:
 }
 
 int
+msp_set_stop_at_local_mrca(msp_t *self, bool stop_at_local_mrca)
+{
+    self->stop_at_local_mrca = stop_at_local_mrca;
+    return 0;
+}
+
+int
 msp_set_discrete_genome(msp_t *self, bool is_discrete)
 {
     self->discrete_genome = is_discrete;
@@ -1030,6 +1037,7 @@ msp_alloc(msp_t *self, tsk_table_collection_t *tables, gsl_rng *rng)
     self->store_migrations = false;
     self->additional_nodes = 0;
     self->coalescing_segments_only = true;
+    self->stop_at_local_mrca = true;
     self->avl_node_block_size = 1024;
     self->node_mapping_block_size = 1024;
     self->segment_block_size = 1024;
@@ -3787,6 +3795,13 @@ msp_merge_two_ancestors(msp_t *self, population_id_t population_id, label_id_t l
         goto out;
     }
 
+    uint32_t min_overlap;
+    if (self->stop_at_local_mrca) {
+    min_overlap = 2;
+    } else {
+    min_overlap = 0;
+    }
+
     x = a;
     y = b;
     /* Keep GCC happy */
@@ -3862,7 +3877,7 @@ msp_merge_two_ancestors(msp_t *self, population_id_t population_id, label_id_t l
                 node = avl_search(&self->overlap_counts, &search);
                 tsk_bug_assert(node != NULL);
                 nm = (node_mapping_t *) node->item;
-                if (nm->value == 2) {
+                if (nm->value == min_overlap) {
                     nm->value = 0;
                     node = node->next;
                     tsk_bug_assert(node != NULL);
@@ -3870,7 +3885,7 @@ msp_merge_two_ancestors(msp_t *self, population_id_t population_id, label_id_t l
                     r = nm->position;
                 } else {
                     r = l;
-                    while (nm->value != 2 && r < r_max) {
+                    while (nm->value != min_overlap && r < r_max) {
                         nm->value--;
                         node = node->next;
                         tsk_bug_assert(node != NULL);
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1411,7 +1411,7 @@ class Simulator(_msprime.Simulator):
         start_time=None,
         end_time=None,
         num_labels=None,
-        stop_at_local_mrca=None,
+        stop_at_local_mrca=True,
     ):
         # We always need at least n segments, so no point in making
         # allocation any smaller than this.
@@ -1462,6 +1462,7 @@ class Simulator(_msprime.Simulator):
             gene_conversion_tract_length=gene_conversion_tract_length,
             discrete_genome=discrete_genome,
             ploidy=ploidy,
+            stop_at_local_mrca=stop_at_local_mrca,
         )
         # Highlevel attributes used externally that have no lowlevel equivalent
         self.end_time = np.inf if end_time is None else end_time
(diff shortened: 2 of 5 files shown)


lib/msprime.c
```diff
@@ -3787,6 +3795,13 @@ msp_merge_two_ancestors(msp_t *self, population_id_t population_id, label_id_t l
         goto out;
     }
 
+    uint32_t min_overlap;
```
jeromekelleher, line 3798: local style here says that we keep variable declarations in the first block of the function. So, this would join the other delcarations above

```
uint32_t min_overlap = self->stop_at_local_mrca? 2: 0;
```

lib/msprime.c
```diff
@@ -5149,7 +5164,7 @@ msp_run_coalescent(msp_t *self, double max_time, unsigned long max_events)
         ca_pop_id = 0;
         for (avl_node = self->non_empty_populations.head; avl_node != NULL;
              avl_node = avl_node->next) {
-            pop_id = (tsk_id_t)(intptr_t) avl_node->item;
+            pop_id = (tsk_id_t) (intptr_t) avl_node->item;
```
jeromekelleher, line 5167: Seems to be a bunch of formatting noise here - which version of clang-format are you using?

msprime/ancestry.py
```diff
@@ -1078,6 +1079,26 @@ def _parse_sim_ancestry(
     random_seed = _parse_random_seed(random_seed)
     random_generator = _msprime.RandomGenerator(random_seed)
 
+    if stop_at_local_mrca is None:
+        stop_at_local_mrca = True
+    elif not isinstance(stop_at_local_mrca, bool):
```
jeromekelleher, line 1084: We probably don't need to force it to be a bool (unless that's what we do with other params). Something like
```
if stop_at_local_mrca is None:
    stop_at_local_mrca = True
if not stop_at_local_mrca:
    if end_time == math.inf or end_time is None:
           ValueError()
```

msprime/ancestry.py
```diff
@@ -1078,6 +1079,26 @@ def _parse_sim_ancestry(
...
+            "stop_at_local_mrca must be a boolean value, or None which defaults to True."
+        )
+    elif not stop_at_local_mrca:  # when set to False
+        if end_time == math.inf or end_time is None:
+            raise ValueError(
+                "You have to specify an end_time when using stop_at_local_mrca, "
+                "otherwise the simulation will run indefinitely."
+            )
+
+        if (recombination_rate in (None, 0)) and (gene_conversion_rate in (None, 0)):
```
jeromekelleher, line 1095: Why do we require this? It's not going to do any harm if someone specifies for a single locus sim is it?

jeromekelleher (commented):
Looks good!

Some minor comment above.

Also needed:

1. Some C unit tests. We should be able to add a test or two to test_ancestry.c to cover the functionality. Nothing elaborate, just making sure the code is covered by valgrind.
2. We also need to cover the other models. This will require updating msp_merge_ancestors in a similar way. Best to start with algorithms.py on this.

msprime/ancestry.py
hossam26644: It runs indefinitely otherwise

jeromekelleher:
I'm going to mute this thread @hossam26644 - please ping me when you'd like a review

hossam26644:
@jeromekelleher check now please, and please check the implementation of merge_ancestors. It works and passes the tests that I put.. but I am not sure if I am missing something

tests/test_ancestry.py
```diff
@@ -890,6 +890,35 @@ def test_str(self):
...
+            model=model,
+        )
+
+        old_time = None
+        for tree in ts.trees():
+            assert len(tree.roots) == 1  # otherwise the test is not valid
+            u = tree.roots[0]
+            assert (
+                tree.time(u) >= end_time
+            )  # makes sure that local simulations only stop at end time
```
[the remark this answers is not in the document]
jeromekelleher: More concisely:
```
root_times = [tree.time(tree.root) for tree in ts.trees()]
assert len(set(root_times)) == 1
assert root_times[0] >= max_time
```
calling tree.root when there's multiple roots raises an error, so you don't need to worry about that.

tests/test_ancestry.py
```diff
@@ -2530,6 +2559,44 @@ def test_bad_additional_node_flag(self):
...
+                2, stop_at_local_mrca=False, recombination_rate=0.1, sequence_length=10
+            )
+
+        with pytest.raises(
+            ValueError,
+            match="stop_at_local_mrca is only supported for simulations with "
+            "recombination or gene conversion.",
+        ):
+            msprime.sim_ancestry(2, stop_at_local_mrca=False, end_time=3)
+
```
jeromekelleher, line 2580: What are these two calls for?

jeromekelleher (commented):
Generally looking good. Needs more models though, in particular the pedigree model. You need to construct a pedigree in which full coalescence will happen and verify that you still go through the rest of the nodes (you'll need store_unary as well I guess, but that's true anyway?)

tests/test_ancestry.py
```diff
@@ -890,6 +890,35 @@ def test_str(self):
         s = str(sim)
         assert len(s) > 0
 
+    @pytest.mark.parametrize("model", ["dtwf", "hudson"])
```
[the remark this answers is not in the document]
hossam26644: Ok, it does not work with sweeps. I will turn it off

tests/test_ancestry.py
```diff
@@ -2530,6 +2559,36 @@ def test_bad_additional_node_flag(self):
...
+    def test_stop_at_local_mrca_no_end_time(self, model):
+
+        with pytest.raises(
+            ValueError,
+            match="You have to specify an end_time when using stop_at_local_mrca, "
+            "otherwise the simulation will run indefinitely.",
+        ):
+            msprime.sim_ancestry(
```
jeromekelleher, line 2570: doesn't look like you're using ``model`` here?

hossam26644: Let's discuss the pedigree model in detail tomorrow. But currently, simulations stop after coalescence with stop_at_local_mrca is False. In the manual, to simulate after pedigree coalescence another model is to be used, dtwf for example. In this case, this implementation is sufficient. 
Practically, in algorithms.py, climb pedigree simulates between individual nodes, so I don't get the meaning of simulating above root in this case --  you can explain it to me tomorrow in our meeting.

jeromekelleher:
Hmm - looks like there's a basic flaw in the logic for the multi-merger case, which wasn't really being tested. I'll need to spend more time on this to get to the bottom of it.

jeromekelleher:
OK, I think this is basically done. Can you review please @hossam26644? I'm going to log some follow-up issues.

jeromekelleher (commented):
OK, thanks for the review. I'll push an update and merge.

jeromekelleher (approved):

jeromekelleher:
Going to let #2432 go through first, then resolve merge conflicts

Issue #2378, opened 2025-07-23
Add gene conversion to smc(k) verification tests


jeromekelleher:
Similar to #2377, we should add a simple test that we get the same distribution of nodes and trees etc when we run the standard smc/smc' models and the equivalent smc(k) implementations.

Issue #2377, opened 2025-07-22
Add multi population SMC_k verification tests


jeromekelleher:
We need to test that we're doing the right thing for the SMC_k when we have some population structure.

We should add a few different demographic models to check this all works OK, with and without recombination.

These should be added to the SmcKTest class. We can just check msprime.SmcKApproxCoalescent against the existing smc and smc_prime models.

Issue #2379, opened 2025-07-23
Add simulating from an initial condition to smc(k) tests


jeromekelleher:
Like #2378 and #2377, check that we get the same distribution of usual values when we complete a simulation from a given starting condition using smc and smc(k) equivalent. Can get that starting condition by running a standard hudson simulation for a short time.

PR #2398, opened 2025-08-11
smck_Verification

hossam26644:
@jeromekelleher check these tests (including a gc)
solves #2377  and #2378


hossam26644:
now should also solve #2379

PR #2400, opened 2025-08-16
Fix asv build error

hossam26644:
For later asv versions (tested on 0.6.4) it fails to build the project without meaningful errors. It is suggested online to add build parameters to asv.conf.json or use version <= 0.6.1 (The second solution is simpler)


hossam26644:
@benjeffery check these now. and the website can be taken down now actually

PR #2410, opened 2025-09-03
Selective Sweep based on reading from a file with rudimentary ability to…

Aalhad-Bhatt:
… to run on multiple demes

Code to run a selective sweep based on reading the events from a binary file in a specified format. The C side also implements nearest neighbour migration with multiple demes. This does not interface with the Python version yet.

### DIFF
--- a/lib/msprime.h
+++ b/lib/msprime.h
@@ -43,8 +43,9 @@
 #define MSP_MODEL_DIRAC 4
 #define MSP_MODEL_DTWF 5
 #define MSP_MODEL_SWEEP 6
-#define MSP_MODEL_WF_PED 7
-#define MSP_MODEL_SMC_K 8
+#define MSP_MODEL_SWEEP_REVERSE 7
+#define MSP_MODEL_WF_PED 8
+#define MSP_MODEL_SMC_K 9
 
 /* Exit codes from msp_run to distinguish different reasons for exiting
  * before coalescence. */
@@ -231,6 +232,11 @@ typedef struct _sweep_t {
     void (*print_state)(struct _sweep_t *self, FILE *out);
 } sweep_t;
 
+typedef struct _sweep_reverse_t {
+    double position;
+    const char *filename;
+} sweep_reverse_t;
+
 typedef struct _simulation_model_t {
     int type;
     union {
@@ -238,6 +244,7 @@ typedef struct _simulation_model_t {
         beta_coalescent_t beta_coalescent;
         dirac_coalescent_t dirac_coalescent;
         sweep_t sweep;
+        sweep_reverse_t sweep_reverse;
     } params;
     /* If the model allocates memory this function should be non-null. */
     void (*free)(struct _simulation_model_t *model);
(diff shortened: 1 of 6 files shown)


lib/tests/test_sweeps.c
```diff
@@ -370,6 +370,61 @@ sweep_genic_selection_mimic_msms_single_run(unsigned long int seed)
...
+
+    // To mimic the verfication.py call
+    msp_set_discrete_genome(&msp, 0);
+    msp_set_gene_conversion_rate(&msp, 0);
+    msp_set_gene_conversion_tract_length(&msp, 1);
+    msp_set_avl_node_block_size(&msp, 65536);
+    msp_set_node_mapping_block_size(&msp, 65536);
+    msp_set_segment_block_size(&msp, 65536);
+
+    ret = msp_set_simulation_model_sweep_genic_selection_reverse(
```
jeromekelleher, line 410: So, here instead of passing in the filename we would pass parameters required to describe the sweep.

lib/msprime.h
```diff
@@ -231,13 +232,19 @@ typedef struct _sweep_t {
     void (*print_state)(struct _sweep_t *self, FILE *out);
 } sweep_t;
 
+typedef struct _sweep_reverse_t {
+    double position;
+    const char *filename;
```
jeromekelleher, line 237: Here we need to include the paramters required to describe the sweep, so from your code above:
```
1. Number of steps
2. Number of demes
3. Pop size
4. Migration rate
5. State after finishing the backward sim
6. state at the start of the backward sim
8. Time, type, and start and end locations for each event
```

So, I think 1-6 are just standard variables and the you have 4 arrays to describe the events? This would looks something like
```
typedef struct _sweep_reverse_t {
    double position;
    // other params
    tsk_id_t *end_deme;
}
```

So, the struct holds pointers, which we can fill in with arrays later as needed.

lib/tests/test_sweeps.c
jeromekelleher: For the C code tests we won't need anything realistic, just some way of generating trajectories that'll test all the different code paths.

jeromekelleher (commented):
I think we have to remove the reliance on a file before we can merge here @Aalhad-Bhatt as the code is currently untestable and we'll end up doing a bunch of work to accomodate the file pattern than we later undo.

For the first pass, I think what we want is

1. Remove the reliance on a file, where we pass in the arrays when we call set_simulation_model., which we take *copies of*. 
2. For the C tests, we just need some example trajectories that'll help us test the code. Nothing realistic is needed.

Given that, I think It would be easier to back out of the Python level changes for the moment, and try and get the C side of things basically ironed out. If we have full test coverage on the new code, then I think we can merge happily enough as it doesn't affect other models.

Following that, we can then add Python support and perhaps some ways of generating useful trajectories from Python.

Does this sound like a good path?

lib/msprime.h
Aalhad-Bhatt: This is super helpful. Thanks!

jeromekelleher:
LMK when you'd like some input here @Aalhad-Bhatt

Aalhad-Bhatt:
@jeromekelleher I'm happy to receive feedback whenever is good for you. Feedback now should spare me from making the same mistakes while writing tests for the spatial case (which are more complex to write anyway). But I can also wait and make changes all at once.

PR #2437, opened 2025-12-05
Update asv

hossam26644:
Updated the code to work with msprime 1.x syntax (sim_ancestry instead of simulate), assuming that benchmarking against msprime 0.x is not longer necessary. 
Included time and memory tests for stop_at_local_mrca=False
updated the hashes list in the file, and included a line to run from a hashfile instead of just going all commits in order for an hour


benchmarks/benchmarks.py
```diff
@@ -51,16 +51,16 @@ def setup(self):
         # Stuff that depends on the recomb_map_chr22 will fail
         if stdpopsim_available:
             species = stdpopsim.get_species("HomSap")
-            genetic_map = species.get_genetic_map("HapMapII_GRCh37")
```
jeromekelleher: Why are you changing this? I think what you're doing will result in a uniform map, which takes away the main point

benchmarks/benchmarks.py
```diff
@@ -127,16 +127,31 @@ def time_many_replicates(self):
     def peakmem_many_replicates(self):
         self._run_many_replicates()
 
+    def _run_human_chr22_simulate_above_root(self):
+        msprime.sim_ancestry(
+            samples=100,
+            population_size=10**4,
+            recombination_rate=self.recomb_map_chr22,
+            random_seed=234,
+            stop_at_local_mrca=True,
```
jeromekelleher: shouldn't this be false?

jeromekelleher (commented):
Spotted some issues.

I was also hoping that we could do basically all of the benchmarks with an object-oriented approach, where we subclass like
```
class Hudson(LargeSimulationBenchmark):
     def _large_sample_size_params(self):
            return {"sample_size": 10**6, ...}
      def _run_large_sample_size(self):
             self.run(self._large_sample_size_params())
      def run(self, params):
             return msprime.sim_ancestry(**params)

class HudsonOverRoot(Hudson)
       def run(self, params):
             return msprime.sim_ancestry(**{stop_at_local_mrca=False, **params})
```

That way we'll get decent coverage on the performance implications (which are not obvious to me - it's a deep change).

hossam26644:
Very neat suggestion. I just took the liberty to propose a small structural change. Let me know what do you think.

hossam26644:
Ready now

PR #2441, opened 2025-12-11
Documenting the performance hit of stop_at_local_mrca=False closes #2429

### LINKED ISSUE #2429: Characterise performance hit for stop_at_local_mrca=False
When we run with ``stop_at_local_mrca=False`` we use a more expensive test to check for termination. This might end up being significant - we should check, and update the documentation if so.

It might be easiest to do this by doing the existing benchmarks with stop_at_local_mrca=False/True, and seeing if this makes any difference.

hossam26644:
closes #2429


hossam26644:
It rendered fine on my computer. Let me reconfirm that again (paranoia) and put a screenshot here before merging.

hossam26644:
image
This is how it looks locally. but I added an empty line after for consistency.

PR #2442, opened 2025-12-11
Fix/statistical tests above root closes #2428

### LINKED ISSUE #2428: Statistical tests for stop_at_local_mrca=False
We need to do some tests to make sure this is working properly. I think the most powerful test here is to run some simulations with stop_at_local_mrca=True and stop_at_local_mrca=False, and then run ``simplify()`` on the ``stop_at_local_mrca=False`` tree sequences. We can then compare the distributions of the usual stuff across the two sets of replicates (which should be the same). This would detect any biases we're introducing by tracking the additional segments.

I can't think of any other things to test - any ideas?

hossam26644:
First implementation of statistical tests for  simulations above root. 

@Jerome
Can You check that these tests are sufficient before I tidy it up a bit?

### DIFF
--- a/verification.py
+++ b/verification.py
@@ -58,6 +58,7 @@ import attr
 import daiquiri
 import dendropy
 import matplotlib
+import matplotlib.colors as mcolors
 import numpy as np
 import pandas as pd
 import pyslim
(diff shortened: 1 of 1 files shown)


verification.py
```diff
@@ -5504,6 +5505,348 @@ def test_simulate_from_demography(self):
...
+    def plot_above_root_stats(self, df):
+        df_above_root = df[df.stop_at_local_mrca == False]
+        df_stop_at_local_mrca = df[df.stop_at_local_mrca == True]
+        for stat in ["tmrca_mean", "num_trees"]:
+            plot_qq(df_above_root[stat], df_stop_at_local_mrca[stat])
+            f = self.output_dir / f"{stat}.png"
+            pyplot.savefig(f, dpi=72)
+            pyplot.close("all")
+
+        above_root_breakpoints = all_breakpoints_in_replicates(
```
jeromekelleher: Don't bother with the breakpoints, any problems should come through in the number of trees.

jeromekelleher (commented):
Looks good, but I think we can simplify a lot.

verification.py
```diff
@@ -5504,6 +5505,182 @@ def test_simulate_from_demography(self):
             pyplot.close("all")
 
 
+class SimulateAboveRoot(Test):
+
+    colors = list(mcolors.TABLEAU_COLORS.values())
+
+    def run_above_root_stats(self, **kwargs):
+        df_list = []
+
```
jeromekelleher: Should probably take a copy of kwargs here to avoid updating the paramter.

```
kwargs = kwargs.copy()
```
is fine.

PR #2445, opened 2025-12-12
Drop smck hull_offset defaults

### LINKED ISSUE #2443: Drop the default Hull offset of the SMCK model
Should we change the default hull_offset value of the `SmcKApproxCoalescent` model to **1**?

`SmcKApproxCoalescent(hull_offset=1) `is equivalent to the _SMC′_ model, which is generally considered by many to be “equal” to the _StandardCoalescent_ model. In contrast, setting `hull_offset=0` is equivalent to the _SMC_ model, which is not as good an approximation to the _StandardCoalescent_ model. Moreover, the time penalty when using `hull_offset=1` instead of 0 is negligible (according to the analysis in the in-preparation publication).

Was there a reason it was set to 0, or can we simply change it to 1?

hossam26644:
closes #2443

### DIFF
--- a/msprime/ancestry.py
+++ b/msprime/ancestry.py
@@ -1798,7 +1798,7 @@ class SmcApproxCoalescent(AncestryModel):
         This model is implemented using a naive rejection sampling approach
         and so it may not be any more efficient to simulate than the
         standard Hudson model.
-        We recommend using the ``SmcKApproxCoalescent(hull_offset=0)`` instead
+        We recommend using the ``SmcKApproxCoalescent(0)`` instead
 
     The string ``"smc"`` can be used to refer to this model.
     """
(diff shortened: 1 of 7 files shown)


msprime/ancestry.py
```diff
@@ -1861,7 +1861,7 @@ class SmcKApproxCoalescent(ParametricAncestryModel):
     hull_offset: float
 
     # We have to define an __init__ to enforce keyword-only behaviour
-    def __init__(self, *, duration=None, hull_offset=0.0):
+    def __init__(self, *, duration=None, hull_offset):
```
jeromekelleher: I think we should make hull offset a required positional arg, like
```
def __init__(self, hull_offset, *, duration=None):
    ...
```

I don't think we put much thought into this when it was going in initially, just put something in to get things working.

(unless this breaks other things?)

hossam26644:
@jeromekelleher 
Please check this: the test test_gaps in test models fails for the smc_prime (now mapped to smc(k=1)). This tests overlapping segments in a non-discrete genome. I think smc(k=1) is not equavelent to the smc_prim in the case of a non-discrete genome. What do you think?

tests/test_models.py
```diff
@@ -332,9 +332,6 @@ def test_sweep_genic_selection(self):
             msprime.SweepGenicSelection(1)
 
     def test_smck_coalescent(self):
-        model = msprime.SmcKApproxCoalescent()
```
jeromekelleher: Should resinstate this test to check that it raises a TypeError (or whatever it is)
```
# mandatory positional argument
with pytest.raises(TypeError):
        msprime.SmcKApproxCoalescent()

jeromekelleher (commented):
Looks good. Spotted one thing.

We can probably get rid of the ``hull_offset`` named argument in new code as well and just use ``SmcKApproxCoalescent(0)`` etc.

PR #2447, opened 2025-12-15
change fixed-pedigree example fixes #2

### LINKED ISSUE #2: Provide API for transposed haplotypes
It is often useful to obtain the haplotypes in a per-column format. We should provide an API to do this directly.

### LINKED ISSUE #2440: Error in "tracing ancestry through pedigree" docs example
This is supposed to be an example of where we use ``stop_at_local_mrca=False", but we've set it to "True"

https://tskit.dev/msprime/docs/latest/ancestry.html#tracing-ancestry-through-a-pedigree

The problems is the example doesn't fully coalesce anywhere, so there's no mrca. We need a different example.

hossam26644:
fixes #2440 

- simplified the fixed-pedigree example
- fixed the stop_at_mrca_flag=True (should have been false)

### DIFF
--- a/docs/ancestry.md
+++ b/docs/ancestry.md
@@ -94,7 +94,7 @@ this.
 : Coalescent with recombination ("hudson")
 
 {class}`.SmcKApproxCoalescent`
-: General Sequentially Markov Coalescent 
+: General Sequentially Markov Coalescent
 
 {class}`.DiscreteTimeWrightFisher`
 : Generation-by-generation Wright-Fisher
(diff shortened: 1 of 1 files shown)


hossam26644:
Aw ok. Sorry, my bad

docs/ancestry.md
```diff
@@ -2842,34 +2842,52 @@ combined with other {ref}`ancestry models<sec_ancestry_models>`.
...
+2   3   4   1.0 0
+3   5   6   2.0 0
+4   5   6   2.0 0
+5   .   .   3.0 0
+6   .   .   3.0 0
+"""
+pedigree = msprime.parse_pedigree(io.StringIO(ped_txt), sequence_length=100)
+
 ped_ts = msprime.sim_ancestry(
     initial_state=pedigree,
```
hossam26644: I suggest you move the simulations block to the next code block

PR #2535, opened 2026-08-07
remove SLiM metadata; closes #2534

### LINKED ISSUE #2534: remove SLiM mutation metadata
With the upcoming SLiM, mutations won't have metadata; the equivalent information will go in top-level metadata. So, msprime should just not [add metadata to the mutations](https://github.com/tskit-dev/msprime/blob/79dfc96cf9936a2f3324f3545b8aeda5fff2f666/lib/mutgen.c#L398); and pyslim will have a method that puts appropriate things in top-level metadata. Note that the only piece of actual information that msprime puts into metadata is the [SLiM time](https://github.com/tskit-dev/msprime/blob/79dfc96cf9936a2f3324f3545b8aeda5fff2f666/lib/mutgen.c#L294), which is just a simple transform of the mutation time anyhow, so no information is lost.

petrelharp:
This just rips out all metadata stuff from the `SLiMMutationModel`. We should not merge this until SLiM v6 is released.

Potential breakage: if someone runs
```
msprime.SLiMMutationModel(type=1, next_id=1, slim_generation=10)
```
then they get a warning saying that `type` and `slim_generation` will be ignored.
If someone runs 
```
msprime.SLiMMutationModel(1, 1, 10)
```
or
```
msprime.SLiMMutationModel(1, 10)
```
then they get a TypeError because the only valid positional argument is the first one.
The latter is necessary to exclude because previously if someone ran this they'd get a model with type 1 and next ID 10; now (if both arguments were positional) they'd get a model with next ID 1 and block_size 10.

One possible problem: previously, the only required argument was mutation type, and it was the first argument. So, if someone ran
```
msprime.SLiMMutationModel(k)
```
then they would get a model with mutation type k and next ID 0. Now, they would get a model with next ID k. This will cause no practical problems - both "mutation type" and "next ID" have to be a nonnegative integers, and so this just means msprime will not simulate mutations with SLiM ID 0, ..., k-1 when you might have thought it would.

The alternative to this PR would be to just leave everything alone: msprime puts in unneeded metadata, and we can just safely ignore it downstream. But, this is confusing.


tests/test_mutations.py
```diff
@@ -1576,52 +1564,33 @@ def validate_slim_mutations(self, ts, mutation_type=0, slim_generation=1):
...
         mts2 = py_sim_mutations(
             ts, rate=rate, random_seed=random_seed, model=model, discrete_genome=True
         )
 
         t1 = mts1.dump_tables()
         t2 = mts2.dump_tables()
         assert t1.sites == t2.sites
-        # Drop the mutation metadata - we're validating that elsewhere and
-        # it's not worth complicating the Python generator with it.
         t1.mutations.set_columns(
```
jeromekelleher: This set_columns is redundant now, right?

jeromekelleher (approved):
LGTM, spotted on trivial thing.

Did we not set a metadata schema for the slim mutation metadata?

I don't think we need to wait until SLiM 6 is released to merge, but we shouldn't release msprime until it is, though.

Might be worth noting that this also means that these mutations won't work with SLiM < 6 now (this isn't obvious from the CHANGELOG).

petrelharp:
> Did we not set a metadata schema for the slim mutation metadata?

Huh, yeah, I guess not.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 In development.
 
+**Breaking changes**:
+
+- Changes to ``SLiMMutationModel`` to match changes in SLiM 6.0: this mutation
+  model no longer stores metadata, and no longer has the ``type`` and
+  ``slim_generation`` arguments. Mutations generated by msprime >= 1.4.3
+  will require adding metadata (e.g., with pyslim) before they can be read in by SLiM.
```
petrelharp, line 12: I wrote it this way rather than "won't be compatible with SLiM < 6.0" because the mutations are compatible, they just require annotation (by a previous version of pyslim); and in fact using the mutations with SLiM >= 6 also requires pyslim to add things in as well.

petrelharp:
This should be ready now!

PR #2537, opened 2026-08-26
actually wait add *different* SLiM metadata

petrelharp:
This PR adds to the SLiMMutationModel: now, this puts the list of SLiM mutation IDs also in metadata (in binary).

Why? Over at https://github.com/MesserLab/SLiM/issues/664 we realized that really, SLiM mutations should have the list of SLiM mutation IDs in metadata.

Previously, we had this list of IDs in the derived state, which was always kinda awkward, leading to illegal VCFs and so forth. So a common option is to replace the derived state with nucleotides. This was find when we had the mutation metadata right there. But now that we've moved mutation metadata to top-level, we need to keep that information associated with the mutation: and, the nice clean thing to do is to put the IDs, in binary, in metadata. It's much nicer this way.

And we're keeping the same thing in the derived state. This feels odd, but we do want *something* in derived state with the property that two derived states are identical if and only if they are actually identical-by-state; and this fits the bill.

We could have added a function in pyslim to do this, but that would be very awkward for the usual use case of (1) simulate in SLiM with selected mutations; (2) add some neutral mutations with msprime: in particular, the mutations that msprime would add (without this PR) would not have valid metadata. We could change the metadata schema to accomodate that, but this is much better, and pretty straightforward.


petrelharp:
Whoops, fixing docs and coverage.

petrelharp:
The coverage is fine, actually: it's just not hitting the OOM errors and such.

petrelharp:
Okay, the docs are fixed; screenshot of locally built docs:

Screenshot From 2026-08-26 07-33-21

Screenshot From 2026-08-26 07-33-29

Screenshot From 2026-08-26 07-33-35

Screenshot From 2026-08-26 07-33-40

lib/mutgen.c
```diff
@@ -327,13 +326,27 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
         goto out;
     }
     tsk_bug_assert(len < (int) alloc_size);
+    mutation->derived_state = buff;
+    mutation->derived_state_length = (tsk_size_t) len;
+
+    /* Append to metadata */
+    buff
+        = tsk_blkalloc_get(&params->allocator, parent_metadata_length + sizeof(int64_t));
```
bhaller: I’d suggest putting `parent_metadata_length + sizeof(int64_t)` into a local temporary since it gets repeated below; less verbose (lets you get rid of the weird linebreak here) and eliminates the possibility of a mismatch, down the road, between the two places where it is used.

lib/mutgen.c
```diff
@@ -327,13 +326,27 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
...
+    if (buff == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    memcpy(buff, parent_metadata, parent_metadata_length);
+    memcpy(buff + parent_metadata_length, &params->next_mutation_id, sizeof(int64_t));
+    mutation->metadata = buff;
```
bhaller, line 341: do you need to worry about freeing something that is already here in `metadata`?  same question for the assignment to `derived_state` above.

lib/mutgen.c
```diff
@@ -327,13 +326,27 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
...
+    mutation->metadata_length = parent_metadata_length + sizeof(int64_t);
```
bhaller: use that local temporary that I suggested here also, `new_metadata_length` or some such

bhaller (commented):
LGTM, just little nits

lib/mutgen.c
petrelharp: Good question; the answer here is "no", which I know because that's how it was before and because valgrind tells us there's no leaks; I don't remember exactly where to verify that, though.

petrelharp:
Updated with some comments from @bhaller (including to the docs).

lib/mutgen.c
petrelharp: Let's see: [here](https://github.com/tskit-dev/msprime/blob/f8190fc1869148320a152e16862e7bed3fa24a56/lib/mutgen.c#L1143) is where that function gets called; and in that function the mutations already exist as properties of the `site`; the mutations themselves are initialized [here](https://github.com/tskit-dev/msprime/blob/f8190fc1869148320a152e16862e7bed3fa24a56/lib/mutgen.c#L849), which memsets the struct to 0, so the `*metadata` and `*derived_state` are both initialized to `NULL`.

lib/mutgen.c
```diff
@@ -296,8 +296,7 @@ slim_mutator_choose_root_state(
 static int
```
nspope, line 296: worth updating the comment at line 260 ("Previously we set SLiM metadata ...") as this is technically off-base now

lib/mutgen.c
```diff
@@ -327,13 +327,26 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
...
+    mutation->derived_state = buff;
+    mutation->derived_state_length = (tsk_size_t) len;
+
+    /* Append to metadata */
+    buff = tsk_blkalloc_get(&params->allocator, metadata_length);
+    if (buff == NULL) {
+        ret = MSP_ERR_NO_MEMORY;
+        goto out;
+    }
+    memcpy(buff, parent_metadata, parent_metadata_length);
```
nspope, line 339: I'm trying to understand how `parent_metadata` gets used here.
If I'm understanding correctly,  when `keep=True` in `sim_mutations`, then for the "top mutation" in a stack (with no `parent`) the site metadata gets passed as `parent_metadata`, here:

https://github.com/petrelharp/msprime/blob/42a039f1edfd59305f8cb6ed5f09d203a6655a6e/lib/mutgen.c#L1137-L1142

and the implication is that the site metadata gets propagated down the stack by these memcpy lines, breaking the decode with the recommended schema because the offset no longer lines up with the bytes. For example:
```python
#!/usr/bin/env -S uv run --script
# /// script
# dependencies = [
#   "msprime @ git+https://github.com/tskit-dev/msprime@42a039f1edfd59305f8cb6ed5f09d203a6655a6e",
#   "tskit",
# ]
# ///
import msprime
import tskit

tab = msprime.sim_ancestry(4, sequence_length=10, random_seed=5).dump_tables()
tab.mutations.metadata_schema = tskit.MetadataSchema( # copied from PR, docs/mutations.md
    {
        "codec": "struct",
        "type": "object",
        "properties": {
            "derived_states": {
                "items": {"binaryFormat": "q", "type": "number"},
                "noLengthEncodingExhaustBuffer": True,
                "type": "array",
            }
        },
        "required": ["derived_states"],
        "additionalProperties": False,
    }
)
tab.sites.add_row(position=1, ancestral_state="", metadata=b'UH-OH')
ts = msprime.sim_mutations(
    tab.tree_sequence(),
    rate=1,
    model=msprime.SLiMMutationModel(),
    keep=True,
    random_seed=1,
)
for m in ts.mutations():
    print(
        f"id:{m.id}, derived_state:{m.derived_state}, "
        f"metadata.derived_state:{m.metadata['derived_states']}"
    )
# first few lines:
# id:0, derived_state:0, metadata.derived_state:[0]                        
# id:1, derived_state:1, metadata.derived_state:[1]                               
# id:2, derived_state:1,2, metadata.derived_state:[1, 2]                          
# id:3, derived_state:0,3, metadata.derived_state:[0, 3]                                 
# id:4, derived_state:0,3,4, metadata.derived_state:[0, 3, 4]              
# id:5, derived_state:1,5, metadata.derived_state:[1, 5]
# id:6, derived_state:6, metadata.derived_state:[6]              
# id:7, derived_state:7, metadata.derived_state:[8007147407445]  <<--- broken
# id:8, derived_state:8, metadata.derived_state:[9106659035221]
# id:9, derived_state:9, metadata.derived_state:[10206170662997]
```

Maybe this is a non-issue if SLiM doesn't record site metadata; but feels like a bug waiting to happen. For example, someone simulates something with SLiM, adds site metadata themselves, and simulates neutral mutations with this mutation model.

nspope: also maybe I'm missing something here, but say `keep=False` ... then `parent_metadata=NULL` and `parent_metadata_length=0`. Isn't `memcpy(buff, NULL, 0)` undefined behavior?

tests/test_mutations.py
```diff
@@ -1555,14 +1555,41 @@ class TestSLiMMutationModel:
     Tests for the SLiM mutation generator.
     """
 
+    metadata_schema = tskit.MetadataSchema(
```
nspope, line 1558: Wrt comment above about contamination from site metadata, it isn't being caught in the tests because none are using keep=True, for example in https://github.com/petrelharp/msprime/blob/42a039f1edfd59305f8cb6ed5f09d203a6655a6e/tests/test_mutations.py#L1600

nspope (commented):
This largely looks good to me, although I see one edge case with `keep=True`

lib/mutgen.c
petrelharp: Why does github not have a "horrified face" emoji? I'll look into it.

petrelharp: Looks like if this is a bug (and it seems to be), it was a pre-existing bug (like in the last msprime release).

petrelharp:
Thanks a bunch for pointing that out, @nspope - but, in the end there's nothing wrong (IMO).

What's happening is that, as you say, we append to the inherited state and to the inherited metadata. So if we have a new mutation with ID 23 at a pre-existing site with ancestral state `"foo"` and metadata `b"123"`and no parent mutation, then the derived state will be `"foo,23"` and the metadata will be whatever you get when you stick the int64 version of 23 on the end of `b"123"`. 

This is entirely consistent, and I've got an example in the tests now of doing things with this behavior. And, I've made the C tests use and verify this behavior as well. It definitely could be surprising to someone, so I've added a bit to the docstring explaining this.

How's this look?

jeromekelleher (commented):
Generally LGTM. We should fix the bug (which I don't entirely follow) that Nate seems to have dug out though.

jeromekelleher (dismissed):

jeromekelleher:
I guess one question we should address is "why don't we add a metadata schema"? It's not obvious to me what we'd break by doing this, if it's just for the this mutation model.

petrelharp:
> Generally LGTM. We should fix the bug (which I don't entirely follow) that Nate seems to have dug out though.

Sorry, I wasn't clear enough. There is no bug; it's doing exactly what it says it does. (And before it didn't say exactly what that was clearly enough; now I've fixed that.) I got worried that there was going to be some bad memory management in the case where site metadata wasn't NULL, but everything seems just fine.

> I guess one question we should address is "why don't we add a metadata schema"? It's not obvious to me what we'd break by doing this, if it's just for the this mutation model.

The by far most common use case for this method is (1) simulate in SLiM; (2) add neutral mutations with this model. In this use case, the mutations will already have a metadata schema, and it seems like bad practice to overwrite it. I guess we could put in a metadata schema only if there isn't one, but this is getting too complicated. Besides, in my new tests I've got an example where we actually use this model with [a different metadata schema](https://github.com/tskit-dev/msprime/pull/2537/changes#diff-ffe44719059fe4196914e26af4ef029c498cc5bfa2abc72619d21c858a9a0bbaR1680); currently there's no use case for doing this but we should leave the door open.

lib/mutgen.c
petrelharp: (Resolution: not a bug.)

petrelharp:
But: @nspope - mind seeing if you agree with me on the "not a bug" thing?

nspope:
> we append to the inherited state and to the inherited metadata. So if we have a new mutation with ID 23 at a pre-existing site with ancestral state "foo" and metadata b"123"and no parent mutation, then the derived state will be "foo,23" and the metadata will be whatever you get when you stick the int64 version of 23 on the end of b"123".

What seems like a bug to me (or at least a footgun) is that the inheritance of site metadata breaks metadata decoding with the recommended MutationTable schema for this mutation model ... to get the right integers back, you'd need to first find the length of the site metadata, then take the remaining bytes and decode those, or you'll get gibberish.

petrelharp:
Huh: but, I think rephrasing what you're saying is that if you use the wrong metadata schema then you'll get the wrong information? What if I just give this caveat in the docs after I introduce that metadata schema?  (Note that you don't need to "find the length of the site metadata", you just need the right schema, as in my new test example.)

nspope:
That is effectively what I'm saying, yes ... the site metadata schema would need to be replicated in the mutation metadata schema to get the right decoding, like you do in that test. I'm imagining someone doing something like,
```python
import msprime
import tskit

tab = msprime.sim_ancestry(4, sequence_length=10, random_seed=5).dump_tables()
tab.sites.metadata_schema = tskit.MetadataSchema.permissive_json()
tab.sites.add_row(position=1, ancestral_state="1", metadata={"very": "important"})
tab.mutations.metadata_schema = tskit.MetadataSchema( # copied from PR, docs/mutations.md
    {
        "codec": "struct",
        "type": "object",
        "properties": {
            "derived_states": {
                "items": {"binaryFormat": "q", "type": "number"},
                "noLengthEncodingExhaustBuffer": True,
                "type": "array",
            }
        },
        "required": ["derived_states"],
        "additionalProperties": False,
    }
)
ts = msprime.sim_mutations(
    tab.tree_sequence(),
    rate=1,
    model=msprime.SLiMMutationModel(),
    keep=True,
    random_seed=1,
)
for m in ts.mutations():
    print(
        f"id:{m.id}, derived_state:{m.derived_state}, "
        f"metadata.derived_state:{m.metadata['derived_states']}"
    )
```
(which looks like a reasonable thing to do) and then being surprised when they get gibberish on decode. 

So it sounds like the caveat in the docs would be "if you need to add site metadata, make sure the mutation metadata schema accounts for this"? Or am I making this more complicated than it actually is?

petrelharp:
> (which looks like a reasonable thing to do)

Actually, I think this is *not* a reasonable thing to do (at least, to do and then expect it to painlessly work), because the SLiMMutationModel is supposed to work specifically with SLiM, which has specific requirements on all the metadata. If you start trying to do SLiM things with a tree sequence that doesn't conform to what SLiM expects then it's not going to work smoothly. 

> So it sounds like the caveat in the docs would be "if you need to add site metadata, make sure the mutation metadata schema accounts for this"? Or am I making this more complicated than it actually is?

I agree, the caveat in the docs is something like this. Here's the section on this I'm drafting right now:
```

Both SLiM models simply create the derived state by appending to the inherited state,
and so if you use {func}`.sim_mutations` with ``keep=True`` to add SLiM mutations
to a tree sequence that already has non-SLiM mutations, you may end up with
mutations with odd derived states. For instance, if a SLiM mutations occurs
at the same site as, and inheriting from, a previous mutation
with derived state ``"A"``, the SLiM mutation might have derived state ``"A,345"``.
For a mutation without a parent, this inherited state is the ancestral state
of the site. If the site was not already present in the tree sequence, this model
will create a new site with the empty string as the ancestral state,
but if the site is already present and with a nonempty ancestral state,
then the same thing will happen.

The same thing is true of metadata: metadata is simply appended (in binary)
to the "inherited" metadata, which is from the parent mutation (if any) or from
the site (otherwise). So, if in your tree sequence your sites have nonempty metadata,
then any SLiM mutations that occur on those sites will have site metadata
prepended to their metadata (and so the metadata schema above will not work).
However, these SLiM mutation models are designed specifically to work with SLiM-format
tree sequences; if you're trying to use them in another situation then you probably
don't have the right tool for the job.
```

nspope:
Looks good to me...

PR #2539, opened 2026-08-27
Add new SLiMv6 mutation model

petrelharp:
As discussed on slack, rather than #2537 we should really just add a new mutation model for SLiM version 6.0+. Here's the code from #2537 added on top of what's already there.

There's currently a little code duplication that I might be able to reasonably tidy up.

Just to be clear:
- recommended upgrade usage is to change `SLiMMutationModel` to `SLiMv6MutationModel`, remove the now-uneeded `type` and `slim_generation` arguments (and wrap the whole call to `sim_mutations` in `pyslim.add_mutation_metadata( )`)
- if someone uses `SLiMMutationModel` and tries to load it into the new SLiM then they'll get an informative error
- if someone uses `SLiMMutationModel` on a SLiM-produced tree sequence (the usual case) and tries to look at mutation metadata they'll get tskit "can't decode metadata" errors

The last thing is unfortunate we can't get a more informative error in there, but I don't think there's a reasonable way to do so, and if someone is inspecting metadata then they're reasonably sophisticated anyhow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 In development.
 
+**New features**:
+
+- A new mutation model, ``SLiMv6MutationModel``, is introduced to match changes in
+  SLiM 6.0: this mutation behaves as ``SLiMMutationModel``, but without the ``type``
+  and ``slim_generation`` arguments, and stores different information in metadata
+  (the list of mutation IDs, in binary). ``SLiMMutationModel`` will remain for
+  backwards compatibility, but code for use with SLiM versions 6.0 and above should
+  use ``SLiMv6MutationModel`` ({issue}`2534`, {pr}`2539`, {user}`petrelharp`)
+
 ## [1.4.2] - 2026-05-20
 
 Bugfix release.
--- a/docs/mutations.md
+++ b/docs/mutations.md
@@ -105,8 +105,11 @@ for more information.
 {class}`.InfiniteAlleles` (Integers)
 : A generic infinite-alleles mutation model
 
+{class}`.SLiMv6MutationModel` (Integers)
+: An infinite-alleles model producing SLiM-style mutations for SLiM v6+
+
 {class}`.SLiMMutationModel` (Integers)
-: An infinite-alleles model producing SLiM-style mutations
+: An infinite-alleles model producing SLiM-style mutations for older versions of SLiM
 
 {class}`.MatrixMutationModel` (General finite state model)
 : Superclass of mutation models with a finite set of states
@@ -485,7 +488,8 @@ Here are the available models; they are documented in more detail below.
 - {class}`.TPM`: Two-phase mutation model for microsatellite repeat copy number. DiRienzo et al. ('94)
 - {class}`.EL2`: Two-phase mutation model, equal rate, linear bias model for microsatellite repeat copy number. Garza et al. ('95)
 - {class}`.InfiniteAlleles`: A generic infinite-alleles mutation model
-- {class}`.SLiMMutationModel`: An infinite-alleles model of mutation producing SLiM-style mutations
+- {class}`.SLiMv6MutationModel`: An infinite-alleles model of mutation producing SLiM-style mutations for SLiM v6+
+- {class}`.SLiMMutationModel`: An infinite-alleles model of mutation producing SLiM-style mutations for older versions of SLiM
 
 (sec_mutations_matrix_mutations_models)=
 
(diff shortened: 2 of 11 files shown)


petrelharp:
Okay: this is ready to go. My first commit here just copies everything in; the second commit refactors a little to share some machinery. I think it's pretty clean, but maybe someone will see something sketchy?

petrelharp:
Here's the rendered docs:
Screenshot From 2026-08-27 12-25-55
Screenshot From 2026-08-27 12-26-07
Screenshot From 2026-08-27 12-26-13
Screenshot From 2026-08-27 12-26-19
Screenshot From 2026-08-27 12-26-27

petrelharp:
As before, the codecov errors are about out-of-memory lines and things like that.

docs/mutations.md
```diff
@@ -105,9 +105,12 @@ for more information.
 {class}`.InfiniteAlleles` (Integers)
 : A generic infinite-alleles mutation model
 
-{class}`.SLiMMutationModel` (Integers)
+{class}`.SLiMv6MutationModel` (Integers)
 : An infinite-alleles model producing SLiM-style mutations
```
bhaller: "for SLiM 6"?

docs/mutations.md
```diff
@@ -485,7 +488,8 @@ Here are the available models; they are documented in more detail below.
 - {class}`.TPM`: Two-phase mutation model for microsatellite repeat copy number. DiRienzo et al. ('94)
 - {class}`.EL2`: Two-phase mutation model, equal rate, linear bias model for microsatellite repeat copy number. Garza et al. ('95)
 - {class}`.InfiniteAlleles`: A generic infinite-alleles mutation model
-- {class}`.SLiMMutationModel`: An infinite-alleles model of mutation producing SLiM-style mutations
+- {class}`.SLiMv6MutationModel`: An infinite-alleles model of mutation producing SLiM-style mutations
```
bhaller: again, "for SLiM 6"?

lib/mutgen.c
```diff
@@ -347,9 +347,10 @@ slim_mutator_choose_root_state(
...
 static int
-slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
-    const char *parent_allele, tsk_size_t parent_allele_length,
-    const char *parent_metadata, tsk_size_t parent_metadata_length, mutation_t *mutation)
+slim_both_mutator_transition(mutation_model_t *self, const char *parent_allele,
+    tsk_size_t parent_allele_length, const char *parent_metadata,
+    tsk_size_t parent_metadata_length, mutation_t *mutation,
+    const tsk_size_t new_metadata_length, char **metadata_buffer)
 {
     int ret = 0;
```
bhaller: a comment here talking about how this is called by both of the models, and those models handle bounds-checking and incrementing the id as well as copying in the new metadata, would be helpful.  this is just shared code, factored out.

lib/mutgen.c
```diff
@@ -379,39 +381,130 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
...
+    const tsk_size_t new_metadata_length = SLIM_MUTATION_METADATA_SIZE;
 
+    ret = slim_both_mutator_transition(self, parent_allele, parent_allele_length,
+        parent_metadata, parent_metadata_length, mutation, new_metadata_length, &buff);
+    if (ret < 0) {
+        goto out;
+    }
+
+    // copy the final bit of metadata in (this is different than the v6 model)
+    copy_slim_mutation_metadata(params, buff + parent_metadata_length, mutation->time);
```
bhaller, line 444: why is the mutation time, in particular, going in here?  shouldn't this be putting in a whole SLiM 5.2 mutation metadata chunk, with all the stuff that's in that schema?

lib/mutgen.c
```diff
@@ -379,39 +381,130 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
...
+    }
+    params->next_mutation_id++;
+
 out:
     return ret;
 }
 
 static int
-slim_mutator_free(mutation_model_t *self)
+slim_both_mutator_free(mutation_model_t *self)
```
bhaller: I'd put the "both" stuff together, rather than interleaving code for "both" and particular models; that's a bit confusing.

lib/mutgen.c
```diff
@@ -379,39 +381,130 @@ slim_mutator_transition(mutation_model_t *self, gsl_rng *MSP_UNUSED(rng),
...
     tsk_blkalloc_free(&params.allocator);
     return 0;
 }
 
+/*************************
+ * SLiMv6 mutation model */
+
+/* To be fully SLiM-compatible, there also needs to be some information in
+ * top-level metadata.  But, it is not our job to add it in here, and that
+ * should be added after the fact with the method pyslim.add_mutation_metadata().
```
bhaller, line 463: Oh, right; interesting.  So the SLiMv6 mutation model in msprime is really totally dependent upon pyslim to make things correct.  Hmm.  Are you _really_ sure you don't want to move this mutation model over to pyslim?  It feels really weird that it is in msprime.

msprime/_msprimemodule.c
```diff
@@ -796,6 +801,98 @@ static PyTypeObject SLiMMutationModelType = {
     .tp_new = PyType_GenericNew,
 };
 
+/*===================================================================
+ * Slim mutation model, v6
```
bhaller: Ahem, "SLiM", please, lol

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
 In development.
 
+**New features**:
+
+- A new mutation model, ``SLiMv6MutationModel``, is introduced to match changes in
+  SLiM 6.0: this mutation behaves as ``SLiMMutationModel``, but without the ``type``
+  and ``slim_generation`` arguments, and stores different information in metadata
+  (the list of mutation IDs, in binary). ``SLiMMutationModel`` will remain for
+  backwards compatibility, but code for use with SLiM versions 6.0 and above should
+  use ``SLiMMutationModelv6`` ({issue}`2534`, {pr}`2539`, {user}`petrelharp`)
```
bhaller: SLiMv6MutationModel

bhaller (commented):
OK, lots of python glue here that I skimmed since I don't understand it.  But hopefully my comments are useful.

lib/mutgen.c
petrelharp: Ah, that's been factored out into this `copy_slim_mutation_metadata` function, which isn't visible here since it hasn't changed. Here, I think [this link](https://github.com/tskit-dev/msprime/pull/2539/changes#diff-704c3dd77f01207ccb2ca11f660f3bdf2f3f802b58d9f1f2737beba50234a0c0R289) should work? (The answer is that everything we need to know for the metadata chunk is (a) in params, or (b) the time.

bhaller: OK, weird though since the SLiM 5.2 mutation metadata model has all kinds of stuff in there – origin tick, fixation tick, selection coefficient, etc.  All of that is in "params"?  :-O  But OK, sure.  :->

lib/mutgen.c
petrelharp: Briefly: yes, I am really *really* **really** sure.  A thing I *could* do is to write a wrapper function in pyslim that does `sim_mutations -> add_mutation_metadata`.

bhaller: OK good, just checking!  :->

lib/mutgen.c
petrelharp: Well some of that is just defaults, like selection coefficient is always 0.  And hm: it's actually [not got fixation tick](https://github.com/tskit-dev/pyslim/blob/01fa816b22fbffb635cfc9a8e6a2fa5d01399e47/pyslim/slim_metadata.py#L1576), it's only got mutation type, selection coefficient, nucleotide, origin tick, and origin subpopulation.

msprime/mutations.py
```diff
@@ -144,6 +146,47 @@ def __str__(self):
...
+    inherited state (which is the derived state of the parent mutation, if any,
+    or the ancestral state otherwise). Also, it sets the metadata to be the
+    64-bit integer n appended to the bytes of the inherited metadata
+    (which similarly comes from the parent mutation, if any, or the site).
+    So, if this mutation model is applied to a tree sequence with existing
+    incompatible mutations (especially if the existing sites have nonempty metadata)
+    the result may be surprising.
+
+    To update your code to use this model instead of {class}`.SLiMMutationModel`,
+    remove the ``type`` and ``slim_generation`` arguments from , and follow
```
nspope:
```suggestion
    remove the ``type`` and ``slim_generation`` arguments, and follow
```

tests/test_mutations.py
```diff
@@ -1651,49 +1693,181 @@ def test_slim_generation(self):
...
+                "derived_states": [],
+            }
+            mt.sites.append(s.replace(metadata=md))
+        mt.mutations.clear()
+        mt.mutations.metadata_schema = self.metadata_schema_with_extra
+        for m in mts.mutations():
+            md = {
+                "tag": f"lmnop{m.site}",
+                "derived_states": [int(x) for x in m.derived_state.split(",")],
+            }
```
nspope, line 1827: shouldn't this mutation + metadata be added to mt.mutations, like above? isn't the table staying empty here after the `clear` --- so below will stack mutations on pre-existing sites, but not pre-existing mutations. Is that the intent?

docs/mutations.md
```diff
@@ -931,4 +942,107 @@ This was already present in the tree sequence, so its derived state is not modif
 `0,3`. We can rationalise this, post-hoc, by saying that the type 1 mutation `3`
 has "erased" the type 2 mutations `100` and `102`.
```
nspope: there's no "type=2" any more with the new class's API example above --- so is this still accurate?

nspope (commented):
Docs and tests look good to me (a couple minor points); I haven't gone over the C implementation carefully yet.

jeromekelleher:
I'm away until next week now, and can review and push out a release then. Feel free to merge and do the release before that if you want to get it out though.

tests/test_mutations.py
petrelharp: whoops, good catch - I missed `            mt.mutations.append(m.replace(metadata=md))`

petrelharp:
I did a final pass through and found a couple of nits; merging after this (presumably) passes.

PR #2541, opened 2026-08-28
census node per lineage, not segment: closes #2540

### LINKED ISSUE #2540: Census event coinciding with `end_time` leads to an `abort()` in `msp_insert_uncoalesced_edges()`
Hello! I think I found a bug that previously wasn't reported. I was likely trying to do something silly and hit a weird edge case, but I decided to report it anyway.

Basically, if I stop the simulation at a fixed `end_time` and try to take a census in the same generation, the interpreter crashes (so it's not an exception that can be caught).

I narrowed it down to a very simple minimal reproducing example:
```
In [1]: import msprime

In [2]: msprime.__version__
Out[2]: '1.4.1'

In [3]: dem = msprime.Demography.isolated_model([1000])
   ...: dem.add_census(time=500)
   ...: msprime.sim_ancestry(
   ...:     samples=4, demography=dem, sequence_length=1e6,
   ...:     recombination_rate=1e-8, end_time=500, random_seed=42)
Bug detected in lib/msprime.c at line 6148. Please report this issue on GitHub, ideally with a reproducible example. (https://github.com/tskit-dev/msprime/issues)
```
It doesn't crash with no recombination, the code requires positive recombination rate.

Looking at the [referenced C line](https://github.com/tskit-dev/msprime/blob/fd7c22c6db89c1d654405bfcccee618f8e1ac2a7/lib/msprime.c#L6148), it is in the `msp_insert_uncoalesced_edges(msp_t *self)` function under the comment that says `For every segment add an edge pointing to this new node`. 

I don't know much about `msprime` internals, but likely that adding census nodes breaks some kind of internal invariant? The comment says "for every segment" but [census docs](https://tskit.dev/msprime/docs/latest/api.html#msprime.Demography.add_census) say "add a node to each branch of every tree" -- but there can be multiple trees in a segment?

Let me know if you have any further questions!

Egor Lappo

petrelharp:
I think this was a conceptual error: all segments in a given lineage at a given time point in the simulation are in the same genome (=node=haplosome), and hence in doing a census event we should assign a single node to each *lineage*, not each segment.

If we agree, this needs an update to the CHANGELOG (and maybe another test? I'm not sure if we've a good way to test this).

Also note the test that triggers #2540 (but this - I think - fixes).

### DIFF
--- a/lib/msprime.c
+++ b/lib/msprime.c
@@ -7421,6 +7421,7 @@ msp_census_event(msp_t *self, demographic_event_t *event)
     lineage_t *lin;
     tsk_id_t i, j;
     tsk_id_t u;
+    tsk_node_table_t *nodes = &self->tables->nodes;
 
     for (i = 0; i < (int) self->num_populations; i++) {
         for (j = 0; j < (int) self->num_labels; j++) {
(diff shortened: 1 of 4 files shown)


petrelharp:
Verification that (after fixing it) the new test catches the bug without this fix:
```
$ git checkout main lib/msprime.c
$ make
$ uv run pytest tests/test_demography.py::TestCensusEvent::test_census_at_end_time
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.13.12, pytest-9.1.1, pluggy-1.6.0
rootdir: /home/peter/projects/tskit-dev/msprime
configfile: pyproject.toml
plugins: xdist-3.8.0, cov-7.1.0
4 workers [1 item]      
Fatal Python error: Aborted
```

petrelharp:
What does this affect? So I think it definitely is a bug, in that we're calling things different nodes that aren't really; but this only affects census nodes: it's not going to change anything about the subsequent simulation. So, it's a pretty minor bug. It'd only affect people who wanted to see how many distinct ancestors there were at a given time using a census event

I got curious: how big of an effect is this? Here's a jupyter notebook: 
[census_edges.ipynb](https://github.com/user-attachments/files/31599833/census_edges.ipynb)
(run with `uv run --with plotnine --with jupyterlab jupyter lab`),
and in this simulation of 100 samples from rho=0.01 there's a lot of census nodes that have more than one segment (like, a third of them): here is all the edges below census nodes (of which there are 154), and the color is how many distinct segments a given parent census node has (55 have > 1):
Screenshot From 2026-08-29 08-44-09

petrelharp:
(Note: one might naively hope that adding census nodes literally does not change the output of the simulation, seed-by-seed, except for the census nodes, but this turns out not to be true; I'm not actually sure why.)

jeromekelleher:
> (Note: one might naively hope that adding census nodes literally does not change the output of the simulation, seed-by-seed, except for the census nodes, but this turns out not to be true; I'm not actually sure why.)

This is almost never the case, and it's rarely worth the effort of figuring out exactly why. I wouldn't worry about it.

lib/msprime.c
```diff
@@ -7431,15 +7431,15 @@ msp_census_event(msp_t *self, demographic_event_t *event)
 
             while (node != NULL) {
                 lin = (lineage_t *) node->item;
+                // Add an edge to the edge table.
```
jeromekelleher:
```suggestion
```
This comment is just confusing

jeromekelleher (approved):
I agree, this is the correct approach. Thanks @petrelharp! 

Let's drop the inaccurate comment though

lib/msprime.c
petrelharp: hah I just copied that over, didn't notice

petrelharp:
This code nerd sniped me: thianks to the [code pointed to by #2540](https://github.com/tskit-dev/msprime/blob/fd7c22c6db89c1d654405bfcccee618f8e1ac2a7/lib/msprime.c#L6148) I thought I knew how to make this work with DTWF at integer times. That was wrong: the DTWF generation comes after the census event, so the problem there is elsewhere (and more complicated, looks like). But still, this last commit makes the census event a bit more general: it now won't fail if there happens to be a pre-existing node at the given time. I think this is pretty straightforward and not fraught with peril but if we are nervous I could revert to commit n-1 to just get the comment update.

jeromekelleher (approved):
LGTM, happy to merge once we have test coverage at a reasonable level.

petrelharp:
Got it; we're down to expected misses:
Screenshot From 2026-09-01 09-45-02


Will merge now after a squash.

PR #2546, opened 2026-09-15
change to slimv5 model; deprecate old model

petrelharp:
This is a better resolution than what we had previously: `SLiMMutationModel` is renamed `SLiMv5MutationModel`; the old name still works but throws a FutureWarning. We don't necessarily ever need to actually remove `SLiMMutationModel` though. This is pretty straightforward. Sorry I didn't do it this way the first time around.


jeromekelleher (approved):
LGTM.  Merge away whenever you're happy.

Is the plan to pop out another point release?

petrelharp:
Thanks!

> Is the plan to pop out another point release?

Yep: since it's backwards compatible, this seems fine.  Although maybe I'll wait for something else to happen or SLiM v6 to be released?

jeromekelleher:
I don't have an opinion really, it's up to you when we release

